PR #208, opened 2015-04-13
Fix VaspInput.write_input (TypeError)

ExpHP:
It was producing TypeError when used due to it passing in Incar/Potcar/etc instances to the file.write method.

Also, none of the tests called this method, so I added one.


shyuep:
Thanks.

PR #221, opened 2015-08-05
Multiple dielectricfunction elements in vasprun.xml

migueldiascosta:
When [using the GW routines for the determination of frequency dependent dielectric matrix](http://cms.mpi.univie.ac.at/vasp/vasp/Using_GW_routines_determination_frequency_dependent_dielectric_matrix.html), there are multiple dielectricfuction elements in vasprun.xml, distinguished by element comments, and the Vasprun class is keeping the last one, which is actually the 'screened Coulomb potential'.

This tries to add support for these additional dielectricfunction elements without breaking the default behaviour by 

1. if there is no comment (there is only one dielectricfunction element), keeping the previous behaviour

2. if an element with comment "INVERSE MACROSCOPIC DIELECTRIC TENSOR (including local field effects in RPA (Hartree))" exists, storing it in the default dielectric attribute and 

2. storing any others (e.g., comments "HEAD OF MICROSCOPIC DIELECTRIC TENSOR (INDEPENDENT PARTICLE)", "1 + v P,  with REDUCIBLE POLARIZABILTY P=P_0 (1 -(v+f) P_0)^-1", "screened Coulomb potential") in an other_dielectric dictionary using the comment as key.

For your consideration.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -242,7 +242,19 @@ class Vasprun(PMGSONable):
         The real and imaginary part of the dielectric constant (e.g., computed
         by RPA) in function of the energy (frequency). Optical properties (e.g.
         absorption coefficient) can be obtained through this.
-        The date is given as a tuple of 3 values containing each of them
+        The data is given as a tuple of 3 values containing each of them
+        the energy, the real part tensor, and the imaginary part tensor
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks for the contribution. I made a minor modification to maintain Py3k compatibility.

PR #232, opened 2015-08-22
agl_thermal

ctoher:
Request for code review only. Further modifications may be required to enable operation with MPWorks framework.


shyuep:
Hi Comac,

Can you pls move agl_thermal to under analysis. And have a more informative name than "agl_" (I have no idea what agl stands for). You also should not be commiting the distribute packages.

PR #242, opened 2015-09-09
fixed python3 compatibility, fixed tabs/spaces

sonium0:


mbkumar:
Hello Alex,

Danny forwarded your email. Thank you for adding your modifications and you are always welcome to contribute code and bug fixes.

Blind removal of 6th line in LOCPOT file within pymatgen is a bad idea. The LOCPOT object created by pymatgen ignores the 6th line. If we are using those LOCPOTs (written to a file by pymatgen), then removing 6th line will introduce a bug. So please change the code such that LOCPOT modification is optional.

sonium0:
Hi, actually I'm not modifying the LOCPOT but creating a copy called LOCPOT.sxd in the same directory and submit this to sxdefectalign. The only thing I was wondering was if this should be a) deleted afterwards and/or b) done in a scratch directory.

mbkumar:
Hello Alex,

I understand that. What I mean is the input LOCPOT file may have the 6th
line deleted already. (This happens automatically when pymatgen reads and
writes a LOCPOT file). So, please make the copying of LOCPOT to LOCPOT.sxd
optional.

In case of copying to LOCPOT.sxd, do it in scratch directory. That way you
don't have to delete it explicitly.



On Mon, Sep 14, 2015 at 6:33 AM, sonium <notifications@github.com> wrote:

> Hi, actually I'm not modifying the LOCPOT but creating a copy called
> LOCPOT.sxd in the same directory and submit this to sxdefectalign. The only
> thing I was wondering was if this should be a) deleted afterwards and/or b)
> done in a scratch directory.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/materialsproject/pymatgen/pull/242#issuecomment-140077029>
> .
>



-- 
Bharat
Postdoc
Lawrence Berkeley Lab
510-984-8628

PR #245, opened 2015-09-21
NwOutput : support parsing hessian matrices.

Bismarrck:
* Support parsing hessian & projected hessian matrices from nwchem outputs.
* Support parsing raw frequencies.
* Unit tests are added and tested.


shyuep:
THanks for the nice contribution!

PR #247, opened 2015-09-22
Gauss

gVallverdu:
* fix compatibility issues between gaussian03/gaussian09
* new methods to read potential energy surface computed with gaussian
* new methods to read electronic transitions computed with TD-DFT


shyuep:
Thanks for the contributions!

PR #252, opened 2015-10-09
Converter between AiiDA StructureData object and pymatgen Molecule/Structure objects

merkys:
I present a converter between AiiDA StructureData and pymatgen Molecule/Structure objects. Since testing requires a sophisticated environment, it is done at AiiDA side.

Please check whether the method names comply with the pymatgen's style.


shyuep:
Thanks!

PR #256, opened 2015-10-16
Problem with multiple preambles in XDATCAR

nisse3000:
I've extended the constructor of Xdatcar to enable successful treatment of multiple consecutive preambles in XDATCAR. That case seems rare, but I've just encountered it.


shyuep:
thanks

PR #259, opened 2015-10-30
Enhance: GaussianOutput

Bismarrck:
The class ``GaussianOutput`` now supports parsing:

* Cartesian forces
* Frequencies
* Normal modes

Also add some unit tests.


Bismarrck:
I will fix my codes and then give a new pull request.

PR #264, opened 2015-11-06
Deformed structure set

montoyjh:
Added the `elasticity` package, which includes modules

- tensors.py - `SQTensor` base class for 2-D tensors and relevant matrix operations
- stress.py - `Stress` class for various operations related to stress
- strain.py - `Deformation`, `Strain`, and `DeformedStructureSet` classes for operations related to  structural deformation and strain
- elastic.py - `ElasticTensor` class for operations related to the elastic tensor

Also added unit tests for each module.


shyuep:
Thanks. This looks great in terms of coding, but can you move the entire elasticity branch under the analysis package. The pymagen.X level if reserved for high-level abstraction. Specific property analyses are parked under pymatgen.analysis.

shyuep:
BTW, you do not need to withdraw the pull request. You simplt need to commit and add on the commits to this pull request. Let me know when it is ready.

montoyjh:
Okay, I think it's ready now.

PR #268, opened 2015-11-14
added explicit polar decomposition method to allow for more flexible …

montoyjh:
…scipy compatibility

### DIFF
--- a/pymatgen/analysis/elasticity/tensors.py
+++ b/pymatgen/analysis/elasticity/tensors.py
@@ -179,7 +179,17 @@ class SQTensor(np.ndarray):
         """
         calculates matrices for polar decomposition
         """
-        return polar(self, side=side)
+        if side not in ['right', 'left']:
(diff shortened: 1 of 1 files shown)


shyuep:
Hmm... not quite sure why we need to implement an explicity polar decompositno method?

montoyjh:
We were having an issue with the scipy dependency, since the polar method was implemented in sciyp/0.14.0 and some folks seem to be using older versions.  Since it was the only scipy dependency in analysis.elasticity, I made it explicit to remove the scipy dependency but keep the method, but an alternative would be to just remove the method, since scipy.linalg.polar can be used with the object anyway.

shyuep:
0.14.0 is ancient. I think the folks should upgrade. pymatgen's requirements.txt already stipulates scipy 0.16.0. So there really isn;t a reason for < 0.14.0.

montoyjh:
Ah, I thought scipy was optional.  Probably just leave it, then.

PR #278, opened 2015-12-02
S Correction

Zhuoying:
add S correction and unit test.


shyuep:
This is not the correct implementation. You are changing anything that has a S. That is not correct. What happens with sulfates?

The correct way of dealing with sulfide is not mdify the sulfide itself, but to add it to the compound energies. That way, anything that has reduced_formula = "S" will be modified to the right sulfur energy.

Zhuoying:
Make energy correction on sulfur rather than sulfide.

PR #287, opened 2015-12-15
add as_base_units for FloatWithUnit and ArrayWithUnit. This is helps …

computron:
This adds an ``as_base_units`` functionality to FloatWithUnit and ArrayWithUnit. When I have these two objects, I often want to quickly convert them to base units and would find this function a handy shortcut.


computron:
Note - there is a second commit that helps the unitized decorator function as expected when returning unit objects.

PR #292, opened 2015-12-25
Upstrm

mbkumar:
Renaming zeo.py in pymatgen/io/zeo.pyy to pymatgen/io/zeopp.py to resolve naming conflict.


mbkumar:
Trying to cherrypick a particular commit, but not happening.

PR #299, opened 2016-01-14
Added an order parameter class for analysis of local structural features

nisse3000:
… plus required unit tests.
My thoughts were to put the entire class actually into the structure_analyzer.py, if this is wanted from other people, too.

prototype_structures.py is mainly a helper class to easily set up prototypic structures for the order parameter unit tests.  But they could be generally also valuable to have.


shyuep:
Thanks for the excellent contribution. However, I would like to suggest a few changes.

i. the in_bound method seems to be more easily dealt with as max(lower, min(upper, value)). In fact, a variant is used in pymatgen.utils.num_utils.abs_cap.

ii. While the prototype structure is useful, the implementation is not the most efficient. The standard fcc, bcc and hcp structures can already be generated using Structure.from_spacegroup("Fm-3m", ...) for example. While I recognize the value of these convenience methods, I think it would be best to add them as a classmethod in Structure, e.g., Structure.fcc(a, species). Then a user only needs to specify the desired lattice parameters and species to get an fcc structure.

nisse3000:
Thanks, Prof. Ong.  I'll implement your suggestions ASAP.

nisse3000:
Prof. Ong, can the order parameter class now be merged with pymatgen?  Or, are there any other changes that you suggest?  The checks have passed successfully now that I have addressed problems in the unit tests related to numerical precision.

shyuep:
Hi Niels, have the changes been implemented?

nisse3000:
Yes.  I have however removed the prototype structure class.

PR #302, opened 2016-01-29
Splines

blondegeek:
Added smooth tolerance and band interpolation failures warnings to get_plot method of BSPlotter.


shyuep:
Many thanks Tess!

PR #308, opened 2016-03-04
fix qchem use_pcm method

matk86:
## Summary

* For QChem versions >=4.2 the solvent parameter section is named "solvent" instead "pcm_solvent"
* Fix use_pcm method in QcTask


shyuep:
Thanks.

PR #309, opened 2016-03-04
Substrate Analyzer

shyamd:
## Summary

Added new substrate analyzer module

* Topology search of matching interfaces for a substrate and film pair using Zur and McGill algorithm
* Calculates multi-layer elastic energy due to super-lattice mismatch from topological search


shyuep:
Thanks. But can you also put in the proper doc for all parameters in all public methods. Most of the public methods are not documented with what each parameter means or does.

PR #320, opened 2016-04-07
Adding in Piezo Tensor to Pymatgen

shyamd:
Added in Piezotensor 
Updated Outcar to automatically parse piezo and dielectric tensors
Added in symmetry operations for rank 3 and rank 4 tensors


shyuep:
Great. but instead of having Operations.transform_r2, transform_r3, and transform_r4, isn't it better to simply have one transform method that works for all dimensions?

```python
dim = tensor.shape
rank = dim[0]
assert all([i == rank for i in dim])

ind1 = "abcdefgh"
ind2 = "ijklmnopq"

n = ["%s%s" % i,j for i, j in zip(ind1[:rank], ind2[:rank]]
m = ",".join(n) + ind1[:rank] +"->"+ ind2[:rank] 
args = [self.rotation_matrix] * rank + [tensor]
return np.einsum(m, *args)
```

shyamd:
Joey had the same idea. I've committed his code in. I think we might have to assert that each dimension is size 3, since the rotation matrix is a 3x3. It technically doesn't make sense in any other form.

Issue #325, opened 2016-04-11
Predictor-corrector parsing, and storage in Structure


aykol:
## System

* Pymatgen version: master
* Python version:  2.7.* and 3.*
* OS version: OS X & Linux

## Summary

For the predictor-corrector data, currently pymatgen looks for a chunk after reading the velocities from CONTCAR, and stores them as "site_properties" in the Structure. There are two issues here:

(1) There is a 3-line-long preamble in the "predictor-corrector" chunk, listing a "key", POTIM, and a set of thermostat parameters. Since site_property stores an array of N data points, upon parsing, these 3 lines spuriously occupy the "site_properties" of the first 3 sites. Consequently, the correct correspondence between a site and the "predictor-corrector" data is lost, and more importantly the predictor-corrector information for the last 3 atoms are not stored. Therefore when a POSCAR is generated from this Structure, the missing information causes VASP to complain and halt.

(2) The "predictor-corrector" chunk consists of 3 sets of 3 x N data points corresponding to positions and derivatives. Currently pymatgen only stores 1 set as site_properties (and as I mention in item (1), that isn't done correctly either).

## Example code

```python
p = Poscar.from_file("test_files/CONTCAR.MD", check_for_POTCAR=False)
```

## Error message
The predictor corrector information in the Structure is wrong. For exampe for the above CONTCAR.MD read from test_files, the regenerated Poscar ends with the following predictor corrector chunk, where the predictor-corrector coordinates of last three atoms are missing.
1
2.0
1.291972 0.009838 0.000000 0.000000
0.333878 0.772915 0.367011
0.030832 0.042070 0.287438
..
0.202082 0.545512 0.332244
0.209706 0.018853 0.630065
0.042787 0.718967 0.808148

missing predictor-corrector coordinates:
>> 0.98321567E+00  0.32246042E+00  0.83616357E+00
>>  0.97678703E+00  0.70638938E+00  0.12546431E+00
>> 0.15997023E-01  0.29204035E+00  0.15797571E+00

Besides, the following two 3xN size additional sets of predictor-corrector information is missing too. VASP cannot read the new POSCAR due to missing data:
>> lib-4001 : UNRECOVERABLE library error 
>>  A READ operation tried to read past the end-of-file
>> ...

## Suggested solution (if any)
Please see the pull request I submitted for details. In a nut shell: 
[...]

PR #326, opened 2016-04-11
Suggested fix for the parsing issues related with the predictor-corrector data in VASP output

aykol:
## Summary

This is a suggested solution for the Issue #325, mostly for code review purposes. I can work on a new pull request based on the comments on this one.

* Predictor corrector data is now correctly parsed.
* These 3 sets of 3 xN parameters are stored properly for each site as a site_property.
* inputs.Poscar is modified to now generate the the correct string for POSCAR's with predictor_corrector information.
* A generic "structure_properties" attribute is added to Structure to store any arbitrary structure information that is not site-specific.
* Using that, predictor corrector preamble (key, potim, thermostat params...) is stored as a structure_property
* Structure and IStructure were modified to accommodate "structure_properties" (e.g. addition of properties, structure generation, copying etc.)
* Changes are tested, no errors generated.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -949,7 +949,7 @@ class IStructure(SiteCollection, MSONable):
                                               to_unit_cell=True,
                                               properties=site_props))
             new_sites = sorted(new_sites)
-            return self.__class__.from_sites(new_sites)
+            return self.__class__.from_sites(new_sites, structure_properties=struct_props)
 
     def interpolate(self, end_structure, nimages=10,
                     interpolate_lattices=False, pbc=True, autosort_tol=0):
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -341,7 +341,7 @@ class IStructure(SiteCollection, MSONable):
 
     def __init__(self, lattice, species, coords, validate_proximity=False,
                  to_unit_cell=False, coords_are_cartesian=False,
-                 site_properties=None):
+                 site_properties=None, structure_properties=None):
         """
         Create a periodic structure.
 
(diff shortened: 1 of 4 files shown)


shyuep:
Thanks. I think the new parsing is great. But I think there is really no reason to add a structure_property field. Structure is mutable by default. So you can simply set any new properties as Structure.thermostat = X for example. Also, I think things like thermostat etc. are not really Structure related but rather POSCAR related. So they should actually be stored in POSCAR itself rather than Structure?

aykol:
Thanks for the comments. I will isolate the parsing and submit a new pull request. 

For the preamble storage (key, time-step and thermostat parameters), I agree that the proper place would be Poscar since they are specific to it. But the reason why I suggested storing them in Structure was, because they are tied with the predictor corrector coordinates and derivatives (e.g. those are valid for only the time-step given in the preamble) so I wasn't sure about separating them. Because in that sense, having them together in Structure ensures that when a Structure parsed from a CONTCAR is stored, it is self-sufficient to generate a new Poscar in future, without needing access to the CONTCAR it was generated from (or a stored Poscar object). I hope my point is clear. What do you think? Would you prefer the "Structure.thermostat" style or storing in the Poscar?

shyuep:
Thanks. I agree it is strange that the predictor_corrector are stored in site properties. There is actually a good reason for that. When the properties are related to an overall structure, there is no issue with storing them in the Poscar itself. When there are properties that relates to sites, the problem is that people may sometimes manipulate the structure e.g., swapping atomic ordering etc., after parsing a Poscar and then regenerating that Poscar. Those properties have to be linked to each site so that any ordering works. 

In general, core objects like Site and Structure should be kept free from DFT code specific things. I don't see the site_properties as a major deviation because site_properties was there to begin with. Things like Poscar related parameters should site within the Poscar object itself.

aykol:
Ok, I got it. I'll put the those parameters in Poscar. Thanks.

PR #332, opened 2016-04-13
add a safe guard to monitor whether ion phase appear when requesting a Pourbaix Diagram

xhqu1981:
add a safe guard to monitor whether ion phase appear when requesting a Pourbaix Diagram


xhqu1981:
Lighting fast. I am just talking to Donny. ^_^. Thanks a lot.

PR #333, opened 2016-04-14
Fix important problem with get_primitive_standard affecting rhombohed…

hautierg:
## Summary

* Fix important problem with rhombohedral cells in get_primitive_standard
* Fix a unittest related to get_primitive_standard
* Add a new unites for get_primitive_standard (rhombohedral)

## TODO (if any)

*add more unittests


shyuep:
Thanks.

PR #343, opened 2016-04-18
add parsing of NMR Chemical Shifts and EFG from OUTCAR

xhqu1981:
NMR chemicals shift and electric field gradient tensor are only available in OUTCAR. I have add the ability to parse them from OUTCAR.


shyuep:
Thanks. But the chemical shift notation class doesn't seem to be right in outputs. I think you need to create a proper NMR analysis class. Parsing can still be in Outcar, but the actual analysis should be in a separate module.

xhqu1981:
OK. I will refactor it. Where do you prefer to store the nmr conversion code? Please be noted that this class is not a real analyzing module, instead, it just provides the conversion between different notations. It is pretty sad that the notation used by VASP is different than the most popular notation in the literatures.

shyuep:
Well, just drop it into a pymatgen.analysis.nmr class. Though today you do not have analysis code yet, I am pretty sure you are going to write some in future.

xhqu1981:
@shyuep Done. Please check. Thanks for the good suggestions.

PR #355, opened 2016-04-29
Bztricci

fraricci:
## Summary
- different convergence method in Runner, only based on searching the first couple of lpfac and energy_grid parameters that gives a fermi level for each doping level.
- DOS mode added but it needs a custom BZT version. An error is raised in case that std version is used. It works correctly with partial eigenvalues and for spin polarized cases (get_complete_dos() in Analyzer)
- BANDS, FERMI modes work properly with std v. of BoltzTrap
- BANDS mode gives a complete BS on sym line (get_symm_bands() in Analyzer). It's also possible to specify any k-point line to have interpolated bands along them.
- accuracy check on sym-line-bands for materials with gap (check_acc_bzt_bands() in Analyzer) 
- energy_span_around_fermi auto selected for DOS and BANDS mode
- fermi surface data read from fort.30 file (std version of Boltztrap) or from boltztrap_BZ.cube (custom version of Boltztrap) and stored as a property of Analyzer.
- experimental plot_fermi_surface() in BoltztrapPlotter. skimage package required (ImportError raised in case it's not installed)
- small fix in the procedure that create the boltztrap folder
- test_files folder reorganized to separate each boltztrap mode. New tests added for BANDS, FERMI, DOS modes.

## TODO
- x_trans script of std Boltztrap overwrite .def file written by boltztrap.py
  Hence,  _make_def_file() and _make_proj_files() functions don't works properly
- extend the accuracy band check to metals
- plot_fermi_surface works well only with loose points grid. A better implementation could be done using mayavi (although it's a huge package)


shyuep:
The tests are failing because of the ASE import. See https://circleci.com/gh/materialsproject/pymatgen/4405 . Can you pls make sure you do not introduce new dependencies unless absolutely necessary, especially dependencies that cannot be installed easily via pip? read_cube can be easily written without using ASE.

PR #356, opened 2016-04-30
Structure analysis methods for liquid and amorphous structures

aykol:
## Summary

Structural analysis methods relevant for liquid and amorphous structures are added.

* Class for Voronoi Analysis: statistical distribution of polyhedra
* Class for Radial distribution function
* Helper function to get ensemble average coordination numbers


shyuep:
RDF is already implemented in pymatgen-diffusion package. Does the Voronoi analysis overlap with the chemenv that David recently added to pymatgen?

aykol:
Re: RDF, do you mean the Van Hove analysis? I agree that Van-Hove reduces to static RDF for t = 0. But given the simplicity of RDFs, I feel like accessing them through a “Van Hove” type dynamic analysis class may be a bit complicated (e.g. initialize a VanHove instance for each MD frame and specie, get G’s for delta_t=0, and then average all G’s at the end). Also this route may not be necessarily obvious to all pymatgen users; and means extra work. In my opinion, considering RDF is a pretty common, preliminary structural analysis for any work involving amorphous or liquid structures, allowing access to them through pymatgen with a simpler route in a dedicated class can come in handy for many users. Up to you.

Re: Voronoi: I checked chemenv. It is pretty comprehensive. But I don’t see any straightforward way of obtaining the same type of Voronoi analysis through that, so I would say overlap is not obvious. I think I named the class too broadly for its purpose; it's actually focused on a certain polyhedron tessellation from glass literature; obtained directly by post-processing qhull output. Maybe refactoring VoronoiAnalysis to a more specific name; e.g. VoronoiAnalysisForGlasses might be helpful in differentiating? 
Thanks.

shyuep:
My postdoc will push the RDF to pymatgen-diffusion in a short while. I'd rather not have two pieces of code oding the same thing.

As for Voronoi, it's fine. I will merge after you remove the RDF.

aykol:
Sounds great.

aykol:
Ready to merge.

shyuep:
Thanks!

PR #364, opened 2016-05-08
Refactor outcar

montoyjh:
## Summary

Uses @xhqu1981's table parser for the elastic and piezo tensors.

* Elastic and piezo tensors read in separate methods using read_table_pattern
* Elastic and piezo tensors are stored in outcar.data, rather than as attributes of outcar

## TODO (if any)

It looks like some parts of the lepsilon parsing methods should be refactored as well using the table_parsing method.  I can go ahead and refactor these as seems prudent from the code, but I don't know much about how these methods are used.


shyuep:
Thanks. I guess the epsilon bits would be useful, though this info is already available in the vasprun.xml.

PR #379, opened 2016-05-16
added an exception for unknown "r" element to be read as "Zr". Added …

albalu:
## Summary

* Added an exception in outputs.py to return "Zr" if "r" element is mistakenly detected. 
* Also, added a test in test_files. Please excuse me if I didn't do the test right; in this case I'd appreciate your help regarding the correct place of this test and whether or not it is necessary.


computron:
@albalu looks like the automated tests fail? Please click details and debug. It might just be that some tests need updating since you added a new vasprun.xml to the folder, which affects tests that are trying to automatically parse all the vasprun.xml

PR #380, opened 2016-05-16
refactor eos.py. move from io.abinit to analysis

matk86:
## Summary

Move the eos.py module from io.abinit to analysis. Its usage is more general.
Content in eos.py untouched.


matk86:
done

gmatteo:
Thanks

PR #433, opened 2016-06-30
Rotate_Tensor Method

dongsenfo:
Pushed correct rotate_tensor method and tests which rotates a tensor from an initial structure's basis to a final structure's basis


montoyjh:
Is this working?  It looks like there are still bugs in both the test and the new method.  Also, should it really be called "rotate_tensor"?  We have a "rotate" method already which takes a rotation matrix and transforms the tensor based on that.  I think the new method should have a more descriptive name.

Note that the IEEE standard structure and the conventional standard structure aren't always the same, so testing this using that conversion data might not work.

dongsenfo:
I can change the name, that seems like a good idea. Could you be more specific on what the errors are? I tested the method and tests locally and they seem to pass.

montoyjh:
rotationprim isn't defined on line 131 of tensors.py.

see https://github.com/dongsenfo/pymatgen/blob/master/pymatgen/analysis/elasticity/tensors.py

Maybe your latest local commit isn't pushed?

dongsenfo:
Ah yes sorry I didn't push my last commit, I'll resubmit it.

shyuep:
Is this still active or can this be closed?

dongsenfo:
It can be closed, thanks.

Best,
Handong Ling

PR #442, opened 2016-07-19
Update to REST Api for the substrate analyzer

shyamd:
* Added in get_substrates to the REST API to pull pre-calculated substrate matches from the materials project database based on substrate orientation, matching area, and elastic energy


shyuep:
You need to provide a proper description. I do not merge unless there is one.

dwinston:
@shyamd `orient` should default to `None` (avoid mutable values like `[]`), and if so, elide the "orient" field in the query string.

Also, let's thicken up the docstring. Include that `n=0` means "no limit" (perhaps this should be the default for an MPRester call?). Mention the sort order (increasing elastic energy if available, MCIA if not).

Finally, FYI I put up a public [endpoint](https://materialsproject.org/rest/v2/materials/all_substrate_ids) requiring no login/api_key to access the list of blessed substrate mp-ids currently used by the builder.

PR #444, opened 2016-07-23
Change the implementation of Structure.interpolate_lattices.

sdacek:
## Summary

Change the implementation of interpolate_lattices. Previously, the
parameters for the lattice (a, b, c, alpha, beta, gamma) were all
linearly interpolated to the parameters of the second structure. 
However, this can lead to various issues, including
that the interpolated volumes are not guarenteed monotonic. This is
resolved via interpolating the lattice matrices directly.


wmdrichards:
@sdacek can you fix this?
This implementation breaks when the input lattice is a rotation. See commit 2f60449

PR #446, opened 2016-07-27
Major changes and improvements in chemenv

davidwaroquiers:
## Summary

* New strategy to identify coordination environments
* More convenient way to store and retrieve neighbors (NeighborsSets in StructureEnvironments and LightStructureEnvironments
* Tests are now much faster
* Added new tests for the new strategy and for the LightStructureEnvironments
* Fixed miscellaneous bugs 

None

## TODO (if any)

* Add the connectivity analysis


shyuep:
The tests seems to be failing?

davidwaroquiers:
Hi Shyue,


I think its a compatibility problem with python 3. I'm running with python
2.7, I'll test with python on my machine and make the necessary changes.


David

shyuep:
Ok. I suggest you look into using pyenv. It's highly useful for having multiple python versions on the same machine. In any case, my group has moved to a "default" Py3k system. So everyone uses Py3k unless it is on a cluster where there is no Py3k. So far, no ill effects.

davidwaroquiers:
Hi,


I am using virtualenv. I think I just fixed the bugs and I committed my
changes.


David

PR #457, opened 2016-09-04
Added remove_duplicates method to Structure.

blondegeek:
The Structure.remove_duplicates method can be used for deleting PeriodicSites that are within a tolerance of each other and have identical species_and_occu property. This is useful for deleting degenerate atoms that are generated when creating a Structure.from_spacegroup where the desired structure is to be in a lower spacegroup setting than it would otherwise be refined to be. One such situation that this could be used for is building the endpoints of an interpolation of a phase transition.


shyuep:
Thanks Tess. Can you implement a few unit tests? Also, the clustering code should use scipy hierarchical clustering using the distance matrix. See the point group analyzer for an example of how it is done.

wmdrichards:
I would suggest that it use the same tolerance and clustering code as Structure.merge_sites()

blondegeek:
Hi Shyue Ping and Will, 
I've added two tests for remove_sites, rewritten remove_sites to use scipy hierarchical clustering, and changed the default tolerance for remove sites to be 0.01 AA in my most recent commit on my master branch.
Thanks!
Tess

shyuep:
@blondegeek @wmdrichards I have merged, but devided the two methods are better as 1. I have retained merge_sites, but added a mode parameter. mode="sum" is the default behavior for merge-sites where the occupancies are summed. mode="delete" is same as remove_duplicates, in which only one site is retained.

blondegeek:
Sounds good! Thanks!

PR #464, opened 2016-09-10
Dev elastic

montoyjh:
## Major update to elasticity analysis package

Major update to the elasticity analysis package, including a significant refactoring of voigt-notation tensors and addition of properties to elastic tensor class.  Thanks to @saurabh02 for the initial import of the thermal properties and to Ian Winter for sharing his method of deformation gradient determination.

* Voigt-notation conversion is now abstracted to the TensorBase class, with _vscale attributes of subclasses allowing for scaling of individual elements of the voigt representation.
* Strain objects will construct a deformation gradient via diagonalization of the strain matrix, and no longer require a deformation matrix for many of the associated methods.
* Various thermal properties estimated from the elastic tensor (conductivity, debye temperature, transverse and longitudinal sound velocities) are now included as methods to the ElasticTensor object, with appropriate unit tests derived from former workflow results.

## TODO (if any)

* Many of the new elastic tensor properties depend on structural parameters (density, etc.).  For right now, these methods take a structure object as an argument, but it may be better to allow users to "attach" structures to the tensors in order to allow for simpler syntax.  I'm not yet sure how best to implement this.


shyuep:
There seems to be failing tests. Pls fix before I merge.

montoyjh:
Should pass the tests now, forgot to update the unit tests when I renamed one of the methods.

PR #469, opened 2016-09-15
Miscellaneous small bug fixes in chemenv + self.structures in NEBAnalysis

davidwaroquiers:
## Summary

* Switch from pyhull.VoronoiTess to scipy.spatial.Voronoi
* Fixed small bugs in chemenv
* Added classmethod to MultiWeightsChemenvStrategy in order to ease it's use
* Added self.structures in NEBAnalysis


shyuep:
Can you also as the same time change the use of Pyhull's Voronoi class to the scipy.spatial Voronoi class in chemenv? I would like to start systematically remove use of pyhull. Note that scipy's implementation is faster.

PR #471, opened 2016-09-16
updates

matk86:
## Summary

* feff io: add api to parse the atomic luster as molecule, some cleanup
* refactor and cleanup analysis.eos.py


pymatgen/analysis/eos.py
```diff
@@ -1,96 +1,88 @@
 # coding: utf-8
```
gmatteo: @matk86  @shyuep 

These changes in eos.py have not been discussed and, most importantly, they break 
the previous API provided by EOS.

This piece of code is used in our automatic flows for the generation of the pseudopotentials 
and I've already spent hours to port the other projects we are developing to pmg4.

Now we have reached pmg4.3.1 and I think we should try to avoid drastic changes in the API, 
especially in parts that are used by other members of the community.

Improvements and bug fixes in eos.py are welcome but this is not the correct approach.
If you don't like the API or you want to refactor the code, you should try to maintain compatibility with
the previous API, deprecate methods or, if really needed, replace the old code with another one.

Could you please revert eos.py to the original version and include your fixes/improvements 
without breaking the API?
Thanks
Matteo

matk86:
Sure I can revert Eos.py. I used the unittests as the reference for refactoring. In future please make sure that the unittests actually test the important API so that these sort of issues won't happen. Also I noticed that the original Eos.py module looks almost exactly the same as the one in ase python package(just the class names are changed). https://wiki.fysik.dtu.dk/ase/_modules/ase/eos.html#EquationOfState
Did you copy the module from there? If so I think that must  be mentioned in the module .. any comments on this bhfl @shyuep

shyuep:
Yes, I agree with @matk86 . While I agree with the general principle that API changes should be minimized, especially for core classes, refactoring should be expected. The benchmarking of what the API actually is should be the unittests that actually test against the expected API. Once that is done properly, the person implementing the new scheme can try to do a proper deprecation.

@matk86 I saw that you reverted the EOS changes. I suggest you put the new EOS back in as maybe eos_2.py. That way, @gmatteo can integrate your changes properly and add unittests.

gmatteo:
Yes, I took the equations from ase and the the deltafactor_polyfit from the deltafactor package.  
We use eos.py to compare ground state calculations done with different codes and different
python frameworks hence it's important to have a similar implementation to reduce numerical noise.  
EOS_fit is a wrapper around the different equations and provides extra tools for post-processing.

matk86:
doesn't copying without mentioning the source violate ase's license agreement?

gmatteo:
There was a comment in `deltafactor_polyfit`

    Taken from deltafactor/eosfit.py

that, actually, you removed in your version.
I will mention in the docstring that the version of eos.py is inspired to a similar implementation provided by ase.

matk86:
ase wasn't mentioned anywhere

PR #480, opened 2016-09-21
Use Element attributes to specify element sets for MPRester.parse_criteria

dwinston:
Please do not merge. This PR, though a working implementation, is for discussion/review. Please comment (esp. @shyamd). 

## Summary

Allow filtering on `pymatgen.core.periodic_table.Element` attributes, using a subset of MongoDB filter syntax, in order to specify element sets as part of `MPRester.query` calls. Enables criteria strings such as

```python
'{"group":1}-O'
'{"group":1}-{"group":2}-*'
'{"X":{"$lt":1,"$gt":0.9}}'
'{"$or":[{"is_alkali":true},{"is_alkaline":true}]}-{"group":{"$in":[13,14,15,16]}}'
# etc.
```

Uses `json.loads`, so use `true`/`false` (not `True`/`False`) and `null` (not `None`).

## TODO

* Only a limited subset of MongoDB operators (`$in`, `$or`, `$lt`, `$gt`) currently supported.
* Determine if this is in fact a valuable addition, and if so, weigh this design/approach against alternatives.


dwinston:
@shyuep recent PRs are failing checks due to a missing MAPI_KEY env var. Perhaps this is related to the introduction of .pmgrc.yaml?

shyamd:
Looks good Donny. 

Can we add not in:
```python
if '$nin' in val:
    elt_list = [e for e in elt_list
                if getattr(e, field) not in val['$nin']]
```

shyuep:
Not failing on master branch....

PR #485, opened 2016-09-23
Chemenv improvent + NEBAnalysis improvements

davidwaroquiers:
## Summary

* Removed old and commented code from chemenv.
* Made NEBAnalysis MSONable.
* Made NEBAnalysis's __init__ method independent from vasp (NEB structures and energies could come from any other code).
* Created from_outcars method for NEBAnalysis (to replace old __init__).
* Created test for NEBAnalysis.


shyuep:
Thanks. But I sugegst a few changes:
i. Interpolation order doesn't seem to work in scipy > 0.18? Since it defaults to cubic?
ii. from_outcars should use **kwargs to deal with parameters. That way, it will passthough the parameters to init without having to code specific values for every one that is added in future.
iii. as_dict and from_dict does not seem to be needed. The default implementations should suffice. In future, do not implement as_dict and from_dict unless the default is not working as expected.

pymatgen/analysis/transition_state.py
```diff
@@ -34,12 +37,47 @@
...
 
 
-class NEBAnalysis(object):
+class NEBAnalysis(MSONable):
     """
     An NEBAnalysis class.
     """
 
-    def __init__(self, outcars, structures, interpolation_order=3):
+    def __init__(self, r, energies, forces, structures, interpolation_order=3):
```
shyuep: Interpolation order seems to be useless in scipy 0.18? If so, please remove.

pymatgen/analysis/transition_state.py
```diff
@@ -239,4 +264,35 @@ def from_dir(cls, root_dir, relaxation_dirs=None):
             else:
                 outcars.append(Outcar(outcar[0]))
                 structures.append(Poscar.from_file(contcar[0]).structure)
-        return NEBAnalysis(outcars, structures)
+        return NEBAnalysis.from_outcars(outcars, structures)
+
+    def as_dict(self):
```
shyuep: as and from_dict does not seem to be required since default implementation in MSONable seems to work just fine?

pymatgen/analysis/transition_state.py
```diff
@@ -34,12 +37,47 @@
...
+        if scipy_old_piecewisepolynomial:
+            self.spline = PiecewisePolynomial(
+                self.r, np.array([self.energies, -self.forces]).T,
+                orders=interpolation_order)
+        else:
+            # New scipy implementation for scipy > 0.18.0
+            self.spline = CubicSpline(x=self.r, y=self.energies, bc_type=((1, 0.0), (1, 0.0)))
+
+    @classmethod
+    def from_outcars(cls, outcars, structures, interpolation_order=3):
```
shyuep: Pls use **kwargs to passthrough other arguments to __init__. This way you do not need to add arguments every single time a new parameter is added to __init__.

davidwaroquiers:
Hi Shyue,


I made the changes that you suggested and reopened a pull request but the
circleCI failed for some tests. The thing is I checked which tests failed
and I never touched these files or anything related to that. The tests that
are failing are related to the MPRester so maybe the MaterialsProject was
not accessible at the time of the circleCI's build/tests and I can't find a
way to ask for a "rebuild".


David

PR #490, opened 2016-09-28
DO NOT MERGE: Request for review of Eqn of State Thermal

ctoher:
Pull request for review of equation of state thermal properties code for Materials Project. Updated to re-clone pymatgen source and update fork after changes to pymatgen repo.

FAO: Patrick Huck, Donny Winston, Joey Montoya, Kristin Persson

Code is in pymatgen/analysis/eqn_of_state_thermal

### DIFF
--- a/pymatgen/analysis/eqn_of_state_thermal/__init__.py
+++ b/pymatgen/analysis/eqn_of_state_thermal/__init__.py
@@ -1,0 +1,4 @@
+"""
+The gibbs package uses (E, V) data to obtain an equation of state for a material,
+which is then used to calculate thermal properties
+"""
--- a/pymatgen/analysis/eqn_of_state_thermal/eos_ev_check.py
+++ b/pymatgen/analysis/eqn_of_state_thermal/eos_ev_check.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python
+
+"""
+This module sorts and checks the (E, V) data for use in the MP-Eqn of State Thermal properties module.
+"""
+
+from __future__ import division
+import warnings
+import sys
+import subprocess
+import unittest
+import pymatgen
+import numpy as np
+import os
+from numpy import matrix
+from numpy import linalg
+import math
+
+__author__ = "Cormac Toher"
+__copyright__ = "Copyright 2014, The Materials Project"
+__version__ = "1.0"
+__maintainer__ = "Cormac Toher"
+__email__ = "cormac.toher@duke.edu"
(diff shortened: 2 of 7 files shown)


bocklund:
I have just briefly skimmed this and I have a few questions.

1) Why did you implement a class for an isotropic strain on your structure? Your class does one thing, which is running the `apply_strain()` method and you that transformation in one place. Why didn't you just use the `apply_strain()` method in the first place?

2) Your `ModifiedVolumeStructureSet` class only has `init()` as a method. That is a strong sign that you should just be writing a function that returns a list of modified structures. I can see how you might want to use the `strainfactors` property later, but you don't use it now and you could easily calculate it by comparing the volume of the strained structure to the volume of the relaxed structure. 

I would also suggest that whether or not you make this a function that you change the step size parameter to the maximum strain, for example 10% strain. The reason for this is that no matter if a user wants to create 5 structures or (for some reason) 100, they will always get a realistic volume deformations. Here your code would apply a 50% strain if a user passed in `nstructs=100` and did not change the step size. I would suggest using NumPy's linspace to go from `-strain` to `strain` with `nstructs`, where `strain` is again something reasonable like 10%. Also, 28 structures seems like a strange default, but that is a nit-pick. 

As I was writing this, I went and looked at your MPWorks commits. I would highly suggest taking a look at the [MatMethods project on GitHub](https://github.com/hackingmaterials/matmethods), which has preset workflows already built for thermal properties calculations with both the Debye model and using phonons with the phonopy project. It uses FireWorks, pymatgen, and custodian, and if you are running MPWorks, MatMethods should be as close to a drop-in replacement that is out there.

montoyjh:
Hi Brandon,

Just FYI on your final point, MatMethods is the future of MP production workflows, but MPWorks is likely going to be used for a little while longer, mostly because of changes in DB infrastructure that we're in the process of implementing in order to get MatMethods production ready.

bocklund:
@montoyjh, you're right, I should have mentioned that. I wasn't sure if @ctoher knew about MatMethods and I wanted to help him and others who come across this to know that there are other solutions out there so they don't spend too much time re-implement code that already exists. As you know, MatMethods lets you choose from several EOS, which is not implemented here. That being said, those analysis functions might be more in the scope of pymatgen than MatMethods.

matk86:
PR #592 implements the numerical eos.Thermal properties can be evaluated  using atomate package. So this PR can be closed.

tschaume:
Agreed. Thanks!

PR #493, opened 2016-10-04
Added a MultiStructuresVis to visualize series of structures (used to visualize from e.g. NEB or MD simulations)

davidwaroquiers:
## Summary

Added a MultiStructuresVis to visualize series of structures (used to visualize from e.g. NEB or MD simulations)

* Visualization of multiple structures with vtk (using "n" and "p" keys to go from one structure to another)
* Possibility to have an "animated" movie of the structures (using the "m" key).
* Possibility to tag (e.g. in red) specific sites in the structures in order to "follow" them in the MultiStructuresVis
* Small refactoring of the __mul__ method in Structure (sites were automatically falled back to the unit cell while I needed them not to be falled back for the visualization ... the __mul__ method behavior is not changed)


shyuep:
I do not understand the mymul method for structure. How is that different from structure * [2, 2, 2]?

davidwaroquiers:
Hi Shyue,


It just allows to disable the fallback to the unit cell. I wasnt sure
whether I could add an optional parameter to the __mul__method but I
thought not as it is a "built-in" method.


David

shyuep:
If that is the preference, simply remove the to_unit_cell in __mul__. I am not a fan of having methods that do exactly the same thing with one different sub behavior. Structure already has way too methods as it is.

davidwaroquiers:
Hi Shyue,


I remove the to_unit_cell in __mul__ and removed mymul.


David

PR #494, opened 2016-10-04
Added rotate_sites method for Molecule and option to offset center for get_boxed_structure

blondegeek:
These methods are useful for defining a specific orientation (rotation and translation) of a Molecule in a boxed structure.


shyuep:
Thanks. I cannot merge this request as you are adding back 5 files. I think the only thing that was changed in structure.py? Pls make a copy, refork, and clone, and move structure,py over. Also, for all new functionality, pls implement unittests.

PR #498, opened 2016-10-08
Developed new functionalities for Xdatcar class in pymatgen.io.vasp module

rambalachandran:
## Summary

<Incorporated options to select structures, concatenate new structures with existing one and write the Xdatcar class into a file>

* Added  functionality to provide the starting and ending ionic step, to enable removal of structures in NVT AIMD before equilibration.
* Created a function concatenate new structures from a file into existing Xdatcar. Helps in concatenating multiple Xdatcars  obtained through restart into a single Xdatcar class.
* Created a function to write the structures in Xdatcar class into a file.

The default behavior of Xdatcar was not modified only additional optional parameters were introduced in the `__init__` function. So it must not affect any other testcase or code where Xdatcar class is used
<None>

## TODO (if any)

1. Concatenate function needs validation to ensure that the new file contains the same lattice and atoms as that of the Xdatcar class.
2. The class needs to be made MSONable to be able to export the class as a dict to store in json format.

[XdatcarTest.zip](https://github.com/materialsproject/pymatgen/files/517428/XdatcarTest.zip)

crystal lattice and atoms as that of the Xdatcar class >

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -2776,19 +2776,33 @@ class Xdatcar(object):
     .. attribute:: structures
 
         List of structures parsed from XDATCAR.
+    .. attribute:: comment
+
+        Optional comment string.
+    Authors: Ram Balachandran
     """
 
-    def __init__(self, filename):
+    def __init__(self, filename, ionicstep_start=1,
+                 ionicstep_end=None, comment=None):
         """
         Init a Xdatcar.
 
         Args:
-            filename (str): Filename of XDATCAR file.
+            filename (str): Filename of input XDATCAR file.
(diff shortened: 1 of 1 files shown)


rambalachandran:
I'm not sure as to why the checks have failed. I modified only one class that doesnt seem to be imported elsewhere (except for two test cases  (i) pymatgen/io/vasp/tests/test_outputs.py and (ii) pymatgen/analysis/tests/test_structure_analyzer.py). If they are failing for these test cases, do I need to update them as well?

shyuep:
Thanks. Don't worry about the API key. That is a matter for the maintainer, i.e., me.

rambalachandran:
Thanks Shyue. That helps.

PR #516, opened 2016-11-22
Optionnal possibility to specify that the saddle_point in the NEB sho…

davidwaroquiers:
## Summary

* Possibility to setup the spline in the NEB analysis so that the slope is zero for the saddle point (works only for new versions of scipy).

### DIFF
--- a/pymatgen/analysis/transition_state.py
+++ b/pymatgen/analysis/transition_state.py
@@ -63,13 +63,33 @@ class NEBAnalysis(MSONable):
         # cubic by default) is constrained by the boundary conditions of the
         # energies and the tangent force, i.e., the derivative of
         # the energy at each pair of points.
+        if 'spline_options' in kwargs:
+            spline_options = kwargs['spline_options']
+        else:
+            spline_options = None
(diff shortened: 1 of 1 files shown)


shyuep:
I will merge but I will make some edits.

shyuep:
I pushed my changes. Can I request that we avoid introducing too much craft and code that duplicates. When you have the exact same call to CubicSpline, that is a problem.

Also, when can we remove the old scipy spline? Pymatgen already pins the latest scipy as a requirement. Do we need that?

shyuep:
also, pls implement a unit test for the spline_options code. Right now, I have no way of knowning whether my changes broke anything.

davidwaroquiers:
Hi Shyue,


Ok, I'll implement a unit test tomorrow (already late tonight).

On the other hand, I will re-add the "setup_spline" method if you don't
mind, because we dont store the spline itself in the as_dict(), and we
might want to use a given spline_options in some cases and another one in
some others. In this way, reading back the NEBAnalysis using from_dict()
you can setup the spline option you want.


David

PR #523, opened 2016-12-06
adding parse_intrans method to BoltztrapAnalyzer to obtain scissor value, etc

albalu:
## Summary

a few things are added to BoltztrapAnalyzer:

intrans as an input: it is a dictionary of inputs with which the boltztrap had been run before; it can be obtained via parse_intrans method from boltztrap.intrans file if BoltztrapAnalyzer is called via "from_files" method.

Justification:
when I call BoltztrapAnalyzer on a bunch of files that belong to a boltztrap run, I want to know if the run was on a normal band structure that was untouched or it was on a scissored band structure. In this way, we can also differentiate later in the database between the data from these two types of results.


shyuep:
Can we have a unittest for the new functionality pls?

shyuep:
THanks

PR #532, opened 2016-12-14
add BSDOS plotter

computron:
## Summary

A new band structure plotter that puts band structure and DOS on the same plot, aligned.

* Band structures include spin polarization
* Element projections on band structures supported
* Element and orbital projections on DOS supported
* Code is relatively straightforward to use, understand and modify as needed

None

## TODO (if any)

Things that could be done:
- support element projections for band structures with 4 or more elements (only 3 of those elements would actually be projected, the remaining elements ignored)
- support element+orbital projections for band structures

## Example images

example1

example2

### DIFF
--- a/pymatgen/electronic_structure/plotter.py
+++ b/pymatgen/electronic_structure/plotter.py
@@ -10,10 +10,6 @@ import warnings
 from collections import OrderedDict
 
 import numpy as np
-from matplotlib.collections import LineCollection
-import matplotlib.lines as mlines
-from matplotlib.gridspec import GridSpec
-import matplotlib.pyplot as mplt
 
 from monty.json import jsanitize
 
@@ -1027,6 +1023,9 @@ class BSDOSPlotter():
             a matplotlib plt object on which you can call commands like show() and savefig()
 
         """
+        import matplotlib.lines as mlines
+        from matplotlib.gridspec import GridSpec
+        import matplotlib.pyplot as mplt
 
         # make sure the user-specified band structure projection is valid
         elements = [e.symbol for e in dos.structure.composition.elements]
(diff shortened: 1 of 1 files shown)


computron:
I changed the fontsizes slightly and made them tunable, now it (by default) looks like this:

example2_revised

albalu:
@computron, requests:

1. as a user I'd prefer to set emin and emax directly rather than setting cb_energy_range; sometimes I want to compare plots of 2-4 materials (e.g. doped or substituted) at a fixed emin, emax range. If I am dealing with one plot, I can always calculate the band gap myself and plot again with emin, emax that make the plot look better.

2. It would be nice to have dos_xmax as an argument with a default of None, sometimes, I just want to focus on the band edges. Or for supercells, automatic dos_xmax is too large.

computron:
1. The auto emax is nicer if you just want to auto-plot things without having to manually figure out different parameters for different materials. i.e., to get a nice looking plot without the user needing to input any parameters at all into BSDOSPlotter. For your use case, you can just subtract the band gap from the value you put in cb_energy_range if you really want the same energy axis for multiple plots.

2. Why not just set your energy limits to focus on the band edges in that case instead of having a DOS plot that overruns its x-axis for certain energies? the latter will be ugly IMO. I would just set vb_energy_range to 1 eV and cb_energy_range to 1 eV, for example. Note the DOS x-limit is auto scaled for the energy range you put in, so even if there is a high DOS outside the energy range, it won't affect your scaling for your target energy range and you should have a nice looking plot

computron:
Actually for #1 I realized that you want to retain the same BSDOSPlotter object for all these w/o redoing the settings. OK I will add an option to turn off the gap addition

albalu:
thank you!

shyuep:
Re xlim, I would say that we shouldn't pollute the code with too many options. Just return a pyplot object and the user can then set xlim or ylim or whatever else themselves.

PR #541, opened 2017-01-06
small update of boltztrap.py

fraricci:
## Summary
- bands check works for metals
- update the write_proj function to use properly the Orbital enum and new projections format
- added the k_el option to the get_thermal_conductivity function 
(k_el = True  is the default output k_0 -PF*T, as before. If False k_0 is returned)
- fixed a bug in the FERMI mode
- included a check of ERROR msg in the boltztrap output inside the convergence loop of BoltztrapRunner
- fix kpt_line option for bands mode. Now it works  for both a list of fractional coords as arrays or a list of Kpoint objects.
- fix case lpfac > max_lpfac in case of convergence=False (fermi,bands,dos cases)

### DIFF
--- a/pymatgen/electronic_structure/boltztrap.py
+++ b/pymatgen/electronic_structure/boltztrap.py
@@ -535,6 +535,8 @@ class BoltztrapRunner(object):
 
         if self.run_type in ("BANDS", "DOS", "FERMI"):
             convergence = False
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks. Can you implement a few unittests to make sure it all works as expected?

shyuep:
Thanks

PR #549, opened 2017-01-21
Adsorption new

montoyjh:
## Summary

This PR introduces an adsorption site finder analysis tool in analysis/adsorption.py.  Using it, one can enumerate symmetrically distinct adsorption sites on a slab and generate structures with molecular adsorbates at those sites.

* AdsorptionSiteFinder class, which has an find_adsorption_sites method for finding symmetrically distinct adsorption sites
* A few new transformation classes (SlabTransformation and AddSitePropertyTransformation), which facilitate workflows for adsorption energies.
* Unit tests for AdsorbateSiteFinder
* A fix to structure.py that was preventing derivative structures from inheriting site properties
* A minor edit to VoronoiCoordFinder that allows for the analysis to work on slab geometries, where the coordination analysis results in vertices at infinity


shyuep:
Tests are faiing because of use of pyhull. Pls try to use scipy.spatial where possible.

PR #550, opened 2017-01-24
update cbm module to make 0eV cbm discoverable

miaoliu:
previously, if the conduction band minimum is equal to Fermi level, the code excludes it and return a neighbor data point.


shyuep:
Pls add a unittest. and resubmit

PR #552, opened 2017-01-24
visualize pymatgen structures in jupyter notebook

aykol:
## Summary

* A handy function  for quick visualization of pymatgen structures in jupyter notebook. 

Optional dependency on an external package called chemview (https://github.com/gabrielelanaro/chemview). Will raise a proper import error upon use if chemview is not installed.

### DIFF
--- a/pymatgen/vis/structure_chemview.py
+++ b/pymatgen/vis/structure_chemview.py
@@ -1,0 +1,75 @@
+from __future__ import absolute_import
+
+import numpy as np
+from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
+from pymatgen.analysis.molecule_structure_comparator import CovalentRadius
+from monty.dev import requires
+
+try:
+    from chemview import MolecularViewer
+    from chemview.utils import get_atom_color
+    chemview_loaded = True
+except ImportError:
+    chemview_loaded = False
+
+@requires(chemview_loaded, "To use quick_view, you need to have "
+                            "chemview installed.")
+def quick_view(structure, bonds=True, conventional=False, transform=None, show_box=True, bond_tol=0.2, stick_radius=0.1):
+    """
+    A function to visualize pymatgen Structure objects in jupyter notebook using chemview package.
+
+    Args:
+        structure: pymatgen Structure
+        bonds: (bool) visualize bonds. Bonds are found by comparing distances
+                        to added covalent radii of pairs. Defaults to True.
+        conventional: (bool) use conventional cell. Defaults to False.
+        transform: (list) can be used to make supercells with pymatgen.Structure.make_supercell method
+        show_box: (bool) unit cell is shown. Defaults to True.
+        bond_tol: (float) used if bonds=True. Sets the extra distance tolerance when finding bonds.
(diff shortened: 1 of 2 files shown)


hainm:
@aykol is there any way to convert `Structure` to a pdb file (or string)?

We ([nglview](https://github.com/arose/nglview)) would like to support it too? cheers.

(or even better, PR to `nglview` is welcome: https://github.com/arose/nglview/blob/master/nglview/adaptor.py)

aykol:
@hainm thank you, it will be nice if nglview can support pymatgen ``Structure``.  It can be written to a string using the ``to`` method in "cif", "poscar", "cssr" or as "json".  

If these are not helpful, I'll be happy to look into it in 1-2 weeks and write an adaptor.

hainm:
@aykol `nglview` can read `cif` from rcsb but for some reasons, it (nglview) can not read from `pymatgen`. 

> If these are not helpful, I'll be happy to look into it in 1-2 weeks and write an adaptor.

yes please.

aykol:
@hainm I looked into some of the options. Writing an adaptor doesn't seem straightforward because other than "cif", there seems to be no common format supported by nglview (pdb, gro, cif, mol2, sdf) and pymatgen.Structure. I can verify that "cif" option does not work for some reason.

One option is to create support for pymatgen.Molecule, for which pymatgen provides "xyz", "gjf" and "g03" output formats natively. In this case "pdb" is supported through the optional OpenBabel dependency. And support for pymatgen.Structure could be further provided via creating an intermediate Molecule object for visualization purposes as in e.g.:

`` _pmg_molecule = pymatgen.Molecule.from_sites(pmg_structure.sites)``
``structure_string = _pmg_molecule.to(fmt='pdb')``

However, I personally don't think this option is efficient because of OpenBabel dependency. Are there any plans on nglview side to support "xyz", "gjf" or "g03"?

hainm:
> I can verify that "cif" option does not work for some reason. 

may be the key is here 'to make cif works'?

shyuep:
I agree. Note that pmg's CIF parsing is reasonably robust. If nglview does not work, the likelihood is that ngl's parsing is not robust.

hainm:
ping @arose

arose:
There are different cif dictonaries, ngl does support the macromolecular dictionary (e.g. atom_site cetegory) and the one for chemical components (e.g. chem_comp_atom category). Can you give me an example cif file outputted by pymatgen? Thanks.

What format would retain the most information from pymatgen? I guess xyz would contain the least. Never hear of gjf and g03. I am happy to look into writing parsers if there is specific benefit.

aykol:
@arose here is a sample cif file generated by pymatgen.
[mp-1143.cif.zip](https://github.com/materialsproject/pymatgen/files/891503/mp-1143.cif.zip)

arose:
thanks @aykol, that is a crystallographic cif file using the core dictionary (http://www.iucr.org/resources/cif/dictionaries/cif_core). Need to think about how I want to support fractional coordinates. Maybe for starters just calculate cartesian ones during parsing.

PR #554, opened 2017-01-25
Phonon band structure and dos objects

gpetretto:
## Summary

This PR introduces phonon related objects for archiving and plotting. The objects have been modeled on those contained in the electronic_structure module. No common class hierarchy has been introduced since the attributes of the electronic objects heavily rely on dictionaries containing Spin and Orbitals, that have no meaning for phonons.
PhononBandStructure accepts eigendisplacements (not eigenvectors) as optional argument.

* PhononBandStructure and PhononDos objects
* PhononBSPlotter PhononDosPlotter for plotting
* Parsers for phonopy output
* Tests

Feedbacks from other users interested in phonons about the objects and in particular about the choice of adding the eigendisplacements to the band structure are welcome. Thanks


shyuep:
Thanks.

PR #567, opened 2017-02-16
Enable NPT-AIMD simulation

HanmeiTang:
## Summary

Make DiffusionAnalyzer class compatible with NPT-AIMD diffusion analysis.

* Added a new attribute *lattices*. For NVT simulation, the attribute is a one-item list; for NPT simulation, the attribute includes all lattice matrix, which is used in *get_drift_corrected_structures()*.
* Use fractional coordinates instead of Cartesian coordinates.
* Minor change for the unit test
* Throw warning when *get_framework_rms_plot()* is called during NPT analysis.

None

## TODO (if any)

None


shyuep:
You need to provide unittests that test the NpT functionality. The current pull request does not have that, only the old tests which only tests the NvT ensemble.

Pls run an NpT simulation with a simple small cell like Li2O or something and add the output file for NpT analysis.

HanmeiTang:
Okay.

HanmeiTang:
NpT output file has been added to test_files.

PR #568, opened 2017-02-16
Method to load the averaged core potentials from OUTCAR

utf:
## Summary

Added a new function to the `Outcar` class to load in the average core potentials of each atom at all ionic steps. A test for this method was also added and is passing.

Some blank spaces at the end of lines were removed by vim python-mode- apologies for cluttering the pull request.


shyuep:
The tests are failing. Pls look at the Travis log. You used itertools.izip, which I htink in Python 3 is no longer available. Just use zip. Also, pls upgade to Py3 soon. Py27 support is now on its way out.

PR #570, opened 2017-02-16
get_average_eff_mass improved and small patch for dos run_type

fraricci:
- get_average_eff_mass upgrade: now it returns values for all the energies,
using doping_level=False like other functions
- small patch for DOS run_type


shyuep:
There are failing tests. Pls check the travis log and fix pls. Thanks.

shyuep:
Thank you.

PR #573, opened 2017-02-17
Class for I/O of exciting input

vorwerkc:
## Summary

New class for the input of structure objects from the exciting input*.xml files, and the output of structures and bandstructure paths to exciting files.
## New Features
* `Structure.from_file` includes exciting files
* separate `excitingInput` object with `write_string` and `write_file` routines


shyuep:
Thanks. However, can you make the following changes:
i. This pull request is huge and affects a lot of files. A lot of the changes seem to be changing the permissions of the files. Pls keep your changes to only the specific files taht are modified.
ii. The class names should be CamelCase. So it should be ExcitingInput, not excitingInput.

shyuep:
Great thanks.

PR #574, opened 2017-02-17
New function to plot fermi surfaces using Mayavi tool 

fraricci:
The function plot_fermi_surface of plotter module has been rewritten.
It now uses Mayavi package (mandatory) to calculate and potting the surface.
The dependency from ASE for reading CUBE files has been replaced with a new internal function.
It allows to plot fermi surfaces from cube file produces by both boltztrap (using BoltztrapAnalyzer.fermi_surface_data) or other softwares (reading the file with the read_cube_file function).
It plots different surfaces in different figures or all in one figure.
It returns both the last figure and the module mayavi in order to allow you to plot different data
in the same figure.
Color and transparency of surfaces are tunable.
Labels of kpoints will be shown in the Brillouin zone if a proper dictionary of kpoints coords is provided.

### DIFF
--- a/pymatgen/electronic_structure/boltztrap.py
+++ b/pymatgen/electronic_structure/boltztrap.py
@@ -589,7 +589,7 @@ class BoltztrapRunner(object):
 
                     for c in p.communicate():
                         print(c)
-                        if "STOP error in factorization" in c:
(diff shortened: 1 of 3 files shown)


shyuep:
Great. Many thanks.

PR #575, opened 2017-02-17
Add rgb_legend style to BSDOSPlotter

computron:
## Summary

This PR adds a new style of legend to the BSDOSPlotter, the rgb_legend, which is by default False. If enabled, it can add a colored RGB triangle (ternaries) or RB bar (binaries) to show how the various colors on projected plots map to element ratios.

Note that the alternate (default) style is a more conventional style of legend.

Some examples of the rgb_legend style are shown below

2_el
3_el

Also, thanks to Jan Pohls for sharing his code from which this was based.

None

## TODO (if any)

* Unfortunately, as can be seen in the above diagrams, the band structure plots can overlap the legend diagrams in some systems. In the limited time I had to develop this, I was not able to develop a fix. e.g., simply adding a matplotlib rectangle Patch object to try to cover up the band structure and create more of a "legend" style didn't work.  The overlap can cause things to look a little funny which is one reason the rgb_legend style is not default for now.

This branch can be safely deleted after merge.

### DIFF
--- a/pymatgen/electronic_structure/plotter.py
+++ b/pymatgen/electronic_structure/plotter.py
@@ -10,6 +10,7 @@ import warnings
 from collections import OrderedDict
 
 import numpy as np
+from matplotlib import patches
 
 from monty.json import jsanitize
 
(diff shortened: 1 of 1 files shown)


computron:
note - element labels are now all black b/c the green "K" label in the ternary example is way too low-contrast to read

Issue #581, opened 2017-02-19
Defining the correct usage of HighSymmKpath


gpetretto:
## Summary

I would like to signal a common misunderstanding that I have seen happen to many different people with the HighSymmKpath object. 
HighSymmKpath is basically a hardcoded list of kpoints based on Setyawan-Curtarolo's paper, that uses the input structure just to get its spacegroup. On the other hand, at the first usage most of the people expect that the list of kpoints would be meaningful even if they are passing a structure which is not the standard primitive. On top of that, the symprec default value in HighSymmKpath (1e-2) is larger than the default in SpacegroupAnalyzer (1e-3), so one could have even generated the standard primitive with the analyzer and get the path for a different spacegroup, if the structure is just slightly distorted. An additional source of possible misunderstandings is the fact that in SpaceGroupAnalyzer.get_primitive_standard_structure the default for international_monoclinic is True, while a False value is needed to reproduce the structure from the paper and be used by the HighSymmKpath object (Also, the international_monoclinic option is not documented in SpaceGroupAnalyzer, making the things even more difficult to understand). The problem is quite nasty as of course the produced band structure won't show any relevant problem and only a close inspection of the result can reveal the error.

## Suggested solution (if any)

My first suggestion would be to stress the intended usage of the HighSymmKpath object in the docstring. However, I personally think that it would be better having a warning or even raising an error if the structure does not correspond to the primitive standard obtained using the given symprec and angle_tolerance.

shyuep:
I have no objection to issuing a warning or even raise an error if the structure does not correspond to primitive standard. Can you implement pls? Thanks.

PR #585, opened 2017-02-21
Warning for non standard structures in HighSymmKpath

gpetretto:
## Summary

This PR addresses the problem highlighted in issue #581 adding a warning if the lattice of structure does not match the one of the primitive standard.

### DIFF
--- a/pymatgen/symmetry/bandstructure.py
+++ b/pymatgen/symmetry/bandstructure.py
@@ -59,7 +59,7 @@ class HighSymmKpath(object):
 
         if not np.allclose(self._structure.lattice.matrix, self._prim.lattice.matrix, atol=atol):
             warnings.warn("The input structure does not match the expected standard primitive! "
-                          "The path can be incorrect. Use at you own risk.")
+                          "The path can be incorrect. Use at your own risk.")
 
         lattice_type = self._sym.get_lattice_type()
         spg_symbol = self._sym.get_space_group_symbol()
--- a/pymatgen/symmetry/bandstructure.py
+++ b/pymatgen/symmetry/bandstructure.py
@@ -11,7 +11,7 @@ Created on March 25, 2013
 """
 
 import numpy as np
-
+import warnings
 from math import ceil
 from math import cos
 from math import sin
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks.

computron:
My understanding is that if the lattice/structure doesn't match the primitive, then the output of HighSymmKPath will be incorrect.

If that's the case I would prefer an error, i.e., bad input should result in an error, not a warning. Under no circumstances should a code knowingly return bad output. A warning might simply be neglected by, say, an automatic workflow running the job.

gmatteo:
Hi all,

> the output of HighSymmKPath will be incorrect

Right but the physical results are still ok, in the sense that they represent a set of KS eigenvalues
defined on a set of lines and k-points that may not be of high-symmetry.
What is incorrect it's the labelling of the k-points and therefore the interpretation of the band dispersion given by the user.

HighSymmKPath is used in several workflows to obtain a k-path in an automatic way
and the fact that HighSymmKPath may be incorrect is not critical for the correctness of the workflow (ab-initio codes usually need wavefunctions and energies computed on a mesh for further calculations)
If we raise an exception at this stage, we should then try to figure out how to handle it at the workflow level
What do we do?

- Abort the worflow completely
- Skip the band structure calculation and continue with the other steps of the workflow
- Use a hard-coded path that is just following that Bravais lattice vectors?

There are also cases in which HighSymmKPath is just used to plot band energies with matplotlib.
with minimal user intervention. 
In phonon calculations for example the ab-initio interatomic constants can be interpolated everywhere in q-space and we use HighSymmKPath to get the list 
of points at runtime if the q-path is not given explicitly.
Also in this case, the results are correct but the interpretation may be wrong
Sill I would prefer to show the phonon band dispersion to the user instead of a python backtrace.

A possible solution could be to change the labels of the k-points, e.g. by adding an "?"
or an extra string such as "_wrong_kpath" so that users and scripts can easily detect 
that the k-point path does not correspond to the one given in Setyawan-Curtarolo's paper.
Then I think we should stress this issue to the users and suggest them to use primitive standard cells whenever possible.

computron:
Well, when I call HighSymmKPath, I expect a high symmetry kpath. Saying that is still represents valid kpoints but that are not high symmetry is like saying that it's ok for the math.log() function to not correctly return the logarithm so long as it returns a valid number that can be put into a spreadsheet. 

I am not sure what is the phonon usage of HighSymmKPath. But either the user is intending to force the code to do something strange (in which case, they should set a flag and neither get a warning nor an error since the code is now clearly doing what the user intended), or the user did not tell the code they are trying to do something strange they should get back an error.

Anyway, that is just my thoughts. The pymatgen BFDL @shyuep gets the final vote on such issues in pymatgen.

computron:
(Btw, note that I really prefer "abort the workflow completely". Then I see there is an error in the run, fix the workflow and rerun. Much better than the workflow keeps going and 6 months later I realize 10% of my runs were incorrect and need to be rerun again).

gmatteo:
>  (in which case, they should set a flag and neither get a warning nor an error since the code is now clearly doing what the user intended),

Adding a flag to control the behaviour is a good idea

gpetretto:
On my side the main reason to prefer the warning over the error was that there should be a number of cases where, since the list of kpoints is given in fractional coordinates, the output will still be correct, even if the lattice does not exactly match the one produced by get_primitive_standard_structure.
A stupid example: for a simple cubic where the basis is not 
[a,0,0]
[0,a,0]
[0,0,a]
but 
[0,0,a]
[a,0,0]
[0,a,0]
the path and labelling would correspond in the two representation, but the check as it is now would raise an error/warning. It would be rather cumbersome to sort out all the cases where the result is meaningful and where it is not. My main concern in the end was to make people aware of what is the correct usage, so I opted for a warning. 

If you prefer to control the execution of the check I was thinking of adding the option to skip the check if the given value of is atol<0.

Issue #508, opened 2016-11-07
Adding magCIF support


mkhorton:
## Summary

magCIF is a relatively new standard (2014) for storing structures of magnetic materials. Adding magCIF functionality to pymatgen would be useful to (a) be able to incorporate existing magCIF structures into a Materials Project workflow and (b) to allow easier visualisation of magnetically ordered structures.

I've created this issue just to check to see if anybody had already started working on this, and also to ask for general feedback on the functionality.

## Suggested solution 

I was going to:

(1) Add read support, to go from the magCIF file to a Structure object + its associated MAGMOM tag, provided that the magCIF file has ` _atom_site_moment_` information included. This should hopefully be straight-forward given that pymatgen already supports CIF.

(2) Add partial write support, to visualise Structure objects with their magnetic moments. This would write a magCIF file without any magnetic symmetry information (i.e. with primitive magnetic symmetry), but including `_atom_site_moment_` information. JMol, VESTA, etc. already support magCIF, so this could be useful for visualisation.

(3) Add full write support. To do this, we'd have to add detection of magnetic symmetry groups to pymatgen, which seems like a much larger problem for a future date... (I know spglib has partial support for collinear magnetic symmetry, but adding support for detection of any arbitrary (non-collinear) magnetic ordering seems like it's going to be a bigger task, though worth doing?).

Any feedback is welcome!

All the best,

Matthew

(p.s. To introduce myself briefly: I'll be joining Prof Persson's group at LBNL in February, and will be working on magnetic materials. As such, if there's any magnetic functionality that would be useful to add to pymatgen, please get in touch! MKHorton@lbl.gov)

shyuep:
All this sounds good. I await your pull request!

mkhorton:
Just to update on this, I've got a `MagCifWriter` and `MagCifParser` in testing, and also a `MagSymmOp` and I'm working on a magnetic version of `SpaceGroup` too (have constructed a JSON file from a magnetic space groups data file which we have permission to use) – this will hopefully enable an equivalent `Structure.from_magnetic_spacegroup()` method. When this is all further along in development I'll submit a PR for code review.

However, there have been a few issues that might need clarifying:

* As I can see, there are currently three representations in use for magnetic moments/spin in pymatgen. There's the list of scalar magmoms, there's the vector (list of lists) magmoms used by `MPSOCSet`, and there's the individual spin attribute on a Specie.

[...]

PR #604, opened 2017-03-04
Proposal for Magmom class

mkhorton:
## Summary

New proposal for `electronic_structure.core.Magmom` class to simplify working between scalar and vector-type magnetic moments for collinear and non-collinear electronic structure calculations.

Internally implements a spin quantization axis, following exactly the specification of `SAXIS` tag in VASP documentation.

Should give unsurprising results:

* if you want to use it as a scalar Magmom, e.g. `m = Magmom(5.0)`, this will work as expected with `float(m)` giving `5.0`

* if you want to use it as a list-type Magmom, e.g. `m = Magmom([1.0, 2.0, 3.0])`, this should also work as expected with `m[0]` giving `1.0`

* if you want to use a custom `SAXIS`, you can either use `m = Magmom([1, 2, 3], saxis=[1,1,1])` or you can use `m = Magmom([1,2,3])` and `m.get_moment(saxis=[1,1,1])`.

There also exist static helper methods such as:

* `Magmom.are_collinear(magmoms)` to check if a list of magmoms are collinear

* `Magmom.get_consistent_set_and_saxis(magmoms)` to return a list of magmoms with the same spin axis, and the spin axis itself to be used with VASP's `SAXIS` tag

For existing structures, can update `magmoms` with `[Magmom(m) for m in magmoms]` if any of the new functionality is required.

Would also simplify use and implementation of `MagCifParser` (#508). In particular, `MagCifParser` will always output a Structure with vector magmoms. However, for those structures which have collinear spin, a collinear approximation can be used with scalar magmoms, and `[float(m) for m in magmoms]` will give the correct values in this instance even in the case of e.g. antiferromagnetic structures. Additionally, `MagCifParser` and `MagCifWriter` need to read/write magmoms in terms of crystal axes, which is made possible by this new class.

PR is submitted for comments / code review. Feedback appreciated!

None

## TODO (if any)

* ~~Further testing.~~
* ~~Implement `get_moments_relative_to_crystal_axes` method used by `MagCifWriter`.~~
* Code review.
* (Future) If/when `Magmom` class is more mature, could change `io.vasp` to use it by default.


shyuep:
Many thanks. Sorry for the lateness in merging. I was thinking if there are better ways to structure the class. For now, this looks great!

mkhorton:
No problem at all, I'm not convinced this is the best way of doing it either, but I think it works for now. Thanks for merging!

shyuep:
I have just pushed a bunch of changes to simplify the implementation a little. I understand the rationale for making it immutable, but for now, let's just let it flow a little.

shyuep:
If you really want to make it immutable, you probably should subclass a namedtuple and add the relevant functions needed.

mkhorton:
Okay, that sounds good, thanks for the clean-up! I'm completely fine with it being mutable, I don't have strong feelings either way in that regard.

(The write-only numpy arrays were probably a bit heavier than necessary anyway, I tried using them to simplify performing vector operations elsewhere, but maybe named tuples would have been better.)

PR #619, opened 2017-03-14
remain site_properties if exist when replace Element

specter119:
## Summary

remain site_properties if exist when replace Element

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -2347,6 +2347,7 @@ class Structure(IStructure, collections.MutableSequence):
             frac_coords = self._lattice.get_fractional_coords(coords)
         else:
             frac_coords = coords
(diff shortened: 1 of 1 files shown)


shyuep:
This does not make sense. If you replace an element, the likelihood is that the site properties will change.

shyuep:
Also, the tests failed.

specter119:
Yes, I see it, sorry.

2017年3月15日 00:11，"Shyue Ping Ong" <notifications@github.com>写道：

This does not make sense. If you replace an element, the likelihood is that
the site properties will change.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/materialsproject/pymatgen/pull/619#issuecomment-286472130>,
or mute the thread

PR #620, opened 2017-03-15
Toec new

montoyjh:
## Minor edits to elastic tensor

Fixes a few issues in elasticity related to using a tolerance for independent deformation finding and summary output for derived properties.

* Elastic tensors can return a dictionary of derived properties and structure-derived properties
* Independent strain-deformation finding uses a tolerance criteria, rather than a strict non-zero criteria
* Updated unit tests


montoyjh:
I'm going to close this for now, want to do a bit more refactoring of the Strain functionality.

PR #623, opened 2017-03-20
Removed mutable default parameters.

nisse3000:
Removed mutable default parameters in class OrderParameters and class StructureMotifInterstitial.

### DIFF
--- a/pymatgen/analysis/defects/point_defects.py
+++ b/pymatgen/analysis/defects/point_defects.py
@@ -1530,8 +1530,9 @@ class StructureMotifInterstitial(Defect):
     __supported_types = ("tet", "oct", "bcc", "tetoct")
 
     def __init__(self, structure, inter_elem,
-                 motif_types=["tet", "oct", "tetoct"],
-                 op_targets=[1.0, 1.0, 1.0], op_threshs=[0.5, 0.5, 0.5],
+                 motif_types=None,
+                 op_targets=None,
+                 op_threshs=None,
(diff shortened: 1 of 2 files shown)


pymatgen/analysis/defects/point_defects.py
```diff
@@ -1530,8 +1530,9 @@ class StructureMotifInterstitial(Defect):
     __supported_types = ("tet", "oct", "bcc", "tetoct")
 
     def __init__(self, structure, inter_elem,
-                 motif_types=["tet", "oct", "tetoct"],
-                 op_targets=[1.0, 1.0, 1.0], op_threshs=[0.5, 0.5, 0.5],
+                 motif_types=None,
```
shyuep: There is no need to replace these with None. Won't a tuple be sufficient to make it non-mutable and be clearer?

nisse3000: I got the impression that tuples are not as widely used as using None (for default lists).  They make sense in this context, yes.  But I mainly found online references that would suggest to use None and then set up the list in the __init__.  But if you approve usage of tuples, I'll stick to using tuples in these situations.

shyuep: Yes, use tuples for the ones which are specific. That way, people do not need to look into the code to know the default values. I do not know what the online fora say. They are wrong. The Hulk is right.

nisse3000: I forgot to put the default parameters into the documentation.  Does that work when I highlight the default list in the documentation?  Or, do you still prefer usage of tuples.

PR #626, opened 2017-03-24
Contributions from UCLouvain group

gmatteo:
## Summary

Contributions from UCLouvain group. 

* Add support for Abinit 8.2.2 
* Improved support for DFPT calculations 
* Add support for `angdeg` in Structure.from_dict
* Fix minor py3k portability issues


shyuep:
Many thanks.

pymatgen/electronic_structure/plotter.py
```diff
@@ -2925,23 +2925,24 @@ def plot_points(points, lattice=None, coords_are_cartesian=False, fold=False, ax
 
 
 @add_fig_kwargs
-def plot_brillouin_zone_from_kpath(kpath, **kwargs):
+def plot_brillouin_zone_from_kpath(kpath, ax=None, **kwargs):
```
bocklund: Shouldn't you just be able to use `ax=ax` here and have it be propagated through the `**kwargs`?

gmatteo: The other functions in `plotter.py` already accepts an optional `ax=None` argument 
while `**kwargs` are passed en masse to the matplotlib API in order to customize the plot e.g.

    ax.plot_wireframe(x, y, z,  **kwargs)

This approach is inspired to the API used in pandas and seaborn 
(http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.plot.html)

If we start to use ``**kwargs`` to propagate keyword arguments that are not directly 
related to the `plot` method then we have to remove these keys before calling e.g. `ax.plot(..., **kwargs)` and it's easy to make a mistake.
I think the `ax=None` approach is more readable, easier to document and consistent with the other methods.

bocklund: 👍

PR #644, opened 2017-04-14
Make it easier for user to submit custom LDAUU

mkhorton:
## Summary

Currently, it's non-obvious how a user would add a custom `LDAUU` value to their `user_incar_settings` that would be used in the general case (regardless of what anions are in the input structure).

From my reading of `sets.py` it seems like if a user wants to set (e.g.) a custom `LDAUU` value for Fe, they would need to see if their structure is (e.g.) an oxide and then supply `user_incar_settings={'LDAUU':{'O':{'Fe':5}}}`. This seems non-obvious, and also would mean that `user_incar_settings` needs to be customized for every structure. The only other way I can see to do this is to add a `LDAUU` site_property but, again, this would require modification of every input structure.

This PR would add a fallback option, so that e.g. `user_incar_settings={'LDAUU':{'Fe':5}}` would work as expected in the general case.

None

## TODO (if any)

I'm not sure the best way to fix this, PR has my suggestion which should work.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -201,6 +201,12 @@ class DictSet(VaspInputSet):
             scales with # of atoms, the latter does not. If both are
             present, EDIFF is preferred. To force such settings, just supply
             user_incar_settings={"EDIFF": 1e-5, "LDAU": False} for example.
+            The keys 'LDAUU', 'LDAUJ', 'LDAUL' are special cases since
+            pymatgen defines different values depending on what anions are
+            present in the structure, so these keys can be defined in one
+            of two ways, e.g. either {"LDAUU":{"O":{"Fe":5}}} to set LDAUU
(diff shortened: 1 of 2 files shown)


pymatgen/io/vasp/sets.py
```diff
@@ -276,11 +276,14 @@ def incar(self):
...
                     elif most_electroneg in v.keys():
-                        incar[k] = [v[most_electroneg].get(sym, 0)
-                                    for sym in poscar.site_symbols]
+                        if isinstance(v[most_electroneg], dict):
+                           incar[k] = [v[most_electroneg].get(sym, 0)
+                                       for sym in poscar.site_symbols]
+                    # else, use fallback LDAU value if it exists
                     else:
-                        incar[k] = [0] * len(poscar.site_symbols)
+                        incar[k] = [v[sym] if (sym in v.keys()) else 0 for sym in poscar.site_symbols]
```
mkhorton: Thanks, have updated the PR.

PR #645, opened 2017-04-15
pymatgen.analysis.ferroelecticity module

blondegeek:
This pull request contains the Polarization and EnergyTrend classes useful for evaluating the spontaneous polarization and energy trends of a candidate ferroelectric using polarization calculations for several structures sampled from along a nonpolar-polar distortion.


pymatgen/analysis/ferroelectricity/ZVAL.yaml
```diff
@@ -0,0 +1,305 @@
+VASP_PBE_ZVAL:
```
shyuep: Is there a need for this file. This should be available from the POTCAR itself? In general, I am not a fan of hard coding these because you will need to maintain these with changes in VASP PSP.

pymatgen/core/structure.py
```diff
@@ -764,6 +764,28 @@ def get_neighbors(self, site, r, include_index=False):
                                       include_index=include_index)
         return [d for d in nn if site != d[0]]
 
+    def get_nearest_site(self, coords, site, r = None):
```
shyuep: If this is only used by ferroelectric analysis for now, I suggest you do not add this as a core function. I am pondering how many use cases there are for this function.

blondegeek: I've also found this function useful for getting the minimum distance between sites of two similar structures (with identically ordered sites) that do not have identical lattices (example script below). Should I instead create a PolarizationLattice object that inherits from Structure and add this method?

```python
def get_min_distances(struct1, struct2):
    assert len(struct1) == len(struct2)
    L = len(struct1)
    dists = []
    for i in range(L):
        coord1 = struct1[i].coords
        coord2 = struct2.get_nearest_site(coord1, struct2[i])[0].coords
        dists.append(np.linalg.norm(coord1-coord2))
    return dists
```

pymatgen/analysis/ferroelectricity/ZVAL.yaml
blondegeek: I've been using this yaml with databases created by atomate workflows that by default do not store the ZVAL for POTCARs but just the pseudopotential type and name. However, I recognize that this is a specific use case issue (and I could modify the workflow to store this information). 

I think this yaml is helpful but not necessary. It does demonstrate what information is needed from pseudopotential files to correctly calculate the ionic polarization. These dictionaries are mildly inconvenient to make if one does not readily have the POTCAR used on hand (or a PSP folder on their local machine) but only the type and name of the pseudopotential used.

Perhaps it would be better to have Outcar extract the ZVALs if lcalcpol = True (modifying __init__ and as_dict). That would solve multiple use cases. Should I go ahead and add the Outcar property if lcalcpol = True and delete the yaml and associated lines in polarization.py?

shyuep: Yes, please modify Outcar to get the ZVAL. VASP will always issue new psp and I don't think we should be trying to trying to maintain a moving target.

pymatgen/core/structure.py
shyuep: Ok. I think if you use coord_utils pbc_shortest_vector does the same thing and is probably faster?

blondegeek: They are not the same. pbc_shortest_vector assumes the same lattice. get_min_distances uses the individual lattices to find the closest image, making comparisons with cartesian coordinates. I went ahead and added PolarizationLattice(Structure) to pymatgen.analysis.ferroelectricity.polarization and removed get_nearest_site from Structure.

pymatgen/analysis/ferroelectricity/ZVAL.yaml
blondegeek: I've removed ZVAL and associated functions and added the additions to Outcar mentioned above. Now, if LCALCPOL = True, Outcar will parse the property zval_dict. I've added a test for this property.

shyuep: Thanks.

Issue #648, opened 2017-04-19
Should efermi from Vasprun be rounded to the 4th decimal place?


blondegeek:
In vasprun.xml, the Fermi level is reported to 8 decimal places, while the band energies are only reported to 4 decimal places (and thus rounded to 4 decimal places).

Excerpt from vasprun.xml:
```
...
   <i name="efermi">      2.71196589 </i>
   <total>
    <array>
     <dimension dim="1">gridpoints</dimension>
     <dimension dim="2">spin</dimension>
     <field>energy</field>
     <field>total</field>
     <field>integrated</field>
     <set> 
      <set comment="spin 1">
       <r>   -31.9226     0.0000     0.0000 </r>
       <r>   -31.7774     0.0000     0.0000 </r>
       <r>   -31.6322     0.0000     0.0000 </r>
...
```
However, this can cause errors in determining whether a material is metallic if the valence band maximum is close to this value. I'm attaching the vasprun.xml and EIGENVAL files for such an example (BaTiO3 in space group 160 which is known to have a PBE band gap of ~2.5 eV).

```python
from pymatgen.io.vasp.outputs import Vasprun
import numpy as np

vrun = Vasprun('bto_vasprun.xml')
bs = vrun.get_band_structure()
print bs.efermi
print bs.is_metal()
bs.efermi = np.round(bs.efermi,4)
print bs.efermi
print bs.is_metal()

>> 2.71196589
>> True
>> 2.712
>> False
```

Should the efermi property be rounded to the same number of decimal places by default to avoid this issue?

[bto_example.zip](https://github.com/materialsproject/pymatgen/files/935823/bto_example.zip)

shyuep:
We should not be changing the precision of the data. What we should do is to edit the is_metal method to account for floating point numerical accuracy and allow for up to a 4th decimal place difference.

PR #649, opened 2017-04-19
added 1e-4 tolerance to is_metal method of BandStructure to account f…

blondegeek:
See [https://github.com/materialsproject/pymatgen/issues/648#issuecomment-295325994](https://github.com/materialsproject/pymatgen/issues/648#issuecomment-295325994).

### DIFF
--- a/pymatgen/electronic_structure/bandstructure.py
+++ b/pymatgen/electronic_structure/bandstructure.py
@@ -298,8 +298,8 @@ class BandStructure(object):
         """
         for spin, values in self.bands.items():
             for i in range(self.nb_bands):
-                if np.any(values[i, :] < self.efermi) and \
-                        np.any(values[i, :] > self.efermi):
+                if np.any(values[i, :] - self.efermi < 1e-4) and \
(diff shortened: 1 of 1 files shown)


shyuep:
Shuldn't the second threshold be -1e-4, rather than 1e-4?

blondegeek:
No, the example on the #648 would still be categorized as metallic in that case.

shyuep:
Let's think about the intent of that method. Basically, the code is asking if there are eigenvalues that are both above and below the efermi level in a band. If there is, then it is a metal. Otherwise, it is not. So why would allowing for a +ve numerical error in the first instance and a -ve numerical error second instance lead to a wrong classification?

More importantly, why is this code not simply ```not np.all(values[i, :] - self.efermi < 1e-4)```?

blondegeek:
I think you are right that one tolerance should be negative, but it should be the first one:
`np.any(values[i, :] - self.efermi < -1e-4) and np.any(values[i, :] - self.efermi > 1e-4)`

This ensures that only energies outside the tolerance on the efermi are used to determine if the materials is a metal.

Issue #519, opened 2016-11-25
Missing dependency pybtex


specter119:
<When reporting bugs/issues, please supply the following information. If
this is a feature request, please simply state the requested feature.>

## System

* Pymatgen version: <pip>
* Python version: <2.7.12.2>
* OS version: <Linux.> 

## Summary

should add pybtex to requirement

shyuep:
This is deliberate. Pybtex is only used in one place. There is no need to add it as a required dependency. Also, it uses a primiive 2to3 method of maintaining Py27 and Py3k compatibility, which makes it difficult of pymatgen to create condo packages using it.

PR #652, opened 2017-04-24
Warn MPRester user about need to install matproj.snl extras

dwinston:
MPRester users are having issues decoding SNL objects due to `pybtex` no longer being installed by default by pymatgen. This issue was raised before as #519. I would love to have `pybtex` required, but as an alternative, this PR adds a user warning to install extras.

* Emits `UserWarning` if cannot import `pybtex`
* Gives one-liner `pip install` command to ensure matproj.snl dependencies.


specter119:
@dwinston I currently see this warning when I use python 3.6, but I think `pytex` is dependency only for python 2.x.

PR #654, opened 2017-04-26
Support Multiple Frame XYZ; Minor Changes to QChem

xhqu1981:
* Added class MXYZ to support multiple frame xyz (e.g. MD trajectory)
* Add support for "velocity" section to QChem module to AIMD setup.
* Add class QcNucVeloc to parse QChem AIMD velocity output.


shyuep:
I think MXYZ and XYZ classes should be merged.

xhqu1981:
I was thinking about it. But, are you sure? It will change the API. e.g. molecules (list) vs molecule. Of course, we can do ad-hoc stuff to make it file dependent, then we will force the user to test the type of molecule.

xhqu1981:
I mean the molecule property can be either a list of molecule or just a molecule, but the user will need to test wether it is a list.

xhqu1981:
An alternative would be to have both molecule and molecules properties. molecule returns the last frame in case multiple frame XYZ. Please let me know what is your preference.

xhqu1981:
@shyuep Merged. Please check whether you like it.

PR #669, opened 2017-05-15
modify KPOINTS file comment to more clearly indicate runs w/new kpoin…

computron:


computron:
(just look at the diff)

PR #670, opened 2017-05-17
MagneticSpaceGroup class and Structure.from_magnetic_spacegroup method

mkhorton:
## Summary

* A magnetic version of `SpaceGroup`, with information on all 1651 magnetic space groups. Magnetic space groups are commonly supplied in one of two settings ('BNS' and 'OG'); this class uses BNS as the canonical setting but can also be initialized with the OG label/number. More information on this can be found in the docstrings.

* Subclasses `SymmetryGroup` and, as much as possible, has methods equivalent to those in `SpaceGroup` have been used, and implementations re-used where appropriate.

* Incorporates data from [ISO-MAG, ISOTROPY Software Suite](http://stokes.byu.edu/iso/isotropy.php) with kind permission of Prof. Branton Campbell, BYU. The way the space group data is stored and loaded in the class is based on this data file.

* Example, to create antiferromagnetic MnF2:

```
Structure.from_magnetic_spacegroup("P4_2'/mnm'", Lattice.tetragonal(4.87, 3.30),
                                   ["Mn", "F"], [[0, 0, 0], [0.30, 0.30, 0.0]],
                                   {'magmom': [4, 0]})
```

It's in the Python standard library, but this does use `sqlite3`. A JSON data file was also tried, but loading times were really slow (19 MB), and look-ups were awkward (since the class needs to look up by one of BNS label, BNS number, OG label and OG number). Therefore a small sqlite database was used instead.

## TODO (if any)

* An additional class allowing transformations between arbitrary settings is being developed. This is slightly different from origin shifts in `groups.py`, but follows a standard specification provided by IUCr used in various CIF formats (which specifies transformations in Jones faithful notation). However, bugs are still being worked out with this.

* Not implemented are `is_supergroup()`, `is_subgroup()`

* Data file includes Wyckoff sites. These are currently parsed and printed, but nothing else is done with them.


mkhorton:
(Looks like CI failures are a result of https://github.com/materialsproject/pymatgen/commit/66a78e5d00a394d7c2a9b1365c533d4176b0e27c, since CI doesn't have a MAPI key defined? All tests in this PR should pass okay otherwise.)

shyuep:
Many thanks. I will merge for now. I am not sure why JSON would be slower than sqlite, though I agree the complexity of the query may necessitate the use of sqlite.

shyuep:
Hi, there are a number of test failures. See https://circleci.com/gh/materialsproject/pymatgen/6335 . These have nothing to do with the MAPI key but are related to the new mag symmetry changes.

Can you fix them and submit a PR?

mkhorton:
Sure, thanks, I see them. Looks like it may be a Python 2.x issue. Will submit a PR asap.

shyuep:
There is still one failure. https://circleci.com/gh/materialsproject/pymatgen/6336?utm_campaign=build-failed&utm_medium=email&utm_source=notification 

Think it is a Py2.7 issue.

mkhorton:
Hmm, weird. It works in the console but not in the nosetests.

For some reason `Structure.from_magnetic_spacegroup(id, ...)` fails when `id` is a str/label but not when it's the equivalent number, but only in the tests (e.g. `Structure.from_magnetic_spacegroup("Pn'ma'", ...` fails in the nosetests, but works in the console). Will look into this.

mkhorton:
Sorry, this should hopefully be fixed once and for all now in #673. Have used `six` to fix. Will make to test Py2.7 robustly before future PRs to make sure this doesn't happen again!

PR #677, opened 2017-05-24
Add magCIF support to CifParser and CifWriter

### LINKED ISSUE #508: Adding magCIF support

mkhorton:
## Summary

* Adds read support and partial write support for the magCIF standard (preliminary v1 dictionary available [here](http://comcifs.github.io/magCIF.dic.html)).

* This was developed as a separate class based on `CifParser`, but given that `CifParser` itself has seen recent changes, it seemed sensible to combine them now before they diverge too much.

* CifParser changes:
	* Includes a new CIF `feature_flags` dictionary (extensible for other non-core CIF features), checks to see if a file is magCIF
	* If magCIF, will try to place atoms:
		* First by magnetic symmetry operations, if provided (most common case)
		* If not, then by BNS symmetry group, if provided
	* Will output a Structure with magnetic moments on site properties
		* Note that this means disordered magnetic structures not currently supported
	*  Includes refactor of Pauling/Springer correction code, putting it into a `_sanitize_data()` function – *all current tests pass, but I'm not familiar with the Pauling files, so I'm cautious about this.* The refactor was done to introduce a logical separation between correction code and parsing code, so that parsing code can assume the data conforms to specification, and will make it easier to make changes to the parser in the future
	* Adds simple `get_bibtex_strings()` function to extract reference from CIF file, I found it useful for constructing SNLs

* CifWriter changes:
	* Adds `write_magmoms` kwarg
	* If moments are written, will not perform symmetry analysis
	* Scalar moments are arbitrarily aligned along z (note that magnetic symmetry not well defined with scalar moments), this comes directly from the `Magmom` class
	* Useful for visualization of magnetic moments (works with VESTA, JMOL etc.)

* Tested on Py3.6 and Py2.7 locally

* Closes #508

None

## TODO (if any)

Core functionality is all present, but there are some features to add (these raise NotImplementedErrors for now):

* Add code to parse an arbitrary change of setting, which is included in the spec

* Add support for disordered magnetic structures
    * I don't think there's currently a good way to do this in pymatgen, was going to implement this by adding a `Magmom` to a `Specie` 'spin' property, but this would likely be confusing to the end user (would introduce a fictitious oxidation state, and misuses the spin property) ... open to suggestions here

additionally,

* More graceful handing of `get_primitive_structure()` for magnetic structures, which often gives the wrong results for complex magnetic structures. Will address this with `get_primitive_magnetic_structure()` in a `MagneticStructureAnalyzer` in a future PR, until then advise using `primitive=False` in `CifParser` for magCIF files.

* (Future) Add magnetic symmetry detection support to output full-featured magCIF files containing appropriate symmetry information

### DIFF
--- a/pymatgen/io/cif.py
+++ b/pymatgen/io/cif.py
@@ -277,7 +277,7 @@ class CifFile(object):
 
     @classmethod
     def from_file(cls, filename):
-        with zopen(filename, "rt") as f:
+        with zopen(filename, "rt", errors="replace") as f:
             return cls.from_string(f.read())
 
 
--- a/pymatgen/io/cif.py
+++ b/pymatgen/io/cif.py
@@ -16,6 +16,7 @@ from six.moves import zip, cStringIO
 
 import numpy as np
 from functools import partial
+
 try:
     from inspect import getfullargspec as getargspec
 except ImportError:
@@ -32,6 +33,9 @@ from pymatgen.core.composition import Composition
 from pymatgen.core.operations import SymmOp
 from pymatgen.symmetry.groups import SpaceGroup, SYMM_DATA
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
+from pymatgen.electronic_structure.core import Magmom
+from pymatgen.core.operations import MagSymmOp
+from pymatgen.symmetry.maggroups import MagneticSpaceGroup
 
 """
 Wrapper classes for Cif input and output from Structures.
@@ -242,7 +246,7 @@ class CifBlock(object):
 
 class CifFile(object):
     """
-    Reads and parses CifBlocks from a .cif file
+    Reads and parses CifBlocks from a .cif file or string
     """
 
     def __init__(self, data, orig_string=None, comment=None):
@@ -280,7 +284,6 @@ class CifFile(object):
         with zopen(filename, "rt", errors="replace") as f:
             return cls.from_string(f.read())
 
-
 class CifParser(object):
     """
     Parses a cif file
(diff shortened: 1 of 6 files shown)


mkhorton:
Two additional comments: (1) my CircleCI isn't working yet, so hope this one's okay! tests pass locally though, (2) what's the desired behavior of `occupancy_tolerance` in `CifParser`? It doesn't seem to do anything at present, the line `if 1 < totaloccu <= self._occupancy_tolerance:` looks like it'll always evaluate as `True` ... haven't submitted a fix because I wasn't sure what you wanted it to do. Might suggest replacing it was a `rescale_occupancies` (bool) kwarg instead, that will just rescale occupancies to sum 1.0 if True?

shyuep:
GReat. Thanks!

PR #678, opened 2017-05-24
Add from_file method to PWInput

utopianf:
## Summary

* Added method from_file to PWInput.

* Enabled PWInput type list.
  * Enabled to read tags shown like "list(index) = val"

* PWInput changes
  * pseudo (dict) not required.
    * For spin polarization calculation 
    (nspin=2), Site.properties for each site in Structure class will be needed.
    However, for non-polarization calculation, pseudo(dict) is also OK.

* Added test file for PWInput
  * test_files/sio2.pw.in

None

## TODO (if any)

None


shyuep:
Thanks!

PR #691, opened 2017-06-09
add dielectric function parser for vasp5.4.4

zbwang:
## Summary

In VASP 5.4.4, there are two datasets given in OUTCAR for dielectric function (LOPTICS=True). One is "density-density" and the other is "current-current". The first data set has values of IMAGINARY and REAL as the old VASP version (e.g vasp.5.4.1). This push adds the feature of extracting the first dataset of VASP5.4.4 while keeping the functionality of old read_freq_dielectric.


zbwang:
Sorry, there is an error on testing. I am working on fixing it.

PR #697, opened 2017-06-13
Enable k-point generation from JHU servlet

montoyjh:
## Summary

This PR enables retrieval of a Kpoint object via retrieval from the JHU servlet via a static method from the Kpoints object.

See http://muellergroup.jhu.edu/K-Points.html and Wisesa, McGill, Mueller, PRB 93, 155109 (2017).

* Usage: Kpoints.get_from_wmm(structure, min_distance=29)


shyuep:
Thanks. But I am not sure I want this method in the Kpoint class. In general, I want to keep the core classes and io classes free of entanglements to external sources etc. The idea is that I may want to refactor all the core classes out to a pymatgen-core at some later date. 

Can you create a new package called pymatgen.ext.wmm and put your new method in there? At some point, I will move the matproj interface in there too. In future, we can build other interfaces to other web databases and REST APIs.

montoyjh:
Yep, I agree and think that's a good plan.  Will update later this week.

montoyjh:
Got one of my branches crossed here, going to close and reopen.

PR #701, opened 2017-06-16
You can supply max bond distances to recognize dimensionality of NaCl

kmu:
## Summary

* This PR extends the `get_dimensionality` method in `strucuture_analyzer.py`.
* Users can optionally pass max bond distances to correctly recognize the dimensionality of NaCl, which is problematic by default as commented in `get_dimensionality` by the original author.
* If the max bond distances are not supplied, original `get_max_bond_lengths` method works to detect them.
* This can be helpful if users want to ignore specific types of bonds.
* Unit tests are added to verify that
    * CsCl is recognized as 1D using `get_max_bond_lengths` (default)
    * CsCl is recognized as 3D by passing the max bond distance of Cs-Cl.
* Hope this helps!

None

## TODO (if any)

None

### DIFF
--- a/pymatgen/analysis/structure_analyzer.py
+++ b/pymatgen/analysis/structure_analyzer.py
@@ -605,7 +605,7 @@ def get_max_bond_lengths(structure, el_radius_updates=None):
 
 def get_dimensionality(structure, max_hkl=2, el_radius_updates=None,
                        min_slab_size=5, min_vacuum_size=5,
-                       standardize=True):
+                       standardize=True, bonds=None):
 
     """
     This method returns whether a structure is 3D, 2D (layered), or 1D (linear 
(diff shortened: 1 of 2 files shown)


computron:
Hi @kmu - thanks for this.

Note that the original code should have allowed you to update the bond lengths by overriding the atomic radii via ```el_radius_updates```, but I forgot to pass the setting to a function. I just pushed a fix for that so that the setting works as intended.

Note that your proposed PR ```bonds``` operates on atom pairs rather than atoms, so the functionality is slightly different than ```el_radius_updates```.

@shyuep I think it's good to have both options and this PR is good to go. The only suggested update is to clarify in the docs that ```el_radius_updates``` and ```bonds``` perform similar things and one can do one or the other method, but not both. i.e., if one specifies ```bonds```, the auto-bonding algorithm is turned off and only those bonds are considered.

shyuep:
Can one of you please fix the docs before I merge? THanks.

kmu:
@computron  Thank you for the modification and suggestion! 
@shyuep  I have just fixed the docs.

shyuep:
Thanks! I just merged.

PR #712, opened 2017-07-05
Dev elastic

montoyjh:
## New TOEC functionality

This PR includes several updates to the elasticity package which extends many of the third and higher-order elastic constant functionality, in addition to refactoring several elements of prior functionality.

* Tensors
    - Einstein summation is simplified with einsum_sequence, which enables quick einstein notation inner and outer products
    - Tensors may be constructed from values and indices, with an option to populate the tensor from structure symmetry, e. g. creating the elastic tensor for Cu from c11, c12, and c44.
    - Tensor population from structure symmetry occurs via an initial step in which zero entries are revised for each symmetry operation and then the entire tensor updated until self-consistent.
* Strains
    - Deformation gradients from Strains are constructed via either cholesky decomposition (by default) or the sqrtm method in the convert_strain_to_deformation function
    - DeformedStructureSets are constructed from actual independent strains, rather than independent deformations
    - Removes IndependentStrain object.  All MP-legacy functionality should now be removed.
* Elastic Tensors
    - ComplianceTensor is now an object in order to allow for its proper voigt conversion
    - directional Poisson's ratios, Green-Kristoffel Tensor, directional Young's modulus
    - Gruneisen parameters from third-order elastic tensors, including GGT, TGT, and linear expansion coefficients.  Note that the spherical integration functionality introduces a dataset (e. g. points and weights) for Lebedev quadrature.
    - Initial implementations of yield stress and Wallace tensor determination
    
## TODO:
- Will eventually expand yield stress functionality (e. g. to enable yield surfaces, etc.).
- Am thinking an example notebook with an exploration of this functionality will be useful.


montoyjh:
It's a Lebedev quadrature for integrating the Grüneisen tensor over the surface of a sphere.

I generated it from quadpy.sphere.Lebedev, but wasn't able to use their integration tools directly because of the way I'm using numpy's einsum functionality.  I figured just having the raw data for the default was better than having a new dependency for just one function.

shyuep:
Ok thanks for the clarification.

PR #713, opened 2017-07-07
add charge_balance_combos to Composition

computron:
## Summary

Adds a new feature to a composition to see if a possible charge-balanced solution can be found. Note that you don't need to assign oxidation states a priori - the algorithm will figure out whether the composition can possibly be charge balanced using either pymatgen common oxidation states or user-specified oxidation states. Will return back a list of all possible charge-balance solutions, with each element of the list being a dict of element->avg_oxidation_state.

Likely helpful to detect ionic compounds as well as exclude compositions that are say S6+ instead of S2- based on the composition alone.

* Add method charge_balance_combos() to Composition.
* Add unit tests for that method

None

## TODO (if any)

* Note that the name "charge_balance_combos" may not be ideal. Feel free to rename if you come up with something better.
* The list of pymatgen Element.common_oxidation_states, used as default oxidation state possibilities, is not really complete. e.g., the common oxidation states of V are only 5+ and not 2+,3+,4+,5+. I initially had an option to use Element.oxidation_states if desired, which is much more inclusive, but found it to be much too inclusive and had all sorts of very uncommon oxidation states. Something in between the two oxidation state tables would be more desirable. For now, the user can use the oxidation_override parameter to set their desired oxidation state preferences.


pymatgen/core/composition.py
```diff
@@ -572,6 +575,59 @@ def to_data_dict(self):
                 "elements": self.as_dict().keys(),
                 "nelements": len(self.as_dict().keys())}
 
+    def charge_balance_combos(self, oxidation_override=None):
```
shyuep: Suggestions:
i. Rename to "get_oxi_state_guesses".
ii. kwargs to have oxi_states_override and target_charge=0.
iii. Allow to find solutions that are not charge balanced (by setting sum.
iv. Use not just common_oxidation_states but all oxidation states, but rank those with only common oxidation states higher.
v. Among those that comprise only common oxidation states, rank those that have the lowest mean_square higher. This biases solutions to find low oxidation states solutions, which I think is more physical.

computron:
Ok, so:

1. Done
2. Done
3. As far as  I can tell, this is the same thing as 2? i.e. target_charge. Let me know if not.
4. This one is tricky. The way the algorithm works, you lose the history of how the balance was generated unless you really change everything about it. Note that you also can't really "post-check" solutions since it might be an average of multiple common oxidation states. What I did was just add an option for using all oxidation states. The user can choose to try all common_oxidation_states first, and if they don't get a result, try it with all_oxi_states to get the behavior you describe. (technically, to really get exactly what you want, you filter entries from the first list from the second list and then concatenate the lists together)
5. I added this ranking. I do think it will help on average (especially if one restricts to common oxidation states) but there will of course be problems. e.g., the famous Li2MnNiO4 where it's Ni2+ and Mn4+ rather than 3/3. If you do all oxidation states, you'll likely get non-sensical results.

Btw, to demonstrate why I didn't include the all_oxi_states option in the first PR (only common oxidation states), take a look at what you get back for Mn2O3 with all_oxi_states on:

```
print(Composition("Mn2O3").oxi_state_guesses(all_oxi_states=True))

[{u'Mn': 0.0, u'O': 0.0}, {u'Mn': 0.5, u'O': -0.3333333333333333}, {u'Mn': -0.5, u'O': 0.3333333333333333}, {u'Mn': 1.0, u'O': -0.6666666666666666}, {u'Mn': -1.0, u'O': 0.6666666666666666}, {u'Mn': 1.5, u'O': -1.0}, {u'Mn': -1.5, u'O': 1.0}, {u'Mn': 2.0, u'O': -1.3333333333333333}, {u'Mn': -2.0, u'O': 1.3333333333333333}, {u'Mn': 2.5, u'O': -1.6666666666666667}, {u'Mn': -2.5, u'O': 1.6666666666666667}, {u'Mn': 3.0, u'O': -2.0}, {u'Mn': -3.0, u'O': 2.0}]
```

A lot of these are just silly (Mn: -3, O: +2)?

You can try a bunch of other compositions and see that it just produces non-helpful results.

shyuep:
Thanks. I think the current implementation is fine. I agree the all oxidation states sound a bit silly. 

For the ranking, I agree it is a bit arbitrary. But honestly, we cannot expect better accuracy than that. NMC is well known to be Ni2+, Co3+, Mn4+. The current ranking will return 3+, 3+, 3+. But really, there is no way of knowing whether the former or latter is better. For the longest time, no one really knew what the oxidation states for NMC is likely to be, even with the wisdom of far better chemists with hundred years of accumulated chemistry knowledge. So ultimately, we just have to return both and choose one of them to be ahead of the other based on some criteria. 

Will merge now.

PR #714, opened 2017-07-08
Minor reconfiguring of IEEE conversion and helper function for tensor mapping

montoyjh:
## Summary

Reconfigures IEEE conversion to generate the rotation and explicitly convert the tensor separately, so structures can be set into the IEEE basis if desired (e. g. with a primitive cell).  Also adds a function that's useful for looking up values in tensor-keyed dictionaries (like the ones generated in symmetry reduction) and a corresponding unittest.


montoyjh:
Found an issue with symmetry reduction algorithm that I'd like to resolve before merging, so going to close for now.

montoyjh:
Nevermind, issue I thought was in pymatgen code was elsewhere.  This should be good to go.

PR #715, opened 2017-07-10
add chgcar_ref args to bader calculation

renpj:
The CHGCAR files from the VASP only contain the valance charge density, which may give the wrong bader charge. This can be overcome by using:
>  bader CHGCAR -ref CHGCAR_sum

where CHGCAR_sum is calculated by AECCAR0 + AECCAR2.

(See http://theory.cm.utexas.edu/henkelman/code/bader/ for more details.)


shyuep:
Pls add a unittest for the new functionality. Thanks.

mkhorton:
Does this require manual construction of chgcar_ref? Might be better to have AECCAR arguments in pymatgen and calculated chgcar_ref internally.

renpj:
@shyuep Sure, I can modify the test file. I need generate a new CHGCAR and  AECCAR0, AECCAR2.
@mkhorton Yes, you're right. The only defect is we should add one more arg to the function.

shyuep:
Let me know when this is reasdy to merge.

renpj:
@shyuep All modifications were passed the test, pls merge it.

PR #718, opened 2017-07-10
Update of Specie

mkhorton:
## Summary

All changes are backwards compatible/non-breaking.

**Change:** Oxidation state is now optional
**Motivation:** There's a difference between an oxidation state of 0 and `None`/unknown. Currently, if we want to assign a `spin` property and we do not know an oxidation state, we have to invent a fictious one. This causes problems if the Structure is then serialized, because people might think the oxidation state is 'real' (e.g. from BVAnalyzer or Bader etc.)

**Change:** `Structure` now has `with_specie_property_as_site_property` and `with_site_property_as_specie_property` methods. This allows general switching between `Specie` and `site` properties.
**Motivation:** This is along similar lines as `add_oxidation_state_by_site` except more general, and using site_properties as a store. There have been several instances where these methods have been useful, for example to convert from a Specie Spin property to a magnetic moment on a Site property.  Also, we have different Structures in the MP database with e.g. an oxidation state from BVAnalyzer and also an oxidation state from Bader, and it's been useful to swap between them, or store them both in one canonical Structure with both sets of information annotated in site_properties.

**Change:** Oxidation state is now a property
**Motivation:** This follows from above. This is only an internal change, specie constructor and to/from dict remain the same for compatibility.

None

## TODO (if any)

None


mkhorton:
Closing for now, the species and structure tests are passing fine, but I want to take a look at the failing Voronoi test ...

shyuep:
I don't quite understand the purpose for this. Can you provide an example usage? My problem with this is that it is confusing if the same property can be part of both. There is a reason why oxi state is a Specie property. A site can be disordered with different species of different oxidation states. The idea of having a site oxidation state only makes sense in the standpoint of a completely ordered site.

If the intent is to simply pass through Species properties to site properties, that can be easily done by suitable additions to getattr of Site.

shyuep:
I implemented the passthrough. Check it out first. So you cna now do something like structure[0].oxi_state. As for setting Specie properties, taht is not allowed. Specie is immutable (for very good reason). When oxidation states are changed, you do not actually modify specie but rather create a new Specie to replace the old one.

mkhorton:
Regarding use cases, I agree it's confusing that there are two separate places to store properties. To be clear, I wasn't suggesting oxidation state should be *both* a site and a Specie property, because I agree that it belongs in Specie.

Here are two illustrative uses to explain where this is useful:

* Magnetic moment is currently a site property. Arguably, it might make more sense as a species property instead. It's been convenient to 'convert' a Structure to have the magnetic moment on a Specie spin property instead of the site property: this way, two Specie with different magnetic moments are not treated as equal. Also, disordered magnetic structures are not possible with magnetic moments as site properties.

* There are lots of algorithms that output a value for each site, and so site_properties is a useful store for the outputs of these algorithms. We have several different structures in the MP database, e.g. a bv_structure, a Bader structure, a 'canonical' structure which has forces and co-ordination as site properties, etc. To minimize confusion, we were thinking about tidying this up, and returning a structure that had, say, both a `bv_oxi_state`, `bader_oxi_state` as site properties. But when you decide which oxidation state to use, you have to have a way to apply them to the structure. The existing method `add_oxidation_state_by_site` basically does this, but doesn't address the question of where we store these oxidation states.

So this is more a useful internal algorithm than something I'd expect to be used by an end user. For the magnetic moment case, this would definitely be useful, but I'm not sure if there are better alternatives for the second case?

mkhorton:
For a specific example, consider:

```
s = mpr.get_structures('NiO')[0]
s.get_primitive_structure() # doesn't respect magmom site props, returns primitive NiO
s.site_property_to_specie_property('magmom', 'spin')
s.get_primitive_structure() # now returns correct AFM primitive NiO
```

So for this specific example we could just 'fix' `get_primitive_structure()` to respect magmoms, if that's what we wanted it to do, but more broadly it's this kind of case where it's useful.

shyuep:
The design principle is that the core Element/Specie encapsulates physical information that rarely changes (if ever). Hence, supported properties on Specie is highly restricted. This also makes it logical for Element to be an Enum type and Specie to be a cached type for memory efficiency. Electron spin is a hard definition that takes on integer multiples and is based on a hard definition of an oxidation state and the number of electrons in the Specie (granted that these are not strictly defined).

The properties on Site/PeriodicSite are looser definitions that accounts for the combination of a Specie with a local environment. Hence, there is no definition of supported_properties on Site. Here, anything goes. Magnetic moment, which is different from spin, is given by the specie and the hybridization with the local environment. It takes on non-integer values, usually supplied from a calculation.

As far as I know, there is nothing preventing you from adding a site_property bv_oxidation_state or any other site related property.

For now, I think the best solution is to just add a "add_spin_by_specie" or "add_spin_by_site" method to Structure.

mkhorton:
>  You have an AFM NiO but after reducing to a primitive cell, it is FM only. 

So what would be the suggested method for obtaining the primitive magnetic cell? That is, for the case where you have an ordered non-primitive cell (e.g. supercell or alternative setting), and you want the primitive cell that retains the ordering.

That was the intent of my example, to show that `get_primitive_cell()` doesn't respect magnetic ordering -- which is presumably correct and by design, but there are cases where you do want to respect the magnetic ordering when trying to calculate a primitive cell.

Annotating Specie with the relevant spin in this case seems to work, and seems like the most obvious option.

> Site properties are not accounted for in determining primitive cell and symmetries. ... The magmom being selected is arbitrary based on grouping.

I'd argue that it's potentially misleading to retain an arbitrary site property when producing the primitive cell, *if* site properties differ from otherwise symmetrically-equivalent sites? That's essentially ignoring whatever meaning is present on those site properties.

shyuep:
I just implemented use_site_props option in get_primitive_cell. 
You can do the add_spin methods.
Certainly, we can extend this to symmetry as well, but I don't think people usually consider symmetries based on site properties, unless it is magnetic symmetries. In that case, it is better to be explicit and cast your calculated magmom to a spin. This is because magmom is a float, which makes it prone to tolerance issues, e.g., is 6.999 the same as 7 for magnetic symmetry purposes?

mkhorton:
Okay, thanks very much!

> I recognize there are instances (rare in my view) you do not want the primitive cell to be reduced and some arbitrary site property is retained.

Yeah, I think it probably is quite rare, but it seems to be causing unintended behavior the existing non-FM structures on MP -- this is probably not the place to discuss that though.

> This is because magmom is a float, which makes it prone to tolerance issues, e.g., is 6.999 the same as 7 for magnetic symmetry purposes?

And very true... purely for symmetry, I've been normalizing the orderings (e.g. retain direction of the magnetic moment vector, but normalize to unity) when comparing cells for magnetic symmetry purposes. This is in a MagneticSymmetryAnalyzer for a future PR...

PR #721, opened 2017-07-13
lammps_log parser update

wood-b:
## Summary

I added more fields to be parsed (simulation info and # of dangerous builds). Additionally, I changed the structure of thermo_data from a numpy array to a dictionary.

## TODO (if any)

We need to write a parser for the one and multi thermo_styles. At the moment only thermo_style custom is supported


shyuep:
There is a failing py27 test. https://circleci.com/gh/materialsproject/pymatgen/6476
Pls fix.

PR #725, opened 2017-07-17
Slight changes in StructureMotifInterstitial class

nisse3000:
1) Using new definitions of tetrahedral and octahedral order parameters now, which cannot become negative.

2) Added a function to obtain OP value associated with a given interstitial site.


shyuep:
There are failing tests. And I am not clear why tet is renamed tetalt.

nisse3000:
There are now variants of tetrahedral and octahedral order parameters (tetalt and octalt) that cannot become negative.

nisse3000:
The current pull request should be successfully passing the tests. At least, the unit tests didn't make any trouble on my system. I apologize for not having run the tests before issuing the pull request. My bad.

computron:
I'm not sure "tetalt" will be clear to people. Maybe "tet_v2"?

shyuep:
In general, I think an enum type would work better. I have no idea what kind of order parameters are available without looking at the code. This is generally a bad design from a user perspective. Always assume that the user is going to try to use your method without looking at the body of the code to figure out what parameters are accepted. People need to be able to work it out from purely the method signature alone

nisse3000:
Thanks for the suggestions, Anubhav and Shyue.

I just observed that I forgot to update the documentation of the constructor to reflect the definition of alternative qtet and qoct. So, if I added something like:

"tetalt": alternative Peters-style OP recognizing tetrahedral coordination which importantly avoids negative values.

I thought that it would be good to keep the original OPs. Another way around it is to call the new OPs qtet and qoct and the old ones qtet_orig and qoct_orig. Does that sound better/more clear?

The types argument (list) is the only required argument for the constructor. All other have default values. I could make that description more concise if you think that helps increasing clarity.

computron:
My personal preference is that if users are expected to use the "revised" parameters, to rename the old ones to _legacy or _orig - i.e., to make things clear for new users by just having the new ones be "tet". But that breaks backwards compatibility with any existing code that's out in the wild that references the current "tet". Since the order parameters code is still pretty new it might be OK, but up to @shyuep

mkhorton:
For keeping older OPs, I'd vote for '_legacy' over '_orig'. For 'tetalt' specificly, why not 'tet_peters'? But as long as it's properly described in the method docstring, I'm sure either is fine and won't cause too much confusion.

nisse3000:
Thanks, guys, for the great feedback and suggestions.  I'll go ahead and rename the original qtet and qoct to qtet_legacy and qoct_legacy as well as updating the documentation to explain the issue and change things in StructureMotifInterstitial and other code (e.g., matminer and pycdt).

nisse3000:
BTW, Matt, I like your profile picture. :-)

PR #732, opened 2017-07-25
Make BoltztrapRunner MSONable

shyamd:
Made BoltztrapRunner MSONable 
Makes it easier to store and pass this object around.


shyuep:
Unittest pls...

PR #741, opened 2017-07-29
Bader update + tests for same (v2)

mkhorton:
## Summary

* Can now automatically construct reference file from AECCAR0 and AECCAR2
* Can now calculate magnetic moments using Bader method
* Adds MSONable `BaderResult` to easily save results from Bader analysis
* Now detects version of `bader` binary, and warns if it's too old (there was previously a minor issue with regards to calculating vacuum charge with reference files)
* Adds helper method to automatically run Bader analysis on a typical VASP output directory, or from pymatgen Chgcar objects
* Fixes regression in `BaderAnalysis` from  #715 where analysis would fail if reference wasn't supplied
* Fixes headers ('min' is now 'min_dist', 'dist' is now 'atomic_vol')

None

## TODO (if any)

None, though I might have renamed `BaderAnalysis` to `BaderCaller` because I think it makes more sense, but I left it as-is for backwards compatibility.

Edit: First PR didn't have tests for MSONable, this has been fixed now.

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -8,6 +8,7 @@ import os
 import subprocess
 import shutil
 import warnings
+import glob
 
 from six.moves import map, zip
 from pymatgen.io.vasp.outputs import Chgcar
@@ -208,7 +209,7 @@ class BaderAnalysis(object):
         return structure
 
     @property
-    def result(self):
+    def summary(self):
 
         summary = {
             "min_dist": [d['min_dist'] for d in self.data],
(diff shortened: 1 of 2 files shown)


mkhorton:
Weird, this is passing on my local machine for both py27 and py36, but not in CI, issue is from:

```
  File "/root/pymatgen/pymatgen/command_line/tests/test_bader_caller.py", line 71, in test_automatic_runner
    self.assertEqual(bader_result.__dict__, summary_ref)
AssertionError: {'min_dist': [0.8358, 0.9295, 0.9295, 0.973,[544 chars]705]} != {'magmom': [4.298761, 4.221997, 4.221997, 3.[647 chars]True}
```

This is just comparing two dicts, but in the CI it seems to be treating them as strings?

matk86:
use self.assertDictEqual

pymatgen/command_line/bader_caller.py
```diff
@@ -193,3 +206,175 @@ def get_oxidation_state_decorated_structure(self):
...
+        }
+
+        if self.potcar:
+            charge_transfer = [self.get_charge_transfer(i) for i in range(len(self.data))]
+            summary['charge_transfer'] = charge_transfer
+
+        return BaderResult(summary)
+
+
+class BaderResult(MSONable):
```
shyuep: I am not convinced of the value of this. A dict is already jsonable without modifications. Why do we need a separate class to store the results? I can see the value if you actually want to combine results in some useful way, e.g., incorporating it with the relaxed structure to provide a propertied structure, or provide more sophisticated analyses.

The from_path and from_object methods can easily be standalone functions. 

The methods below also assumes certain names for CHGCAR, AECAR0 etc. It would be better to glob files of CHGCAR* and choose the sorted version. Often, the files we use have suffixes like CHGCAR.relax.2.gz.

mkhorton: Ideally, I would have made the BaderAnalysis itself MSONable, but I can't without changing the constructor and breaking backwards compatibility. I put it into its own class under the assumption it'd have some methods added in future (e.g. like `get_oxidation_state_decorated_structure`), but I'm more than happy to just put `from_path` and `from_object` into their own functions for now instead.

Will modify `from_path` to take a `suffix` argument and glob those, and if `suffix==None` will do a sort as proposed.

shyuep: You can rename BaderAnalysis to BaderCaller. Just add a deprecation stub. 

Perhaps the question I have is why you need to make it MSONable. MSONAble is used when you have an object that cannot be easily serialized to json. A dict of list of values can already be serialized without additional effort.

Also, in general, I prefer if MSONable classes use the default as_dict and from_dict methods without subclass modification. This usually can be done quite easily. Implementing a custom as_dict and from_dict makes it very likely for people to add a new arg/kwarg and then forget to update the as_dict and from_dict methods. Following the typical MSONAble protocol automatically deals with this.

mkhorton: Point well taken about MSONable -- will leave them as plain dicts for now, and possibly refactor in future if it calls for it. Have moved the `from_path` and `from_object` methods into their own standalone functions.

shyuep:
Tests failed. Can you submit a pull request fixing the problem?

Also, the tests depend on bader executables in the cmd_line directory. That might be the cause. The versions might be outdated.

mkhorton:
Looks like the same problem as before, though I'm using `assertDictEqual` now:

```
======================================================================
FAIL: test_automatic_runner (pymatgen.command_line.tests.test_bader_caller.BaderAnalysisTest)
  File "/home/travis/build/materialsproject/pymatgen/pymatgen/command_line/tests/test_bader_caller.py", line 71, in test_automatic_runner
    self.assertDictEqual(summary, summary_ref)
AssertionError: {'min_dist': [0.8358, 0.9295, 0.9295, 0.973,[544 chars]705]} != {'magmom': [4.298761, 4.221997, 4.221997, 3.[647 chars]True}
Diff is 4625 characters long. Set self.maxDiff to None to see it.
```

Tests pass locally. Will look into it.

shyuep:
Instead of doing dict equal, it is better to just compare the values for each key with a loop. That wya you know exactly where the problem is. It is likely a precision issue. These are floats. You probably need ArrayAlmostEqual or something of that sort.

mkhorton:
Yeah, they may not be exactly equal if different bader versions are used -- I'll make the test more loose. Generally, `assertDictEqual` should show the keys where the difference is located though, otherwise there's not much point using it over `assertEqual`.

PR #744, opened 2017-07-31
(Wait for CI) Various magmom handling fixes, improvements for SOC calculations + tests

mkhorton:
## Summary

This touches a lot of files, so I'd like to wait for CI results before merging + read them carefully.

* `Outcar` now detects vector magmoms from SOC calculations (previously would only detect x component)

* `Chgcar` now properly handles charge density from SOC calculations
	* for SOC calculations, `.data` will have `.data['diff_x']`, `.data['diff_y']`, `.data['diff_z']` as the canonical magnetization densities, but will also try to construct a `.data['diff']` referenced to an arbitrary direction to give a scalar density (this uses the same approach as `float(Magmom)`)
	* has new `net_magnetization` property and `is_soc` property
	
* `Structure` `to`/`from_file` now detects `.mcif`
* `Structure` implements `add_spin_by_site`, `remove_spin` (c.f. `add_oxidation_state_by_site`, `remove_oxidation_state`), which was suggested in #718
	* Had to include the minor change to `Specie` that makes oxidation state optional, so we can have Specie with spin even if oxidation state is unknown, also added comma to str representation (so `Ni,spin=5`, `Ni2+,spin=5` vs old `Nispin=5`, `Ni2+spin=5`)

* `CifWriter` and `Incar` includes better handling of Magmom class, more tests

* `Magmom` has minor bug fix to define equality between floats and Magmoms

* ~~~**`Site` now has explicit `magmom` attr**, this follows the same logic as in `io.vasp.sets`, which is to look for site property first and then spin property if not, and will always returns a instance of `Magmom` if magmom is defined~~~ (this was a bad idea, reverted)
	* ~~~this makes reasoning about magmoms a lot easier since it provides a uniform interface to magmom; if we do want to change this behavior in future, it'll be much easier to change it here than hunting everywhere magmoms are handled in pymatgen (I have a few ideas to improve this later)~~~
	*  ~~~I think this is probably the most controversial change here, do we want to keep this?~~~

None

## TODO (if any)

* (In future) Improvements to handling of the spin quantization axis, in particular processing it from `Outcar`, setting the `SAXIS` variable in `Incar`, and smarter handling in `MPSOCSet` (and, by extension, atomate's SOC powerups too), and better tests for non-standard `SAXIS` throughout


shyuep:
There seems to be *a lot* of test failures.

mkhorton:
Magmom's already in electronic_structure.core.

Agreed re. test failures, I tested all the tests for the modules that were changed and they passed, but I suspected there could be failures in unrelated modules, which is why I wanted to see the CI results (I haven't got my own CI set up yet, but will do soon hopefully).

mkhorton:
Will close it for now until test failures are resolved.

shyuep:
You don't need to do your own CI setup. You just need to cd into the pymatgen/core directory, and then run "nosetest". At the minimum, all the errors on the core classes will pop up.

mkhorton:
Have reverted the problematic change (the explicit magmom site attr, I think this was a bad idea!). The other changes should be non-breaking, but let's see...

My nosetests are passing in pymatgen/core, but I often have nosetests fail even on the master branch, and I haven't had time to debug it yet. I think it may be because of an old enumlib.

PR #748, opened 2017-08-04
`BaderAnalysis.from_path`

adengz:
## Summary

Added a Convenient constructor that takes in the path name of VASP run to perform Bader analysis.

## TODO (if any)

code cleanup in `bader_caller.py`, I'll leave this for @mkhorton to decide.

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -234,27 +234,28 @@ class BaderAnalysis(object):
             name_pattern = filename + suffix + '*' if filename != 'POTCAR' \
                 else filename + '*'
             paths = glob.glob(os.path.join(path, name_pattern))
-            if not paths:
-                if filename == "CHGCAR":
-                    raise IOError("Could not find CHGCAR!")
-                else:
-                    if filename == "POTCAR":
-                        warning_msg = "cannot calculate charge transfer."
-                    else:
-                        warning_msg = "interpret Bader results with caution."
-                    warnings.warn("Could not find %s, " % filename
-                                  + warning_msg)
-                    return None
-            if len(paths) > 1:
+            fpath = None
+            if len(paths) >= 1:
                 # using reverse=True because, if multiple files are present,
                 # they likely have suffixes 'static', 'relax', 'relax2', etc.
                 # and this would give 'static' over 'relax2' over 'relax'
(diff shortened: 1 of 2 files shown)


pymatgen/command_line/bader_caller.py
```diff
@@ -40,10 +38,6 @@
 __date__ = "4/5/13"
 
 
-@requires(which("bader") or which("bader.exe"),
```
mkhorton: I wouldn't remove this, it's a useful error message for people who haven't installed the bader binary. I think `@requires(which(...))` probably gave an error when you added the classmethod below, but you should be able to move the `@requires()` to annotate the class `__init__` instead?

pymatgen/command_line/bader_caller.py
```diff
@@ -231,6 +225,62 @@ def summary(self):
 
         return summary
 
+    @classmethod
```
mkhorton: I think this makes sense, but if you're adding `.from_path()` here, I'd delete the `bader_analysis_from_path()` standalone function. No point keeping both!

pymatgen/command_line/bader_caller.py
```diff
@@ -231,6 +225,62 @@ def summary(self):
...
+                # they likely have suffixes 'static', 'relax', 'relax2', etc.
+                # and this would give 'static' over 'relax2' over 'relax'
+                # however, better to use 'suffix' kwarg to avoid this!
+                paths.sort(reverse=True)
+                warning_msg = "Multiple files detected, using %s" \
+                              % os.path.basename(paths[0]) if len(paths) > 1\
+                    else None
+                fpath = paths[0]
+            else:
+                if filename == "CHGCAR":
```
mkhorton: Instead of this nested if/else logic, probably easier just to return `None` from this function if the file isn't found, and below write:

```
chgcar_filename = _get_filepath("CHGCAR")
if chgcar_filename is None:
    raise IOError("Could not find CHGCAR!")
```

etc.

pymatgen/command_line/bader_caller.py
adengz: Added it back. I got stuck here for the whole afternoon...

pymatgen/command_line/bader_caller.py
adengz: Took your suggestion. Thanks.

PR #761, opened 2017-08-14
QuasiharmonicDebyeApprox: Consider anaharmonic contribution to Debye temperature.

bocklund:
The Debye temperature can have anharmonic contributions which can be calculated through the Gruneisen parameter.

This change implements the anharmonic contribution by default (can be turned off) with updates to documentation to reflect the change.

Previous tests were updated to reflect the anharmonic contribution option.

For theoretical details see:
Moruzzi et al. Phys. Rev. B 37, 790–799 (1988). (Eq 20)
Shang et al. Comput. Mater. Sci. 47, 1040–1048 (2010). (Eq 18)

Result: 

image

@matk86 please take a look

### DIFF
--- a/pymatgen/analysis/quasiharmonic.py
+++ b/pymatgen/analysis/quasiharmonic.py
@@ -26,7 +26,7 @@ from scipy.optimize import minimize
 from pymatgen.core.units import FloatWithUnit
 from pymatgen.analysis.eos import EOS, PolynomialEOS
 
-__author__ = "Kiran Mathew"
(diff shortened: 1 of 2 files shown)


matk86:
looks good, could you add a test with  anharmonic_contribution=True?

matk86:
also for backward compatibility set the default value for  anharmonic_contribution to be False

bocklund:
This is ready for another look

matk86:
:+1:

PR #763, opened 2017-08-16
Interface

yihanxiao92:
## Summary

Add interface_reactions.py and its test file to pymatgen/analysis library. This module does analysis for interfacial reactions between two reactants, and plots reaction energy vs. mixing ratio x. The "Interfacial reaction" web app on Materials Project website will use this module for calculations.

* Gets critical mixing ratio where the composition tie lie intersects with phase diagram facets.
* Gets reaction energy and balanced reaction equation for each critical mixing ratio.
* Finds maximum driving force for interface reaction, which is a measure of the reactivity.

### DIFF
--- a/pymatgen/analysis/interface_reactions.py
+++ b/pymatgen/analysis/interface_reactions.py
@@ -1,0 +1,337 @@
+from __future__ import division
+
+"""
+This module provides InterfacialReactivity class for analyzing interfacial reactions 
+and plotting reaction energy vs. mixing ratio between two reactants.
+
+Class Methods:
+
+    -- get_kinks: Gets information on critical reactions: [indices of kinks, critical
+                  mixing ratio x, critical energy, balanced reaction].
+    
+    -- get_products: Gets all potential potential products in interfacial reactions.
+    
+    -- labels: Dictionary version of class method get_kinks.
+
+    -- plot: Plots the reaction energy as a function of mixing ratios of reaction 
+             between electrolyte and cathode. Kinks are labeled.
+
+    -- minimum: Gets the minimum reaction energy E_min and corresponding mixing ratio
+                X_min: (x_min, energy_min).
+
+    -- get_no_mixing_energy: Gets only the reaction energies of reactants from 
+                             equilibration with the external reservoir.
+
+"""
+
+__author__ = "Yihan Xiao"
+__copyright__ = "Copyright 2011, The Materials Project"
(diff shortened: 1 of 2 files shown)


shyuep:
There are a lot of failing tests. Pls fix before I merge.

yihanxiao92:
Dear Prof. Ong,

The errors in test files during the Travis CI build check all come from plot functions using matplotlib. I am not sure it is a failing test. Not only the newly_added "test_interface_reactons.py" suffered from the error, but also the existing test codes such as test_get_plot (pymatgen.phasediagram.tests.test_plotter.PDPlotterTest) and test_bz_plot (pymatgen.electronic_structure.tests.test_plotter.PlotBZTest). In this case, should I remove the test of plot function in my test file, or is anything could be fixed in the Travis CI build check?

Thanks,
Yihan

mkhorton:
@yihanxiao92 if you read the Travis CI logs, it looks like all the interface reaction tests are failing with this error from [this line](https://github.com/materialsproject/pymatgen/pull/763/files#diff-c4a0a20f0c27251532b028f5b1c2895bR70):

`'Please provide grand phase diagram to compute no_mixing_energy for reactants!`

I suspect something might be going wrong in your test `setUp`?

shyuep:
For tests that use plot, you need to implement matplotlib.use("pdf"). There is no DISPLAY on travis CI. If the backend is not switched to a non-display version, there will be an error.

yihanxiao92:
@mkhorton Yes, you are correct. I was trying to test AssertionError with the code in the following format:
```
with self.assertRaises(Exception) as context:
             #Do something that create AssertionError
         self.assertTrue('Error message' in context.exception)
```
This worked as expected in the nosetests. But it seems the Travis CI misunderstood context.exception to be an iterable and then reported the error. So I modified the line to:
```
with self.assertRaises(Exception) as context:
             #Do something that create AssertionError
         self.assertTrue('Error message' == str(context.exception))
```
Hope that could fix the problem.

shyuep:
@yihanxiao92 I looked at the interface code in greater detail. Unfortunately, the docstrings do not conform to pymatgen standards, and also, the style is definitely not pep8. 

Pls fix all these changes in a pull request. Otherwise, I will need to revert the entire commit until they are fixed.

yihanxiao92:
Sorry about that. I will fix them as soon as I can. Expect to finish in 1-2 days.

PR #771, opened 2017-08-17
updates to oxi_state_guesses and valence table

computron:
## Summary

This PR adds many new functionalities and improvements regarding oxidation states

* The periodic table "Element" class has a new attribute called icsd_oxidation_states. These are oxidation states found in ICSD that have at least 10 examples *and* at least 1% of entries for that element. I have found this to be much superior to the current oxidation state tables, which are either too restrictive or too loose.
* oxi_state_guesses() uses the icsd_oxidation_states by default (not common_oxidation_states, e.g., the latter only has V5+ whereas the new version has V2+,3+,4+,5+)
* the ranking of solutions in oxi_state_guesses() is much improved. Instead of min sum squared of the oxidation states, which can lead to bad solutions, solutions are ranked based on prior probabilities from the bond valence table.
* Structure has a new convenience method called "add_oxidation_states_by_guess". Just give it a Structure and it will decorate it with reasonable oxidation states. Note that it may add fractional / float oxidation states to a site, though.

## TODO (if any)

The loading of prior probabilities is done every time one calls Composition.oxi_state_guesses(). I'd prefer to do this at the top of the composition.py module, but didn't want to slow down the load of this common module for this one feature. So for now, things are a little slower than they could be.


shyuep:
Tests not passing on py3. Pls fix.

PR #776, opened 2017-08-17
Initialization of spectrum module

czhengsci:
## Summary

Initial commit of pymatgen.analysis.spectra module

* Add Spectrum base abstract class
* Add Xanes class
* Add unittest for Xane


pymatgen/analysis/spectra/spectrum.py
```diff
@@ -0,0 +1,161 @@
...
+    @classmethod
+    @abc.abstractmethod
+    def from_file(cls, filename):
+        """
+        Initiate Spectrum from a filename
+        """
+        pass
+
+
+class Xanespectrum(Spectrum, MSONable):
```
shyuep: CamelCase pls. XANESSpectrum.

pymatgen/analysis/spectra/spectrum.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+
+        self.formula = formula
+        self.absorption_specie = absorption_specie
+        self.edge = edge
+        self.structure = structure
+        self.spectrum = spectrum
+
+
+class XanesFEFF(Xanespectrum):
```
shyuep: Why does this need to be a separate class from XANESSpectrum? Spectrum classes should not contain any calculation specific information. It shouldn't matter whether it comes from FEFF or VASP or whatever code.

czhengsci: Actually, that is what I thought when I design the pattern. I agree that the Spectrum class should not contain any calculation specific information, that is why I create a separate XANESSpectrum at first. Thanks for the comments, now it seems like removing the XANESSpectrum class and design individual XANES class for each calculation might be a better idea, i.e. XANESFeff, XANESVasp, EXAFSFeff, XANESEels....

shyuep: No, that's not what I meant. Just keep XANESSpectrum (or even just call it XANES and EXAFS). As far as I can tell, that so called XANESFeff has not FEFF specific info. So why do you want to label it with FEFF? Does FEFF provide something?

shyuep: Note that things like dir_name and FEFF input parameters generally should not be stored in the class. Since when do we do this for any pymatgen analysis object? If you wish, you can have the XANES object take in a `provenance` optional kwarg. That can store things like whether it is calculated by FEFF or whatever code with what parameters. Most of the methods within the XANES object should not operate on provenace data.

Just think about the Dos object. Do I classify it as a VASP DOS? I don't! Density of states is a generic concept that can be generated by VASP, ABINIT and many other codes. The DOS object needs to be flexible enough to deal with all of this. The computational code specific stuff should be part of the io parsing, but not pollute analysis. The way I would analyze a VASP generated DOS is the same as I would do for a ABINIT generated one.

Based on your strategy, you are going to have tends of XANES object each catering to one type of computational code and repeating the exact same methods. That is ridiculous.

czhengsci: Indeed, the XanesFEFF object has some specific information related to the db entry we generated in our database construction. The absorbing_atom_index field and mp-id field value we extracted from the XAS database entry's are pretty useful in generating ensemble spectrum. Some functions we tend to implement regarding spectrum processing rely on these few field. Intrinsically, those specific fields' information more db related. Also, some input parameters like S02 might be helpful in later analysis, I will keep the XANES class as a generic. For the more db specific spectrum classes construction, should I inherited the XANES class from pymatgen and keep them in private repo? Thanks a lot,

shyuep: You are not getting it. Pymatgen has not DB-related code. Provenance info such as MP-ids and running parameters have no place in a generic class for Spectra analysis. You should not design a Spectra class with the intention of containing all info for insertion into the database. By your logic, I would need to create a VASPStructure, ABINITStructure, etc. That is plainly ridiculous. Leave the FEFF or computational methodology specific stuff out of the spectra classes. at best, they can be left in the provenance key, which should not be used in any way in analysis.

shyuep:
Is this ready for merging or are you still working on it? The tests are failing. Also, Spectrum should be MSONAble, not XANES. That way, all subclasses of Spectrum are MSONAble.

czhengsci:
Hi Shyue Ping, I fixed the code and make the Spectrum class MSONAble. All tests passed. I am working on the rest on the todo list now. Thanks a lot

shyamd:
Would it make more sense to have spectrum_data be abstract in the Spectrum class and then XANES can just construct the data on the fly? 

Also, there should be some sort of concat method for XANES that lets you easily add them together, since it's possible to have multiple unique sites for one species that contribute differently to the overall spectrum.

shyuep:
It will be like DOS. There are __add__ methods etc. that allows adding of spectra.

pymatgen/analysis/spectra/spectrum.py
```diff
@@ -0,0 +1,135 @@
...
+        :return: Gaussian-smeared densities
+        """
+
+        from scipy.ndimage.filters import gaussian_filter1d
+        diff = [self.x[i + 1] - self.x[i] for i in range(len(self.x) - 1)]
+        avg_eV_per_step = np.sum(diff) / len(diff)
+        rv = gaussian_filter1d(self.y, sigma / avg_eV_per_step).tolist()
+        return rv
+
+    def ensemble_average(self, spect_list, spect_multi):
```
shyuep: This is not what you are supposed to do. Look at how DOS does it. You just need to implement `__add__` and `__mult__` and you automatically can do linear combination of spectrum. E.g., A *spec1 + B * spec2. That will automatically allow you to do ensemble averages. 

I said many times - LOOK at the DOS object. I don't know how many times I have to write this.

shyuep:
Ok, I will merge and make a few changes.

PR #777, opened 2017-08-17
add read fermi contact shift

paulgholin:
## Summary
Add read_fermi_contact_shift. (Currently Pymatgen can only do: read_chemical_shifts.)


shyuep:
This is fine, but do not use a gzipped OUTCAR. Just a normal outcar. Git can handle file compression.

shyuep:
Also, paste here a sample output for the fermi contact shift data.

PR #778, opened 2017-08-17
Added local_env.py for near-neighbor finding

nisse3000:
In the future, the OrderParameter class from structure_analyzer.py will be moved to this module under the proposed name LocalStructOrderParas to avoid confusion with order parameters in a more global materials science-related sense. Is that desirable, @shyuep?


shyuep:
Great. Can you put deprecation messages on the relevant classes in structure_analyzer? Say they will be removed with pymatgen > 2018.

nisse3000:
Will do.

nisse3000:
Maybe pymatgen > 2017 (so, in 2018)?

shyuep:
Ok... forgive my lack of precision. pmg >= 2018.

nisse3000:
OK. :-)

PR #783, opened 2017-08-18
Minor bugfix for periodic heatmap plotting

johnson1228:
## Summary

This pull request fixes the bug when max_val becomes negative or min_val becomes positive.


shyuep:
You need to merge the master first. Right now, you are reintroducing the code in the periodic_table.py

shyuep:
Look at the files changed. THer should only be one file, but there are three.

PR #785, opened 2017-08-18
add vasp input sets for BSE calc

zbwang:
## Summary

Add GW/BSE vasp input sets with the newest GW psp (PBE_54) as default.


pymatgen/io/vasp/sets.py
```diff
@@ -1044,6 +1044,185 @@ def __init__(self, structure, potim=0.015, **kwargs):
         self.config_dict["INCAR"].pop("NPAR", None)
 
 
+class MVLGWscDFTSet(DictSet):
```
shyuep: Just call this MVLGWSCSet. No need DFT. All the input sets are DFT

pymatgen/io/vasp/sets.py
```diff
@@ -1044,6 +1044,185 @@ def __init__(self, structure, potim=0.015, **kwargs):
...
+                      "LPEAD": True, "OMEGAMAX": 40})
+
+        return incar
+
+    @classmethod
+    def from_prev_calc(cls, prev_calc_dir, copy_wavecar=True,
+                       nbands_factor=5, **kwargs):
+        """
+
+        :param prev_calc_dir:
```
shyuep: These should be Google style docs. You can copy and paste from MPStaticSet to save time.

pymatgen/io/vasp/sets.py
```diff
@@ -1044,6 +1044,185 @@ def __init__(self, structure, potim=0.015, **kwargs):
...
+        """
+        Generate gamma center mesh k-points grid for GW calc.
+        :return: gamma centered k-points
+        """
+        return Kpoints.automatic_density_by_vol(self.structure,
+                                                self.reciprocal_density,
+                                                force_gamma=True)
+
+
+class MVLGWDiagDFTSet(MVLGWscDFTSet):
```
shyuep: Remove DFT.

pymatgen/io/vasp/sets.py
shyuep: It does not seem like you need four input sets. All the input sets are very similar, and the only difference is the incar default settings. I suggest that you just have either one or two input sets, and one of them can have a "mode", which can stipulate whether it is a SC, Diag, G0W0 or BSE calculation, and then set the incar_settings accordingly.

shyuep:
@setten @hautierg @gmatteo @computron (this is a somewhat random list of people I think might be interested in this)

Pls note that I intend to delete the GWvaspinputsets.py. That file is extremely out of date, and also, there are methods in there which writes SLURM job submission scripts. Such things really shouldn't be in pymatgen. If necessary, they should be in FireWorks.

So pls take the opportunity to review these inputs sets that Zhenbin has written, and if you have comments, pls say them now. You can also help improve on these new sets if you wish. Also, note that the default POTCAR for GW is switched to the PBE_54 series for VASP, not the standard PBE. These are necessary to make sure the calculations work as intended.

Issue #12, opened 2012-12-11
Issue with supercells in new StructureMatcher


shyuep:
Go to "test_files" and do 

../scripts/matgenie.py compare Li2O.cif POSCAR.Li2O

Both are Li2O. Second one is a big-ish supercell. The code fails with a TypeError after a while, given below:

Traceback (most recent call last):
  File "../scripts/matgenie.py", line 423, in <module>
    args.func(args)
  File "../scripts/matgenie.py", line 282, in compare_structures
    for i, grp in enumerate(m.group_structures(structures)):
  File "/Users/shyuepingong/Documents/Aptana Studio 3 Workspace/pymatgen/pymatgen/analysis/structure_matcher.py", line 271, in group_structures
    if s2_ind == -1 and self.fit(s_list[i], s_list[j]):
  File "/Users/shyuepingong/Documents/Aptana Studio 3 Workspace/pymatgen/pymatgen/analysis/structure_matcher.py", line 202, in fit
    if s2_cart[ind] != []:
TypeError: list indices must be integers, not NoneType

This is likely the issue of finding the primitive cell of the supercell.

PR #786, opened 2017-08-18
Adding GB input set

Tinaatucsd:
This is to address the issue #12 GB input set.


pymatgen/io/vasp/sets.py
```diff
@@ -1125,6 +1125,90 @@ def kpoints(self):
         return kpt
 
 
+class MVLGBVaspInputSet(MPRelaxSet):
```
shyuep: MVLGBSet will do.

pymatgen/io/vasp/sets.py
```diff
@@ -1125,6 +1125,90 @@ def kpoints(self):
...
+    """
+    Class for writing a vasp input file for grain boundary, slab or bulk.
+    Args:
+        structure(Structure): provide the structure
+        k_product(int): kpts[0][0]*a.
+        default to 40
+        bulk(bool): Set to True if the structure is bulk supercell. Default is False.
+        dope(bool): Set to True if your structure is doped structure. Default is False.
```
shyuep: Remove dope parameter. This can be achieved by user_incar_settings={"ISIF": 2}

pymatgen/io/vasp/sets.py
```diff
@@ -1125,6 +1125,90 @@ def kpoints(self):
...
+        slab(bool): Set to True if your structure is slab. Default is False.
+        manual_kpoints(list): e.g. manual_kpoints=[4,4,1]
```
shyuep: What cannot be already achieved by user_kpoint_settings that we need manual_kpoints??

pymatgen/io/vasp/tests/test_sets.py
```diff
@@ -596,5 +595,48 @@ def test_batch_write_input(self):
             shutil.rmtree(d)
 
 
+class GBinputsetTest(unittest.TestCase):
```
shyuep: Test class name should follow the actual class name. E.g, MVLGBSetTest.

shyuep:
Note that there are failing tests. Instead of introducing a new Mo structure, I suggest you just change it to a Li structure instead since the POTCAR for Li is already available.

Issue #767, opened 2017-08-16
Workfunction analysis


shyuep:
These should be made publicly available. @richardtran415 to take charge.

PR #787, opened 2017-08-18
Added tag_surface_sites and surface_sites_equal

richardtran415:
## Summary

This pull request partially adresses issue #767
Added the tag_surface_sites and surface_sites_equal class methods to the Slab class. These methods will be useful later on for further analysis of surface properties such as work function and surface energy as well as finding viable adsorbate sites.

* Feature 1: tag_surface_sites is a class method of Slab, it finds surface sites based on undercoordination relative to the bulk and then adds site properties indicating whether a site is on the surface.

* Feature 2: surface_sites_equal checks if the sites on both surfaces are equivalent. This is useful for defining equal surfaces in slabs that do not have Laue symmetry.

* All functions that looks for Laue symmetry in the surface.py and its test file uses is_laue() method in SpaceGroupAnalyzer class.

## TODO

An algorithm is needed to find sites in a slab equivalent to sites in the bulk.


shyuep:
@richardtran415  There are failing tests. Pls fix.

pymatgen/core/surface.py
```diff
@@ -437,6 +436,97 @@ def from_dict(cls, d):
             site_properties=s.site_properties, energy=d["energy"]
         )
 
+    def tag_surface_sites(self, decimal=5):
```
shyuep: Do we need to tag, or will returning a list of indices of the surface sites suffice?

Also, how do we know the surface site is on the top or bottom surface? I think there are instances where you want to know that

pymatgen/core/surface.py
```diff
@@ -437,6 +436,97 @@ def from_dict(cls, d):
...
+                    properties.append(True)
+                else:
+                    properties.append(False)
+            except RuntimeError:
+                # or if pathological error is returned, indicating a surface site
+                properties.append(True)
+
+        self.add_site_property("is_surf_site", properties)
+
+    def surface_sites_equal(self):
```
shyuep: I think the naming should be changed. Maybe "have_equivalent_surfaces"?

Issue #765, opened 2017-08-16
LAMMPS io to be improved to support structures


shyuep:
LAMMPs io is primarily coded for molecules/polymers. Need to support periodic structures, e.g., for MD simulations and descriptor calculations.

PR #789, opened 2017-08-18
io.lammps updates

matk86:
* added more general lammps data file parser
* enable data file generation for periodic structures. fixes issue #765 
* added from_structure method to inputset
* tests for all new additions


shyuep:
I want to discuss this code. @chc273 is supposed to implement the Structure version. In general, I don't understand why LAMMPSData needs to take in molecules. It should just take in a structure and have a static method that takes in a molecule with box sizes. As far as I can tell, the LAMMPS input file is also periodic, for which a structure is the de facto object.

@chc273 @adengz Discuss with Kiran properly and merge the one that you have in our internal code base to the one in pymatgen. In future, this kind of thing should be done in pymatgen. I don't understand why you implemented it in pymacy in the first place.

matk86:
doesnt LammpsData.from_structure  do what you are suggesting? i wasnt aware of your group working on it..anyway i would rather do it myself than wait for someone...

shyuep:
I am not asking you to wait. I am saying you need to coordinate so that the needs of you and others are handled. I don't care who implements it. But I do care that it is implemented in a way that does not box us into one specific mode of operation.

mkhorton:
I'm not involved in this at all, but have used LAMMPS before. Just a comment on:

> As far as I can tell, the LAMMPS input file is also periodic, for which a structure is the de facto object.

LAMMPS can have a variety of boundary conditions, including mixed boundary conditions (some periodic, some non-periodic), though I don't know how often people use these. http://lammps.sandia.gov/doc/boundary.html

shyuep:
@matk86 Pls check that the merged file is correct in the resolution of conflict. Once it is done, I will merge the complete file. At some point, I want a unified LAMMPSData.

matk86:
checked, no merge conflicts ..looping in @wood-b too

PR #792, opened 2017-08-18
Add LMPData class to convert a pymatgen structure into a charge-type lammps input data. 

chc273:
## Summary

* Read periodic pymatgen structure and convert it to lammps data
* The cell box size is calculated from pymatgen structure
* Enable non-orthogonal cells 

## TODO (if any)
LMPData supports only the specific "atom_type charge" lammps input data and are for crystal structures. It perhaps can be merged with the molecule-specific input data class in the same data.py file.


matk86:
other than setting the charge = oxidation state i dont see the new LMDdata class doing anything different from LammpsData

shyuep:
The old version idd not support non-orthogonal cells. Since the new one that you pushed today did, I have deleted the new LMPData.

PR #798, opened 2017-08-23
minor io.lammps updates

matk86:
* some code simplifications and cleanup


pymatgen/io/lammps/output.py
```diff
@@ -2,22 +2,7 @@
...
-import numpy as np
-
-from monty.json import MSONable
-
-from pymatgen.core.periodic_table import _pt_data
-from pymatgen.core.structure import Structure
-from pymatgen.core.lattice import Lattice
-from pymatgen.analysis.diffusion_analyzer import DiffusionAnalyzer
-from pymatgen.io.lammps.data import LammpsData, LammpsForceFieldData
+from __future__ import division, print_function, unicode_literals, absolute_import
```
shyuep: pymatgen sets margins at 80, not 100. I don't care what the other code bases do, but pymatgen sticks to 80.

shyuep:
I will merge first. But I will make some changes.

pymatgen/io/lammps/output.py
matk86: are we talking about the same pymatgen that contains the chemenv and io.abinit modules.. ..haha

shyuep: I will get to those at some point in time (e.g, the way I got to the GW input sets). 

The fact that there are some code that do not obey the rules does not mean that all code do not obey rules. :-)

P.S. Try maintaining a full code one day and you will see the difficulty.

shyuep: Also, pls do not reverse changes I previously made. E.g., the imports should be at the top of the file. These commits moved them back below.

PR #806, opened 2017-08-24
Bug fix min_freq() in phonon/bandstructure.py

marcoesters:
## Summary

The method min_freq in the PhononBandStructure object returns the wrong q-point. In its current form, the method would return the q-point with the index of the band that contains the minimum instead of using the index of the minimum frequency.

Fix: Change self.qpoints[i[0]] in line 106 to self.qpoints[i[1]]


shyuep:
Thanks!

PR #808, opened 2017-08-24
fix parsing BSE vasprun error

zbwang:
## Summary

When Vasprun() parses vasprun.xml , it generally first checks the convergence status (electronic and ionic). However, there is no electronic or ionic step in BSE calc. This fix is to ignore the convergence checking of BSE vasprun.xml.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -389,7 +389,7 @@ class Vasprun(MSONable):
         try:
             if "BSE" in self.incar["ALGO"]:
                 pass
-            if not self.converged:
+            elif not self.converged:
                 msg = "%s is an unconverged VASP run.\n" % filename
                 msg += "Electronic convergence reached: %s.\n" % \
                        self.converged_electronic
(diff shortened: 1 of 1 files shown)


shyuep:
There is a test error. You need to check your tests and fix.

pymatgen/io/vasp/outputs.py
```diff
@@ -386,14 +386,17 @@ def __init__(self, filename, ionic_step_skip=None,
             if parse_potcar_file:
                 self.update_potcar_spec(parse_potcar_file)
 
-        if self.incar["ALGO"] == "BSE":
+        try:
+            if "BSE" in self.incar["ALGO"]:
```
shyuep: This should not be a try-except. self.incar.get("ALGO") == "BSE" is better.

zbwang:
I am not sure how to deal with this situation in a pythonic way. The fix I wrote looks not professional. You may improve it. Thanks.

pymatgen/io/vasp/outputs.py
zbwang: I see.

shyuep: self.incar.get("ALGO", "") == "BSE"

PR #809, opened 2017-08-24
Converting LMPDATA to POSCAR

ucsdlxg:
## Summary

Add the code to transform from the format of the lammps Data to vasp POSCAR.

* Feature 1
* Feature 2
* Fix 1
* Fix 2

## TODO (if any)

to be done

* Feature 1 supports a, but not b.


shyuep:
The code should not convert LAMMPS to POSCAR. It should just allow a person to get a structure.

shyuep:
There is never a reason to write code to convert one file format to another. If you are able to get a structure from LAMMPSDAta, you can already pass the structure to the Poscar class to get the Poscar object. The Structure object is the thing that links everything.

ucsdlxg:
I understand. I just found that the lammps data file is closed to the POSCAR format. Thus it is easier to transform it to POSCAR first, and then use Poscar class to get the structure, which can link to everything. I have already obtained the POSCAR. Just use Poscar class to get the structure. Can I do it this way?

shyuep:
No,you should not convert to poscar first. It is straightforward to parse the lammps input data to a structure.

shyuep:
Thanks.

shyuep:
@ucsdlxg See the version that I wrote. You should not be introducing a new method and reparse the Data file. There is already a LAMMPData.from_file to parse the data file. All you need is to convert the LAMMPSData to a structure. See the `@structure` property in the latest pymatgen. 

However, for some reason, I cannot retrieve even the Li2O structure. Pls debug.

shyuep:
Nvm, I think the code works fine.

PR #811, opened 2017-08-25
Bug fix for LMPDATA to Structure object

ucsdlxg:
## Summary

I add a variable "atom_style" to tell the LammpsData format.
The difference of atomic mass for some elements is small, e.g. Pt, Ce. Now set the criteria to 0.05. 

* Feature 1
* Feature 2
* Fix 1
* Fix 2

## TODO (if any)

to be done

* Feature 1 supports a, but not b.


shyuep:
Pls run Pycharm -> Reformat code first. Your code does not conform to Python PEP8.

PR #815, opened 2017-08-28
add self.ref_energy which is useful for combining neb plots

paulgholin:
add self.ref_energy which is useful for combining neb plots


shyuep:
I don't see a unit test.

PR #760, opened 2017-08-14
ENH: Implemented get_symmetrised_molecule + helper functions; Improved performance of generate_full_symmops

mcocdawc:
# Summary
Closes materialsproject/pymatgen#746

New public functions:
* ``PointGroupAnalyzer.symmetrize_molecule``
* ``PointGroupAnalyzer.get_equivalent_atoms``

Changes in public functions:
* ``analyzer.cluster_sites`` now supports an optional boolean keyword argument ``give_only_index``, which defaults to ``False`` (previous behaviour)
* ``analyze.generate_full_symmops`` was optimised numerically and does not use recursion anymore. In the case of icosahedral point groups (tested on fullerene) the new implementation is 72 times faster.

New private functions (helper functions for ``PointGroupAnalyzer.get_equivalent_atoms``)
* ``PointGroupAnalyzer._get_eq_sets``
* ``PointGroupAnalyzer._combine_eq_sets``

PR #817, opened 2017-08-29
MAINT: Numerical stability for symmetrise + finer control over tolerance

mcocdawc:
## Summary

Followup to #760 

* Improved numerical stability in the symmetrise code. Previously if *O(i, j)* denotes the the symmetry operation to map atom *i* unto *j* it could end up with a chain of  *O(3, 4) O(2, 3) O(1, 2)*  even if *O(1, 4)* was already available. Mathematically this does not matter, but in longer chains of matrix multiplications the errors could accumulate.

* Iterative symmetrize supports two keywords now. epsilon and tolerance. The tolerance is the same as in the other functions in the same module and epsilon is the maximum allowed difference between two subsequently symmetrized molecules. Before tolerance was used in both cases.

* A second test was added


shyuep:
Thanks. However, there seems to be a test error? You can click on the travis-ci above to see the error.

mcocdawc:
Thanks as well for accepting ;)

PR #818, opened 2017-08-31
Parsing energies from gen_scfman module in Qchem 5

wood-b:
## Summary

In Qchem 5 there is new scf module called gen_scfman, unfortunately the scf output is slightly different from the default module. Also, I added .eggs/ to the git ignore file. I saw a few egg related ignore so hopefully that is alright.

* parse energies from gen_scfman output
*unittest added

## TODO (if any)

There are other incompatibilities with the new output. I'm unsure how many exists but for instance the scf_iter_pattern no longer matches.

* find and fix other incompatibilities

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,6 @@ setuptools*
 .ipynb_checkpoints
 .cache
 .tox
-
+.eggs/
 gulptmp_4_1
 .coverage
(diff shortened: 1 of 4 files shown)


shyuep:
Thanks!

PR #819, opened 2017-09-01
Bug fix for Hubbard-U in Vasprun, and added basic vdW support

marcoesters:
## Summary

Fixed a bug for Hubbard calculations in Vasprun, and added basic van-der-Waals support.

* Hubbard: If LDAUJ is not specified in INCAR, Vasprun.as_dict() will raise a VaspParserError because the lengths of the LDAUU and LDAUJ lists are different. This problem has already been resolved in Vasprun.hubbards, but hasn't been ported to the as_dict() methods. An easy way to fix this is to just set d["hubbards"] = self.hubbards, which does the same thing as what the current code intends to do.

* vdW: VASP supports a lot of vdW functionals with different flags. I added a convenient way to add these parameters to an Incar object. For the vdW-DF functionals, VASP adds the information to the vasprun.xml file, so I updated Vasprun.run_type() to detect them. run_type now also detects PBEsol, and LDA calculations.

## TODO (if any)

* The way Vasprun checks if the calculation is DFT or DFT+U is not completely fool-proof. If a user has LDAUJ parameters in the INCAR file with less elements than in the LDAUU tag, VASP won't abort the calculation. Instead, it will do a standard DFT calculation but still write all the U parameters into vasprun.xml. The only place to see whether DFT or DFT+U is used is the OUTCAR file. Maybe there's a clever way to determine that and warn the user?
* Vasprun cannot read vdW functionals that use the IVDW tag because it's not written in the vasprun.xml file. Besides reading the OUTCAR file, I don't know how to determine the vdW functional.


shyuep:
Thanks! However, I don't think the use_vdw and del_vdw should eb within the Incar object. In general, the Incar etc. input files are generically defined to support any functional or calculation type. For actual settings, these should be implemented in the vasp input sets (in pymatgen.io.vasp.sets.py). Can you implement a VDWSet using your proposed paramters? Thanks.

marcoesters:
Yes, I can do that instead.

marcoesters:
I have another question before I implement this: van-der-Waals interactions can also make a lot of sense for some slab calculations, but implementing an additional VDWSlabSet class seems a little excessive.

I could instead add the kwarg vdw=None to the existing MVLSlabSet and add the parameters to slab_incar if vdw is not None. If you don't think that's a good idea, I will just implement the VDWSet class.

shyuep:
I would suggest this - given that vdw is a relatively basic parameter, you can add that to DictSet. Based on the vdw, you can modify the self._config_dict["INCAR"] to include the vdw settings.

Because all sets inherit from DictSet, the vdw = "XXX" will work on all input sets immediately without further mods.

marcoesters:
Okay, that should do it. Thank you for your patience, and please let me know if you want me to change anything else.

shyuep:
Great thanks!. I will merge and make a few minor changes. Generally the user_incar_settings should not be modified. Whatever is specified by the user is the paramount setting, regardless of the defaults.

PR #822, opened 2017-09-03
Kpath fix

katherinelatimer2013:
## Summary

Modifications to account for space group naming conventions changes in spglib

* Fix in pymatgen.symmetry.bandstructure.HighSymmKpath class
* Fix in pymatgen.symmetry.analyzer.SpacegroupAnalyzer class
* Tests for both modifications


shyuep:
Pls fix the failing tests.

gpetretto:
Sorry, I have found a few problems in this PR.

- Simply adding the test for the presence of "A" in the spg symbol does not guarantee to give a proper standard primitive cell for the face centered orthorhombic spacegroups:
```
from pymatgen import MPRester

s = MPRester().get_structure_by_material_id('mp-4587')

print len(s)
#out: 4

spga = SpacegroupAnalyzer(s, symprec=1e-2)

print spga.get_space_group_number()
# out: 38

print spga.get_space_group_symbol()
#out: Amm2

std_prim = spga.get_primitive_standard_structure()

print len(std_prim)
#out: 8   --> Not a primitive

print len(std_prim.get_primitive_structure())
#out: 4

```

- Is there any reason to remove all the comments and warnings indicating that a primitive standard structure should be passed to the HighSymmKpath object? It doesn't seem that this PR changes this point.

- According to its docstring the method _symmetrize_molecule_ should return a dict with a specific set of keys. However a check has been added so that the method may return a list of other objects instead.

shyamd:
@gpetretto 
I think @katherinelatimer2013  forgot to rebase before she made this PR. I'll help her fix that.  I don't believe she actually changed most of the warnings, etc. The primary purpose of this PR was to fix the 'C'/'A' issue and add in unit tests so that we are not blind-sided in the future, while she works on a more general method of finding the HighSymmKpath based on the spacegroup operations.

katherinelatimer2013:
Sorry for the errors; I am going to close this and resubmit after fixing my local configuration.

gpetretto:
Thanks. I will expand a bit on the previous comment, hoping that this can be of some help. In case you are going to add the unit tests for the primitive structures as well, pay attention that both the current implementation and the one checking on the presence of the 'A' in the symbol will likely give you the wrong structure. The spacegroups 38 to 41 are all affected from this issue, as now in spglib the preferred notation is A-centered rather than C-centered for these spacegroups. (see https://arxiv.org/pdf/1506.01455.pdf)

katherinelatimer2013:
@gpetretto  Yes; the confusion arose because I didn't realize that the call to "get_refined_structure" in "get_conventional_standard_structure" uses the spglib convention (which you reference). I will make that fix, which should be sufficient for everything else for now as "get_primitive_standard_structure" and the initializer for HighSymmKpath both call the get_conventional method.

I think the mixing of conventions could be confusing for general users, since there is not much explicit documentation that I've been able to find on the various MP sites regarding which one we are using for which purposes...

PR #826, opened 2017-09-05
resolving two small issues with periodic table data

albalu:
## Summary

both changes are very small:

* Electrical resistivity for Te is 1000010<sup>-8</sup> in periodic_table.json but it should be 10000 10<sup>-8</sup> equal to 1e-4 ohm.m consistent with here ( http://periodictable.com/Elements/052/data.html ) for example

* to have every Element property as float, I set the "&gt;" to its lower bound.


shyuep:
Pls fix the values not in the json, but rather the yaml version in pymatgen/dev. That is supposed to be "base data" that we edit on. We then convert that yaml to json using the scipt in pymatgen/dev.

albalu:
ok I just changed this value inside pymatgen/dev_scripts/periodic_table.yaml

ardunn:
@shyuep The script does not seem to fix the issue when working with `Element`, which uses the `pymatgen/core/periodic_table.json`. The file in `pymatgen/dev_scripts` is updated but the one in `pymatgen/core` is not.

For example, both @albalu and I notice that `Element("Se").electrical_resistivity` returns `u'about 10000010<sup>-8</sup> &Omega; m'`, not `1e-08`.

Should we change `pymatgen/core/periodic_table.json` file or the script?

shyuep:
You need to copy the generated json to the core folder. I just did that.

Issue #827, opened 2017-09-05
Possible bug in local_env.py or PeriodicSite/Lattice, incorrect jimage?


mkhorton:
## System

* Pymatgen version: 2017.9.1 (development)
* Python version: 3.6.0
* OS version: macOS 10.12.6

## Summary

Would appreciate a sanity check here. I'm looking at a single layer of MoS2 from a .cif file ([attached](https://raw.githubusercontent.com/mkhorton/pymatgen/e807247f43153a71b6f3389120c623c3d7e13c25/test_files/critic2/MoS2.cif)), where site 0 is Mo, site 1 is S, site 2 is S. From a visual inspection in VESTA, I see the Mo site has the following neighbors in the periodic images specified:

0 -> 1, (0, 0, 0)
0 -> 1, (0, 1, 0)
0 -> 1, (-1, 0, 0)
0 -> 2, (0, 0, 0)
0 -> 2, (0, 1, 0)
0 -> 2, (-1, 0, 0)

However, using `MinimumDistanceNN()` (which is very straight-forward, most logic comes from `PeriodicSite`/`Lattice`), I get:

0 -> 1, (0, 0, 0)
0 -> 1, (0, 1, 0)
0 -> 1, (1, 1, 0)  # this line is different
0 -> 2, (0, 0, 0)
0 -> 2, (0, 1, 0)
0 -> 2, (1, 1, 0)  # this line is different

This seems wrong to me, but I'm having difficulty tracking down the bug, I get the same results with `MinimumOKeeffeNN` too ... @nisse3000 do you have any ideas?

## Example code

```python
from pymatgen.core import Structure
from pymatgen.analysis.local_env import MinimumDistanceNN
s = Structure.from_file('MoS2.cif')
nn = MinimumDistanceNN()
nn.get_nn_info(s, 0)  # neighbors of Mo site
```
## Files (if any)

MoS2 CIF file [here](https://raw.githubusercontent.com/mkhorton/pymatgen/e807247f43153a71b6f3389120c623c3d7e13c25/test_files/critic2/MoS2.cif).

mkhorton:
A little bit closer to diagnosing this, but still don't understand what's going on here...

Using the same `lattice`* from the CIF file,

```
f1 = [0.3333, 0.6666, 0.21329]
f2 = [2/3., 1/3., 0.12356]
lattice.get_distance_and_image(f1, f2)  # returns (2.4169088271902091, array([0, 0, 0]))
```

Alternatively,

[...]

PR #835, opened 2017-09-11
Corrected image location computation in local_env.py

nisse3000:
I've changed the way the images locations are computed in local_env.py according to @mkhorton and @shyuep's comments in issue #827


shyuep:
Can we have a few unittests confirming the new definition pls? @mkhorton example of MoSi2 would be a good one.

PR #839, opened 2017-09-18
Gibbs free energy of a material with respect to Pourbaix stable domains

as2362:
Added functions related to computation of Gibbs free energy of a material with respect to Pourbaix stable domains. 

Some revisions and simplification to existing routines.


shyuep:
I am not sure the __eq__ test for phase diagrams is a good idea. In fact, I am not even sure tehre should be an __eq__ implementation for a PD. When would such a function be used?

montoyjh:
It's not for the pd itself, it's just for a single PDEntry.  We use it primarily to test if a list of entries contains a specific entry (e. g. `if entry in list_of_entries`), which is an integral part of the multi-entry processing that's in the current pourbaix diagram implementation.  We could isolate it to the PourbaixEntry object, if it's a problem for the general case.

shyuep:
The problem here is that the current `__eq__` makes no sense to me even for PDEntry. Generally, any sort or search algo computes the hash first. Here, your hash is the id. The #1 rule of thumb in programming is that all equal objects must have the same hash, though objects with the same hash need not be equal. Your code violates the construct used in things like sets and dicts. For example,

```
class Val:

    def __init__(self, val):
        self.val = val

    def __eq__(self, other):
        return self.val == other.val

    def __hash__(self):
        return id(self)

    def __repr__(self):
        return "Val: %s" % self.val

a1 = Val(1)
b1 = Val(1)

print(id(a1))
print(id(b1))

print(a1==b1)
print(set([a1, b1]))
```

The final set contains two items, not one. That is wrong behavior. Similarly, if you use the PDEntry in a set or dict, it will lead to very strange behavior.

montoyjh:
Ah, I understand the problem now.  I've just tested it after deleting the new eq and hash and it doesn't seem to break the new unittests, so we can just scrap it.  I'm not sure why I thought it was necessary in the first place.

PR #841, opened 2017-09-18
use significant_digits in poscar lattice str

ExpHP:
## Summary

* Fix formatting of lattice in `Poscar.get_string` to respect the `significant_digits` argument


shyuep:
Thanks.

PR #848, opened 2017-09-22
add combine_neb_plots method

paulgholin:
add combine_neb_plots method.
Due to this method:
move test_files/neb_analysis/neb1/start/OUTCAR to test_files/neb_analysis/neb1/neb/00/
move test_files/neb_analysis/neb1/end/OUTCAR to test_files/neb_analysis/neb1/neb/04/
Finally, remove test_files/neb_analysis/neb1/start and test_files/neb_analysis/neb1/end


pymatgen/analysis/transition_state.py
```diff
@@ -303,3 +305,92 @@ def as_dict(self):
...
+                neb1_energies = neb1_energies[0:len(neb1_energies) - 1] \
+                                + [(neb1_energies[-1] + neb2_energies[0]) / 2] \
+                                + neb2_energies[
+                                                                  1:]
+                neb1_structures = neb1_structures + neb2.structures[1:]
+                neb1_forces = list(neb1_forces) + list(neb2.forces)[1:]
+                neb1_r = list(neb1_r) + [i + neb1_r[-1] for i in
+                                         list(neb2.r)[1:]]
+
+        if (x.fit(neb1_start_s, neb2_start_s) == True
```
shyuep: Why does this need to be == True? Why not just x.fit(..)? Does the == mean anything? 

Same with the ones that are == False.

This is really bad programming practice. The if statement is already a True/False test. There is no need to == True/False.

pymatgen/analysis/transition_state.py
```diff
@@ -303,3 +305,92 @@ def as_dict(self):
                 'energies': jsanitize(self.energies),
                 'forces': jsanitize(self.forces),
                 'structures': [s.as_dict() for s in self.structures]}
+
+def combine_neb_plots(root_dirs, arranged_dirs=False, reverse_plot=False):
```
shyuep: This method should not take in a list of rootdirs. It should simply take in a list of NEBAnalysis. Why assume that people are doing the analysis from a set of files?

shyuep:
When is this going to be done?

paulgholin:
This might be merged. The error of ModuleNotFoundError: No module named 'pip._vendor.requests.adapters' seems not related to this code.

PR #851, opened 2017-09-26
Added many more OPs and made normalization procedure more robust

nisse3000:
Added order parameters for:
  - single-bond case
  - trigonal planar
  - square planar
  - pentagonal planar
  - square pyramid (previous one (sq_pyr_legacy) recognizes any pyramidal motif with neighbors on the basal plane being on a circle)
  - trigonal pyramid
  - pentagonal pyramid
  - hexagonal pyramid
  - T-shape
  - cuboctahedron
  - see-saw

Also changed normalizations to a numerical procedure via a counter because the trigonal bipyramid OP could become >1 before that (which is not desirable).

Rewrote parameter assignment section (more concise now).


shyuep:
At this juncture, I think you need to rethink your approach. Basically, there is far too many permutations of order parameters to just hard code them into if statements. I suggest storing the parameters in some sensible json/yaml file, and the *short* code then uses the data from the file without the many if statements. @computron comment pls.

computron:
From casual inspection of the code I agree with @shyuep . It seems possible to:

- group together order parameters that have the same functional form, but different parameters, into a single code block. If the functional forms are similar but not exactly the same, it might still be possible to combine.
- The parameters can then be read from a file. I would suggest JSON (for speed). Also, make sure the file is read only ONCE and the data is cached. We don't want to be reading the parameter file every time the OrderParameter class is instantiated, just the first time. e.g., you can look at ``periodic_table.py`` which loads the data at the top of the module so that it's available upon import.

You should also use Google-style docstrings - i.e., clearly explain the parameters and their expected type (array, dict, etc) and also the return type. There is a lot of description of the order parameters themselves but not much "programmer" information about what kind of format is expected for the various parameters.

nisse3000:
@computron & @shyuep, I think I have addressed all your concerns regarding the OP class restructuring. The one failing build test doesn't seem to have to do anything with pymatgen-related things, right? Pull request "Molecular Orbitals Functionality #863" by @dyllamt seems to have the same issue.

shyuep:
Thanks!

PR #852, opened 2017-09-29
Reverted slab argument to structure in MVLSlabSet

montoyjh:
Can we revert the constructor syntax of MVLSlabSet such that it takes a "structure" kwarg instead of a "slab"?  My reasoning for this is twofold:

* Input structures to MVLSlabSet aren't always slabs, e.g. when doing reference bulk calculations, but they are always structures
* Having "slab" as the input argument name changes the structure of the dictionary resultant from as_dict() such that it's different from every other input set, which makes it unwieldy when trying to do some db operations or modifications on the serialized object, for example.


shyuep:
This is fine with me.

PR #860, opened 2017-10-05
Made adsorb_both_surfaces() into a class method and more flexible

richardtran415:
## Summary

This pull request removes the need for differentiating two surfaces when building an adsorbed slab with adsorbates on both sides. Therefore the use of COM in the AdsorbateSiteFinder() class as well as the option to adsorb at the bottom or top is also remove. The algorithm will rollback to Joseph Montoya's previous method as the new method for adsorb_both_surfaces() will not require any edits to the original algorithm.

* adsorb_both_surfaces() is now a class method of AdsorbateSiteFinder()


shyuep:
Pls do not keep adding to the same pull request. merge the current origin to your fork properly and submit NEW pull requests in future.

PR #869, opened 2017-10-23
Fix bug in POSCAR parsing

shyamd:
Bug in POSCAR parsing from file looking for POTCAR by name only
Misses extensions such as .gz .orig, .relax1.gz , etc. 
Fixed by looking for POTCAR in filename

### DIFF
--- a/pymatgen/io/vasp/inputs.py
+++ b/pymatgen/io/vasp/inputs.py
@@ -228,7 +228,7 @@ class Poscar(MSONable):
         names = None
         if check_for_POTCAR:
             for f in os.listdir(dirname):
-                if f == "POTCAR":
+                if "POTCAR" in f:
                     try:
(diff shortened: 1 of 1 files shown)


mkhorton:
This fix seems to be causing a lot of existing tests to fail? https://travis-ci.org/materialsproject/pymatgen/builds/291690908

The test file [here](https://github.com/materialsproject/pymatgen/blob/master/test_files/POSCAR) claims it's LiFePO4 but it's actually Fe4P4O16, I'm not sure if that's intentional or not, but I think that's where the failures are coming from.

shyuep:
This seems to be a Py3 vs Py2 issue. The Py3 test has a failure that is not related to this change. But the Py2 test has a huge number of failures. I am reverting this pull requset. Pls fix it properly first.

PR #873, opened 2017-10-24
fix jhu issue

montoyjh:
Should fix JHU servlet issue.  Was using a static INCAR from the test files, rather than generating it on the fly, and JHU updated their system to parse magmoms for symmetry.


shyuep:
Thanks

PR #875, opened 2017-10-27
Feff io DirectoryInput

czhengsci:
## Summary

Add PostFEFFSet class to pymatgen.io.feff that allows post-processing of FEFF input files from an existing FEFF input files directory. Unittest files have been added accordingly


shyuep:
This is not a good name. SEe how VaspInput is defined.

czhengsci:
I renamed the class name to PostFEFFSet to make the name more specific about the class' function, similar to the class names defined in the VaspInput.

PR #877, opened 2017-10-27
surface_analysis.py update

richardtran415:
## Summary

Removed import of MPRester to calculate chemical potential range. Temporarily set chemical potential range for a nonstoichiometric slab as an optional parameter. A future release will handle the calculation of the chemical potential range automatically. Added class method to plot binding energy vs clean surface energy for class SurfaceEnergyPlotter(). Added more unit tests, specifically for the case of adsorption (credits to Joseph Montoya for providing the files). 

* gbind_vs_gamma() class method to plot binding energy vs clean surface energy
* class WulffShape() method get_plot() parameter color_set can now be the name for a color set or a dictionary of user specified rgb list for each facet (hkl).

## TODO (if any)
    -Only works for monatomic adsorbates
    -Only works for 1 reference element if we are dealing
        with non-stoichiometric slabs.
    -Need a method to automatically get chempot range when
        dealing with non-stoichiometric slabs
    -Simplify or clean the stable_u_range_dict() method
        for the SurfaceEnergyPlotter class
    -Combine chempot_vs_gamma_clean(), chempot_vs_gamma_plot_one()
        and chempot_vs_gamma_facet() into one class method.
    -Annotations for which configuration is on a plot


shyuep:
I said this before - merge the origin with your fork and submit a new pull request. This pull request contains all the code from previous pull requests.

richardtran415:
Will do, sorry about that

PR #880, opened 2017-11-01
Order-disorder transformation tools

computron:
## Summary

Adds some new features for order-disorder transformation

* A new DiscretizeOccupanciesTransformation that will discretize the site occupancies in a structure, e.g., 0.62, 0.19, 0.19 to 0.6, 0.2, 0.2 - according to tunable tolerances.
* Test for above
* A new option for EnumerateStructureTransformation that lets you set max_disordered_sites instead of max_cell_size. The former can iterate over multiple max_cell_size values until either finding a solution or hitting the max number of disordered sites.
* Test for above

None

### DIFF
--- a/pymatgen/transformations/advanced_transformations.py
+++ b/pymatgen/transformations/advanced_transformations.py
@@ -242,6 +242,70 @@ class MultipleSubstitutionTransformation(object):
         return True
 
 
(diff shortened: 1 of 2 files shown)


computron:
Should wait for tests ... for some reason py2 tests were failing when I created the branch - even before I made any changes. Could not figure out the reason for this.

PR #882, opened 2017-11-02
Fix a bug in is_element in computed entry

shyamd:
This bug causes problems in evaluating if an elemental composition is an element. 
Fixed by switching to the logic used in PDEntry.


shyuep:
Can you also add a test to ensure this doesn't happen again.

Issue #886, opened 2017-11-05
problem when reading magnetic cif files


cyz11:
I had problems when reading a magnetic cif file generated by ISOCIF.
It seems the 'parse_magmoms' method in cif.py have some problems.

First, the keys in data
'_atom_site_moment_crystalaxis_x'
'_atom_site_moment_crystalaxis_y'
'_atom_site_moment_crystalaxis_z'
'_atom_site_moment_label'
should be changed into
'_atom_site_moment.crystalaxis_x'
'_atom_site_moment.crystalaxis_y'
'_atom_site_moment.crystalaxis_z'
'_atom_site_moment.label'
to support the new standard magnetic cif files.

Secondly, it parse_magmom gets the magnetic moments in Cartesian coordinates, which turns out to be a problem in _get_structure method. Because the symmetry operation in _get_structure is done in relative units, using Cartesian coordinates first will give the wrong moment for symmetry related sites. I think the correct order is first do the symmetry operation to get all the sites and then transfer into Cartesian coordinates, or you can just leave it as relative units.

Hope the bugs could be fixed soon.

shyuep:
@mkhorton Pls handle this. Thanks.

mkhorton:
Will do, thanks for the report.

@cyz11 do you have a test file to share please?

mkhorton:
Regarding the first issue, it should work regardless of if an underscore or period is used as the delimiter -- this was because the magCIF standard changed during development, but there are already files out there in the wild using the older format. However, there may be a bug, so I'll check up on this.

Regarding the co-ordinate system, we want them in Cartesian in the generated structure, since I believe this is the format VASP expects for non-collinear calculations, though I think relative would make more sense generally. We could well be changing co-ordinate systems at the wrong point however, so I'll investigate -- again, any specific example is very helpful.

mkhorton:
Ok, the first issue should be fixed. The best reference I have for magCIF right now is http://comcifs.github.io/magCIF.dic.html -- this claims that both `_atom_site_moment.crystalaxis_x` and `_atom_site_moment_crystalaxis_x` are correct (the latter is an alias). I've updated the parser to support both.

cyz11:
I have attached a cif file I am working on. It report an error for the first issue if I use get_structures. Even if I fixed the first issue, I found it did not give the correct magnetic moment for symmetry related points(You can check it with the image in VESTA).
Thanks!

[...]

PR #887, opened 2017-11-06
Adds aliases for common keys in magCIF standard

mkhorton:
See http://comcifs.github.io/magCIF.dic.html for more information

Fixes issue 1 of 2 in #886


shyuep:
Pls add tests for the new fixes.

mkhorton:
Will add the test file from @cyz11

mkhorton:
Both issues should be fixed now, tests added.

PR #888, opened 2017-11-07
move two functions to BandStructure instead of BandStructureSymmLine

albalu:
I mistakenly defined get_sym_eq_kpoints and get_kpoint_degeneracy under BandStructureSymmLine but they work with any BandStructure object so I moved them under BandStructure.


albalu:
If this is merged, could you please release a new version of pymatgen so that I can use this updated feature in matminer?

PR #890, opened 2017-11-08
LAMMPS data rewritten

### LINKED ISSUE #865: Improvements needed for io.lammps.data
A few issues with the current implementation I would like to discuss with whoever involved in this module.

* A list of lists of floats is now used as the data structure for each long section like "atoms", "bonds", etc. Since each column has its specific meaning, can we store the data using a list of dicts instead? Though this means more efforts in refactoring the parser, it will make formatting (e.g., float rounding) a lot easier. The current code after last time fix will output integers with 6 digits after the decimal point as well, which is ridiculous.
* With the dict setting above, the "atoms" section can be formatted according to an external variable `atom_style`.
* Is it possible to merge the two objects? The parent class `LammpsData` is a special case where only "atoms" is required.

adengz:
## Summary

`io.lammps.data` has been completely rewritten to fix the formatting issue. Closes #865 

* `LammpsForceFieldData` has been merged into `LammpsData`, with `force_field` and `topology` as optional args.
* Data in each line is now stored in a dict rather than a plain list in order to distinguish floats from intergers. Will be of great help for handling a mixture of data, especially in Atoms section. 
* For the format of Atoms section, `atom_style` can be set to control the output or the way a file is parsed. Implementing a new supported style should be easy-peasy.
* Code tuning box for either molecule or structure has been completely REMOVED.
* Moved `ForceField` and `Topology` into `io.lammps.data` as they only serve `LammpsData`.
* `Topology.from_molecule` is now `Topology.from_bonding`. Angle and dihedral searching in the latter method is index-based, 10x faster than before. Also fixed an incomplete bug in dihedral searching.

No

## TODO (if any)

to be done

* (optional) It would be even more beneficial to use `pandas.DataFrame` for 2d arrays in a LAMMPS data file. This would make breaking down `LammpsData` to `ForceField` and `Topology` much more efficient (code with similar spirit in `io.lammps.output.LammpsRun`). The price to pay is introducing pandas as a new dependency.


shyuep:
I will merge now. BUut go ahead and introduce pandas as a dependency. I foresee we will be doing so soon enough given that there are clear advantages to using pandas for many types of analysis.

PR #892, opened 2017-11-08
Add PostFEFFSet

czhengsci:
## Summary

Add PostFEFFSet used for post-processing of FEFF input from a directory contains FEFF input files. Useful when existing FEFF input needs some adjustment.


shyuep:
This is truly a nonsensical class name and approach. It even violates what an "InputSet" as defined by pymatgen is. An InputSet is not a collection of input; it is a collection of RULES that create input. 

Do you see the equivalent class in pymatgen.io.vasp.sets? Of course not. If you just want to collect the input files, you use the pymatgen.io.vasp.inputs.VaspInput class. Only if you have new rules (e.g., to generate a BandStructure set of inputs) that you have a MPBSSet with corresponding from_prev_calc.

I am rejecting this PR. Rethink the approach.

PR #893, opened 2017-11-10
DiscretizeOccupanciesTransformation: fix_denominator

josuav1:
added fix_denominator (bool) to DiscretizeOccupanciesTransformation: 

if True, will enforce a common denominator for all species. This prevents a mix of denominators (for example, 1/3, 1/4) that might require large cell sizes to perform an enumeration.

added Test; successful


pymatgen/transformations/standard_transformations.py
```diff
@@ -696,13 +696,18 @@ def __init__(self, max_denominator=5, tol=0.25):
         self.max_denominator = max_denominator
         self.tol = tol
 
-    def apply_transformation(self, structure):
+    def apply_transformation(self, structure,fix_denominator=False):
```
shyuep: This breaks the API for apply_transformation, and violates PEP8. If anything, this parameter should be part of the init.

pymatgen/transformations/standard_transformations.py
```diff
@@ -716,7 +721,9 @@ def apply_transformation(self, structure):
                 old_occ = sp[k]
                 new_occ = float(
                     Fraction(old_occ).limit_denominator(self.max_denominator))
-                if round(abs(old_occ - new_occ), 6) > (
+                if fix_denominator: 
+                        new_occ = round(old_occ*self.max_denominator)/self.max_denominator
```
shyuep: This is a bad implementation. What if the rounding is out of tolerance and fix_denominator is true? That will round the occupancies no matter what and by pass the check in the elif statement.

pymatgen/transformations/standard_transformations.py
josuav1: Thanks for the comment, fixed that.

pymatgen/transformations/standard_transformations.py
josuav1: It is unlikely someone wants to use a fixed_denominator only within a given tolerance, but I changed this now so there is no bypass of the tolerance anymore. Hope I fixed the bad implementation that way.

shyuep:
Thanks for the fixes.

shyuep:
There are failed tests on Py2 or 3. Pls fix.

https://circleci.com/gh/materialsproject/pymatgen/7263?utm_campaign=build-failed&utm_medium=email&utm_source=notification

PR #898, opened 2017-11-13
Make the conversion battery module useful for multivalent cations

miaoliu:
Applied the same treatment as we did for insertion electrodes. When the ion oxidation state is 2+, the voltage reduced to half, and the capacity multiplied by 2.

## Summary

* Previously, the code works for LIB. However, it ignores the valence state of multivalent cations such as Mg and Ca, and treat it like Li.  The fixed code, take the oxidation state of the working ion into account. 

* N/A

## TODO (if any)

* someone may add nosetest for it


shyuep:
Pls add a test first. I do not merge things without tests.

miaoliu:
I will close this pull request and send off a new pull request with unittest.

PR #902, opened 2017-11-15
add timeout option to BoltztrapRunner

albalu:
add timeout option to BoltztrapRunner (default 7200 seconds) to avoid infinite loop and wasting CPU time.


shyuep:
This PR contains too many unrelated commits because you did not sync your fork with the origin. Pls do so and resubmit a new PR.

PR #920, opened 2017-12-01
added more unittests for reconstruction

richardtran415:
## Summary

* Added option to include reconstructions when using generate_all_slabs
* Slab class has property called is_reconstructed. Can be a bool or a str (i.e. the name of the reconstruction).
* Made sure reconstructions that inherit from another base_reconstruction will properly be built.
* Added unittest for generate_all_slabs to make sure it includes existing reconstructions.


shyuep:
I don't think having is_reconstructed having dual meanings of boolean and string is a good idea. I think it is better to have a property called "reconstruction", which is a string. If it is None, it means it is not reconstructed.

richardtran415:
Understood, I will make this change in my next pull request instead.

PR #923, opened 2017-12-01
fix voltage sign for anion voltage profile

HanmeiTang:
## Summary

Fixed the sign of calculated voltage of `InsertionVoltagePair` when anion is intercalated.

* Fix the wrong sign of the calculated voltage (revised script, line 431)
* Minor formatting.


shyuep:
Pls add a test using a hypothetical compound.

PR #928, opened 2017-12-05
Crystal Orbital Hamilton Populations (COHPs)

marcoesters:
## Summary

This pull request introduces a new module for Crystal Orbital Hamilton Populations (COHPs). It can also be applied to Crystal Orbital Overlap Populations (COOPs). The objects are roughly modeled after DOS objects in order to be more intuitive for pymatgen users. It also creates modules to read the output files of the two codes that I used to calculate them (Stuttgart LMTO-ASA and LOBSTER).

* **electronic_structure.cohp** introduces a Cohp and a CompleteCohp object, similar to Dos and CompleteDos. The big differences are that the Cohp objects also include the integrated populations, and that CompleteCohp objects can be directly loaded from an output file. Cohp objects can also be used for COOPs via the Boolean attribute cls.are_coops
* **electronic_structure.plotter** got an additional CohpPlotter class, which works similar to a DosPlotter object. The main difference is that COHPs can be positive and negative, so Spin.up and Spin.down are plotted using solid and dashed lines, respectively. Additionally, it is common to plot -COHPs instead of  COHPs, so the plotter needs to be fairly flexible. The user can adjust all that with Booleans.
* **io.lobster** contains a Cohpcar and a Icohplist class that read COHPCAR.lobster and ICOHPLIST.lobster files, respectively (and of course COOPCAR.lobster and ICOOPLIST.lobster files).
* **io.lmto** contains the classes LMTOCtrl, which handles the CTRL files for the LMTO code. It currently only contains the tokens for the structure, so it is mostly used to read the structure for a CompleteCohp object. It can also be used to write a new bare-bones CTRL file, but see caveat in the TODO! It also contains a class (LMTOCopl) to read a COPL file, which contains all the COHP data.
* **core.structure** has been updated to recognize LMTO CTRL files and read the structure.
* Of course, I also added unit tests and plenty of test files. Since this is a new module, I may have gone a little overboard though. The tests files be found under test_files/cohp.

## TODO (if any)

* Orbital-resolved COHPs are not supported yet. I will implement those when I have more time.
* Stacked COHP plots are not implemented yet.
* The COOP files in the Stuttgart LMTO-ASA code are written in binary format, so I haven't implented COOPs for LMTO yet.
* This one is lower priority: the CTRL files written with the LMTOCtrl object aren't always interpreted correctly by the LMTO-ASA code, mainly because LMTO isn't particularly tolerant when it comes to the precision of atomic positions. For example, even though the CTRL file contains the correct structure, LMTO may assign the wrong spacegroup. Fixing this requires more knowledge of the LMTO code and its symmetrizers than I currently have, so use that feature with caution. Reading CTRL files works reliably though.

### Additional comment

I think the unit tests I wrote for the CohpPlotter object could be easily applied to other plotter objects. That way, we'd have tests that are a little more solid than the current bare-bones tests.


shyuep:
Thanks! There seems to be some failing tests (you just need to check the Travis-CI one). It is related to the plotter tests, and seems probably like some form of a numerical or dict sorting issue. Can you fix them pls? Thanks.

marcoesters:
I found the issues with the code and solved them. The AppVeyor test failures are not because of my code.

The biggest issues were incompatibilities between python2 and python3 (I forgot to test with python3 - lesson learned). For example, the error with the keys are not because of sorting but because I needed to convert the keys into a list whereas python2 lets me get away without doing that.

There were also some numerical issues with the LMTO COHPs, so I needed to make the test a little more detailed. Maybe we (or I) need to add a "assertDictAlmostEqual" to the PymatgenTest class?

PR #930, opened 2017-12-06
Bug fix for DiffusionAnalyzer with NPT runs

bjmorgan:
The `DiffusionAnalyzer` `from_structures()` method calculates cartesian displacements using

    c_disp = [np.dot(d, m) for d, m in zip(f_disp, l)]

`f_disp` has dimensions `[site, time step, axis]`.
`l` has dimensions `[time step, axis, axis]`

This means the lattice used for calculating the cartesian displacements is selected based on site number, not on time step.

Displacements calculated for NPT simulations therefore contain errors (equivalent to assuming NVT, and then taking a different lattice for each site).

This commit fixes this, by replacing the calculation of `c_disp` with

    c_disp = []
    for i in f_disp:
        c_disp.append([np.dot(d, m) for d, m in zip(i, l[1:])])

The new calculation also addresses an off-by-one error, where the displacement at time step t was calculated using the lattice at time step t-1.

This commit includes a test to check `self.disp` is correct after initialising using `from_structures()`.

### DIFF
--- a/pymatgen/analysis/diffusion_analyzer.py
+++ b/pymatgen/analysis/diffusion_analyzer.py
@@ -602,7 +602,9 @@ class DiffusionAnalyzer(MSONable):
         dp = p[:, 1:] - p[:, :-1]
         dp = dp - np.round(dp)
         f_disp = np.cumsum(dp, axis=1)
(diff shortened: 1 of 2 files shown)


shyuep:
THanks.

PR #938, opened 2017-12-09
SCAN input set

shyamd:
This a simple SCAN input set. All potentials have been benchmarked to work with SCAN. The key here is the default to PBE_52 potcars. I'm not sure if this is the best way to go about this?


shyuep:
Erm, what's the difference with the MVLSCANSet?

shyamd:
What MVLSCANSet? I don't see it.

shyamd:
NVM. found it. That works too!

shyamd:
Two quick questions. I'm not a big fan of the ISYM, EDIFFG, nor NELM parameters. I'm running SCAN for MP using ISYM on, EDIFFG ~ -0.05, and NELEM at default of 100 with decent success. Can i just go ahead and change these parameters or do you want a separate MP input inherited from MPRelaxSet?

shyuep:
You can remove ISYM. EDIFFG is necessary, and I am fine with changing it to 0.05. I have no view on NELEM. Go ahead and change those.

shyuep:
ISYM = 0  is default anyway if you don't specify anything. @HanmeiTang

HanmeiTang:
Will do.

Issue #940, opened 2017-12-12
density-density and velocity-velocity dielectric functions in VASP 5.4.4


bjmorgan:
VASP 5.4.4 calculates the frequency dependent dielectric response in two ways, using a "density-density" response function, and using a "velocity-velocity" response function (see the comments in `linear_optics.F`. The latter is marked in `OUTCAR` as `current-current (2nd set in vasprun.xml)`, but is not labelled in `vasprun.xml`.

These two dielectric functions can have different imaginary components (in my tests, this appears as omega → 0.)

The present XML parser for `Vasprun` objects overwrites the first set of data with the second.

If both data sets are present in `vasprun.xml`, should these both be stored in the `Vasprun` object?

shyuep:
YEs, these should be stored in the Vasprun object.

PR #946, opened 2017-12-14
Dielectric 5.4.4

bjmorgan:
Addresses issue https://github.com/materialsproject/pymatgen/issues/940.

VASP 5.4.4 calculates the frequency dependent dielectric response in two ways, using a "density-density" response function, and using a "velocity-velocity" response function (see the comments in `linear_optics.F` in the vasp source code). The latter is marked in OUTCAR as `current-current (2nd set in vasprun.xml)`, but is not labelled in `vasprun.xml`.

This commit addresses an issue where the second unlabelled dielectric function would overwrite the first.

For consistency with older versions of vasp, the "density-density" linear response function
is read into `self.dielectric`.
The "velocity-velocity" linear response function (described as "current-current" in OUTCAR)
is read into `self.other_dielectric['current-current']`.

This commit includes a test that if `vasprun.xml` contains a second unlabelled set of dielectric
repsonse data, this is read into `self.other_dielectric['current-current']`.

If a third unlablled set of dielectric data is present in the `vasprun.xml`, a `NotImplementedError`
is raise (also tested).


shyuep:
I would suggest we be explicit and we keep all dielectric data in self.dielectric_data["density"] and self.dielectric_data["velocity"]. And then map self.dielectric to self.dielectric_data["density"].

bjmorgan:
That makes sense. What should be the behaviour for older VASP versions? Read directly into `self.dielectric` or into `self.dielectric_data[“density”]` and map?

shyuep:
Yes, old and new should have the same behavior. Only that newer VASP versions have an additional self.dielectric_data["velocity"]. That way, people can write code for both VASP versions transparently.

mkhorton:
Hi @bjmorgan, this is useful, thank you! I just came across this myself. However I read the `linear_optics.F` comment and was none-the-wiser. I read that both methods should be equivalent and agree, but read elsewhere they can differ practically and especially in the low frequency limit. Do you have a reference on the difference between these, and when one is preferred over the other?

bjmorgan:
@mkhorton I don't think the discussion with @shyuep was that these two methods in VASP should give the same output when processed. Rather that `pymatgen` should give the same result for VASP versions pre and post 5.4.4. when `self.dielectric` is used (which now is a mapping of `self.dielectric_data["density"]`).

As to the difference in the physics of how these two are calculated, I am not able to help, and it has been a while since I looked into this.

mkhorton:
No worries, but thanks for the context!

PR #948, opened 2017-12-14
add plot_element_profile method in PDPlotter

Zhuoying:
## Summary

* Add a method in PDPlotter to draw plot using data from Phasediagram.get_element_profile
* Unittest added


shyuep:
Write a proper PR summary.

Zhuoying:
@shyuep Done with the summary editing

shyuep:
This is not good enough. Why do you make people supply an evolution instead of the same parameters as pd.get_element_profile and just call that method to get the evolution? How does anyone know what "evolution" means? Why are the profile steps not labelled at all? Surely this is not what we publish in our papers?

You have been in the group for several years. Surely you understand the importance of proper functional argument and code design. Fix all these problems by today. I need it for a demo next week.

PR #957, opened 2017-12-16
Update MAGMOM setting for Ln3+ ions

w6ye:
## Summary

* Feature 1 Update MAGMOM setting for Ln3+ ions, the values are calculated as
                    number of ground state unpaired electrons of Ln3+ * 1.5


pymatgen/io/vasp/VASPIncarBase.yaml
```diff
@@ -1,17 +1,31 @@
 MAGMOM:
+  Ce: 5
+  Ce3+: 1.5
```
shyuep: Why is Ce3+ 1.5? Shouldn't it be 1?

pymatgen/io/vasp/VASPIncarBase.yaml
```diff
@@ -1,17 +1,31 @@
...
   Co: 5
   Co3+: 0.6
   Co4+: 1
   Cr: 5
+  Dy3+: 7.5
+  Er3+: 4.5
```
shyuep: Same with all these. The MAGMOMS are not integer values.

shyuep:
And finally, why no unittests at all for these?

shyuep:
@w6ye It should not take this long to write a few tests and push them.

w6ye:
1. From VASP default is usually the experimental magnetic moment multiplied by 1.2 or 1.5. I followed it by multiplying the experimental ground state magnetic moment (No. of unpaired spin) by 1.5. I think it's fine to have non-integer MAGMOM
2. I have added test

shyuep:
No, use the actual integer values. There is no need to multiply by 1.2 or 1.5. Even if we wanted to do that, we will make it an operation that is applied across all MAGMOMs, not hard code it in the yaml file. Use some sense in writing code. The other magmom values do not have 1.2 or 1.5 x. Why would the Ln be treated differently?

w6ye:
Updated. Use the true number of unpaired spin instead of x1.5 for all Ln3+.

Issue #972, opened 2017-12-27
Charged Structures and Ewald Summation


WardLT:
I am unable to reproduce the Ewald summation result from `test_ewald.py`. For the same structure file, I get an energy of -1123 eV using LAMMPS, GULP, and [mint](http://github.com/materials/mint), yet the Ewald summation value listed in the unit test is -1116 eV/atom.  

Am I correct in assuming that this difference between pymatgen and these codes is the result of pymatgen not supporting charged cells? I am basing this assumption based on the warning listed in `total_energy`.

## System

* Pymatgen version: master
* Python version: 3.6
* OS version: Windows

## Summary

* Ewald Summation returns different results than other codes
* Documentation unclear about whether it supports charged structures

## Suggested solution

* Add explanation about charged structures and pymatgen in documentation

shyuep:
I am not sure how LAMMPS and GULP deals with charged structures. In pymatgen, the implementation using a jellum background to maintain neutrality. One way to check is to comment out the jellum interaction and sees if it leads to the same answer.

WardLT:
Are you sure that pymatgen uses a Jellium correction? I don't see it implemented in the code, but could certainly be mistaken. 

GULP uses a charged cell correction. When I add the same correction to EwaldSummation, I get the same answer as GULP and LAMMPS.

I will soon open a PR with the code that adds the charged-cell correction to EwaldSummation.

PR #973, opened 2017-12-28
Added charged cell correction to EwaldSummation

WardLT:
## Summary

Added charge cell correction to Ewald summation method. Ewald summation now exactly matches results for Ewald calculations from LAMMPS and GULP.

See discussion in #972 

None


WardLT:
I see that this fails some unit tests. Sorry for not catching it. I'll clean that up soon.

shyuep:
Thanks!

PR #974, opened 2017-12-28
Add charges to Composition based on oxidation state guesses

WardLT:
Added a function that returns a copy of a Composition decorated using oxidation states determined from `oxi_state_guesses`. Also refactored the `oxi_state_guesses` to move the computation of likely oxidation states to a private utility function so that the guessing logic could return both the average oxidation states for each element and the oxidation states for each site.

As part of this modification, the `oxi_state_guesses` function now returns a tuple of operations rather than a list but is otherwise unchanged. I could easily modify the code to still return a list, but decided it was unnecessary since both collections are iterable and addressable.

## Summary

* Added operation for adding guessed charges to Composition object

None

### DIFF
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -615,14 +615,88 @@ class Composition(collections.Hashable, collections.Mapping, MSONable):
                 composition is not charge balanced, an empty list is returned.
         """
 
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks.

PR #975, opened 2017-12-30
Some bug fixes for Gaussian

marcoesters:
## Summary

The current Gaussian module cannot handle multi-parameter keywords and multi-line route cards in input and output files (see the l-cysteine test files for examples). This pull request fixes these bugs and has some minor changes.

* Fixed the multi-parameter keyword issue. Multi-parameter keywords such as opt=(tight, maxcycles=20) are now of the form {"opt": {"tight": None, "maxcycles": 20}}.
* Fixed the multi-line issue for input and output files.
* Fixed another bug that raised a KeyError when there was no space between a z-matrix variable name and the equal sign.
* Added test files (l-cysteine.inp and l-cysteine.out) and unit tests for these cases.
* Implemented a way to recognize the title of a Gaussian run from its output file.
* The method to_input was changed to not force the user to write the returned object to file. If the user wants to write the input file, they can always do it with the write_file method of the returned input object. Also, the default title is not "restart" anymore, but the title of the output file.
* Added an underscore to all the helper functions and regex patterns that shouldn't be used by the user, but would otherwise appear as methods/attributes when e.g. using ipython.
* Established more consistency between the attribute names of input and output objects. E.g. the route card is "route_parameters" for GaussianInput and "route" for GaussianOutput, which is less intuitive to users than having the same names.
* Some PEP8 clean-up.

EDIT: After running codacy, I fixed an old issue where a dictionary was called "dict". There is one more old issue, but I'd have to go through a lot of code to figure out how to fix it. I'll do it in a later PR.

## TODO

This is only semi-related to this PR, but I would like to get feedback before coding. The GaussianInput has a method get_zmatrix(). However, I think the z-matrix is important enough that it should actually be a method for the molecule and not just limited to GaussianInput. I think the best way to do this would be to actually have a Zmatrix(MSONable) class to allow for greater flexibility. What would be the best place for it? core.structure, core.site, or a new module core.zmatrix?


marcoesters:
Would it also be possible to squash-merge so that my old commits don't keep appearing? Thanks!

shyuep:
Thanks. I actually do squash and merge. But you need to merge latest changes from origin before submitting a PR in order for the old commits not to appear.

As for Zmatrix, I agree. I suggest you start a core.molecule module and put it in there. I have wanted to separate MOlecule and Structure for some time. I can easily do this in a backwards compatible manner.

shyuep:
Also, pls look at the codacy review. There seems to be new issues introduced.

marcoesters:
I like the idea of creating a new Molecule class. I will get to that when I have time.

I may be wrong, but I think Codacy only flagged the issue as new because I modified the line during the  PEP8 clean-up. The issue it was complaining about is still present in the original code. I found a way to fix the issue, so Codacy should be happy now either way.

shyuep:
Oh you misunderstood. I am not asking to create a new molecule class. I am merely asking for you to create the zmatrix class in a molecule module. We will move the existing Molecule class there later.

marcoesters:
That's what I meant. I have no idea why I wrote "class" instead of "module"...

PR #977, opened 2018-01-02
Wavecar

mturiansky:
## Summary

This pull request adds support for reading WAVECAR output files from VASP. It is essentially a python version of [WaveTrans](https://www.andrew.cmu.edu/user/feenstra/wavetrans/) by Feenstra and Widom with some added functionality.

* adds a Wavecar class to pymatgen.io.vasp.outputs
* 100% test coverage of the added code

* N/A

## TODO (if any)

While I don't consider this a work in progress, the following may be desired (and I would be willing to look into upon request):

* spin-polarization support
* partial charge density plotting from the wavefunctions


shyuep:
Many thanks for your contributions!

PR #986, opened 2018-01-06
Avoid `inspect.getfullargspec` to maintain compatibility with py2.7 

gmatteo:


shyuep:
This PR contains far too many commits. Also, we are cutting Py2.7 support from 2018.

gmatteo:
The majority of the commits have been already merged in #961 
and indeed the changes introduced by this PR are minimal as can be seen in https://github.com/materialsproject/pymatgen/pull/986/files
Perhaps github reports 65 commits because you have reverted the merge in
4a6db8133d35f6ed2349d9ac3cc8baed31146853

For what concerns py2.7 support:
The PR address a regression introduced in 38122517b13e0e039254977d6cdaab132f8df451
as `inspect.getargspec` is deprecated in py3.6 while `inspect.getfullargspec` is not 
available in py2.7. 
So we have refactored the code to avoid using introspection at run time and 
we consider the new approach much safer.

 The pymatgen documentations explicitly states:

>From v2018.1.1, new features implemented in pymatgen no longer need to support Py2k (i.e.,  unittests do not need to pass Py2k testing), though existing features will still be Py2k > compatible.`

Here we are not adding a new feature, we are just refactoring a preexisting API that 
stopped working after the recent modifications.
This piece of code  is being used in production with py2.7 and py3.6 and we need to maintain
py2k compatibility for this existing feature.

shyuep:
Fine, I will merge this time. But note that we will not support Py2 going forward. This means that it is entirely possible that some features will be introduced in core classes that will break Py27 compatibility. I would suggest you move to Py3k. We have given ample notice since 2017 that we will be making this move. And in fact, future versions of numpy and scipy will be py3k only.

PR #994, opened 2018-01-14
Fix bug in task.reset

gmatteo:


shyuep:
Pls submit a PR that does not include past commits. it is impossible to know what is happening with this PR.

PR #998, opened 2018-01-17
Minor update to extract references from CIF files + test

mkhorton:
## Summary

Now uses pybtex, more robust. Updated test.

None (pybtex already requirement).

## TODO (if any)

None


pymatgen/io/cif.py
```diff
@@ -37,6 +37,8 @@
 from pymatgen.core.operations import MagSymmOp
 from pymatgen.symmetry.maggroups import MagneticSpaceGroup
 
+from pybtex.database import BibliographyData, Entry
+
```
dwinston: `pybtex` is technically not a core install requirement of pymatgen (see [setup.py](https://github.com/materialsproject/pymatgen/blob/d465f584b4cd3f85544bbd9d280250c085bec3a4/setup.py#L111)), so would want a try/except here or perhaps import within `get_bibtex_string`? But the CI tests pass, so I don't know...

mkhorton: Ah thanks. It's listed in the 'required dependencies' in the documentation (http://pymatgen.org/installation.html), I think because it's used by SNLs. Will defer to @shyuep  on this.

shyuep:
Pybtex is an optional req. I would prefer to keep it that way given that it is not very Py3 friendly.

mkhorton:
Ok, have updated the import accordingly.

PR #1007, opened 2018-01-25
Added several new OPs and optimized parameters involved in OP computation.

nisse3000:


shyuep:
Can you do a rebase before sending a PR? This contains commits from Aug.

shyuep:
Also, the tests failed.

nisse3000:
I did rebase. It's this weird github behavior that the entire PR history would pop up again.  We had this discussion before (in December) I think.

I know that the tests failed. I'm working on that.  It takes a little longer because I take the opportunity to clean up all the warnings that some classes would be shuffled aroudn with pymatgen >= 2018.

shyuep:
I will merge it this time. But I am not going to next time unless the commits properly tell me what is being changed. There are others who have submitted PRs and they don't seem to encounter this issue.

nisse3000:
@shyuep, I know that you can avoid that if you always get a new fork.  But is that really sensible that you'd require that people do that every time---when you really want people to participate and contribute in as easy a way as possible?  On the long run, I fear that we'd just make people walk away from pymatgen if we'd require them to always make new forks.  Don't you agree that that could hurt us as an active developer community that tries to make this library as attractive, because as open and welcoming to new people as possible?

shyuep:
I am not insisting on new forks. But I don't think that's the only way to sync the fork. I never had these issues on other repos that I contribute to (e.g., FireWorks). https://help.github.com/articles/syncing-a-fork/ almost always works for me.

I cannot believe your claim that creating new forks is the only way to do this. Surely Github cannot be that badly designed that the only alternative is for the lead maintainer to tolerate looking through > 100 and growing commits from every developer.

shyuep:
@nisse3000 I have ascertain the issues with why the commits are showing up. Essentially, it is a difference between using "Squash and merge" vs "Merge commits". For future PRs, I will do a merge commit and that should avoid duplicate commits. What I would suggest you do now is to do a trivial change (e.g., some minor doc fix), and resubmit a PR for that trivial change. I will then do a proper merge commit and that should resolve all these issues with the commit hisotry.

nisse3000:
@shyuep, OK.  Just made a test pull request.

PR #1008, opened 2018-01-25
Add timeout kwarg to enumlib + test

mkhorton:
## Summary

Adds a timeout kwarg to enumlib to automatically terminate if no enumerations are found within a specified length of time.

None

## TODO (if any)

None


mkhorton:
Py2 tests fail because TimeoutError is apparently new to Py3, but given the new Py2 deprecation policy I'll leave this as-is unless asked otherwise.

shyuep:
That's fine. We are not going to care about py2.

Issue #1002, opened 2018-01-19
Standalone function for d_hkl to define size of structures


richardtran415:
Is there a stand alone function that returns d_hkl given a structure and a Miller index? I know the class XRDPattern() has something that gets d_hkl, but I don't that can be used outside the class right? This would be useful for things like if we wanted a standard nlayer of atoms of slab thickness or gb thickness thats not defined by length scale where the number atoms would vary across different systems (e.g. a 10 Angstrom slab/gb of Cs would only have n layer of atoms while the same size slab/gb for Fe would have a larger number of layer of atoms.

shyuep:
Don't think there is right now. By all means implement it in the Slab class.

mkhorton:
A function to go from three PeriodicSites to output a (h,k,l) tuple would be nice to have too.

richardtran415:
@mkhorton  Just to clarify, do you mean something that would generate a plane with any 3 PeriodicSites and then return the hkl of that plane?

mkhorton:
@richardtran415 yes, that's exactly what I meant -- to my knowledge this isn't in pymatgen currently?

richardtran415:
@shyuep 
Actually I think a function for getting d_hkl might be more at home as a method of the Lattice class, its not standalone, but at least its accessible outside xrd.py, but I agree the function to describe slab thickness in terms of nlayers should be part of class Slab().

shyuep:
that's fine. Implement then.

PR #1027, opened 2018-02-01
Function to get Miller index from three sites

richardtran415:
New function in surface.py called miller_index_from_sites() that takes in the supercell_matrix and three Cartesian coordinates to return a Miller index associated with the plane formed by the three points (credits to @wenhaosun ). This concludes issue #1002

* Minor fixes to surface_analysis.py
* Added unittesting for miller_index_from_sites


mkhorton:
Thanks for adding `miller_index_from_sites`, will be useful!

PR #1028, opened 2018-02-01
Removed bug in _get_image function of NearNeighbor class.

nisse3000:


pymatgen/analysis/local_env.py
```diff
@@ -299,8 +299,15 @@ def get_nn_info(self, structure, n):
...
         gives lattice image from provided PeriodicSite."""
-        return [int(f) if f >= 0 else int(f - 1)
-                for f in frac_coords]
+        images = [0,0,0]
+        for j, f in enumerate(frac_coords):
+            if f >= 0:
+                images[j] = int(f)
+            else:
+                images[j] = int(f - 1)
+                if -f % 1 == 0:
```
mkhorton: Why is there a minus sign in this line? Doesn't `f % 1 == 0` == `-f % 1 == 0` ?

Maybe worth adding a comment here to explain what the bug is and how it was fixed. As I read the code, it looks like we're rounding towards the origin cell now when trying to calculate the image vector?

There may be more robust (technically correct) way of doing this which is to just to subtract the frac coords of the image site from the original site (which we can find using `_get_original_site()`),  which would give the image vector directly ... but if this works, let's leave it as it is for now.

mkhorton: Thanks for fixing this btw :)

nisse3000: Hey @mkhorton , welcome.  ;-)

The minus sign can be removed.  Found it's way there from other fix trials.

We WERE rounding towards the (0,0,0) point.  Now, we are rounding towards the left-side interval boundary of each image.  That is, 1.1 is located in image 1, value of 1 itself also, 0.9 is located in image 0, as is a value of 0, -0.1 is located in image -1, -1 also, but -1.1 is in image -2.  Right?  On the left interval side it's an inclusive boundary, whereas on the right interval side it's an open boundary. So, in mathematical terms it should be: [-2, -1[, [-1, 0[, [0, 1[, [1, 2[, etc.  Does that make sense and, more importantly, is that correct?

PR #1032, opened 2018-02-05
Minor fix in lammps.output

adengz:
## Summary

Fix `unique_atomic_masses` in two methods.

None

## TODO (if any)

This bug was reported in a user post in pmg google group. It was caused by rewriting `lammps.data`, but I would not take the blame because it would have been found if there were unit tests for those methods. So, if the developer/user is still using them, can you update your code and add comprehensive tests, please? Thank you.

### DIFF
--- a/pymatgen/io/lammps/output.py
+++ b/pymatgen/io/lammps/output.py
@@ -325,8 +325,7 @@ class LammpsRun(MSONable):
         end = (step + 1) * self.natoms
         mol_vector_structured = \
             self.trajectory[begin:end][self.mol_config[mol_id]][param]
-        new_shape = mol_vector_structured.shape + (-1,)
-        mol_vector = mol_vector_structured.view(np.float64).reshape(new_shape)
+        mol_vector = np.array(mol_vector_structured.tolist())
         return mol_vector.copy()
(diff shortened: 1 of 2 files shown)


shyuep:
Since you fixed the bug, then write a unittest for it so that it will not happen again.

adengz:
Only a few simple logic tests were added. Honestly it is the developer's responsibility to write his/her own unit test.

shyuep:
@adengz This is not the right attitude. Of course it is the developer's responsibility to write tests. At the same time, people sometimes miss things. The code base only improves when people give time to write new functionality, including fixing bugs and writing missing tests. If you fix a bug that was not found before, you help write the test to prevent it from recurring.

If you really want to be that calculative on the time spent, how about you evaluate how much you have benefited from the code of other developers of pymatgen?

adengz:
I think there is a misunderstanding. I am more than happy to fix something and ensure it never happens again. But for a piece of code with no test at all, it makes this kind of contribution difficult, and takes extra amount of time. 

This bug is not totally avoidable at the first time if there are failing tests at the beginning. Unfortunately, there are not.

PR #1039, opened 2018-02-12
Added a unit test.

nisse3000:


nisse3000:
@shyuep, if the checks pass successfully, can you release the next stable version soon?  Would be greatly appreciated because I'm currently changing/completing matminer site-fingerprint classes that depend however on the most recent changes to the LocalStructOrderParas classes and associated yaml files.

PR #1040, opened 2018-02-12
Implement Nscf task from external files with paral_kgb 

gmatteo:
## Summary

* NscfTask can now be initialized from external DEN files and executed in parallel with paral_kgb
* Added new methods to produce graphs in the DOT language for flows/works/tasks 
   (graphviz module is optional)


shyuep:
Pls remove assert.

PR #1057, opened 2018-03-12
An alternative matrix in Lattice.from_parameters

kmu:
## Summary

* An alternative Cartesian lattice matrix compatible with VESTA, GULP, and cif2cell.
* Useful for processing Cartesian trajectories of molecular simulations under periodic boundary conditions.
* The alternative matrix is off by default, but can pass most of the unittests.

Unittests are added to ensure:

* The conventional matrix cannot handle the xyz file exported by VESTA, while the alternative one can correctly interpret it.
* If `alpha == 90` and `beta == 90`, the alternative matrix and the conventional matrix are identical.


shyuep:
Thank you very much for the contribution.

PR #1075, opened 2018-03-22
Fix incar list_key parse error

HanmeiTang:
## Summary

This commit fixes incar write error and one example is shown below. This error is triggered if a structure contains elements in ["F", "O", "S"] but the most electro-negative element is not within the list.

Example:
```python
>>> from pymatgen.ext.matproj import MPRester
>>> from pymatgen.io.vasp.sets import MITRelaxSet
>>> rest = MPRester()
>>> SCl = rest.get_structures("SCl")[0]
>>> vis = MITRelaxSet(SCl)
>>> print(vis.incar)

TypeError
~/repos/pymatgen/pymatgen/io/vasp/sets.py in incar(self)
--> 341         has_u = hubbard_u and sum(incar['LDAUU']) > 0
TypeError: unsupported operand type(s) for +: 'int' and 'dict'
```

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -330,7 +330,7 @@ class DictSet(VaspInputSet):
                     # else, use fallback LDAU value if it exists
                     else:
                         incar[k] = [v.get(sym, 0)
-                                    if isinstance(v.get(sym, 0), float)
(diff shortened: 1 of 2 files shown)


shyuep:
Write a test.

HanmeiTang:
Unittests have been added.

PR #1076, opened 2018-03-22
Bug handling 0 during writing VolumetricData

frssp:
## Summary

* Fix the Bug handling 0 during writing VolumetricData

before fix
-..000000000EE+01
after fix 
0.00000000000E+01


shyuep:
Thanks.

PR #1082, opened 2018-03-27
surface_chempot_range_map()

richardtran415:
WorkFunctionAnalyzer now takes a structure, list of locpot along c and fermi energy. There is a from_files method still available to build the analyzer from paths to the CONTCAR, LOCPOT and OUTCAR instead.

ReconstructionGenerator will apply operations to all possible terminations instead of determining which one before hand.

For each reconstruction type, check exact structure is generated in unittest

There is no reason not to generate the LOCPOT at the end of each slab calculation to get the workfunction, as such LVTOT is set to True by default in MVLSlabSet

Fix the scaling of Wulff shapes with a user given effective radius in the NanoscaleStability class. Also added option to scale the formation energy of the particle with number of atoms.

A method to check if the work function is converged based on its electrostatic potential along the c direction of the slab is added to the WorkFunctionAnalysis class. 

Added a surface_chempot_range_map() method to the SurfaceEnergyPlotter. Provides a 2D map similar to that of the PDPlotter except the "phases" shown are different terminations, adsorption coverages, segregation etc. with stability determined by their surface energy. 

TODO:
-ReconstructionGenerator() with simpler methods of building reconstructions
-morphology_evolution_chempot_range_map for analyzing change in Wulff shape
-Make surface_chempot_range_map() more efficient by solving up to three system of eqns.


shyuep:
Pls resolve the codacy issues.

shyuep:
Fix failing tests.

PR #1084, opened 2018-03-29
 add scattering coefficient of deuterium

resnant:
## Summary

Add X-ray scattering coefficient of deuterium (D).

The value was copied from hydrogen (H).
This change enables us to calculate XRD of deuterium-substituted materials.


shyuep:
Thanks!

PR #1094, opened 2018-04-04
add bandgap option for FermiDos

albalu:
add bandgap option for FermiDos enabling scissoring of the gap + testing. No scissoring by default.


albalu:
Thanks @shyuep, Would it be possible for you to release a new pymatgen version that reflect these changes?

PR #1099, opened 2018-04-09
NMR Updates

shyamd:
## Summary

Refactored the NMR tensor and added a NMR InputSet for VASP

- Updated periodic table to have NMR quadrupolar moments for isotopes
- Added an MPNMRSet for chemical shielding and EFG calculations
- Fixed inheritance issues in from_prev for various InputSets
- Updating parsing of chemical shielding and electric field gradient tensors in Outcar
- Switch the NMRChemicalShiftNotation class to an actual tensor class
- Updated nomenclature to reflect the fact that VASP actually calculates chemical shielding even though it says shift in several places in its own output
- Added in parsing of EFG tensors and a class for them


shyuep:
Pls fix codacy issues and check the other failing tests.

shyamd:
I fixed all the issues related to the NMR code. For some reason Hanmei's fixes that i merged in are showing up as changes and Codacy is complaining about them. They are structural, e.g. the use of asserts and random, so I don't think changing them is a good idea.

PR #1100, opened 2018-04-10
Small bug in tetragonal conventional structure method when a = c

tom-wood:
## Summary
The `get_conventional_standard_structure` method currently unexpectedly switches a and c for tetragonal space groups if a = b = c. This is obviously not desirable and so has been removed, while still allowing for switching a and c if a != b = c (the original intention). 

None

## TODO (if any)
This only removes unexpected behaviour for a = b = c, but doesn't determine which is the conventional c axis in that case.


shyuep:
I would suggest you add a test to make sure this is done correctly.

PR #1104, opened 2018-04-12
CifParser updates

gpetretto:
## Summary

This PR mainly addresses the problem of having element symbols in capital letters, as in several cif files present in the COD database. This happens both in the _atom_site_type_symbol and in the _atom_site_label keys. Notice that in the latter case a simple str.title() does not solve the problem as the label sometimes contains additional letters.

List of changes:
* parsing of element symbols in capital letters
* added an `errors` attribute to CifParser. It stores all the warnings and errors issued during the parsing.
* explicitly require pybtex for functions requiring it, instead of crashing (is an optional dependence)
* tests

While this increases the number of cif files parsed in the COD database, this can maybe introduce some regressions. Any comment about possible issues with the current modification?


pymatgen/io/cif.py
```diff
@@ -79,6 +81,14 @@ def _get_cod_data():
     return _COD_DATA
 
 
+upper_case_elements = ['AC', 'AG', 'AL', 'AM', 'AR', 'AS', 'AT', 'AU', 'BA', 'BE', 'BI', 'BK', 'BR', 'CA',
+                       'CD', 'CE', 'CF', 'CL', 'CM', 'CO', 'CR', 'CS', 'CU', 'DY', 'ER', 'ES', 'EU', 'FE',
+                       'FM', 'FR', 'GA', 'GD', 'GE', 'HE', 'HF', 'HG', 'HO', 'IN', 'IR', 'KR', 'LA', 'LI',
+                       'LR', 'LU', 'MD', 'MG', 'MN', 'MO', 'NA', 'NB', 'ND', 'NE', 'NI', 'NO', 'NP', 'OS',
+                       'PA', 'PB', 'PD', 'PM', 'PO', 'PR', 'PT', 'PU', 'RA', 'RB', 'RE', 'RH', 'RN', 'RU',
+                       'SB', 'SC', 'SE', 'SI', 'SM', 'SN', 'SR', 'TA', 'TB', 'TC', 'TE', 'TH', 'TI', 'TL',
+                       'TM', 'XE', 'YB', 'ZN', 'ZR']
```
shyuep: Do we really need to specify upper case elements? Can't we simply check Element.is_valid_symbol(string.title())?

pymatgen/io/cif.py
```diff
@@ -533,6 +544,36 @@ def _sanitize_data(self, data):
...
+
+            for i, sym in enumerate(data.data['_atom_site_label']):
+                match = r.match(sym)
+                if match:
+                    new_sym = sym.replace(match.group(), match.group().title())
+                    msg = "label {} converted to {}".format(sym, new_sym)
+                    warnings.warn(msg)
+                    self.errors.append(msg)
+                    data.data['_atom_site_label'][i] = new_sym
+
```
shyuep: I really wouldn't add a whole section of if statements to deal with this. I would rather deal with such issues in the section that parses the element symbols and allow all caps symbols there, with a warning if needed.

pymatgen/io/cif.py
gpetretto: I agree that the listing of all the symbol is far from an elegant solution (notice that only two letters elements are liste here). Unfortunately there are cases that have labels like "MgT" (e.g. http://www.crystallography.net/cod/view.php?f=9003921.cif , but there are several similar examples). In such cases Element.is_valid_symbol(string.title()) will fail. 
Since in the _atom_site_label one can apparently find all sort of notations and symbols, I ultimately decided to go for the explicit matching with the elements. Any other suggestion is welcome.

shyuep: I think you should basically just stipulate that you parse the first two letters, and title case them. Of course, all these should come with a warning. I am just not a fan of writing this kind of specific conversions which is very fragile and not general.

gpetretto:
I refactored the parse_symbol function to handle the symbols in capital letters and removed the explicit usage of a list of elements. I modified the list of special cases to handle the possible problematic symbols I have encountered so far. This should comply with the requests.

I also moved the function as a method of the object so that it can be explicitly tested (is there any particular reason to keep it as a nested function?)

PR #1105, opened 2018-04-13
split Brunner algos into three classes

hillarypan:


shyuep:
I need tests for all of them.

shyuep:
Note that there are some conflicts. You need to do a merge and fix them first.

PR #1109, opened 2018-04-16
AFLOW crystal prototype matcher + DisorderOrderedTransformation + pretty printing for disordered formulae

mkhorton:
## Summary

* A class to match Structures against crystal prototypes from the 'AFLOW Library of Crystal Prototypes', http://aflowlib.org/CrystalDatabase/ -- the terms of use on their website is to cite them if using their encyclopedia so I've made sure to include citations in the docstrings, and also in the JSON file containing the prototype structures as SNLs.

* A `DisorderOrderedTransformation` to merge sites to create a disordered structure from an input ordered structure -- this is probably quite a niche transformation, but the intent here was to try and find a way to 'go back' to plausible disordered structures for the purposes of crystal prototype matching. This is useful if the disordered structure would structure match against a crystal prototype, while the ordered structure would not.

* A string util to create pretty formula like `Cu_{x}Au_{1-x} x=0.25` from a disordered Structure.

* None

## TODO (if any)

* Better matching of ordered crystal structures derived to their original disordered prototype.


shyuep:
I am not sure why we would want to bother. We already have our own prototypes and matchers. Is the AFLOW any better?

mkhorton:
I think the advantage of a formal prototype library like this is mostly pedagogical, e.g. it's useful to know if a structure is a "fluorite" or a "perovskite" (especially if it's a less common crystal structure you may be previously unfamiliar with, but which has a common mineral name that can be easily googled).

The problem is that many structures don't StructureMatch against a prototype from the AFLOW library anyway -- either because there's been a partial species substitution, or there's too much local distortion even with loose structure matching tolerances, or maybe it's a layered structure and the interlayer spacing differs by too much, so it isn't perfect. But when it does work it can be useful.

PR #1112, opened 2018-04-17
Implementation of a new NN method

computron:
## Summary

This PR implements a new NearNeighbor method to the list already available in local_env.py

The method is my own custom algorithm for determining near neighbor. Briefly, it is based on post-processing Voronoi solid angle weights with additional considerations for site distance as well as optional preference for bonds with higher electronegativity difference.

Without any knowledge or "fitting", this method beat all other local_env methods currently in pymatgen on a benchmark test set of compounds including metals, minerals, and porous structures by a factor of two or more. The test set, which is still being expanded, is available here:

https://github.com/hillarypan/MaterialsCoord

## TODO

The method is complete and usable. However, it is possible we might fine tune the method later on if we see any issues that we didn't see in the current round of benchmarks. There are interested parties in this code that want to test this iteration of the NearNeighbor algo within pymatgen at the present time to help in the benchmarking process for diverse compound sets.

### DIFF
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -2660,11 +2660,10 @@ class CrystalNN(NearNeighbors):
             for i in range(length):
                 cn = i+1
                 if cn not in cn_scores:
-                    cn_scores[cn] == 0
-                    cn_info[cn] == []
+                    cn_scores[cn] = 0
+                    cn_info[cn] = []
 
         return self.NNData(nn, cn_scores, cn_info)
-
 
     def get_cn(self, structure, n, use_weights=False):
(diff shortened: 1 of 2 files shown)


computron:
FYI there are failing tests but pretty sure they have nothing to do with this PR

computron:
the problem was the EnumlibTest timing out (unrelated to this PR)

Restarting the CircleCI test fixed it. Going to try rebuilding Travis as well.

Maybe you want to bump up the timeout in pymatgen.command_line.tests.test_enumlib_caller.EnumlibAdaptorTest#test_timeout? Not sure why it's so low.

PR #1115, opened 2018-04-19
Add new class for calculate neutron diffraction pattern

### LINKED ISSUE #1070: Neutron diffraction pattern calculator
Hi, now I research magnetic materials using neutron beam.
I am writing a program to calculate the neutron diffraction pattern based on pymatgen.analysis.diffraction.xrd, for the case of nuclear scattering.
I will prepare the code according to the contribution guidelines, so please consider merge it.
Do you have an interest?

resnant:
## Summary
Closes #1070 

Add powder neutron diffraction (ND) calculator based on analysis.diffraction.xrd.XRDCalculator.
The change of code is slight.
Main changes by using neutron instead of X-ray are as follows:
 1. Atomic scattering length is a constant.
 2. Polarization correction term of Lorentz factor is unnecessary.

### Fix peak indexing of forbidden reflection
The diffraction peak appearance is checked after group it by 2theta, in the present implementation of this code and XRDCalculator.
This means that a forbidden reflection seems to appear when the other reflection occurs at the same 2theta position. This behavior may confuse non-professional users.

### Make the plot function more generic
Now plot functions are implemented respectively in XRDCalculator and NDCalculator.
I think these should be integrated as a generic function because they are same except the name.


shyuep:
Many thanks Yuta. I will merge this now, but I will be making some changes. The XRD and NMR code shares a lot of common pieces. So I think it would make sure to have a parent class and have XRD, NMR etc. inherit from that. I will implement this shortly.

shyuep:
@resnant I have just pushed my changes. Basically, I merged the two XRDPattern and NDPattern classes, moved the get_plot and other functions into a parent module. Right now, most of the differences are in the get_pattern method. There is still some duplication there, but it is something we can live with for now. Note that the names of some of the methods have been changed to make them more generic. Pls pull the latest changes and your NMR pattern should work. Thank you very much for the contribution.

resnant:
@shyuep Thank you for your quick response. I pulled and checked the [latest code](https://github.com/materialsproject/pymatgen/commit/56d75aeecfd1ec7d078fe846baf16302fdb413c1). 
It was nicely simplified, and the result looks good.

When I come up with a solution of the issue of peak indexing, I will send pull-request. But now I don't need strict peak indexing in my research purpose; the fix might be procrastinated.

PR #1125, opened 2018-04-25
Removed default dictionaries from sets

wood-b:
## Summary

default dictionaries weren't referenced properly, and I think it makes sense to put all the defaults in the __init__

* Fix 1 defaults dictionaries
* Fix 2 pep8 line length

## TODO (if any)

Unittests need to added @samblau


pymatgen/io/qchem_io/sets.py
```diff
@@ -37,9 +39,9 @@ def __init__(self, molecule, job_type, basis_set, SCF_algorithm, DFT_rung=4, PCM
             raise ValueError('molecule must be a Pymatgen Molecule object!')
```
shyuep: Pls remove this test. In general, it is a VERY bad idea to do type testing. The whole point of Python is to use duck typing. What if someone supplied something which effectively is like a molecule but not called a molecule?

wood-b: I'll remove it, thanks. FYI @samblau

PR #1127, opened 2018-04-25
 Adaption to anticipated use of x_diff_weight 

nisse3000:


shyuep:
If you removed bugs, that means that these bugs were not tested originally. Pls include tests to show that the bugs are removed.

nisse3000:
@shyuep, the class is from @computron. I'm currently testing in-depth and finding these bugs. Will be added to unit tests.

shyuep:
Ok. But pls submit PR only when you are more or less ready.

nisse3000:
Sure. But "Removed 2 bugs" and the newer "shortcomings" are two separate things. Didn't anticipate more clean up, when I issued the first PR.

Issue #1126, opened 2018-04-25
Feature request: bond order of covalent bonds


liwt31:
Function `is_bonded` in `bonds.py` is able to find if 2 sites are bonded according to bond length data stored. It's great. I think it's natural to automatically judge the bond order between the 2 sites, and `is_bonded` can return an int bond order instead of a bool. Further, why not make `bond_order` a member variable of the `CovalentBond` class and set it when initializing a `CovalentBond` object?

I can submit a pull request if the community think the feature is reasonable.

shyuep:
Yes, this feature would be great. Thank you in advance and please submit a pull request (with unittests) when you are ready.

mkhorton:
Note that you can now also get bond orders from `pymatgen.analysis.local_env.OpenBabelNN` also where bond orders are returned as neighbor weights.

PR #1129, opened 2018-04-26
Add bond order property of a CovalentBond

liwt31:
## Summary

Add bond order property of a CovalentBond (Issue #1126) . Initially I plan to change `is_bonded` to return an int indicating bond order. After some reflection I think this is a bad idea:
* According to the semantics, `is_bonded` should return a Boolean instead of an int.
* In some cases, like in benzene, the bond order can not be described by an int.

So what I finally do is to build another function `get_bond_order` which returns a float indicating the bond order. The bond order is calculated by the ideal bond lengths of different bond orders and the actual site distance. For example, if C-C bond is 1.54 A and C=C bond is 1.34 A and we have a distance of 1.4 A (benzene). Then the bond order is 2 - (1.4 - 1.34) / (1.54 - 1.34) = 1.7. I think this is enough to provide some qualitative information.

What I did is listed below in detail:
* Add a function `obtain_all_bond_lengths` as a wrapper of the bond length database. 
* Add a function `get_bond_order` to calculate bond order
* Add a member method `get_bond_order` for class `CovalentBond` to indicate the bond order of the covalent bond.
* Refractor `get_bond_length` and `is_bonded` to make them more concise using new functions.
* Improve the behavior of `get_bond_length`. If data for the bond order requested is not available, a warning will be shown and the radius sum will be returned, instead of `None`, which is the same as the behavior when no bond order data is available for the species.
* Make bond lengths json file indented to provide convenience for inspection.
* Add tests for the bond order related functions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 before_script:
   - python setup.py --quiet develop
 # command to run tests, e.g. python setup.py test
-script: travis_wait 18000 "nosetests -v pymatgen"
+script: travis_wait 14400 "nosetests -v pymatgen"
 notifications:
   email:
     recipients:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 before_script:
   - python setup.py --quiet develop
 # command to run tests, e.g. python setup.py test
-script: travis_wait 14400 "nosetests -v pymatgen"
+script: travis_wait 18000 "nosetests -v pymatgen"
 notifications:
   email:
     recipients:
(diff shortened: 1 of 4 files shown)


liwt31:
I found there are so many tests that with some bad luck travis-ci will time out. Is it possible for new PR's to only test codes that are modified? In ideal condition maybe we should test over the whole project. But currently we are so close to the time limit of travis-ci. Any ideas?

shyuep:
Do not worry about the timeout. You just need to make sure that the tests for your new code passes. I usually take a look at the reasons for failure.

liwt31:
Great. That makes me feel a lot better.

Issue #1137, opened 2018-05-03
Random test errors in elasticity


shyuep:
The elasticity test have errors due to some numpy issue. There are also a number of invalid values in numpy.

See https://circleci.com/gh/materialsproject/pymatgen/7783

Pls fix. @montoyjh

PR #1138, opened 2018-05-04
Elastic fix

montoyjh:
This addresses the primary issue in #1137.

The invalid value issues are related to negative arguments of a square root function in sound velocity and thermal conductivity models.  Currently these functions return np.nan if, say, G_VRH is negative because of bad input data and issue the warning.  I can change this, but wanted to get some advice on what I might do, should I

(1) - issue a different warning that's more descriptive of the issue, but still return np.nan
(2) - raise an error

Let me know.

### DIFF
--- a/pymatgen/analysis/elasticity/tests/test_elastic.py
+++ b/pymatgen/analysis/elasticity/tests/test_elastic.py
@@ -3,10 +3,8 @@ from __future__ import print_function
 from __future__ import division
 
 import unittest
-import os
 
-import numpy as np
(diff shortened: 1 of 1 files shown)


shyuep:
I would suggest raising an error as the safest route. Then end users are forced to explicitly handle these errors. Such errors can cause significant problems in analysis.

montoyjh:
Thanks for the feedback.  I've implemented as suggested with a wrapper for the relevant functions that will throw an error if the tensor is unphysical.

PR #1140, opened 2018-05-07
Small bug fix: Missing multiplication

JaGeo:
## Summary
Fixes a broken multiplication. Without it, mu_so does not work correctly and always returns "None".

### DIFF
--- a/pymatgen/analysis/magnetism/jahnteller.py
+++ b/pymatgen/analysis/magnetism/jahnteller.py
@@ -33,7 +33,7 @@ class JahnTellerAnalyzer:
 
         self.spin_configs = {
             "oct":
-                {   # key is number of d electrons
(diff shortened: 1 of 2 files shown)


shyuep:
Fix the exception being caught to a less extensive one. And if you fix this bug, pls include a unittest to ensure it does not happen again.

JaGeo:
I will work on it. It might take some time.

JaGeo:
I included a test for the static method mu_so. Moreover, only AttributeErrors are caught in the mu_so method. get_crystal_field_spin raises them, if the element is not a metal or the oxidation state does not fit. I decided not to catch ValueErrors because they are raised if an invalid coordination or spin state are used.   
I hope this is sufficient.

shyuep:
Thanks!

PR #1142, opened 2018-05-09
extend the XYZ class

gVallverdu:
## Summary

some float formats were not recognized when reading an xyz file. Only this one was read: 0.500e-09 and 0.7766e3. That floats were not recognized by the regex and are now read correctly: 

* 0.100E+01
* 0.100D+02
* 0.100d+2


shyuep:
Can you implement a few simple tests pls? Thanks.

gVallverdu:
I did it and push a new commit. It seems it appears already here.

shyuep:
Thanks.

Issue #665, opened 2017-05-10
Proposal for new k-point generation scheme


shyuep:
## People who definitely need to look at this code and comment
* @computron @kristinpersson @hautierg @zbwang @johnson1228 @montoyjh @wmdrichards 

I will leave this in the branch for three days. After that, I will merge if there are no comments.

## Summary

* Today's kpoint generation scheme is highly inefficient, with rounding to integers occurring at multiple steps along the way. 
* The major problem is that even for lattices where you have lattice lengths of 13 A or more, the code assigns a 2x2x2 grid with the MPRelaxSet. This makes calculations of such large cells extremely expensive. The following code gives

```python
from pymatgen import Structure
from pymatgen.io.vasp.sets import MPRelaxSet
import numpy as np
for a in np.arange(3, 17, 1):
    s = Structure.from_spacegroup("Pm-3m", np.eye(3) * a, ["Fe"], [[0, 0, 0]])
    v = MPRelaxSet(s)
    kpts = v.kpoints.kpts[0]
    print("For a = %.1f, k = %d x %d x %d" % (a, *kpts))
```

```
For a = 3.0, k = 8 x 8 x 8
For a = 4.0, k = 6 x 6 x 6
For a = 5.0, k = 6 x 6 x 6
For a = 6.0, k = 4 x 4 x 4
For a = 7.0, k = 4 x 4 x 4
For a = 8.0, k = 4 x 4 x 4
For a = 9.0, k = 4 x 4 x 4
For a = 10.0, k = 2 x 2 x 2
For a = 11.0, k = 2 x 2 x 2
For a = 12.0, k = 2 x 2 x 2
For a = 13.0, k = 2 x 2 x 2
For a = 14.0, k = 2 x 2 x 2
For a = 15.0, k = 2 x 2 x 2
For a = 16.0, k = 1 x 1 x 1
```

Obviously, a 2x2x2 grid for a 15 A cell is ludicrous, especially for insulators (this is the main class of materials the MPRelaxSet handles - metals are a separate class we deal with separately). We have grids of 4x4x4 all the way up to a 9A cell. 

## New proposal

* We will fix this in several areas. 
* In the Kpoints.automatic_density_by_vol method, kppa will no longer be rounded to an integer.
[...]

PR #1145, opened 2018-05-11
denser kpoints + more strict EDIFF independent of nsites

albalu:
## Summary
Making EDIFF independent of number of atoms is mainly to avoid errors in large unit cells. I tested the EDIFF=1e-5 with 120 ABC structures and had structures more consistent with the Materials Project both in terms of lattice constants and energy. Even if it is a slight overkill, I think it's good insurance against things going wrong down the line.

Also related to atomate structure optimization discussion: https://github.com/hackingmaterials/atomate/issues/19

### DIFF
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -11,6 +11,7 @@ import six
 import ruamel.yaml as yaml
 import os
 import json
+from copy import deepcopy
 
 from pymatgen.analysis.molecule_structure_comparator import CovalentRadius
 from pymatgen.core.sites import PeriodicSite
@@ -1399,14 +1400,14 @@ class LocalStructOrderParams(object):
 
         self._params = []
         for i, t in enumerate(self._types):
-            d = default_op_params[t].copy() if default_op_params[t] is not None \
+            d = deepcopy(default_op_params[t]) if default_op_params[t] is not None \
                 else None
             if parameters is None:
                 self._params.append(d)
             elif parameters[i] is None:
                 self._params.append(d)
             else:
-                self._params.append(parameters[i].copy())
+                self._params.append(deepcopy(parameters[i]))
 
         self._computerijs = self._computerjks = self._geomops = False
         self._geomops2 = self._boops = False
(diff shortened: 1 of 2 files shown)


albalu:
I forgot to change the kpoints density: not ready to merge yet.

shyuep:
You really shouldn't be modifying MPRelaxSet unless you have the consensus of the whole MP team. @computron @montoyjh Is this something that was discussed?

albalu:
We haven't really discussed this. This is one solution. I do get inconsistent/unconverged structure with the looser k-points (density=64) so I think it's better to change it in MPRelaxSet.

Another way of course would be to just change it in atomate. That's fine by me but I do think the k-point density is too low so better to just change it in MPRelaxSet.

shyuep:
I do not disagree, but this is not a decision that you can take.

montoyjh:
I'm confused about why you're getting different results, as far as I know, we've been using the same parameters the whole time (aside from the recent potcar issue and a minor change at some point in how k-point meshes were determined for very large unit cells).  We should compare these results directly before making any changes.

computron:
A lot of this has been discussed before

I think people should take a look at #665

shyamd:
Did you ever test just making EDIFF per atom more strict? Also your structure inconsistencies might have been potcar version dependent. We have 2001 potcars in MP.

albalu:
@shyamd , no I am just "request"ing to get rid of EDIFF_per_atom that likely would cause issues with large unitcells but my tests are limited to comparing these settings with some old runs (yes I checked with MP potcars). At this point, I am not sure what is a good test set. There should be a set of structures (5-10 maybe) that were run recently on MP so then I can compare the results of current defaults + some stricter settings. Based on my previous tests I say that the current defaults (atomate and MPRelaxSet) should change but change to what exactly I like to show it with a test set that everybody agrees on (i.e. everyone thinks the MP version of those structures is well converged).

shyuep:
Let us discuss this and settle this problem once and for all. Can one of you schedule a call with the key players? I will participate.

albalu:
This could be done sometime in June 6-8 when we have our next coding session if you are available. I can also join and see if I can help. I have put together a document that compares some outputs from atomate defaults (w/ correct POTCARs) and MP data. They are generally consistent but with some clear outliers. The cases where they are not consistent (either structure or bandgap) requires a more careful investigation.

[compare_with_MP.csv.zip](https://github.com/materialsproject/pymatgen/files/2018792/compare_with_MP.csv.zip)

In the meantime, please feel free to close this PR until I have a more solid test set and the reason for inconsistencies.

computron:
I think we should have a meeting before the coding session to discuss general strategy. The purpose of the coding session can be to implement that strategy

PR #1153, opened 2018-05-19
Updates to LAMMPS modules

adengz:
## Summary

For historical reasons, non-orthogonal simulation box was not properly treated in the original LAMMPS interface. This PR addresses all the issues left with those less used boxes. 

* Added a `LammpsBox` class, useful for building the data input file and parsing simulation box info in  dump files.
* Simplified the parser for dump files. Instead of having an extra long list (Natoms * Ntimesteps) for atoms data, all data now is separated for each timestep. 
* Fixed coordinate translation in `LammpsData` with non-orthogonal box using `LammpsBox`. The old constructor from a single structure is now `LammpsData.from_structure`. Since this is a high level constructor, this change is the only one that is backward compatible.

* None

## TODO (if any)

to be done

* The LAMMPS log parser seems buggy in my attempts using it. Should be easy to fix or even rewrite. 
* Will look at other building blocks to see if any improvements needed.


adengz:
Test fixed on AppVeyor. The failed test on Travis is not relevant to this PR.

PR #1160, opened 2018-06-04
Using now copy's deepcopy in local_env instead of shallow copy

nisse3000:


shyuep:
Can we have some tests pls?

nisse3000:
Sure thing, Shyue Ping.
(Kinda rhymes :-) ).

PR #1162, opened 2018-06-05
QChem updates for Atomate

samblau:
## Summary

Lots of polishing, formatting, and small to medium tweaks that were found to be necessary when building QChem functionality into Atomate and testing workflows. 

* All parsed files are accessed with zopen to avoid issues with gzipped files. This precipitated compatibility issues with python 2.7 that were then resolved.
* Outputs parses walltime and cpu time and has an as_dict method.
* Inputs now makes sure all dictionary keys and values are lowercase and unique to avoid the same key being set to multiple different values due to differences in key capitalization.
* Small changes in default DFT parameters.
* Tests updated and improved throughout. 

We would greatly appreciate if this could be merged as soon as possible and that a new version could be released immediately afterwards so that we can do a pull request for Atomate before Codebusters tomorrow. Thank you!


samblau:
All checks have passed. I believe this is ready to be merged in. We would greatly appreciate if it could be merged and a new version could be released this evening so that we're ready to go for Codebusters tomorrow morning. Thanks!

PR #1163, opened 2018-06-06
Fix nelect implementation for disordered structures and fix a few basic bugs in vasp sets

dbroberg:


pymatgen/io/vasp/sets.py
```diff
@@ -364,7 +364,7 @@ def incar(self):
             incar['NUPDOWN'] = nupdown
 
         if self.structure._charge:
```
shyuep: Why is a private variable being accessed?

shyamd: @shyuep I originally implemented it that way to ensure that people using charge decorated structures didn't get accidental changes in NELECT when they were expecting the structure -> VaspInputSet process to basically ignore charge decorations.

shyuep: yes, but why not access the public variable charge instead of the private variable _charge?

shyamd: The public variable first checks the private variable and then if it's not set uses the sum of the oxidation states. The previous behavior ignored this and always ran a charge neutral structure regardless of input structure oxidation states, so I wanted to maintain that behavior. We can change that to use the public method if that change in default behavior is alright with you.

shyuep: Maybe I am not quite understanding. What's the difference between the private and public variable in what it returns? Pls give a specific example.

shyamd: Here are the two relevant pieces of code. This is in SiteCollection:
```python
    @property
    def charge(self):
        """
        Returns the net charge of the structure based on oxidation states. If
        Elements are found, a charge of 0 is assumed.
        """
        charge = 0
        for site in self:
            for specie, amt in site.species_and_occu.items():
                charge += getattr(specie, "oxi_state", 0) * amt
        return charge
```

Then in IStructure, we override with this:
```python
    @property
    def charge(self):
        """
        Overall charge of the structure
        """
        if self._charge is None:
            return super(IStructure, self).charge
        else:
            return self._charge
```

Prior to the addition of `_charge` to IStructure, `charge` was only defined by the oxidation states, and VaspInputSets ignored it completely. Thus I could have a CIF Fe2O3 structure with oxidation states and if it wasn't  labeled with proper oxidation states such that the net charge was 0, the resulting VASP input would still be for the neutral structure. Thus the expected pattern was that regardless of what the structure's charge was, you'd get a neutral VASP input unless you modified it yourself.

`_charge` is simply a way to force the structure to have an overall charge without specifying a site to contain that. This is important for defects, for instance a charged vacancy.  Checking for `_charge` ensures that someone can continue to use the badly decorated Fe2O3 structure and still use VaspInputSet in the same way with the same expected behavior as before, but if someone knowingly sets `_charge`, or passes in `charge` when creating the structure, they get a VaspInputSet that properly sets `NELECT` to have a charged structure in the VASP input.

shyuep: I think we have to be careful about use cases here. Generally people expect neutral vasp inputs -  setting a charge is actually a very deliberate thing that you want to do for special situations. That is why the VaspInputSets ignore anything set in Structure - there is no guanrantee someone will have an oxidation state decorated structure and in most instances, I would assume people are supplying something they expect to be neutral.  Instead of trying to shoehorn this into the Structure object, I think it is better to allow VaspInputSets to take in a net_structure_charge which will be used to set nelect. Then remove the charge parameter for structure.

mbkumar:
I concur with Shyue on this. Using an extrinsic property stored in
Structure to influence the behavior of VaspInputSets is not a good idea.
You are adding a specific use case and pushing it on unsuspecting
users/developers. These use cases are hard to identify later on and
understand in general.

IMO, subclassing VaspInputSet is a better approach. In pycdt,
DefectRelaxSet specifically addresses this situation.

pymatgen/io/vasp/sets.py
shyamd: @shyuep I agree we have to be careful to ensure users can continue with their previous behavior which is why the implentation is the way it is currently. It may be a bit cryptic but it satifisfies a rational need. I don't agree that charge should be with VaspInputSet. The charge is for the structure not the calculation. If I have a structure with a charged vacancy, the structure itself has a net charge. There should be no need to carry a VaspInputSet to have this information with the structure. A similar issue arises for charged surfaces. 

I suggest a compromise with making the net charge more explicit in structure by changing the private variable`_charge` to the public variable `net_charge` and having VaspInputSet look at that to modify NELECT.

shyuep: @shyamd I understand why Structure has a charge, but no, it does not mean that setting charges should be done via Structure. As I mentioned, most people manipulate structures without consideration of charges, and when they output a set of VASP (or any other inputs), they expect a charge-neutral. Setting a charge should be an extraordinary step that requires thought. So please make the necessary changes such that overall charges for structures are done in VASP input set. As a compromise, you can set a boolean which says "use_structure_charge" but it needs to default to False.

shyuep:
thanks

PR #1164, opened 2018-06-07
set nkpoints to the actual length of kpoint list

albalu:
## Summary

previous implementation would return 0 if kpoints is auto-generated and 1 otherwise both of which are not the expected behavior.


shyuep:
Pls implement tests.

PR #1178, opened 2018-06-20
Fixed Charge/ Magnetization Parsing

sivonxay:
## Summary
* Fix 1 - Outcar parsing error : "ValueError: could not convert string to float:  "
Occasionally, reading the outcar results in an error when the "average electrostatic potential at core" contains values <-100. This is caused by the charge parsing incorrectly finding the electrostatic potential at core.

Example of this issue can be found by trying to create an Outcar object with the following file: [OUTCAR.gz](https://github.com/materialsproject/pymatgen/files/2117308/OUTCAR.gz)

* None

## TODO (if any)

* None


shyuep:
can you put in a unit test pls?

mkhorton:
@sivonxay in this PR, you're then skipping parsing electrostatic potential completely ? Would it be possible to fix parsing instead? (e.g. I think if every `-` sign in that block was replaced with ` -` this would fix it?)

mkhorton:
Thanks @sivonxay !

Issue #1103, opened 2018-04-12
_get_image() in pymatgen.analysis.local_env doesn't give correct image


computron:
## System

* Pymatgen version:  master

## Summary

* pymatgen.analysis.local_env.NearNeighbors#_get_image doesn't give the correct image
* Looks like @mkhorton did an initial implementation and @nisse3000 made the most recent changes to it

## Example code

```python
    mpr = MPRester()
    s = mpr.get_structure_by_material_id("mp-685291")

    cnn = MinimumDistanceNN(tol=0.5)
    nn = cnn.get_nn_info(s, 71)
    for x in nn:
        if x["site"].specie.symbol == "Al":
            print(x["site"], s[71].distance(x["site"], x["image"]))  # gives the WRONG distance for Al - way too far!!
            print(x["site"], s[71].distance(x["site"], [0, 0, 0]))  # gives the CORRECT distance for Al - this is the right image!
```

## Error message

none

## Suggested solution (if any)

haven't looked at the code

## Files (if any)

none

mkhorton:
There was some discussion on this on the PR Nils made for the most recent
change to get_image too. If you encounter further issues I’ll go back and
take another look.

computron:
From what I can tell, the neighbors returned by VoronoiNN, MinimumDistanceNN, and perhaps all NN algorithms give back a Site objects that have the correct cartesian coordinates but the wrong image (e.g., fractional coordinates that don't correctly represent the image they are in). But, I don't have much time to debug. Just the code example above, here modified a bit to show the Cartesian distance at the end and how it is in agreement with a "hacked" image but not the reported image:

```python
from pymatgen import MPRester
from pymatgen.analysis.local_env import MinimumDistanceNN
import numpy as np

mpr = MPRester()
s = mpr.get_structure_by_material_id("mp-685291")

cnn = MinimumDistanceNN(tol=0.5)
nn = cnn.get_nn_info(s, 71)
for x in nn:
    if x["site"].specie.symbol == "Al":
        print(x["site"], s[71].distance(x["site"], x["image"]))  # gives 15.85  - WRONG!
        print(x["site"], s[71].distance(x["site"], [0, 0, 0]))  # gives 1.96 - RIGHT!
        print(x["site"], np.linalg.norm(x["site"].coords - s[71].coords)) # gives 1.96 - RIGHT!
```

As far as I can tell the root of the image problem is the fractional coords of the site

mkhorton:
[...]

PR #1181, opened 2018-06-21
Faster Voronoi Tessellations

WardLT:
## Summary

Added methods to accelerate the computation of Voronoi tessellations for an entire simulation cell. Our current implementation computes the tessellation for each atom, this new implementation computes it once for the entire cell. As the time to compute the Voronoi tessellation scales with `O(N log(N))` (`N` being the number of atoms), this method is faster for larger simulation cells.

The chart below shows the performance of this new code on a series of simple cubic supercells. The cross-over point in performance is at ~2 atoms/cell. So, my revised implementation of `get_all_nn_info` only uses the "compute all at once" method for cells with > 2 atom.

image

None

## TODO (if any)

None


shyuep:
Pls fix the conflicts first. Also, if the limit is just 2 atom/cell, I wouldn't bother with making a conditional thing just for it. It introduces code complexity for a very minor case. People who are just doing one atom/cell can just wait a bit more.

WardLT:
Not yet ready to merge. While making the merges, I realized that VoronoiNNFiltered is unnecessary as the functionality for filtering neighbors is already built into the base class. 

I'm going to make a few changes to address that.

pymatgen/analysis/local_env.py
```diff
@@ -641,7 +640,62 @@ def get_voronoi_polyhedra(self, structure, n):
...
+
+        # Run the tessellation
+        qvoronoi_input = [s.coords for s in sites]
+        voro = Voronoi(qvoronoi_input)
+
+        # Get the information for each neighbor
+        return [self._extract_cell_info(structure, i, sites, targets, voro)
+                for i in range(len(structure))]
+
     def _get_elements(self, site):
```
WardLT: Also, the `_get_elements` method seems out of place here as it seems like this would be useful for a number of other operations. If you agree, should I open a separate PR to move it to Site?

WardLT:
Ok, got that taken care of. If the test passes, I think this is ready for review.

WardLT:
Actually, I found another bug when running this on a larger dataset. Numerical issues with `_get_image`. Sorry, I should have run this on a large dataset first.

WardLT:
This PR is ready for review again, if the error in Travis is indeed unrelated. 

In case it's relevant, the problem on the large dataset had to do with issues resolving the supercell image of atoms with fractional coordinates near 0 or 1. I avoided this problem by altering `Structure.get_all_neighbors` to return the supercell index used to generate each neighbor - bypassing the need to re-compute the supercell image. So, the issue listed in #1103 remains unfixed.

Another outstanding issue is that `get_all_neighbors` can return the supercell image, but `get_neighbors` cannot. Should I add a "feature request" Issue to make both operations have the same functionality?

shyuep:
I will merge. For get_neighbors, just implement it.

Issue #1085, opened 2018-03-29
Unable to Determine CN # For Mixed Site Occupancies


GerardMattei:
## System

* Pymatgen version: 2018.2.13
* Python version: 3
## Summary
Hello,
I am running into an error message when attempting to determine the CN # of mixed occupancies using the Voronoi NN class (sample code attached below). If I attempt to determine the CN # around an pymatgen periodic site with a single species, the .get_cn() method successfully returns a CN # of 12.

But if I then use the same structure, but replace the Pt site with Pt0.5Pd0.5, the method instead returns an error message. Since the structures for both are the same, this should also return a CN # of 12. A similar issue is encountered when using the .get_nn() method for mixed site occupancies.

If it helps, I just updated to the newest version of pymatgen (2/13/18). The VoronoiCoordFinder from Version 2017.8.21 was able to successfully return CN #'s, for either single or mixed occupancies.

Cheers and thank you for the help,
-Gerry

## Example code
from pymatgen.analysis.local_env import VoronoiNN
from pymatgen.ext.matproj import MPRester
from pymatgen.core.structure import Structure

with MPRester() as m:
    Pt = m.get_structure_by_material_id('mp-126')
    PtPd = m.get_structure_by_material_id('mp-126')
    PtPd['Pt'] = {'Pt': 0.5, 'Pd': 0.5}

print(Pt, PtPd)
print('Pt Coordination: ', VoronoiNN().get_cn(Pt, 0))
#This successfully returns a CN # of 12 around Pt[0].
print('PtPd Alloy Coordination: ', VoronoiNN().get_cn(PtPd, 0)) # This should also return a CN # of 12, but instead produces an attribute error.

nisse3000:
Hi @GerardMattei,

thanks for indicating this issue.  A couple of quick questions.

I'm trying to reproduce that the previous pymatgen version that you used would indeed give expected results. I tried one pymatgen version from August 20 (5e6f0f755a41016975c8e1137fd11c6810c39766) and one from August 21 (764da43fca24ba71978c331c9ecdf1e77a4cd13e).  With neither of these, I got it to work.  Have you used exactly this testing code in the past, too?  Or, is there a possibility that the partial occupancy was actually not set so that the previous pymatgen version didn't complain at that point.  I'm just trying to understand the problem completely.

In any event, the NearNeighbor class should be able to successfully deal with partial occupancies, and I'm working on getting that fixed.

nisse3000:
[...]

PR #1182, opened 2018-06-21
 Fixed issue with partial occupancies in VoronoiNN and added corresponding new unit test.

nisse3000:
Pertains to Issue #1085.


nisse3000:
@shyuep, the Travis build failed, but the reason is not the local_env module. Merge?

PR #1184, opened 2018-06-26
Add Supercell Image to Neighbor-Finding Routines

WardLT:
## Summary

Adding the ability to output to all routines that determine nearest neighbors. Detecting which image a neighbor comes from given its coordinates (Cartesian or fraction) can be complicated for neighbors near the edges of cells, and this change allows algorithms to avoid needing to perform such operations.

None

## TODO (if any)

~~Need to add the image outputs to neighbor finding routines in Structure. I'll be working on those shortly, but wanted to open a PR to run unit tests because changing Lattice touches many parts of pymatgen.~~

### DIFF
--- a/pymatgen/analysis/diffraction/neutron.py
+++ b/pymatgen/analysis/diffraction/neutron.py
@@ -133,7 +133,7 @@ class NDCalculator(DiffractionPatternCalculator):
         peaks = {}
         two_thetas = []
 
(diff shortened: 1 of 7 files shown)


WardLT:
Note, this may provide a route to fix #1103

WardLT:
I've added "return supercell" to all of neighbor-finding routines I could find. This PR is ready for review.

PR #1186, opened 2018-06-29
add method get_critical_original_kink_ratio to interface_reactions module

yihanxiao92:
## Summary

Added method get_critical_original_kink_ratio to interface_reactions module. The point is to provide users with the critical mixing ratios between ORIGINAL reactant compositions, not normalized ones. These original mixing ratio would more useful and straightforward for experimentalists.

* Added method get_critical_original_kink_ratio to interface_reactions module, and the test of the function.


yihanxiao92:
@shyuep Hi Prof. Ong, it seems that the checks are all completed but the status is still pending. May I know what I could do to get this pull request merged? Thanks!

PR #1187, opened 2018-06-29
Fractional NELECT

mturiansky:
## Summary

In the DictSet provided in pymatgen.io.vasp.sets, fractional NELECT values are not allowed. For almost all purposes, this is not a problem. However, I am using the DictSet to prepare calculations that use the fractional hydrogen pseudopotentials, giving an incorrect NELECT value. I added a flag 'allow_fractional_nelect', which defaults to False, to the DictSet class.

I added relevant code to MITMPRelaxSetTest.test_nelect for code coverage.

None.

## TODO (if any)

None.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -261,7 +261,7 @@ class DictSet(VaspInputSet):
                  constrain_total_magmom=False, sort_structure=True,
                  potcar_functional="PBE", force_gamma=False,
                  reduce_structure=None, vdw=None,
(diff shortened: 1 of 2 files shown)


mturiansky:
I see that the tests failed with 'OSError: You do not have the right POTCAR with functional PBE and label H1.25 in your VASP_PSP_DIR'. I was under the impression that these fractional hydrogen pseudopotentials were included with VASP as they were included in mine, and the test completed fine on my computer.

I can change the test so it uses a non-fractional pseudopotential but still runs the relevant code.

PR #1208, opened 2018-07-18
QChem Refactor and Update

samblau:
* QChem IO functionalities have been refactored to be named consistently with other IO packages
* Default sets have been improved, including changing the default SCF strategy from DIIS to GDM
* Multiple new functionalities have been added to the output parser, including the ability to classify different types of structural changes encountered while optimizing
* Tests have been udpated and supplemented accordingly


shyuep:
There are still failing tests. Pls fix. Thanks.

samblau:
Pretty sure this will fix the appveyor / travis-ci build, but it required that I put an unused attempted import openbabel just to know whether or not to include that test, which now makes codacy upset. If this isn't acceptable, please advise. Thanks!

PR #1215, opened 2018-07-23
Elastic Revision

montoyjh:
* Tensors are MSONable, with optional flag for using voigt notation dict representation (this stores them more efficiently, but would only be used if the user explicitly chose to do so, as information may be lost)
* Revises determination of coefficient of linear expansion to use compliance tensor rather than inverse of bulk modulus, this is more rigorous and provides more directional information
* Fixes minor issue in finite-difference fitting of TOECs related to division by zero if equilibrium strains are included in fitting data
* Adds a set method for tensor mapping, primarily for dealing with tensor-paired values
* Fills in a few missing methods for tensor expansion objects that present for tensors
* Adds associated tests for prior bug, MSONable functionality, new alpha


montoyjh:
Need to fix a few tests.  Will work through and update when done.

montoyjh:
I think this is ready on my end.  The travis tests seem to be failing because of an enumlib issue which I believe is unrelated to this PR.

PR #1222, opened 2018-07-29
Changes to JMolNN class

jdagdelen:
## Summary

Jmol's spelling was incorrect (the developer explicitly says not to capitalize "mol") and there was a redundant constant used to determine bond lengths.

* JMol -> Jmol
* One tolerance instead of two.


shyuep:
tests are failing.

mkhorton:
@jdagdelen there was a query from Hillary/Anubhav recently about whether the Jmol emulation in pymatgen faithfully matched the behavior in Jmol itself; were you able to find a good reference for Jmol's algorithm? If so, would be good to add a reference to the docstring.

The best I was able to find was this memo, but I don't know if it's up to date: https://sourceforge.net/p/jmol/mailman/message/5402874/

jdagdelen:
Oh man, I forgot to refactor the name in the rest of pymatgen. That was a dumb oversight, will fix and submit another PR.

jdagdelen:
@mkhorton I didn't find any references for their algorithm so I was just comparing with Jmol's source code to verify the behavior.

PR #1226, opened 2018-08-01
io.lammps updates

adengz:
## Summary

Major updates to `io.lammps`. The entire module is now flexible enough to handle most lammps related tasks. 

* Simplified code for handling input script, which should just behave just like a string template. 
* Revised log parser, and added many tests. 
* Minor changes to dump parser. Now dump data for a single snapshot is a `LammpsDump` instance. 
* Removed several badly designed modules (`lammps.sets` is no different than `lammps.inputs`) and classes (`LammpsRun` is very specific for certain type of runs).

None

## TODO (if any)

None


shyuep:
Good. But are there supposed to be some sample templates/scripts? I don't seem to see them in there.

adengz:
Sure, there will be examples on the usage of input script template updated in matgenb separately.

shyuep:
This is better, but still not what I meant. THink about it from a new user perspective - I am clueless about LAMMPS and all I want to do is to use pymatgen to generate an input file for me to run a simple structure. At the minimum, the example templates should allow for specification fo arbitrary structures.

What I want is basically something that allows the following code:

```
lr = LammpsRun.md(structure, various parameters...)
lr.write(output_dir)
```

adengz:
Personally, I am not a fan of this setting. The transferability (in many aspects) for lammps simulations is highly limited. Finding a proper potential and coding it right in lammps is already non-trivial. It might help users with completely no knowledge, but is hardly useful for users starting to do production runs. 

Anyway, @shyuep let me know if this is the thing in your mind. And we surely can add more in the future.

shyuep:
Doesn't LAMMPS itself come with some potentials? The argument cannot be that "this is non-trivial" and so we do not need to provide any help or gudiance for users. By that logic, I can argue that VASP runs are very difficult, with so many parameters like smearing, energy cutoff, kpoints, psuedopotentials, etc. So you might as well work on your own input files. That is now how it works.

So let me lay this out simply. I am a new LAMMPS user. I want to generate a LAMMPS MD calculation and analyze it. I am not very sophisticated - so I will just live with the LJ or buckingham potential that is built into LAMMPS. Can I generate the input file and do the analysis in a few lines of code?

Of course you also need to cater to the highly sophisticated user who wants to use their custom potential.

adengz:
I agree that the interface for LAMMPS should be designed similar with the one for VASP here. The milestone achieved here for LAMMPS io is mainly the components for handling various files, like `Incar` and `Vasprun` for VASP. 

The `VaspInputSet` counterpart is effectively the input script in LAMMPS. I have been more than grateful for all these input sets, allowing me to do calculations from a structure without any DFT knowledge a few years ago. 

But the fundamental difference is the **transferability**, not only on the level of potential energy surface, but also software features. As I have shown you earlier, a script for LAMMPS can have infinite possibilities and can hardly be handled with coded rules. Similarly, analyses to be performed rely on the availability of output data, which is again highly customizable in the script. And that is why `lammps.outputs` might look way too simple at first glance. It has to be general for all the possibilities. 

The template is definitely not perfect, but enough as a simple example. I will leave it as is. Improvements and addition of templates will be long-term based on more feedbacks from others, as I am just a solely LAMMPS user only familiar with my own day-to-day tasks.

PR #1233, opened 2018-08-09
from_file: recognize ".vasp" extension as POSCAR-type

ruixingw:
POSCAR-type files with `.vasp` extension can be read by VESTA but not by pymatgen. One has to rename the file because the `from_file` method doesn't provide a `fmt=` flag.
This change makes life easier.


shyuep:
Unittest pls. I know this change seems trivial, but I will still need tests for it.

PR #1234, opened 2018-08-09
Added shortcut if structures is equal to self

WardLT:
## Summary

Determining whether a structure is equal to itself can be very expensive for large structures. A structure with 100 atoms can take seconds to verify, and the cost scales with the square of the number of atoms. This change adds a check in the equality operation if the "other" class refers to the same object in memory to drastically lower the cost of "x == x"

This problem has implications for the caching methods we use in matminer.

None

## TODO (if any)

None


WardLT:
I'm guessing the timeout failure in the enumlib test is unrelated?

PR #1239, opened 2018-08-13
Include reading and evaluation of DOSCAR.lobster

JaGeo:
## Summary

I have included a class to read "DOSCAR.lobster" created with Lobster and VASP and another class to evaluate the results so that one could plot "DOSCAR.lobster" with the help of the class DosPlotter.

I have also written tests for both classes.


JaGeo:
What else do I need to do so that one can merge this? The Travis CI check seems to fail because of a time limit.
Thanks in advance.

Issue #1072, opened 2018-03-20
Adding Functional Group substitution to periodic crystals


computron:
## System

* Pymatgen version: master

## Summary

* Currently, one can only add a functional group to a site for a Molecule object (substitute() method)
* Sometimes, you want to do functional group replacement for molecular crystals (i.e. Structure objects)
* From what I can tell, this is easy to do. The substitute method essentially works out of the box, you just have to change:

```py
del self[index]
for site in func_grp[1:]:
    self._sites.append(site)
```

to

```py
del self[index]
for site in func_grp[1:]:
    s_new = PeriodicSite(site.species_and_occu, site.coords, self.lattice, coords_are_cartesian=True)
    self._sites.append(s_new)
```

Thoughts on:
* do you want to add this?
* what's the best way to do it to avoid code repetition, e.g. putting it in SiteCollection, etc.

shyuep:
You can add this. If you can think of a general enough way to add it to SiteCollection, that would be good. But I think the Site vs PeriodicSite might be tricky.

The other thing we may want in future for both Mol and Structure is to be able to replace a functional group (e.g., -CH3) with a functional group (-NH2). Conceptually it isn't hard. There is already a method for bond breaking in Molecule. But doing it in a way that is robust is not easy.

mkhorton:
I've been thinking of ways to do this also -- using the bond graph should make this relatively tractable, and this works with periodicity, might be overkill though. Will have a think about how to do this gracefully (i.e. integrating a common solution into SiteCollection). @espottesmith do you have any thoughts on this?

espottesmith:
Warning: since I've been working pretty much exclusively with non-periodic Molecules, I'm not sure I have a complete understanding of the differences that would be involved with Structures.

So, I think that there are a couple of different levels of complexity here, and I'm not sure there's a single solution that elegantly addresses all of them.

- For simple functional group substitution, I agree with @mkhorton that using the StructureGraph/MoleculeGraph classes is a bit overkill. I haven't read the SiteCollection class that closely, but I think it should be possible to extend & generalize Molecule.substitute().
[...]

PR #1247, opened 2018-08-22
Improve Functional Group Substitution/Replacement, aiming to resolve Issue #1072

espottesmith:
## Summary

Previously, functional group substitution was only possible for Molecules, not Structures. In addition, functional group replacements (substituting a functional group for another functional group, rather than an atom for a functional group) was impossible across both classes. This pull request adds functional group substitution to Structures as well as StructureGraphs, and adds functional group replacement to MoleculeGraphs. In addition, this establishes near feature parity between the MoleculeGraph and StructureGraph classes.

## TODO

Functional Group replacement could be added to StructureGraph, if there is a significant need. It was not added yet because it would likely require a significantly different algorithm than that used for MoleculeGraph in order to identify the extent of the functional group.


espottesmith:
No errors have arisen, but it seems that Travis timed out. What is the best way to move forward and resolve this problem?

PR #1248, opened 2018-08-23
Fix Outcar's MSONable pattern

shyamd:
## Summary

Outcar is subclassed from MSONable, but is not intended to fully implement this pattern as its not designed to instantiated from the dictionary it generates. This ensures that Monty doesn't try to auto convert the dictionary into an Outcar object, for instance when using loadfn from a jsonified Outcar or task document. 

- Ensures Outcar as_dict dictionaries don't have "@class" or "@module" keys
- Ensures Outcar from_dict throws an error

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -2569,8 +2569,7 @@ class Outcar(MSONable):
         return aps
 
     def as_dict(self):
-        d = {"@module": self.__class__.__module__,
-             "@class": self.__class__.__name__, "efermi": self.efermi,
+        d = {"efermi": self.efermi,
              "run_stats": self.run_stats, "magnetization": self.magnetization,
              "charge": self.charge, "total_magnetization": self.total_mag,
              "nelect": self.nelect, "is_stopped": self.is_stopped,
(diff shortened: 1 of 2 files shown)


shyamd:
note the travis-ci test failed due to a timeout. Should we increase this to get all the tests to complete?

mkhorton:
This is tricky, because this breaks the MSONable contract. I'm assuming you want to keep the MSONable subclass so that dumpfn will correctly use its as_dict() method.

(Note that I believe MontyEncoder will automatically add the `@class` and `@module` if they're not present anyway, so I'm not sure if the fix proposed in this PR would even work.)

Ideally we'd just change Outcar to be properly MSONable but I imagine that would break a lot of other code at this point.

shyuep:
Question: Why does Outcar need to be MSONAble? The idea behind an MSONable is that you can convert to and from json/yaml format. Here, the conversion is one way.

mkhorton:
Yes, I agree -- we could just make Outcar not subclass MSONable. This would fix @shyamd's problem of loadfn not working, while keeping the current as_dict() behavior.

shyamd:
@mkhorton  That's right. I want to dump and load files that contain Outcar dicts somewhere in them. @shyuep  I'm fine with just changing Outcar to just inherit from object. In which case, It might be easiest if we just got rid of this PR and you just made the simple change.

shyuep:
Done.

PR #1266, opened 2018-09-10
Add function to remove atoms when atoms are too close

ucsdlxg:
The user now can set a criteria to remove atoms if the atoms are too close with each other.
This function is very useful for tilt GB generation.


ucsdlxg:
I read the merge_site code in structure.py. It seems like that it is still different from what I want if I understand it correctly. e.g. if we have three atoms a, b, c, distance of ab and bc are within the threshold. Then the merge_site code will merge these three as one atom. In my case, I just need to remove b, since ac is still normal. The pairs have closer distance are always one from bottom and one from top grain. So ab belong to different grains. While ac belong to the same grain. We should not merge these three to one.

For the many changes, 
1) I also add one more gb properties, 'sigma_from_site_prop'. This is to determine the sigma value from site properties. This is help to determine the site_properties determined from this code is correct or not. The accurate sigma value is from the property of 'sigma', which are determined from the volume ratio of the oriented cell and primitive cell. If the 'sigma_from_site_prop' is different from 'sigma', then the site_properties are wrong. The user may need to adjust the 'tol_coi' to find the correct coincident sites.
2) I also fix a bug in the previous code in the function reduce_mat.

mkhorton:
@ucsdlxg would it also be possible to rename this class to something more descriptive, e.g. `GrainBoundaryGenerator` or `GrainBoundary` instead of `Gb` ? (likewise for the submodule name)

mkhorton:
`Structure.merge_sites` currently drops any site properties, which would cause problems in the Gb class (we'd lose the grain labels), so in this instance I think this PR is ok. Ideally, `Structure.merge_sites` should work for this though.

shyuep:
How about we fix merge_sites to preserve site_properties? That said, we need some way to deciding what site properties to preserve. Maybe we take average of values or something. In either case, we need to issue a warning when this is done.

shyamd:
@ucsdlxg 
Please use the template in the future and write out what you're changing and why.  For instance, it looks like you're only changing GB when you're initial description sounded like you were changing structure.

Example for this PR:
## Summary

Added some features to GB class:
* Add tolerance to remove sites rather just merging clusters using structure.merge_site
 This prevents groups of sites : a,b,c from being merged into one site and instead will remove just one site to satisfy the distance tolerances
* Calculate Sigma from site properties

ucsdlxg:
Hi @mkhorton, I have changed the merge_sites in structure.py to keep the site properties, and use it in GB generator.  I also changed the class names.

One issue I am still a little worried is that if we use merge_sites, it need the distance matrix for the whole sites. But the overlap atoms only locate near the interface of the two grains. Therefore, if the GB is super large (e.g. tens of thousands atoms), this maybe time-consuming.

pymatgen/core/structure.py
```diff
@@ -3124,14 +3124,18 @@ def merge_sites(self, tol=0.01, mode="sum"):
...
                 sp = self[i].species_and_occu
                 if mode == "s":
                     species += sp
                 offset = self[i].frac_coords - coords
                 coords += ((offset - np.round(offset)) / (n + 2)).astype(
                     coords.dtype)
-            sites.append(PeriodicSite(species, coords, self.lattice))
+                if self[i].properties != props and props[list(props.keys())[0]] is not None:
+                    props[list(props.keys())[0]] = None
+                    warnings.warn("Sites with different site_prop are merged. Site_prop is set to none")
```
mkhorton: Can you change the warning to "Sites with different site property {} were merged, so property set to None." with the specific property name? As written, this doesn't seem like it'd support multiple site properties properly.

mkhorton:
Thanks for the changes @ucsdlxg. Regarding the scaling of merge_sites, have you tested it? It looks like it uses some algorithms from scipy.spatial which will hopefully be quite efficient.

ucsdlxg:
Hi @mkhorton, I modified the structure.py to enable the multiple site_properties.

In addition, I add the joining plane information to the GB object. In principle, a GB has five freedoms. With the rotation axis (2 freedoms), angle (1), and the plane (2 freedoms) of one grain (GB plane), one can determine the GB structure (That is how the code generate the GB structure). But people usually also want to know the plane of the other grain (joining plane), so I also add that information to the GB object. For twist and symmetrical tilt, the GB plane and joining plane are the same. For a general GB, these two planes can be different.

ucsdlxg:
@shyuep , Sure I will fix it. I also plan to write a function so that it can enumerate all possible 'twist', 'symmetric tilt', 'normal tilt', 'mixed' GBs for a given rotation axis and angle with a plane miller index cutoff for cubic system at least. This is very useful for high-throughput calculations.

shyamd:
There are a number of ops that could be vectorized by using numpy. There's at least 4 sets of nested loops that you could vectorize with numpy and get significant speed up.

shyuep:
I cleaned up some of the gb code to remove uses of np.matrix. You should pull the latest changes before continuing.

I also notice a lot of print statements. That should not be the case. You should use a proper logging and allow people to specify the verbosity level. By default, all the logs should be at info level and turned off. Otherwise, it is very noisy. This is the system used in most of pymatgen.

ucsdlxg:
Or maybe we can divide the GB into two pseudo structures. One use merge_site, the other keep the same.

shyamd:
Theres nothing wrong with overriding merge_sites in GB so that its specific for the GB. You might want to do something like build a temp structure with just the relevant sites. Call Structure.merge_sites there. Then copy the resulting sites back over.

shyuep:
Is this ready for merging? If so, Matt can do a final check before doing so.

ucsdlxg:
@shyuep, I think there is still one more issue I am trying to solve now. Thas is not related the GB structure but about how to determine the coincident sites. Right now, I just check the overlap sites between the top and bottom grains. But I am thinking we should also need to consider the symmetry operations. If I use Spacegroupanalyzer to obtain the symmetry operations, for some big systems, it is really slow. Do you have any suggestions?

ucsdlxg:
Hi @mkhorton, I have optimized the GB codes thoroughly and ready to merge.  Can you help do a final check?  Thanks!

mkhorton:
Will do, thanks @ucsdlxg

PR #1270, opened 2018-09-12
circleci fixed

fraricci:
- circleci edited to compile boltztrap

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,11 @@ jobs:
             conda update --quiet --all
             mkdir coverage_reports
             pip install --upgrade pip
-            pip install --quiet --ignore-installed -r requirements.txt -r requirements-optional.txt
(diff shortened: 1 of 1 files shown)


mkhorton:
Would it be possible to notify the BoltzTrap2 maintainers of this as well? A colleague was trying to install BoltzTrap2 recently via pip and was unable to do so / didn't realize additional CFLAGS had to be set.

fraricci:
Yes, it's a better solution actually. It can be useful in general.
I'll do that so. Thanks

shyuep:
@fraricci I think you need to fix the tests. Right now, the boltztrap2 tests take so long that they fail on CircleCI. Tests need to be reasonably quick (less than a minute preferably, but certainly no more than 10 mins).

fraricci:
I'll check that. It sounds weird because in my computer test for btz2 run quickly.

mkhorton:
Thanks!

Issue #1236, opened 2018-08-10
Band structure fails to load if vasprun is gzipped and KPOINTS isn't


ajjackson:
## System

* Pymatgen version: 2018.7.23
* Python version: 3.6
* OS version: Linux Mint 17.3 64-bit

## Summary

* Using the `BSVasprun` class to import a band structure calculation
* If vasprun.xml and KPOINTS files are present, it loads ok
* If both vasprun.xml and KPOINTS files have been gzipped (i.e. vasprun.xml.gz), it loads ok
* If only one file has been gzipped, an error is raised complaining about lack of KPOINTS file. This is quite disorientating if there is in fact a KPOINTS file...

## Example code
```bash
gzip vasprun.xml
```

```python
from pymatgen.io.vasp.outputs import BSVasprun
vr = BSVasprun('vasprun.xml.gz')
vr.get_band_structure(line_mode=True)
```

## Error message

```
  File "/home/adam/.local/lib/python3.6/site-packages/pymatgen/io/vasp/outputs.py", line 749, in get_band_structure
    raise VaspParserError('KPOINTS needed to obtain band structure '
pymatgen.io.vasp.outputs.VaspParserError: KPOINTS needed to obtain band structure along symmetry lines.
```

## Suggested solution

* This behaviour is caused by the following line in _pymatgen.io.vasp.outputs_:
```python
        if not kpoints_filename:
            kpoints_filename = self.filename.replace('vasprun.xml', 'KPOINTS')
```
Presumably the "replace" expression is intended to maintain the first part of the path (i.e. the parent directory). However, if `self.filename` was _/some/path/to/vasprun.xml.gz_ then the `replace` operation will leave _/some/path/to/KPOINTS.gz_.

* Suggested change

```python
        if not kpoints_filename:
            vasprun_dir = os.path.abspath(os.path.join(self.filename,
                                                       os.path.pardir))
            for kpoints_filename in ('KPOINTS', 'KPOINTS.gz'):
                if os.path.exists(os.path.join(vasprun_dir, kpoints_filename)):
                    kpoints_filename = os.path.join(vasprun_dir,
                                                    kpoints_filename)
                    break
            else:
                kpoints_filename = ''

```

computron:
Not sure why this is assigned to me (pretty sure I had no part in the original code)?

But anyway, I am not sure what os.path.pardir refers to in the above example.

[...]

PR #1272, opened 2018-09-12
Allow mixed (un)compressed files from VASP band structure

ajjackson:
## Summary

Pull request corresponding to Issue #1236 

Monty's `zpath` is used to find (un)compressed KPOINT files corresponding to (un)cimpressed vasprun path. This was the solution suggested by @computron and it seems to cover all the permutations; you can even use gzip for one file and bzip2 for the other.

* None. Uses existing dependency Monty

## TODO (if any)

I looked at adding a test but the existing tests rely on passing specific filenames and to test this we need to either
a) add some redundant test data files or
b) copy/zip some files to a temporary directory and delete them. Probably overkill?


ajjackson:
A meaningful test will need to open files called "KPOINTS" and "vasprun.xml.gz". Where should these be located? Is there a convention in this project for making temporary files/folders during testing?

shyamd:
try using monty's ScratchDir
```
from monty.tempdir import ScratchDir

with ScratchDir("./"):
    Copy files
    Test
    Gzip some files
    Test more
```

shyuep:
Thanks!

PR #1275, opened 2018-09-13
- vasprun test corrected

fraricci:


fraricci:
weird... Is it loading boltztrap2 package?

fraricci:
I'm not sure how to set a variable in windows.
Could you take care of that?
We could also skip installing boltztrap2 on windows, I added an IF that should skip the test in case
it's not installed

PR #1283, opened 2018-09-18
Non-integer subscripts in MolecularOrbitals class

dyllamt:
## Summary

* Semantically, `MolecularOrbitals` should only handle compositions with integer-valued subscripts.
* The doc strings for this class have been updated to reflect the intended use.
* A feature has been added that raises a `ValueError` when non-integer subscripts are used.

(the issue of non-integer subscripts propagated into _matminer_, where this issue was first identified)


shyuep:
Can you implement some unittests to confirm the new functionality pls? Thanks.

PR #1285, opened 2018-09-20
added parsing of volumetric data as dictionaries

jmmshn:
Changed
VolumetricData(object) -> VolumetricData(MSONable)
added tests for as_dict and from_dict


mkhorton:
Thanks @jmmshn, will take a look at this when tests pass.

PR #1289, opened 2018-09-26
Change default VoronoiNN cutoff

### LINKED ISSUE #1281: VoronoiNN causes Python to crash
Calling `pymatgen.analysis.local_env.VoronoiNN.get_voronoi_polyhedra()` on a structure obtained via `MPRester` is causing Python to crash. Note that Python itself is crashing, not just returning an error message.

The structure in question is: https://materialsproject.org/materials/mp-978085/. It is a relatively small primitive cell with only 4 sites.

## System

* Pymatgen version: 2018.9.12
* Python version: Tested on 3.6.2 and 3.7.0
* OS version: macOS 10.13.6 (High Sierra)

## Example code

```python
from pymatgen.analysis.local_env import VoronoiNN

mpr = MPRester()
struct = mpr.get_structure_by_material_id('mp-978085')

voro = VoronoiNN()
voro.get_voronoi_polyhedra(struct, 2)  # sites 0, 1, and 3 work fine
```

## Error message

macOS crash reporter produces an error message saying:

```
malloc: *** error for object 0x1062909e8: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Python quit unexpectedly.
```

image

This appears to be a memory allocation error. Obviously Python itself should not crash under any circumstances but I wonder if we can do something on our end to circumvent this issue.

utf:
## Summary

Change default cutoff of `VoronoiNN` from 10.0 to 13.0. This fixes #1281 for all 85,000 materials currently in the materials project database.

I have also changed the way the max_cutoff is calculated. Previously, the max cutoff was incorrect for cells with angles not equal to 90 deg (e.g. for the example in #1281). The max cutoff is now set to the largest of the 4 cell diagonals. I'm not confident that my implementation is the cleanest way to do this but it works.

Lastly, I moved the `self._extract_cell_info()` call to within the try-catch block, as I found that just increasing the cutoff fixed this error in most cases (maybe all cases). Just to be sure, I altered the error handling to make sure the right error message is thrown.

### DIFF
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -590,7 +590,7 @@ class VoronoiNN(NearNeighbors):
             (default: 0).
         targets (Element or list of Elements): target element(s).
         cutoff (float): cutoff radius in Angstrom to look for near-neighbor
-            atoms. Defaults to 10.0.
+            atoms. Defaults to 13.0.
         allow_pathological (bool): whether to allow infinite vertices in
             determination of Voronoi coordination.
         weight (string) - Statistic used to weigh neighbors (see the statistics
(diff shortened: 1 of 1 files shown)


mkhorton:
Excellent, thanks @utf !

PR #1298, opened 2018-10-08
Insertion batt js

jmmshn:
## Small changes to the insertion battery module
* updated the artificial shift in energy to the non-working-ion elements so that we can always have a convex hull of with just the working-ion changing
* made the dicts for adj_pairs and all_paris return the actual list and not just the iterator.


mkhorton:
Test for a case that previously wasn't working?

PR #1300, opened 2018-10-13
Chgcar insertion

jmmshn:
added capability to identify site by their integrated charge density


shyuep:
What's the difference between this method and Chgcar.get_integrated_diff?

jmmshn:
This is meant to look at the data['total'] and just suggest the best site to insert an interstitial atom.
The Chgcar.get_integrated_diff is really meant for analysis of the data['diff'].
There is mention of get_integrated_diff being time-consuming in the documentation, I assume that is for the binning process.  We do not need to do the binning here.
Other than that, I think I can replace the masking parts of my routines with the `lattice.get_points_in_sphere` so there is less redundant code.

HanmeiTang:
Hi @jmmshn, I think this is similar to `ChargeDensityAnalyzer` class, which locates local minima using charge density read from CHGCAR or similar files. See the example code below.

```python
from pymatgen.analysis.defects.point_defects import ChargeDensityAnalyzer as CDA

ca = CDA.from_file("CHGCAR_sum") # or CHGCAR
s = ca.get_structure_with_nodes()
s.to(filename="output.cif")
```

Issue #1301, opened 2018-10-19
miller_index_from_sites gives wrong miller index


utf:
The `pymatgen.core.surface.miller_index_from_sites` method gives incorrect results. It also seems quite temperamental about the choice of coordinates. E.g.

```python
from pymatgen.core.structure import Structure
from pymatgen.core.surface import miller_index_from_sites

structure = Structure(
    [10, 0, 0, 0, 10, 0, 0, 0, 10],
    ['Fe', 'Fe', 'Fe'],
    [[0.5, 0.8, 0.8], [0.5, 0.4, 0.2], [0.5, 0.3, 0.7]]
)

miller_index_from_sites(structure.lattice.matrix, structure.cart_coords)
```

Gives `[-1.04,  0.  ,  0.  ]` rather than `[1,  0.  ,  0.  ]`.

Also, changing the atom coordinates slightly to:

```python
[[0.5, 0.8, 0.8], [0.5, 0.6, 0.6], [0.5, 0.3, 0.3]]
```

Gives `[nan, nan, nan]`.

richardtran415:
Looking into it

utf:
I've got a fix for this. Just doing some testing but will submit a PR tomorrow.

utf:
Fixed in #1302

PR #1302, opened 2018-10-19
miller_index_from_sites now supports more than 3 sites

utf:
## Summary

Fixed bugs in `miller_index_from_sites` detailed in #1301. 

I also extended `miller_index_from_sites` to support passing more than 3 sites. In this case, the "best" plane that minimizes the distance from all atoms to the plane will be calculated using singular value decomposition. This behavior is the same as provided in the VESTA visualization tool.

Other changes include:
- Support for fractional or cartesian coordinates.
- Improvements in getting the integer miller index.
  - The number of negative indexes is minimized. E.g. `(-1 -1 1)` -> `(1 1 -1)`.
  - If there are negative indexes, the largest number should be possitivie. E.g. `(-2 1 0)` -> `(2 -1 0)`.
  - Other improvements to reliability.
- Tests for the failing case described in #1301.
- Tests for finding the plane with more than 3 sites.

One thing to note is that previously `miller_index_from_sites` required the lattice matrix as input, whereas now it requires a `Lattice` object. This is so I can calculate the fractional coordinates.

### DIFF
--- a/pymatgen/core/lattice.py
+++ b/pymatgen/core/lattice.py
@@ -5,6 +5,17 @@
 from __future__ import division, unicode_literals
 import math
 import itertools
+import warnings
+
+from functools import reduce
+try:
+    # New Py>=3.5 import
+    from math import gcd
+except ImportError:
+    # Deprecated import from Py3.5 onwards.
+    from fractions import gcd
+
+from fractions import Fraction
 
 from six.moves import map, zip
 
@@ -30,7 +41,6 @@ __status__ = "Production"
 __date__ = "Sep 23, 2011"
 
 
-
 class Lattice(MSONable):
     """
     A lattice object.  Essentially a matrix with conversion matrices. In
@@ -39,7 +49,6 @@ class Lattice(MSONable):
     """
 
     # Properties lazily generated for efficiency.
-
 
     def __init__(self, matrix):
         """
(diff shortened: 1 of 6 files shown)


mkhorton:
> One thing to note is that previously miller_index_from_sites required the lattice matrix as input, whereas now it requires a Lattice object. 

I know this was discussed in a previous PR, but since this requires a `Lattice` object, perhaps it should be a method on Lattice? I think a convenience method on Structure might also be really nice, `.miller_index_from_site_indices()` maybe.

Thinking about discoverability for new users and where they'd expect to find this information, since even though it's useful for surface, there are many applications not including surfaces too.

utf:
My use case for multisite support is actually nothing to do with surfaces, so moving the code to Lattice would make sense to me. Especially, as the index for a set of sites is specific to a particular lattice.

Happy to implement the convenience method in structure too.

mkhorton:
That'd be great, thank you!

This may be beyond the scope of this present PR, but I think there'd be value of a MillerPlane class inside lattice.py -- I could imagine methods like `.get_in_hexagonal_notation()`, `.get_family()`, `to_latex_string()`, `.get_cartesian_plane(lattice)` (this last method would be *very* useful for our visualization tools), `.get_d_spacing(lattice)`, etc.

mkhorton:
Great, thank you! Is this ready to merge once tests pass?

utf:
Yep, it should be.

mkhorton:
Thanks for docstring / backwards compat improvements. Can see AppVeyor test issues are unrelated so will merge once Travis passes.

PR #1304, opened 2018-10-23
Fix defect_generate_structure issue with velocity site property

dbroberg:
Previously a Substitutional or Interstitial defect structure could be generated (from generate_defect_structure) with a velocity property which was not consistent with the initial bulk_structure. 

This PR fixes this issue for the Substitution and Interstitial objects and adds a unit test to ensure the fix works.

### DIFF
--- a/pymatgen/analysis/defects/core.py
+++ b/pymatgen/analysis/defects/core.py
@@ -205,11 +205,26 @@ class Substitution(Defect):
         defect_structure = self.bulk_structure.copy()
         defect_structure.make_supercell(supercell)
 
+        # consider modifying velocity property to make sure defect site is decorated
+        # consistently with bulk structure for final defect_structure
+        defect_properties = self.site.properties.copy()
+        if ('velocities' in self.bulk_structure.site_properties) and \
+            'velocities' not in defect_properties:
+            if all( vel == self.bulk_structure.site_properties['velocities'][0]
+                    for vel in self.bulk_structure.site_properties['velocities']):
+                defect_properties['velocities'] = self.bulk_structure.site_properties['velocities'][0]
+            else:
+                raise ValueError("No velocity property specified for defect site and "
+                                 "bulk_structure velocities are not homogeneous. Please specify this "
+                                 "property within the initialized defect_site object.")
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks Danny!

Could you imagine this issue recurring with other site_properties, not just velocity?

mkhorton:
Also saw that all tests pass here, but we're just started getting errors on AppVeyor from the defects module. Seems pretty weird, related to matplotlib, so might be completely unrelated to this PR. Do you have any ideas?

https://ci.appveyor.com/project/shyuep/pymatgen/builds/19763331

```
File "C:\projects\pymatgen\pymatgen\analysis\defects\tests\test_corrections.py", line 14, in <module>
2011    from pymatgen.analysis.defects.corrections import FreysoldtCorrection,\
2012  File "C:\projects\pymatgen\pymatgen\analysis\defects\corrections.py", line 13, in <module>
2013    import matplotlib.pyplot as plt
...
ImportError: cannot import name 'get_backend' from 'matplotlib' (c:\miniconda36-x64\lib\site-packages\matplotlib\__init__.py)
```

dbroberg:
I don't know what the issue is with the matplotlib loading issue. I haven't seen that before.

With regard to other site_properties, I don't anticipate any other issues.

Honestly, this was a hack to try and fix a larger issue we have run into when using atomate/custodian related functionalities which require loading the VaspInput set before running a VASP calculation (ex. [here](https://github.com/materialsproject/custodian/blob/d6276f37009eafd3806b6d73ed29c53acf1bbadd/custodian/vasp/jobs.py#L200)). The problem is that if the POSCAR has an incomplete velocities set then the VaspInput loading breaks (Note that VASP doesn't care about velocities if they are all zero). This is why I issued a "[read_velocities](https://github.com/materialsproject/pymatgen/blob/945d82c9118356aaec68fdbc4a9d118fe9c39518/pymatgen/io/vasp/inputs.py#L198)" fix a while back. Perhaps a better fix would be for the VaspInput set to have "read_velocities=False" for default POSCAR loading? Open to other strategies as well.

PR #1307, opened 2018-10-25
Boltztrap2 for emmet

fraricci:
BandstructureLoader supports single spin component in order to calculate the relative dos.
A merge function to get a complete (up/down) dos has been added too.

This commit allow to run Doses for MP with Boltztrap2DosBuilder in Emmet.


shyuep:
Need unittests...

shyuep:
The test files are not in the right place. The tests are failing because of that.

PR #1313, opened 2018-11-03
fixed the accidental overide of get_min_instability

jmmshn:
fixed the accidental override of get_min_instability
I also realized that there is no test for the get_min_instability and get_max_instability functions, and I will write them as part of this pull request, I just want to let you know that a typo was made on the last one that was not caught by the tests.


mkhorton:
Ok, thank you.

jmmshn:
The tests have been added.

PR #1317, opened 2018-11-05
CrystalNN takes into account ionic radii

computron:
## Summary

If a structure has oxidation states set, CrystalNN will try to use ionic radii instead of covalent/atomic radii.

* PeriodicTable has two new attributes per element: ``average_cationic_radius`` and ``average_anionic_radius``. These are used in the new CrystalNN, and for elements like N, P, S etc that can act as both anions and cations, it is useful to have an average radius that doesn't mix together the positive and negative valence states.
* CrystalNN: if ``distance_cutoffs`` are set, CrystalNN will throw a warning if oxidation states are not set, although it will still work without oxidation states (just generally poorer results)
* CrystalNN: if ``distance_cutoffs`` are set, CrystalNN will use oxidation state data from the structure to set better ionic radii for those distance cutoffs.

### DIFF
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -3370,6 +3370,15 @@ class CrystalNN(NearNeighbors):
 
     @staticmethod
     def _get_default_radius(site):
+        """
+        An internal method to get a "default" covalent/element radius
+
+        Args:
+            site: (Site)
+
+        Returns:
+            Covalent radius of element on site, or Atomic radius if unavailable
+        """
         try:
             return CovalentRadius.radius[site.specie.symbol]
         except:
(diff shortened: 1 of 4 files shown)


computron:
Note: the failing test on CircleCI looks to be due to gulp_caller and unrelated to this PR:
```
======================================================================
ERROR: test_get_energy_relax_structure_buckingham (pymatgen.command_line.tests.test_gulp_caller.GlobalFunctionsTest)
  File "/root/pymatgen/pymatgen/command_line/tests/test_gulp_caller.py", line 229, in test_get_energy_relax_structure_buckingham
    enrgy, struct = get_energy_relax_structure_buckingham(self.mgo_uc)
  File "/root/pymatgen/pymatgen/command_line/gulp_caller.py", line 612, in get_energy_relax_structure_buckingham
    gout = gc.run(gin)
  File "/root/pymatgen/pymatgen/command_line/gulp_caller.py", line 539, in run
    raise GulpError(err)
pymatgen.command_line.gulp_caller.GulpError: GulpError : STOP GULP terminated with an error
```

PR #1319, opened 2018-11-07
allow choosing element to add

hello-underworld:
For the ChargeDensityAnalyzer function, I was trying to insert a certain specie to the structure but original version only allows a dummy specie 'X'. Changed the default to 'X' but allowed choosing an element to add as well. Thanks.


shyuep:
I need a unit test for this functionality. Thanks.

PR #1324, opened 2018-11-10
Add query-chunking capability to MPRester.query

dwinston:
## Summary

Add a `bulk_query` method to `MPRester`, which wraps `query` to provide a better user experience for fetching a large sum of data across many materials. Fixes issue where user requests not too much data per material, but too much data overall in one query because the number of materials matching criteria is large.

* Chunk a query into multiple queries by first fetching a list of material ids satisfying the criteria.
* Re-try fetching chunks a configurable number of times in case of 5xx (e.g. timeout) errors.
* Improves user experience when fetching many material documents by providing a progress bar (uses `tqdm_notebook` if `tqdm` is installed and user is in Jupyter Notebook, falls back to plain `tqdm` if installed, and otherwise uses simple `print`-based progress indication.
* Puts general chunking and progress-bar sequence utilities in `pymatgen.util.sequence`.
* Adds unit tests for the above.

### DIFF
--- a/pymatgen/ext/matproj.py
+++ b/pymatgen/ext/matproj.py
@@ -9,6 +9,7 @@ import itertools
 import json
 import re
 import warnings
+from time import sleep
 
 from monty.json import MontyDecoder, MontyEncoder
 from six import string_types
@@ -25,6 +26,8 @@ from pymatgen.entries.computed_entries import ComputedEntry, \
 from pymatgen.entries.exp_entries import ExpEntry
 
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
+
+from pymatgen.util.sequence import get_chunks, PBar
 
 """
 This module provides classes to interface with the Materials Project REST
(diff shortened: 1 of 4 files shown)


dwinston:
FYI two of the three Codacy checks seem misleading. They complain about unused imports, but I think they are only unused by the Codacy environment. If a user has `tqdm` installed, one of the imports will be used, and if a user is in a Jupyter notebook with `tqdm` installed, the other import will be used.

As for the third Codacy check, I formatted the module doc as a multiline string rather than comment lines because this is the style for all existing modules in the `pymatgen.util` package -- I think refactoring all to comment lines would be appropriate for a separate PR/commit.

shyuep:
Is there a reason why this cannot be just simply be part of the query method?

shyuep:
E.g., chunk size can be zero by default, which is usual query behavior. And anything larger will be a bulk type query.

shyuep:
I leave it to your best judgement

fxcoudert:
@dwinston in my testing over network, it seems chunk sizes between 200 and 500 are “safe” (they never time out) and are rather optimal (the overhead of queries is negligible)

dwinston:
Okay @shyuep I'll leave the chunk_size default as 500. You can merge. Thanks for the feedback @fxcoudert.

PR #1327, opened 2018-11-13
Class to handle IR spectra

henriquemiranda:
The new class contains the IR oscillator matrix elements.
Allows plotting the IR spectra with user-defined broadenings.
The oscillator matrix elements are computed with DFPT as implemented in Abinit.
It will be possible to start this class from the MPRester once the endpoints are created.


shyamd:
Hi Henrique,

A few things that would make this easier for everyone else to use:
1.) Could you change the variable names to be more explicit? It took me a while to figure out what epsinf was until I realized it was the high-frequency dielectric limit. Still, this would have been easier if it was just named epislon_infinity or dielectric_hf or even dielectric_high_frequency.

2.) Try subclassing the core Spectrum class for your IRSpectrum

3.) It might be easier to eliminate the generator class and just have a class method in the spectra to generate it from the oscillator strength, phonon modes, etc. You could even store these in the spectrum and have a method to generate new ones using different broadening coefficients. 
This makes the IRSpectrum more descriptive. I think we want to stick to Material Science concepts where possible and only introduce CS terminology when useful for the application.

henriquemiranda:
Hi Shyam,

Thanks for your comments.
1) I changed epsinf to epsilon_infinity and added a docstring with a link to the paper explaining what it actually is.
2) The IRSpectra is not one single spectrum but a 3x3 frequency dependent tensor. I don't see how to subclass Spectrum in a proper way. I could create multiple instances of the Spectrum class inside IRSpectra but it is not worth it since I am not going to expose any methods of Spectrum. At most I can add a method to return a Spectrum instance for a single component of the tensor.
3) I can merge the two classes into one if that is what you mean. I would prefer that but then we don't have the plotting class separated from the main class as seems to be the rule in other parts of pymatgen. That's the reason I separated in two classes in the first place.

mkhorton:
I second @shyamd's comments, especially with respect to adopting the `pymatgen.core.spectrum` class. A lot of spectra require a common set of base features (e.g. broadening etc.), so it makes sense to build as much as possible from this core class. This is especially true since it'll mean that creating our future web apps will be easier if you adopt this base class.

henriquemiranda:
@shyamd 

1) I am sorry but I have to disagree with changing phfreqs_gamma to phonon_frequencies_gamma. It's the name of a variable in a code, not a textbook sentence. It is customary to use abbreviations. I added docstrings that were missing in the previous commit and explain the meaning of the variable. Adding long names of variables will cause problems with maintaining the number of characters per line below a reasonable number. (We can discuss this better during the next MP update meeting if you want).

2) Thanks for pointing it out. I addressed these issues in the last commits.

3) I added checks for the components should be in 'xyz'.
With respect to adding all the spectra, what exactly do you mean? Adding all contributions from the different phonon modes? This is what the code is doing already with numpy arrays.

With respect to whether we should show the real or imaginary part, which type of experiments do you have in mind where one measures the real part of the dielectric screening? If the experiment is IR absorption then the relevant part is the imaginary part. See for example here: https://aip.scitation.org/doi/10.1063/1.3690064

I modified the IRDielectricTensorGenerator to IRDielectricTensor and I am now creating multiple instances of the Spectrum class and returning a SpectrumPlotter instance. I also added a function to plot directly using the SpectrumPlotter class.

@mkhorton The broadening implemented in the Spectrum class (convolution with a Gaussian function) is different from the one we are using in IRSpectra. The broadening is related to the phonon lifetimes and can be different for each mode. As such we have implemented the option to pass a list of broadenings. This is not possible with the smear function implemented in the Spectrum class.

mkhorton:
> @mkhorton The broadening implemented in the Spectrum class (convolution with a Gaussian function) is different from the one we are using in IRSpectra. The broadening is related to the phonon lifetimes and can be different for each mode. As such we have implemented the option to pass a list of broadenings. This is not possible with the smear function implemented in the Spectrum class.

@henriquemiranda duly noted! Makes total sense you'd need to subclass + add additional IR-specific functionality, I just wanted to encourage use of the base class where possible + adding any features that might be appropriate for other spectrums to the base class directly. But of course this will not always be sensible.

shyuep:
I would suggest modifying the Spectrum's class smear function to enable different smearing options, rather than just ignoring the existence of the Spectrum class altogether. There is a good reason to use the same basic framework as it simplifies a lot of development for things like additions, plotting etc. Smearing is hardly a unique thing to phonons.

As for variable names, I am fine with clear abbreviations, but I would suggest adding an underscrore e.g., `ph_freq_gamma`.

henriquemiranda:
@shyuep We already discussed that above and I explained why I did not inherit from the Spectrum class.

shyuep:
Ok. But can you modify the variable names?

mkhorton:
I'm sorry but I still do not understand why not to use the `Spectrum` class? If you need to over-ride the broadening with your own broadening that's fine, or even if the base class itself needs to be modified to be more versatile that also seems fine.

This class is already, in fact, creating a `Spectrum` object on the fly in order to be able to plot it, yet the end user will have no access to the object itself, which seems frustrating.

henriquemiranda:
@mkhorton What don't you understand?

mkhorton:
* `get_ir_spectra` could return a `Spectrum`
* `get_plotter` could then be simplified
* In any case, if you want to customize the keyword args you give to `get_ir_spectra` *and* also plot the resulting spectra, there seems to be no way to do this (e.g. to change the broadening)
* The properties `max_phfreq`, `nph_freqs`  have no docstrings, and it's not obvious what they mean since they're heavily abbreviated

henriquemiranda:
Well, I don't know how to implement what you are asking, sorry.

mkhorton:
Ok, firstly, thanks for adding the docstrings!

Regarding my comment, to be more specific, you have a function `get_ir_spectra` which returns a tuple of x, y values, and I'm asking that it return `Spectrum` object instead. This is what the `Spectrum` class is designed for, as a light-weight object to hold x, y values (and labels). It offers some convenience methods etc. but the benefit is to make the intent of the code clearer, especially as we're adopting static type checking more. It's going to be useful if all plottable spectrum-like data in pymatgen inherits from a common base class, since it's going to make it easier for us to refactor this functionality in the future.

mkhorton:
Ah, wait, so the second return parameter is a tensor, so perhaps this would have to be split between two separate methods (one for the tensor, one to return the `Spectrum` for specific component). My point more is that you *are* already constructing the `Spectrum` object on line 172 anyway, it's just that this is hidden inside the body of a method.

mkhorton:
Would the pymatgen.core.tensor.Tensor class be useful for you?

henriquemiranda:
@mkhorton I am still a bit confused sorry. Shall I use the pymatgen.core.tensor.Tensor class?

mkhorton:
@henriquemiranda, ideally, yes. From an architecture point of view, it is better if a method that returns a tensor actually returns a `Tensor` type rather just an ndarray, and if a method returns a spectra (a set of x, y values + x, y labels) then it can return a `Spectrum` type. This makes it easier if we want to add common functionality, plotting, type checking etc.

I have no concerns about code quality or correctness with this PR -- in that respect, it looks very nice.

henriquemiranda:
@mkhorton I think the last commit addresses your concerns about using the Spectrum class. As for me the branch is ready to be merged.

mkhorton:
Thanks @henriquemiranda !

PR #1329, opened 2018-11-13
Tools for EPH calculations

gmatteo:
## Summary

* Add workflows for EPH calculations (require dev-version of Abinit)


shyuep:
I accidentally clicked merge. I have reverted this PR. Pls implement proper unittests to show that all new functionality work as intended.

shyuep:
Hi Matteo,
I would suggest we separate out the workflow portions of the abinit code from the analysis part. In other words, code that is required to generate ABINIT input or read ABINIT output and do analysis from ABINIT should be in pymatgen. Code that is basiclly workflows for ABINIT should be in atomate (or abipy if you prefer). This is the structure we have used for VASP, Qchem and other electronic structure codes. For example, the automation of electron-phonon calculations sounds like a workflow to me.

Looping in @mkhorton and @computron for comments.

computron:
Hi

I would suggest abipy, since I think you will have more control over the code approval process / etc. inside abipy than atomate, and atomate workflows follow their own standard on how to construct workflows.

mkhorton:
I think this is difficult. It's obviously not feasible to run a full abinit software stack inside the pymatgen CI (or the atomate CI for that matter). I agree workflow code should not be in pymatgen + we should stick to file I/O and analysis only. If this is just analysis or plotting code, that's fine, but workflows are difficult to maintain.

gmatteo:
Hi all,

I agree with your analysis and the fact that workflow code should not be in pymatgen
but implemented in higher-level packages.
This is indeed the approach we are using now to develop abiflows that depends on the abipy/pymatgen/fireworks/mongodb software stack.

Unfortunately, we didn't use this philosophy when we started to work on the low-level API for Abinit workflows provided by pymatgen.io.abinit.

We kept using the old approach in order not to break code relying on pymatgen.io.abinit but I understand that the situation is far from being optimal and that a serious refactoring of the abinit modules is really needed.

At this point, I would suggest to merge my PR so that I can start to work on the separation between 
the workflow code that will be moved to AbiPy and the low-level tools related to IO that will be kept in pymatgen.

The refactoring will take some time because these changes will affect other packages.
I will send you another PR with the changes you are proposing when I'm done with 
the refactoring in pymatgen and all the other packages have been updated.

shyuep:
OK, sure. Can you resubmit the same PR? BEcause I reverted the earlier PR. SEnd it to me again, and I will just merge it for now since it does not impact code in other modules. And we can take some time refactoring out some of the workflow pieces.

PR #1337, opened 2018-11-20
Fix bug in analysis.graphs

mkhorton:
Ensure `jimage` kwarg is respected in `get_connected_sites`


mkhorton:
Thanks to @utf for reporting

PR #1338, opened 2018-11-21
Implementation of VASP's PAW PSP recommendations

HanmeiTang:
## Summary

VASP recently added the recommended PAW potentials for DFT calculations using vasp.5.2 to the manual on Oct 28, 2018. This PR adds a new input set class `MVLRelaxSet` to utilize the recommended potentials.

* Added config file "MVLRelaxSet_52.yaml" to implement VASP's PAW PSP recommendations, while INCAR & KPOINTS settings are consistent with `MPRelaxSet` (credit to @zbwang)
* Added `MVLRelaxSet` class
* Added unittests
* Minor documentation change and reformatting

Links:
1. [Recommended PAW potentials for DFT calculations using vasp.5.2](http://cms.mpi.univie.ac.at/vasp/vasp/Recommended_PAW_potentials_DFT_calculations_using_vasp_5_2.html)
2. [Release notes](http://cms.mpi.univie.ac.at/vasp/vasp/New_features_added.html#new-features)


shyuep:
Can this already be done by supplying a PSP argument to MPRelax? Other than the PSP choice, what other parameters are changed?

HanmeiTang:
The PSP choice is the only change made, which results in 21 elements different from`MPRelaxSet`. This class uses either `PBE_52` or `PBE_54` according to the VASP manual and it is equivalent to the following block:

```python
from pymatgen.io.vasp.sets import DictSet, _load_yaml_config
vis = DictSet(structure, _load_yaml_config("MPRelaxSet_52"), potcar_functional="PBE_52")
```
I tried to find easier ways to apply the new settings, such as:
1. Adding the config file only and let the user create a temporary DictSet when using; 
2. Using "user_potcar_settings" where it is needed, etc.

But my concern is this isolated config file will be easily neglected by users, so I think it is better to wrap it in a class. Indeed it will be an issue if VASP continue to release new recommendations.. but so far the frequency seems to be very low. I agree that it is not the best way to implement these settings and I can make changes if you have any suggestions.

shyuep:
So how do you use this if you need to compare with MP data for phase stability?

HanmeiTang:
I don't think this is a problem since only a few standard settings are compatible with MP. Calculations using SCAN, HSE or with vdW corrections cannot be compared with MP entries directly. Now we don't have any benchmark for these settings, but if it turns out to a significantly improved performance, we can get some correction to make it compatible with MP with enough test cases.

shyuep:
Not just config. Also the name of class.

HanmeiTang:
Sure. It is done.

PR #1343, opened 2018-11-28
ScaleToRelaxedTransformation

richardtran415:
New Transformation class. This transformation takes an initial unrelaxed structure and a corresponding relaxed structure. It determines a matrix containing the percent change in each scalar in going from the initial lattice matrix to the relax lattice matrix. 

When using apply_transformation(), a similar structure (in regards to crystal, spacegroup, site positions) is taken in as well as a species map. Using the percent change matrix as well as the relaxed site positions, the lattice and site positions of the input structure will be scaled according to the relaxed structure. 

This is useful for finding an initial structure closer to its final structure assuming the input unrelaxed/relaxed structure is the same type of structure as the structure one wishes to apply the transformation to.


mkhorton:
Hi @richardtran415, this looks useful! My only concern is that the `apply_transformation()` method signature is different from the other transformations. This should either just be `apply_transformation(structure)` or `apply_transformation(structure, return_ranked_list=...)` (in the case of one-to-many transformations). This is what Codacy picked up on too.

I understand this is awkward however ... you could either move `species_map` to the `__init__()`, or maybe try and automatically calculated the species_map using `StructureMatcher`? Not sure if you have any ideas.

richardtran415:
Hi @mkhorton ,

Thanks a lot for the feedback and suggestions. I decided to go with the best of both worlds, the user can now input an optional species_map dict in __init__() which is None by default. If none, StructureMatcher will then do the mapping for the user when calling apply_transformation

mkhorton:
Sweet, sounds great! Thanks

mkhorton:
@richardtran415 are you happy for this to be merged? (if not, can you add [WIP] to the title until you are happy with it)

richardtran415:
@mkhorton 

I'm satisfied with it for now, it should be ready for merging. Thanks!

PR #1345, opened 2018-11-29
Properly catch non 0D graphs in zero_d_graph_to_molecule_graph

utf:
## Summary

An exception is now thrown when trying to convert non-zero dimensional graphs to `MoleculeGraph`s.

A test to ensure this behavior has been added.


utf:
Please don't merge this as I have some further commits.

utf:
If the tests pass then this should be ok to merge.

mkhorton:
Cool, thanks! It'd be useful to add [WIP] to the title of future PRs if they're not ready to merge, just to communicate explicitly that you're still working on them.

PR #1349, opened 2018-11-29
Update MPRester to load endpoint from PMG_MAPI_ENDPOINT

mkhorton:
## Summary

To support setting the MPRester endpoint via an env variable (for testing purposes).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,8 @@ jobs:
     working_directory: ~/pymatgen
     docker:
       - image: materialsvirtuallab/circle-ci-pmg-py3:0.0.2
+        environment:
+          PMG_MAPI_ENDPOINT: https://zola.lbl.gov/rest/v2
     parallelism: 4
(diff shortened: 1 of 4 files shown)


shyuep:
Given that we are going to expose the endpoint, we might as well just hard code the endpoint in the tests themselves? i.e., initialize MPRester with endpoint=zola?

mkhorton:
I set it as an environment variable because there are some tests that use the MPRester that aren't in ext.matproj (e.g. the Pourbaix pipeline tests). So it seemed easier just to set it as an environment variable so all tests would use the alternative endpoint by default.

That said, arguably it's bad form that we're using the MPRester in tests outside ext.matproj, maybe the Pourbaix tests should be serialized somehow instead.

mkhorton:
In terms of exposing the endpoint, if we do end up having load issues on this endpoint or people using it when they shouldn't, perhaps we can restrict access to just a subset of API keys/users.

PR #1352, opened 2018-12-06
Add IUPAC ordering for Composition formulae

utf:
## Summary

The IUPAC red book details a system for ordering elements in chemical formulae. The ordering is detailed in Table IV shown below. The ordering produces formula slightly different to ordering by electronegativity. For example, `H6CN` goes to `CNH6`. 
screen shot 2018-11-29 at 3 21 58 pm

This PR:

- Adds the IUPAC ordering to the periodic table. I.e. Each element now has an `iupac_ordering` attribute.
- Adds the attribute `iupac_formula` to `Composition`.
- Adds the option `use_iupac_ordering` to `get_reduced_formula_and_factor` and `get_integer_formula_and_factor`. This defaults to `False`.
- Adds tests for the functionality in both `test_composition` and `test_periodic_table`.
- Adds a script to regenerate the IUPAC ordering in the `periodic_table.json`.

One note about the implementation is that (as per the IUPAC red book) polyanions are still determined by their absolute electronegativity. For example, the formula "Li3Fe2(MoO4)3"  remains the same when using the IUPAC ordering. If the table above was strictly followed the formula would instead be written "Li3Mo3(FeO6)2", which is incorrect.


shyuep:
Isn't this almost like ordering by Mendeleev number? The Pettifor Mendeleev number effectively goes in the same manner (specific orderings may differ a little).

utf:
It is very similar but the main difference is the position of Hydrogen which is exactly what this ordering aims to fix.

PR #1358, opened 2018-12-07
Bug fix in core.composition

jmmshn:
to_data_dict['element'] returned <dict_keys> which is a generator and cannot be pickled


shyuep:
Can you add a unitest. THanks.

PR #1359, opened 2018-12-07
Remove logging.basicConfig call in graphs.py

utf:
## Summary

`graphs.py` currently has a call to `logging.basicConfig` to set the logging level. This will affect the top level logger rather than the module level logger. The code now sets the logging level of the module level logger.


mkhorton:
Thanks!

PR #1360, opened 2018-12-08
Fix integer miller indices

utf:
## Summary

The `get_integer_index` function for miller indices now correctly returns integer indices when possible (if an integer index cannot be determined a floating point index is given).

E.g. `get_integer_index((2, 2, 0))` now gives `(1, 1, 0)` rather than `(1.0, 1.0, 0)`.


mkhorton:
Thanks!

PR #1362, opened 2018-12-10
Add option to return graph for structure components

utf:
## Summary

Added the `inc_graph` option to `get_structure_components()` to allow returning of networkx graphs for structure components.

The graphs have their nodes indexed to match the returned component `Structure` objects rather than initial structure passed to `get_structure_components`.

### DIFF
--- a/pymatgen/analysis/dimensionality.py
+++ b/pymatgen/analysis/dimensionality.py
@@ -27,7 +27,9 @@ import numpy as np
 
 from collections import defaultdict
 
-from pymatgen.analysis.graphs import MoleculeGraph
+from networkx.readwrite import json_graph
+
+from pymatgen.analysis.graphs import MoleculeGraph, StructureGraph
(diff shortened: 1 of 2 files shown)


mkhorton:
Doesn't the MoleculeGraph itself contain the networkx graph as an attribute?

utf:
The molecule graph is only returned when the component is zero-dimensional. I agree that this method is getting a little messy. I’m tempted to rewrite it so that instead of a structure it always returns a StructureGraph.

mkhorton:
Yeah, I'd prefer not to directly expose the networkx graph if possible for exactly that reason (messiness). The design intent between structure/molecule graph is that the node indices will always be matched to the indices of the underlying structure/molecule too, so modifying the networkx nodes shouldn't be necessary.

utf:
Ok, the new implementation is a little cleaner.

mkhorton:
Thanks! Will merge when tests pass

PR #1363, opened 2018-12-11
Added all_in_polar option for Polarization class

blondegeek:
I added an all_in_polar option for the Polarization class and updated the tests to be for the option where all_in_polar= False. all_in_polar uses the polar lattice to calculate "proper" polarization differences. 

There is usually little or no difference between the output of the Polarization class methods using all_in_polar=True vs. all_in_polar=False (because we choose the polarization branch closest to [0., 0., 0.] in the polarization lattice), but using all_in_polar is more correct. The biggest differences will be seen for examples where the nonpolar structure has a half quantum of polarization along any lattice direction.

I have kept the tests with all_in_polar=False to match the old test values and because all_in_polar=True makes certain tests trivial (get_lattice_quanta).


mkhorton:
Thanks! Is there a test for the new functionality? (i.e. with `all_in_polar=True`)

blondegeek:
Additional tests added!

blondegeek:
Looks like continuous integration is failing, but at least from my read of the error it is not related to my code.

mkhorton:
Thanks @blondegeek , it's not clear to me why the AppVeyor tests are suddenly failing either but I agree I don't think it's your code so will go ahead and merge this

PR #1366, opened 2018-12-17
Added Trajectory class for MD trajectory analysis 

sivonxay:
## Summary

Added a Trajectory class which decreases memory overhead for working with MD trajectories in pymatgen. Here, the trajectory is stored as a singular structure and a set of displacements, which allow the positions at any timestep to be reconstructed. 

* Added Trajectory Class
* Added MPMDSet


shyuep:
It is unclear to me what this is supposed to do, given that diffusion_analyzer pretty much handles a lot of this. And the construct of this Trajectories class violates a lot of python internal constructs for lists/sequences. In particular, having functions like `change_index` to change indices is definitely not normal behavior.

shyuep:
I am not disputing there is a need for a trajectories class. What I mean is that DiffusionAnalyzer already stores it as a Structure + Displacements as a numpy array. That is already an extremely efficient mechanism. So I'd rather this be a class under the diffusion analysis for now, and DiffusionAnalyzer be modified to work with this new trajectories class.

The comment remains that this class, though it is essentially a sequence, violates standard python protocols for sequences. This needs to be fixed too.

shyamd:
@shyuep ,

DiffusionAnalyzer suggest it is there just for diffusion. That's about all people read and move on, e.g: we knew about DiffusionAnalyzer and chose to write a trajectory class because we wanted to do sampling for amorphous configurations and RDFs, etc. It's just not designed for generic trajectory analysis. Trajectory should be a core class in pymatgen just like structure and then DiffusionAnalyzer should employ Trajectory. One is an analysis method and the other is a data type. 

I agree we could fix the python paradigm. My proposal is this:

Trajectory 
- inherits from numpy array, but also decorates a structure
- need some way for attribute trajectories as well
- provides an iterator that yields a pymatgen structure for each step
- provides accesssor for a unique index
- provides basic concat and slice functionality

shyuep:
Fine. Just a note that RDF analysis is already implemented in pymatgen-diffusion. Whatever the case may be, I am fine with putting trajectories in core and moving some of the RDF (and van Hove analysis) into pymatgen. But we should not be reinventing the wheel.

mkhorton:
@sivonxay Discussion of RDF aside for the moment, as far as I can tell this class assumes the lattice remains constant? Could we also allow this to change by storing a list of `.deformations`? As far as I know, some of your MD runs do not have constant volume.

sivonxay:
Yeah, I think storing deformations is a good idea. I'll include it in the new class.

mkhorton:
Thanks, that'd be great :)

sivonxay:
Closing this pull request. Will submit a new PR without the unintended changes.

PR #1369, opened 2019-01-07
Add defect sublattice generator and point defect matcher

dbroberg:
This PR adds a few essential features for defect analysis which relate to the use of a new sub-lattice generator.  The newly added create_saturated_interstitial_structure function generates the sub-lattice of an interstitial defect object via a "saturated interstitial structure". This is done by generating a new structure with a fully occupied sub-lattice for the defect (under the bulk_structure's space group symmetry). 

This is useful for determining the multiplicity of an interstitial defect (for thermodynamic analysis) as well as for comparing interstitial objects which might be at different cartesian coordinates, as is implemented in the PointDefectComparator class (also included in this PR). 

Also adding explicit as_dict and from_dict methods for DefectEntry objects.


dbroberg:
Sorry - having git issues. I will reopen this when I fix it.

PR #1372, opened 2019-01-12
PARCHG from WAVECAR

mturiansky:
## Summary

Adds the function "get_parchg" to the Wavecar class. This function outputs a Chgcar object that corresponds to the charge density of the specified wavefunction. This emulates the PARCHG output from VASP, but it does not include augmentation from the PAWs.

* PARCHG from a Wavecar object
* For real wavefunctions, adds the ability to visualize the phase of the wavefunction.

N/A

## TODO (if any)

N/A


mkhorton:
This is really nice, thank you 👍

PR #1374, opened 2019-01-17
set point_group attribute from sgencodings and test point_group attri…

blondegeek:
The SpaceGroup.point_group attribute was not being set if int_symbol was not in SYMM_OPs. This would cause an error when calling for the point_group attribute for space groups with underscores in their name, for example, space group 14 with symbol 'P12_1/c1'.


mkhorton:
Thanks! Will merge once tests pass

PR #1375, opened 2019-01-18
Polarization cartesian fix

blondegeek:
VASP reports dipole moments in along Cartesian directions. This pull requests adds the ability to specify when creating a Polarization object whether ionic and electronic contributions to the polarization are given in Cartesian or lattice directions. The default for ionic contributions is lattice directions (since it is recommended to use the calc_ionic function as described in the documentation) and the default for electronic contributions is Cartesian (since these are read directly from VASP). I have added a test checking that the conversion between Cartesian to lattice directions is computed correctly and added additional documentation detailing these conventions.

### DIFF
--- a/pymatgen/analysis/ferroelectricity/polarization.py
+++ b/pymatgen/analysis/ferroelectricity/polarization.py
@@ -29,19 +29,25 @@ the distortion.
 See Nicola Spaldin's "A beginner's guide to the modern theory of polarization"
 (https://arxiv.org/abs/1202.1831) for an introduction to crystal polarization.
 
+VASP reports dipole moment values (used to derive polarization) along Cartesian 
+directions (see pead.F around line 970 in the VASP source to confirm this).
+However, it is most convenient to perform the adjustments necessary to recover
+a same branch polarization by expressing the polarization along lattice directions.
+For this reason, calc_ionic calculates ionic contributions to the polarization
+along lattice directions. We provide the means to convert Cartesian direction 
+polarizations to lattice direction polarizations in the Polarization class. 
+
 We recommend using our calc_ionic function for calculating the ionic
-polarization rather than the values from OUTCAR.
-
-We find that the ionic dipole moment reported in OUTCAR differ from
-the naive calculation of \\sum_i Z_i r_i where i is the index of the
-atom, Z_i is the ZVAL from the pseudopotential file, and r is the distance
-in Angstroms along the lattice vectors.
-
-Compare calc_ionic to VASP dipol.F. SUBROUTINE POINT_CHARGE_DIPOL.
-
-We are able to recover a smooth same branch polarization more frequently
-using the naive calculation in calc_ionic than using the ionic dipole
-moment reported in the OUTCAR.
+polarization rather than the values from OUTCAR. We find that the ionic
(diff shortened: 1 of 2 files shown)


utf:
The `convert_cart_to_latt` looks like a really useful function. I know I've wanted something like this in the past (for something unrelated to polarization).

It seems to me like it could be a function in the Lattice class instead?

@mkhorton Thoughts?

shyuep:
That method is already in Lattice. It is called get_catesian_coord

—
Shyue Ping

utf:
As far as I know, neither `get_cartesian_coords` or `get_fractional_coords` do the same as `convert_cart_to_latt`.


- `get_cartesian_coords` takes a vector in fraction coordinates and converts it to cartesian coordinates.
- `get_cartesian_coords` takes a vector in cartesian coordinates and converts it to fractional coordinates.
- Whereas `convert_cart_to_latt` takes a vector in cartesian coordinates and converts them to be along lattice directions.

E.g.: If the cartesian and lattice directions align then the input and output vectors will be the same but if the cell has angles != 90 degrees, then the vector will different.

mkhorton:
Yes, I added something very similar to `convert_cart_to_latt` in the `Magmom` class:

https://github.com/materialsproject/pymatgen/blob/ac210ba17db2d3b8c88d988bed6275c421e47dc5/pymatgen/electronic_structure/core.py#L398

https://github.com/materialsproject/pymatgen/blob/ac210ba17db2d3b8c88d988bed6275c421e47dc5/pymatgen/electronic_structure/core.py#L415

This has similar requirements to the dipole moments, so having common functionality in `Lattice` might be sensible.

Generally, working with vectors along lattice directions is more useful scientifically, it just seems like VASP prefers using Cartesian (I'm also thinking of revisiting that Magmom class to rework some of it, and in particular to switch to using vectors along lattice directions by default instead of Cartesian).

mkhorton:
(Also perhaps I'm missing something obvious, but I agree that this is a different use case to `get_cartesian_coord`)

blondegeek:
I agree with @utf and @mkhorton that `convert_cart_to_latt` is different than the other two functions mentioned. It could very easily be modified to be based on `get_fractional_coords` and just multiply by lattice vector lengths (you can see an example in the test_polarization.py file [here](
https://github.com/materialsproject/pymatgen/blob/master/pymatgen/analysis/ferroelectricity/tests/test_polarization.py#L41)). I can try modifying and make a pull request if that is preferred. Let me know.

shyuep:
I am a bit confused at what convert_cart_to_latt is supposed to do. My guess is that it is just the dot projection of the polarization in the three crystal directions?

blondegeek:
So it’s different than the projection of the polarization in the three crystal directions. As you guessed, the projection could simply be expressed as `np.einsum(‘ax,x->a’, lattice_mat, vector)` where `lattice_mat` is the lattice matrix and `vector` is the Cartesian vector. I’m not sure the best way to say what exactly it means to convert a Cartesian vector to be along lattice vectors, so here’s an image instead showing how it differs from a standard projection for a 2D lattice. Let me know what you think.
7b061efa-8c04-441f-a2ae-7f68043445ae

shyuep:
Yes. Thanks.

PR #1378, opened 2019-01-24
move convert_cart_to_latt to method get_lattice_coords of Lattice class

blondegeek:
See discussion on #1375.

### DIFF
--- a/pymatgen/analysis/ferroelectricity/polarization.py
+++ b/pymatgen/analysis/ferroelectricity/polarization.py
@@ -109,20 +109,6 @@ def get_total_ionic_dipole(structure, zval_dict):
     return np.sum(tot_ionic, axis=0)
 
 
-def convert_cart_to_latt(vector, structure):
-    """
-    Convert a vector along Cartesian directions to be along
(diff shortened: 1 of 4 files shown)


mkhorton:
Thanks @blondegeek ! Haven't merged yet because I'm just mulling over that `get_lattice_coords` method name, I'm not sure if that might cause confusion for people? I'm not sure what a better name for it might be though. The inverse method might also be helpful (get lattice coords from cart coords).

mkhorton:
Actually, for the method name, instead of renaming it maybe just adding an illustrative example to the docstring would help, something like "for example, this may be used to calculate a polarization vector..."

shyuep:
I think the method should be renamed. Lattice coords is far too confusing.

mkhorton:
I think that name's reasonable, I can't think of a better one in any case. As long as it avoids confusion with eg `Lattice.get_cartesian_coords` I think that's the main thing.

blondegeek:
Pending integration and code quality checks...done!

mkhorton:
Thanks!

PR #1379, opened 2019-01-24
Added Trajectory class for MD trajectory

sivonxay:
Reopened #1366 due to branch contamination.

@shyamd @mkhorton


mkhorton:
Is this ready to merge? Will have a read over if so.

sivonxay:
Yes, should satisfy the criteria @shyamd listed on the last PR.

mkhorton:
Ok, this looks like it'd be really useful to have in pymatgen! It seems like the comments from the last thread were addressed, thanks.

There are a few last items that need to be addressed before merging, most importantly docstrings need to be added. There is also some error checking commented out (the exceptions in `extend` -- these could also be `raise ValueError` for a more specific exception type). Finally, `combine_attribute` should probably be marked private (leading underscore) and put as a static method on `Trajectory` unless there are plans to use this elsewhere, in which case it would also need a test.

mkhorton:
Thanks @sivonxay !

PR #1383, opened 2019-01-31
Some odds and end fixes

shyamd:
## Summary

Few odd fixes
* Add Metagga types to run_type in Vasprun
* Compatability shouldn't index error for weird elements like Americium.


mkhorton:
Is this WIP ? Indentation error in some tests. (Also due credit stub was included; I think this is a good idea, I just wasn't sure whether it was intentional since it wasn't mentioned in the commit log)

shyamd:
Sorry, Duecredit is for another PR. I think I fixed that indentation.

Issue #1384, opened 2019-02-04
Bug in StructureGraph.remove_nodes method


lan496:
**Describe the bug**
StructrureGraph.remove_nodes method  do not seem to change attributes of a graph.

**Environments**
python 3.6.4
pymatgen==2019.1.24
networkx==2.2

**To Reproduce**
```
structure = Structure(Lattice.tetragonal(5.0, 50.0), ['H'], [[0, 0, 0]])
square_sg = StructureGraph.with_empty_graph(structure, edge_weight_name="", edge_weight_units="")
square_sg.add_edge(0, 0, from_jimage=(0, 0, 0), to_jimage=(1, 0, 0))
square_sg.add_edge(0, 0, from_jimage=(0, 0, 0), to_jimage=(-1, 0, 0))
square_sg.add_edge(0, 0, from_jimage=(0, 0, 0), to_jimage=(0, 1, 0))
square_sg.add_edge(0, 0, from_jimage=(0, 0, 0), to_jimage=(0, -1, 0))

square_sg.insert_node(1, "O", [0.5, 0.5, 0.25])
square_sg.add_edge(1, 0, from_jimage=(0, 0, 0), to_jimage=(0, 0, 0))
square_sg.add_edge(1, 0, from_jimage=(0, 0, 0), to_jimage=(1, 0, 0))
square_sg.add_edge(1, 0, from_jimage=(0, 0, 0), to_jimage=(0, 1, 0))
square_sg.add_edge(1, 0, from_jimage=(0, 0, 0), to_jimage=(1, 1, 0))

print(square_sg.graph.nodes)  # >>> [0, 1]
square_sg.remove_nodes([0])
assert len(square_sg.graph.nodes) == 1
>>> IndexError: list index out of range
```

I think this bug is due to the default behavior of networkx.relabel.relabel_nodes.
By default, relabel_nodes returns a relabeled graph object and does not relabel the nodes in place.
And I think remove_nodes is not set up to relabel in place.

ref: https://networkx.github.io/documentation/stable/reference/generated/networkx.relabel.relabel_nodes.html#networkx.relabel.relabel_nodes

PR #1385, opened 2019-02-04
fix relabel_nodes parameter in StructureGraph.remove_nodes

lan496:
Fix https://github.com/materialsproject/pymatgen/issues/1384

### DIFF
--- a/pymatgen/analysis/graphs.py
+++ b/pymatgen/analysis/graphs.py
@@ -523,7 +523,7 @@ class StructureGraph(MSONable):
         for correct, current in enumerate(self.graph.nodes):
             mapping[current] = correct
 
-        nx.relabel_nodes(self.graph, mapping)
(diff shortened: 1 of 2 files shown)


shyuep:
Pls add a unit test to ensure this bug does not recur. Thanks.

lan496:
I've added tests for checking if (StructureGraph|MoleculeGraph).(remove_nodes|insert_node) correctly update a graph.

In writing these tests, I also noticed that insert_node method also may have a bug about networkx:
(1) networkx.relabel.relabel_nodes does not relabel nodes in place by default.
(2) networkx.MultiDiGraph.nodes is not guaranteed to return nodes by in ascending order.

About (1), I've changed the code to set copy=False at networkx.relabel.relabel_nodes in remove_nodes and insert_node methods.

About(2), I've changed the code to sort explicitly the order of nodes, since insert_node method assumed self.graph.nodes is in ascending order.

PR #1398, opened 2019-02-19
Fix outrageous performance bug in Poscar.get_string

ExpHP:
```python
#!/usr/bin/env python3

import numpy as np
from pymatgen import Structure

N = 100000
s = Structure(np.array([[1,0,0],[0,1,0],[0,0,1]]), ['C']*N, np.zeros((N, 3)))
Poscar(s, velocities=np.zeros((N, 3)).tolist()).get_string()
```
Prepare to wait several millenia.

pymatgen has some bad cases of property abuse, but this takes the cake.  Writing a file had O(N^2) performance because `Poscar.selective_dynamics` is a property implemented in terms of another property that transposes a list of dicts for all atoms.

(and it *still* computes this transpose of dicts way more times than it should have to)

### DIFF
--- a/pymatgen/io/vasp/inputs.py
+++ b/pymatgen/io/vasp/inputs.py
@@ -478,11 +478,12 @@ class Poscar(MSONable):
             lines.append("Selective dynamics")
         lines.append("direct" if direct else "cartesian")
 
+        selective_dynamics = self.selective_dynamics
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks. Perhaps you can also highlight where there are othe property abuses, and we can try to fix those.

PR #1399, opened 2019-02-20
How best to handle PointGroupAnalyzer error when list from `_get_smallest_set_not_on_axis` is empty.

blondegeek:
This pull request has a question at the bottom of the post.

For the following Molecule, with the default parameters, PointGroupAnalyzer throws an error from the list returned [`self._get_smallest_set_not_on_axis(axis)` being empty](https://github.com/materialsproject/pymatgen/blob/master/pymatgen/symmetry/analyzer.py#L1079) but gives the correct point group when increasing `eigen_value` to `2e-2`.

```
8
H2 C6
C 0.043026 -3.170664 -0.017826
C 0.025377 -1.961567 -0.006592
C 0.006889 -0.608728 0.002115
C -0.009166 0.608634 0.009211
C -0.026800 1.961505 0.013524
C -0.042680 3.170676 0.016064
H 0.059658 -4.231919 -0.031944
H -0.056304 4.232063 0.015448
```

```
mol = Molecule.from_file('mol.xyz')
pga = PointGroupAnalyzer(mol, tolerance=0.3, eigen_tolerance=1e-2)
pga.sch_symbol
>> ValueError: min() arg is an empty sequence 
```

```
mol = Molecule.from_file('mol.xyz')
pga = PointGroupAnalyzer(mol, tolerance=0.3, eigen_tolerance=2e-2)
pga.sch_symbol
>> "D*h"
```

This molecule should be classified as "D*h" (a linear molecule with inversion symmetry). The reason why it fails the linear check is that the eigenvalues fail the `eigen_zero` check [in line 896](https://github.com/materialsproject/pymatgen/blob/master/pymatgen/symmetry/analyzer.py#L896) for default parameters because the eigenvalue closest to zero is ~2e-6. If `eigen_tolerance=2e-2`, the molecule's point group is resolved correctly without error.

It seems that the case where the `eigen_zero` check fails and proceeds to check `self._get_smallest_set_not_on_axis(axis)` should be handled with at least a warning. 

One approach is to simply return an empty list when `min` will fail. This would then result in this molecule being classified in the "Ci" point group. This is the fix I have implemented. However, it might be good to also implement a warning that says something like -- "possible linear molecule resolved to lower symmetry due to eigen_tolerance being too small". 

Which is better? An error with an updated warning message, or a fix with a warning message but a possibly lower symmetry point group?

### DIFF
--- a/pymatgen/symmetry/analyzer.py
+++ b/pymatgen/symmetry/analyzer.py
@@ -1076,7 +1076,7 @@ class PointGroupAnalyzer:
             if len(valid_set) > 0:
                 valid_sets.append(valid_set)
 
-        return min(valid_sets, key=lambda s: len(s))
+        return min(valid_sets, key=lambda s: len(s)) if len(valid_sets) > 0 else []
 
     def _check_rot_sym(self, axis):
         """
--- a/pymatgen/symmetry/tests/test_analyzer.py
+++ b/pymatgen/symmetry/tests/test_analyzer.py
@@ -478,6 +478,23 @@ class PointGroupAnalyzerTest(PymatgenTest):
         mol = Molecule(["C", "H", "N"], coords)
         a = PointGroupAnalyzer(mol)
         self.assertEqual(a.sch_symbol, "C*v")
+        coords = [[ 7.19818282e-02, -5.13850399e+00,-2.78587036e-02],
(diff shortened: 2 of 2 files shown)


shyuep:
I think the way the eigen_zero is tested can be changed. There is no reason for the **3. It is rather arbitrary. I just pushed a fix.

Issue #1321, opened 2018-11-08
Bug in NearNeighbors.get_all_nn_info


utf:
When the fractional coordinates of a site contain `1.`, the bonding to the site is duplicated.

**To Reproduce**

```python
from pymatgen.core.structure import Structure
from pymatgen.analysis.local_env import CrystalNN

cnn = CrystalNN()

sites =  [[0., 0.2, 0.2], [0, 0, 0]]

structure = Structure([7, 0, 0, 0, 7, 0, 0, 0, 7], ['I'] * len(sites), sites)

bonded_structure = cnn.get_bonded_structure(structure)
bonded_structure.get_connected_sites(0)
```

Returns 1 connected site, as expected. Changing the fractional coordinate of the site from `0.` to `1.`, should not affect the bonding as the site is in the same position. However:

```python
sites =  [[1., 0.2, 0.2], [0, 0, 0]]

structure = Structure([7, 0, 0, 0, 7, 0, 0, 0, 7], ['I'] * len(sites), sites)

bonded_structure = cnn.get_bonded_structure(structure)
bonded_structure.get_connected_sites(0)
```

Returns 2 connected sites. One with `jimage = (0, 0, 0)` and the other with `jimage = (1, 0, 0)`.

This seems to be due to the behavior of `_get_image` method in the `NearNeighbors` class, which does not consider wrapping of fractional coordinates:

```python
    @staticmethod
    def _get_image(frac_coords):
        """Private convenience method for get_nn_info,
        gives lattice image from provided PeriodicSite.

        Args:
            frac_coords ([float]*3): Fraction coordinates
        Returns:
            ((int)*3) Lattice image
        """
        # TODO: This is not numerically stable. Also, move to PeriodicSite? -WardLT, 23Jun18
        frac_coords = frac_coords
        return tuple(map(int, np.floor(frac_coords)))
```

This could also be related to #827

mkhorton:
Agree with @WardLT's comment here that this should probably just me moved to be a `image` property on PeriodicSite (as well as fixing the underlying bug).

mkhorton:
There also seems to be a larger issue with the construction of Structure and if fractional coords should be wrapped to [0,1) by default, which might help avoid some of these issues.

WardLT:
[...]

PR #1402, opened 2019-02-23
Fix NearNeighbors._get_image to work properly for sites not in unitcell

lan496:
## Summary
- fix behavior of `NearNeighbors._get_image` to deal with cases such that initial sites are not in unitcell(fix https://github.com/materialsproject/pymatgen/issues/1321)

An image is defined as displacement from original site in structure to a given site.
For example, if structure has a site at (-0.1, 1.0, 0.3), then (0.9, 0, 2.3) -> jimage = (1, -1, 2).

Note that this definition returns the same result with the previous implementation of `NearNeighbors._get_image` for atoms in unit cell(i.e. when fractional coordinates are contained in [0, 1)^3)

### DIFF
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -481,17 +481,25 @@ class NearNeighbors:
         return list(all_sites.values())
 
     @staticmethod
-    def _get_image(frac_coords):
+    def _get_image(structure, site):
         """Private convenience method for get_nn_info,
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @lan496 !

mkhorton:
(p.s. this fixed an unrelated issue I hadn't had chance to track down yet, so thanks again! saved me a lot of time)

PR #1403, opened 2019-02-25
Various bug fixes (graphs, phase diagram, magnetism)

mkhorton:
## Summary

Fixes MSONable for PhaseDigram when non-PDEntries used. Fixes graph bug (thanks Jack Sundberg). See individual commit messages for more, includes code reformatting. Will merge once CI passes.


mkhorton:
Note to self: tests can fail due to `muellergroup.jhu.edu:8080` being offline for `pymatgen.ext.tests.test_jhu.JhuTest`, will try to handle this more gracefully.

shyuep:
I suggest adding a pymatgen.util.typing module with the useful type hints like Vector.

shyuep:
That said, I would suggest no need to go overboard. Let's just do hints for the five core modules (periodic table, sites, compostion, lattice and structure). then we see how useful they are. And we can expand gradually to other modules as needed.

mkhorton:
Agreed, I'd be happy to see hints just in core for now to see how useful they are.

Adding comprehensive tests to Lattice was already seemed helpful, both for live coding (PyCharm will highlight if you try to call a method with the wrong type), and also to highlight hidden complexity (eg very long type signatures are interesting). Personally, I've found it most useful so far just to know when a method is returning an np.ndarray vs a list/tuple.

shyuep:
Ok. I suggest you go ahead and merge first and just work on master. I already implemented a few hints here and there. It might be easier to deconflict. in general, the co-pymatgen dictator does not need to use pull requests for simple bug fixes or doc (though for big features, it is still a good practice).

mkhorton:
Noted!

As mentioned, I will keep to PRs for larger features for record keeping/good practice.

PR #1412, opened 2019-03-12
Fix Python 3.5 compatibility

jan-janssen:
To prevent `TypeError: invalid file: PosixPath()`


shyuep:
Thanks. Though people should start moving to py36/37....

jan-janssen:
Btw: You are at TMS right?

shyuep:
I was. Had to leave on Monday as I have to teach.

PR #1414, opened 2019-03-12
[WIP] Add user-agent header to MPRester session requests

dwinston:
Add a user-agent header of [conventional syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) to identify pymatgen and its version to the MP server.

* Allows server, if it is running a later version of pymatgen than the user, to suggest upgrading (e.g. through adding a note in the response "warning" field).
* Allows server to collect statistics on pymatgen versions in active use for gathering MP data, and on what proportion of API use is via pymatgen.

### DIFF
--- a/pymatgen/ext/matproj.py
+++ b/pymatgen/ext/matproj.py
@@ -14,7 +14,7 @@ from monty.json import MontyDecoder, MontyEncoder
 
 from copy import deepcopy
 
-from pymatgen import SETTINGS
+from pymatgen import SETTINGS, __version__
 
 from pymatgen.core.composition import Composition
 from pymatgen.core.periodic_table import Element
@@ -115,7 +115,7 @@ class MPRester:
                               "you should install dependencies via "
                               "`pip install pymatgen[matproj.snl]`.")
         self.session = requests.Session()
-        self.session.headers = {"x-api-key": self.api_key}
+        self.session.headers = {"x-api-key": self.api_key, "user-agent": "pymatgen/"+__version__}
 
     def __enter__(self):
         """
--- a/pymatgen/ext/matproj.py
+++ b/pymatgen/ext/matproj.py
@@ -77,7 +77,9 @@ class MPRester:
             can be changed to other urls implementing a similar interface.
         include_user_agent (bool): If True, will include a user agent with the
             HTTP request including information on pymatgen and system version
-            making the API request. Set to False to disable the user agent.
+            making the API request. This helps MP support pymatgen users, and
+            is similar to what most web browsers send with each page request.
+            Set to False to disable the user agent.
     """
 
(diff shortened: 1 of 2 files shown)


mkhorton:
Excellent idea!

mkhorton:
Do you think it might be useful to add Python version also? It might also be a good idea to mention this in the docstring to ensure that users are appropriately informed that this information is being shared.

mkhorton:
I'm on the fence for this from a user privacy perspective. In favor, I think this is very standard information that is included in, for example, any other kind of HTTP request in a user agent string (such as from a web browser). As long as it's disclosed appropriately in the docstring for MPRester I think this is ok, but I would welcome other input.

For a more expanded version, we could have:

```
import sys
import platform

... f"pymatgen/{__version__} (Python {sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}) {platform.system()} {platform.release()}"
```

Which would give a user agent like `pymatgen/2019.2.28 (Python 3.7.0) Darwin 18.0.0`.

I think this would be incredibly useful information in aggregate from a development perspective.

dwinston:
Sounds good to me. I amended your disclosure in the docstring to clarify the purpose and normalcy of user-agent headers.

However, I'm not sure this is enough, and it gave me an additional idea that I believe could help tremendously for communicating important messages on pymatgen upgrades. This may be best done via a separate PR, but the idea is to store a flag e.g. `described_changes` in a user's `~/.pmg.rc` indicating whether a summary of this version's changes have been communicated to the user (I'd suggest via `warnings.warn`). We also want an e.g. `last_version` flag (default `None` on the first install).  If `described_changes==False`, the relevant changelog sections are output, the flag is set to true, and `last_version` is updated. This way, we can e.g. clearly communicate the user-agent header addition in the case of users who do not consult the docstring, and in general, we can have more confidence that more users are aware of changes upon upgrade.

I realize the above is not standard practice even the most important packages -- standard practice is to insist that users RTFM. So, I expect resistance, and I'd be okay with merging as is. However, in our case, many users may not even know to look for a changelog, releases are often packed with interesting new features that some users become aware of months/years later, and I think many contributors would love to know that their acknowledged contributions (via name- and @-mentions in the changelog) are announced to the entire userbase in what I feel is a suitable time and place.

Wow, that was long.

mkhorton:
This will need tests also (to check that disabling the user agent does indeed delete it from the headers at least), for some reason `__version__` doesn't seem to be working on the AppVeyor tests.

dwinston:
Excellent, your case for being co-BDFL grows stronger by the day...

mkhorton:
Thanks @dwinston, will merge once tests pass!

PR #1422, opened 2019-03-20
Read WAVEDER file

knc6:
Class for reading WAVEDER file. The WAVEDER file is produced invoking LOPTICS tag in VASP.


shyuep:
Thanks for the contribution. However, please provide a unittest with an example input file of how this will be used. Thanks.

knc6:
Hi @shyuep ,

I guess I added a simple unittest in the test_outputs.py.
Could you please check if its good enough?

Thanks

shyuep:
Thanks!

PR #1426, opened 2019-03-22
Propagate site properties to orthogonal slab

sivonxay:
## Summary

Previously site properties in slabs were not passed to the new slab when getting an orthogonal c slab.


shyuep:
CAn we have unittests pls? Thanks.

sivonxay:
Unit tests added! :D

PR #1429, opened 2019-03-27
Implement new OrderDisorderElementComparator

utf:
## Summary

The current OrderDisorderElementComparator will match in unexpected cases. E.g.,
1.  A structure with the fractional composition Sn0.2 O0.8 will match a structure with the composition Sn0.5 O0.5.
2.  A structure with a 50:50 Sn:Pb site would match with a structure with a 50:50 Sn:O site.

This PR fixes both scenarios and should allow proper matching of disordered structures to their ordered counterparts.

### DIFF
--- a/pymatgen/analysis/structure_matcher.py
+++ b/pymatgen/analysis/structure_matcher.py
@@ -242,17 +242,16 @@ class OrderDisorderElementComparator(AbstractComparator):
         Returns:
             True always
         """
-        set1 = set(sp1.element_composition.keys())
-        set2 = set(sp2.element_composition.keys())
-        if set1.intersection(set2):
-            return True
-        return False
+        set1 = set(sp1.elements)
+        set2 = set(sp2.elements)
+        return set1.issubset(set2) or set2.issubset(set1)
 
     def get_hash(self, composition):
-        """"
-        No hash possible
         """
-        return 1
+        Returns: Fractional composition
+        """
+        return composition.fractional_composition
+
 
 class OccupancyComparator(AbstractComparator):
     """
--- a/pymatgen/analysis/tests/test_structure_matcher.py
+++ b/pymatgen/analysis/tests/test_structure_matcher.py
@@ -667,6 +667,21 @@ class StructureMatcherTest(PymatgenTest):
         self.assertTrue(sm_sites.fit(s1, s2))
         self.assertFalse(sm_atoms.fit(s1, s2))
 
+    def test_disordered_to_disordered(self):
+        sm_atoms = StructureMatcher(ltol=0.2, stol=0.3, angle_tol=5,
+                                    primitive_cell=False, scale=True,
+                                    attempt_supercell=True,
+                                    allow_subset=False,
+                                    comparator=OrderDisorderElementComparator())
+        lp = Lattice.orthorhombic(10, 20, 30)
+        coords = [[0., 0., 0.], [0.5, 0.5, 0.5]]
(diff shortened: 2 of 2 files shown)


mkhorton:
Thanks @utf! Will review and merge once tests pass

mkhorton:
Can you think of any reason where the old behavior would be desirable? (I can't, just curious if this is an explicit bug if there was a use case I wasn't considering)

utf:
I couldn't think of any scenarios where you'd want the old behavior. I think it is just an oversight.

Note that this comparator won't allow matching disordered structures with partial occupancies (e.g. sum(site occupancies) != 1).

I have an idea for a future PR to make this possible.

dkitch:
The reason for the old behavior was that disordered entries (for example in the ICSD) are often very approximate in their composition so you want to match partial occupancies with a lot of leeway. It is very easy to check the average composition either of the whole structure, or each individual site, after using StructureMatcher to check coordinates. It is hard to go the other way.

One solution would be to add a tolerance on disordered composition, but this is so easy to do on a case-by-case basis after the fact that I personally very much prefer the old behavior.

mkhorton:
Thanks for the input Daniil! We can certainly add an additional comparator
with fuzzy matching/with a tolerance. There’s also the discretize
occupancies transformation as a preprocessor that might be suitable. It
seems clear we do need a comparator that does compare occupancies as in
this fixed version however, and I would probably argue that’s its expected
behavior.

dkitch:
Fair enough, having a occupation fraction tolerance flag would make the behavior more transparent. I personally have a lot of code that relies on very loose occupation fraction matching as per the old behavior, but I'll just preserve the old comparator in my code internally. I agree that matching composition plus/minus a reasonable tolerance would be the expected behavior for a general user who is not relying on historical hacks.

mkhorton:
Ok, that’s fair. Apologies for breaking backwards compatibility for your
code!

shyamd:
@dkitch 
To be clear are you talking about situations such as matching a Sn0.45 Pb0.55 O2 with SnPbO4 , ie the ordered counterpart only differs slightly from the disordered composition, or are you referring to something more extreme?

dkitch:
The difference in the two implementations is if we have a

site1: occupied by some combination of A and C
site2: occupied by some combination of A and D

they used to match while now they wont.

The situation where one might want the old behavior is if C and D are defect elements that don't matter for matching

Example 1: Si p-doped with 0.1% B on the Si site should match to Si p-doped with 0.1% Al on the Si site, since they're both structurally Si. This issue comes up with processing experimental CIFs which often have defect elements listed as part of site occupancies. 

Example 2: Matching structural frameworks across different kinds of stabilizer elements. For example, you want to match prototypical yttria-stabilized zirconia to anything else that stabilizes cubic zirconia, since what you care about the the cubic zirconia not the stabilizer.

Writing this up, I realize that whether or not you'd want this behavior depends entirely on whether or not you treat the dopant chemistry as important or just as "there is some dopant". In both cases, the alternative solution would be to round off low-concentration elements for the purposes of structure matching, or replace dopants with dummy species.

shyamd:
Ok,

This helps a lot as we're thinking about connecting the disordered ICSD entries to MP entries. I think there should be an explicit DisorderedWithDopantMatcher for these two cases, since it should only happen in the "dilute" limit which can be a user configurable parameter.

utf:
Hi @dkitch, apologies for breaking your code.

I agree with Shyam that there are two separate use cases here. I think the new behavior of OrderDisorderElementComparator makes sense and will be less confusing to users but I can see the utility of a more tolerant comparators that allows for dopant species.

PR #1435, opened 2019-04-05
Hops dev

jmmshn:
## Summary
    
New methods for analyzing the migration path of a material using the following scheme:
1. start with an empty structure and multiple structures with a single cation
2. Map the positions of the relaxed structures back onto the empty host lattice
3. Apply symmetry operations of the empty lattice to obtain the other positions of the intercollated atom
4. Get the symmetry inequivalent hops
5. Get the total charge density that each inequivalent hop in the system has to move through
6. Map all the values obtained for these unique hops onto the full set of hops

The end results of this analysis a dataframe that contains all the hops in the lattice and total amount of charge density that each hop as to move through.


shyuep:
Pls check the pymatgen-diffusion package and make sure it is not already implemented there.

shyuep:
Even if some parts of it is not implemented there, I would prefer if those portions be integrated there rather than having duplicate code. E.g., I think charge density is not implemented (though our experience is that the charge density actually does not work that well). But unique path enumeration is already there.

mkhorton:
@shyuep there's already a path finder in pymatgen also? in pymatgen.analysis.path_finder, there's definitely already some charge density analysis in pymatgen

@jmmshn maybe a more descriptive PR title? :)

shyuep:
The pathfinder in pymatgen does not do enumeration of paths. It merely finds a path based on given start and end points and a potential. Pymatgen-diffusion's pathfinder enumerates ALL points, and there are other potentials implemented like the IDPP relaxer (which we find to be a lot more reliable than the Static Potential).

jmmshn:
Thanks for the feedback.    I think the path_finder in pymatgen is for getting the geometry of the individual hop.     It can be incorporated into this workflow to come up with a better scalar descriptor for a given hop.    
  

  
I will look at pymatgen-diffusion more to see if there is a better place to put this.
  

  
@shyuep can you elaborate on your experience with using the charge density?
  
I had trouble using the charge density alone — so using the local minima as the initial and final positions.     But using the relaxed positions seems to work much better but I’m still in the testing phase so any input will be greatly appreciated.
  
  
  
 -Jimmy     
  

  
  
>

mkhorton:
Understood

shyuep:
@jmmshn In general, the charge density is tends to be very shallow. Even if you use local minima, that minima is spread across a large spatial region. Further, cation-cation repulsion plays a big role in stabilization of sites. While the method may work marginally for dilute limits, e.g., single Li in FePO4, it does not work well when you have non-dilute diffusion.

As a general schematic of how I think we can merge all these efforts, I suggest the following:

i. Let's keep diffusion stuff in pymatgen-diffusion. I am happy to add you as a co-contributor. We have no plans of writing a paper based on that code base, but you will be recognized.
ii. I would prefer to keep the current schematic of how pymatgen-diffusion implements path enumeration, but with some changes. The path enumeration code should be modified to include both vacancy and species diffusion. The enumeration identifies start and end points. We can then implement an abstract Path maker class which can be subclassed to construct different paths. E.g., LinearInterpolationPath, IDPPPath, ChargeDensityPath, etc.

jmmshn:
@shyuep 

>  In general, the charge density is tends to be very shallow. Even if you use local minima, that minima is spread across a large spatial region. 

I was actually running into this exact problem. In my experience, the local minima is a good way to suggest initial positions to try but you really have to use the relaxed position to map out the migration path.

> i. Let's keep diffusion stuff in pymatgen-diffusion. I am happy to add you as a co-contributor. We have no plans of writing a paper based on that code base, but you will be recognized.

Thanks!

> ii. I would prefer to keep the current schematic of how pymatgen-diffusion implements path enumeration, but with some changes. The path enumeration code should be modified to include both vacancy and species diffusion. The enumeration identifies start and end points. We can then implement an abstract Path maker class which can be subclassed to construct different paths. E.g., LinearInterpolationPath, IDPPPath, ChargeDensityPath, etc.

This sounds like a good path forward. I will look at pymatgen-diffusion and figure out how to do this.
I will close this PR for now.

jmmshn:
> ii. I would prefer to keep the current schematic of how pymatgen-diffusion implements path enumeration, but with some changes. The path enumeration code should be modified to include both vacancy and species diffusion. The enumeration identifies start and end points. We can then implement an abstract Path maker class which can be subclassed to construct different paths. E.g., LinearInterpolationPath, IDPPPath, ChargeDensityPath, etc.

PR #1440, opened 2019-04-10
allow the MimimumDistanceNN class to only use the cutoff distance as …

jmmshn:
## New option in MimimumDistanceNN
Allow the MimimumDistanceNN class to only use the cutoff distance as the filter.
So you can ignore the tolerance value.


mkhorton:
Thanks @jmmshn, will merge once a test has been added

PR #1443, opened 2019-04-18
Speed up get_ir_reciprocal_mesh function

utf:
## Summary

Currently the implementation of `SpacegroupAnalyzer. get_ir_reciprocal_mesh` is woefully slow for very large meshes (due to unnecessary repeated calls to `list.count`). This pull request massively speeds up the implementation.

GaAs using a 80x80x80 k-point mesh, the time to calculate the ir mesh went from 120 seconds to less than 1 second.


mkhorton:
Fantastic, thank you!

PR #1447, opened 2019-04-24
Plot fatbands from Lobster and option to force the reading of hybrid-like bandstructure for other functionals

JaGeo:
##
I included classes and tests to plot FATBAND files from Lobster. I had to add several classes to do so. I also added tests. 
Furthermore, I added an option to the Vasprun class to make it possible to read in self-consistent band structure calculations for every type of functional. I have already needed this several times before.

I am happy to rename options or work on the documentary again if anything is unclear.


JaGeo:
I had to fix another test (a borg test). Not sure it was the correct thing to do.

pymatgen/io/lobster.py
```diff
@@ -901,3 +907,215 @@ def _get_all_info_lines(self, data):
                     infos.append(" ".join(splitrow[1:]))
 
         return infos
+
+
+class Fatband(object):
```
mkhorton: Small note, you don't need to subclass from `object` in Python 3

mkhorton:
Thanks for this PR, the new functionality looks very useful.

It's not clear to me why the Borg test needed to be changed because of the new functionality you've added? How does the ComputedEntry differ from previously?

JaGeo:
It changed from 8 to 11. I have included a lot of new test data, e.g. 3 new "vasprun.xml" files.

mkhorton:
Oh, understood, it's just that there are extra vaspruns in the test dir now. Thank you!

JaGeo:
Okay, great. I will soon demonstrate the functionality in the jupyter notebook with examples for plottig lobster files with pymatgen.

JaGeo:
Can it be merged or should I do something else? Thanks.

mkhorton:
No, I think this is good to go! Thanks.

mkhorton:
Oh, actually there seems to be a large commented out Kpoint test in this PR?

JaGeo:
Sorry, my fault. I excluded it for running the test faster and did not include it later. I hope it is okay now.

mkhorton:
All ok! Thanks again for such a nicely-documented and tested pull request.

shyuep:
There are failing CircleCI tests when this was merged. Basically there is a segfault.... Can you two look into this.

mkhorton:
@JaGeo have you seen this before? Test log is here https://circleci.com/gh/materialsproject/pymatgen/8474 If not, we may have a larger issue, since this isn't the first time we've seen segfaults in Circle recently where the same tests were passing locally and on Travis.

JaGeo:
I haven't seen this before but I will have a look at it again.

JaGeo:
It seems that some of my new tests run differently on different operating systems. They worked fine on one but fail on the other. I will open a new pull request with the adapted tests. I also had some other problems while running the tests locally. I hope this will resolve the issue but I am not sure yet.

mkhorton:
Thanks @JaGeo!

PR #1449, opened 2019-04-25
NMR symmetrize tensor update 

dongsenfo:
## Summary

Symmetrize the tensor before returning the principal axis system

Uses the symmetrize method from the tensor class


dongsenfo:
I'm was having trouble managing my local branches, hopefully this only changes the one nmr file?

mkhorton:
Yes, that’s correct, this only changes the one file

mkhorton:
Hi @dongsenfo, tests need to be updated to reflect this change since they're currently failing.

dongsenfo:
I've updated the test, hopefully things should work now!

mkhorton:
Thanks @dongsenfo !

PR #1451, opened 2019-04-30
Updated publication info for dimensionality module

pmla:
## Summary

The paper cited in the "get_dimensionality_larsen" function has been published.  I have replaced the arXiv reference with the published one.

I have also made a two small changes to the function, which improves the speed by 2-3x.


pmla:
@utf Nice implementation!

The rank calculation is the bottleneck here.   The contents of `seen_comp_vertices` is always full-rank, so we can skip recalculating its rank.

Post-publication it occurred to me that we do not need to add nodes to the queue which are not added to a component.  This change (line 234) also reduces calculation time.

shyuep:
Thanks.

utf:
Thanks for this @pmla. Congrats on the publication. This is a really nice method, so I'm glad to see the performance increase by so much.

Issue #1436, opened 2019-04-07
GulpCaller crashes because cell input string too long


LuciusV:
**Describe the bug**
Gulp cannot parse cell if it is too long

**To Reproduce**
`from pymatgen.command_line.gulp_caller import GulpCaller, BuckinghamPotential, GulpIO

from pymatgen import Structure
struct = Structure.from_str("""Mg2 Al4 O8
1.0
5.003532 0.000000 2.888790
1.667844 4.717375 2.888790
0.000000 0.000000 5.777581
O Mg Al
8 2 4
direct
0.736371 0.736371 0.736371 O
0.263629 0.263629 0.709114 O
0.263629 0.709114 0.263629 O
0.709114 0.263629 0.263629 O
0.736371 0.290886 0.736371 O
0.290886 0.736371 0.736371 O
0.263629 0.263629 0.263629 O
0.736371 0.736371 0.290886 O
0.125000 0.125000 0.125000 Mg
0.875000 0.875000 0.875000 Mg
0.500000 0.500000 0.000000 Al
0.500000 0.500000 0.500000 Al
0.000000 0.500000 0.500000 Al
0.500000 0.000000 0.500000 Al""", fmt='poscar')

bp = BuckinghamPotential(bush_lewis_flag="bush")
gio = GulpIO()
input = gio.buckingham_input(struct, ['relax conp'])
caller = GulpCaller()
gout = caller.run(input)`

**Expected behavior**
If I copy input generated by buckingham_input but cut cell parameters to reasonable amount of characters after . gulp runs without problems

Instead I got

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!! ERROR : Insufficient cell parameters in input
!!       : Error is apparently on line      3
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

LuciusV:
This workaround makes code above run without problem
`def prepare_input(bad_input):
    input = ''
    for n, line in enumerate(bad_input.split('\n')):
        if n == 2:
            input += '{:.6f} {:.6f} {:.6f} {:.6f} {:.6f} {:.6f}'.format(*map(float, line.split())) + '\n'
        else:
            input += line + '\n'
    return input`

PR #1452, opened 2019-04-30
A small bug fix for gulp_caller

kmu:
## Summary

A bug addressed in #1436 was fixed.
Old code can generate a too long line for cell parameters, which cannot be handled by GULP.
This patch fixes this issue.
A unit test was added to reproduce the bug.


mkhorton:
Thanks @kmu!

shyuep:
Many thanks.

PR #1454, opened 2019-05-02
Updates to defect modules and FermiDos scissor operator fix

dbroberg:
## Summary

A number of major changes and fixes for point defect modules.

* Update multiplicity approach for Defect class such that Vacancy and Substitution defect types raise early ValueErrors if a defect_site is fed to it which isn't in the bulk structure
* Remove overly complicated approach for accommodating structures with velocities. If this is really desired, a defect structure can be initialized with velocity after generation by the Defect object. 
*  Incorporate Kumagai correction functionalities from PyCDT, with associated utils requirements
* Make use of logging in DefectCompatibility and DefectPhaseDiagram
* Update default parameters for DefectCompatibility based on recent benchmark work
* Add DefectPhaseDiagrams capabilites for: (a) suggesting larger supercells to be rerun, (b) calculating non_equilibrium fermi levels, (c) calculating dopability limits of formation energy plot.
* Fix DefectPhaseDiagram approach for grouping defects to allow for non-identical defects with the same name to be differentiated. 
* Fix DefectPhaseDiagram hull for cases with large negative or positive formation energies
* Add plotting capabilities for DefectPhaseDiagram
* Fix to FermiDos scissor operation. Previously, if the fermi level index was the same as the vbm index, then the scissor operator did not result in desired band gap. 
* Updates to unit tests to account for all of above.
* Updates to READMEs and variable names for clarity

No additional dependencies introduced.


mkhorton:
Thanks @dbroberg 

Regarding the velocities issue, I remember there was some discussion of this previously...can you remember where that was, or who wanted it?

dbroberg:
It was me reporting the issue and the previous fix I made was overly complicated. This fixes the same issue but in a more simple manner.

mkhorton:
Ah, fantastic :-) Understood -- will merge once the CI tests pass.

PR #1455, opened 2019-05-02
Repair failing tests

JaGeo:
I have adapted some of my new tests so that the CircleCI tests will hopefully pass. 
I also had some problems when running "test_cohp.py" and other files locally because there was no "@version" key in my dicts.


shyuep:
Note that it is still segfaulting on CircleCI. Is something particularly memory intensive in these tests?

JaGeo:
I will look into this again.

PR #1456, opened 2019-05-02
Smaller test file + different testing of bandgap

JaGeo:
I think the test_outputs.py is the one that is failing. Thus, I decreased the file size for my test file and also changed the test a little. I hope this helps. If not, I am out of ideas for the moment.


JaGeo:
It failed again. Any other idea where this could come from? I don't think it is a memory problem.

shyuep:
Yeah, honestly I don't know. It passed when you were running it. But failed when I merged it. I think I have an idea... Let me push a fix.

mkhorton:
What makes you think this isn’t a memory problem?

shyuep:
Ok, force-reinstall of phonopy works in removing the segfault. I think the old phonopy (or rather new phonopy v2+) was still in the system, and we previously encountered segfaults with that. 

However, there are still failures with the fatbands test. See https://circleci.com/gh/materialsproject/pymatgen/8485 .

shyuep:
I would strongly advise against direct comparison of dicts. It is more useful to compare the contents.

JaGeo:
I will correct this in one of the next days.

PR #1457, opened 2019-05-03
Fix for fatband tests

JaGeo:
I replaced all AssertDictEqual tests with other tests. I hope this fixes the current test problems.


JaGeo:
It is finally working. Thanks for your help. And, I will avoid assertDictEqual tests in the future...

Issue #1453, opened 2019-05-01
MPNonSCFSet k-point generation incompatible with tetrahedron smearing


utf:
Currently, `MPNonSCFSet` with `mode=uniform` creates an explicit list of k-points using `SpacegroupAnalyzer.get_ir_reciprocal_mesh`. As this list of k-points does not include tetrahedron information, the generated k-points file is incompatible with setting `ISMEAR = -5`. A recent update to the MPNonSCFSet set the default smearing parameter to `ISMEAR = -5`, so **all uniform NSCF jobs in atomate are currently failing.** See hackingmaterials/atomate#283.

Is there a reason for generating the list of k-points manually rather than letting VASP automatically generate the k-points based on a mesh size (e.g., 3x3x3)? VASP will automatically add the tetrahedron information when `ISMEAR = -5`, so this issue is avoided. 

I really can't see the benefit of EVER generating the list of k-points manually, except for line-mode band structures but maybe I am missing something?

hautierg:
This was a while ago but I remember that the reason we went for explicit generation of k-points is symmetry. VASP uses a certain way of setting symmetries that is sometimes incompatible with other assignment for instance in spglib and from what I remember not super clear from the source code. That means spglib will tell you a certain k-point mesh in the irreducible brillouin zone and VASP another one. When you want to do things outside VASP (I think Boltztrap treatment was one example this would lead to inconsistencies). The MPNonSCFSet method was never intended to be used with the tetrahedron method and I think it used to not set ISMEAR=-5.

But as I said this was a while ago for specific purpose and if we want to switch to something else I am fine with it. However, we should keep this issue of symmetry in mind.

Geoffroy



On 01 May 2019, at 20:20, Alex Ganose <notifications@github.com<mailto:notifications@github.com>> wrote:


Currently, MPNonSCFSet with mode=uniform creates an explicit list of k-points using SpacegroupAnalyzer.get_ir_reciprocal_mesh. As this list of k-points does not include tetrahedron information, the generated k-points file is incompatible with setting ISMEAR = -5. A recent update to the MPNonSCFSet set the default smearing parameter to ISMEAR = -5, so all uniform NSCF jobs in atomate are currently failing.

Is there a reason for generating the list of k-points manually rather than letting VASP automatically generate the k-points based on mesh size (e.g., 3x3x3)? VASP will automatically add the tetrahedron information when ISMEAR = -5, so this issue is avoided.

[...]

PR #1458, opened 2019-05-03
Add boltztrap mode to MPNonSCFSet

utf:
## Summary

As discussed in #1453, currently the MPNonSCFSet k-point generation routines are incompatible with tetrahedron smearing for uniform calculations. Historically, this decision was made to enable compatibility with Boltztrap calculations.

As we are transitioning to using tetrahedron smearing for higher quality density of states, I have updated the MPNonSCFSet "uniform" mode to be compatible with tetrahedron smearing. In addition, I have added a "boltztrap" mode which reproduces the legacy behavior.

I have added tests for the new uniform behavior and for the old boltztrap behavior.


mkhorton:
Fyi, the `test_optics` tests are failing with this change.

mkhorton:
Thanks @utf

Issue #1460, opened 2019-05-07
Use 0 based index for CifWriter


kmu:
I'm wondering why `CifWriter` employs 1 based index, while `Structure` itself uses 0 based index.

https://github.com/materialsproject/pymatgen/blob/19f56d4d918c3b7fdc445aaca7d56d0d9bc3ddb6/pymatgen/io/cif.py#L1283-L1292

I'm always frustrated when I export a structure to a CIF file, open by VESTA, and try to inspect a specific site.
I would like to fix this if there is no compatibility problem or other issues for it.

mkhorton:
If this isn't mandated by the core CIF spec, I'd be on board with changing it.

kmu:
I tested 0 based CIF file and 1 based CIF file by [checkCIF](http://journals.iucr.org/services/cif/checkcif.html) and got essentially identical results.
Both files can be opened by CrystalMaker and VESTA without problem.

Any integer seems to be allowed for `_atom_site_label ` according to the example shown in [the specification of CIF](http://scripts.iucr.org/cgi-bin/paper?pii=S010876739101067X).

mkhorton:
OK, I trust checkCIF on this, in that case you would be very welcome to make this change, thank you!

PR #1461, opened 2019-05-07
Use 0 based index for CifWriter

kmu:
## Summary

`CifWriter` now uses 0 based index for the consistency with `Structure`.
You can compare the site index in VESTA and that for pymatgen without confusion.
See #1460.


shyuep:
thanks.

Issue #1464, opened 2019-05-09
Changing the signature of vaspinputset.from_prev() 


computron:
Currently, the from_prev() method in the VaspInputSets is a wild west of parameters - i.e., there is no consistency in the signature of from_prev() between different VaspInputSets.

This ends up causing a lot of downstream complexity in atomate. Although the atomate FireTasks can take in a VaspInputSet object to help users control input settings, they usually also mirror the from_prev() parameters to allow users to also control the behavior of the from_prev() method for the VaspInputSet.

The current situation in atomate is unambigously a mess. There are two potential solutions:
i) Atomate uses a single **kwargs style option for updating from_prev() settings in a VASP input set. Thus, at least it would not be mirroring a dozen parameters.
ii) Pymatgen VASP input sets are modified so that *all* parameters (including parameters currently set independently in the from_prev() method) are in the constructor. The method signature of from_prev() is the same for all VaspInputSets, which will be ```my_vaspinputset.from_prev(directory=".")``` - i.e., the only parameter for from_prev is the directory, every other parameter is set in the constructor.

We (atomate discussion team) think (ii) is the much better solution, since it means that once you create an input set, all the rules/flags/etc on the operation of that inputset are determined and can be serialized, etc. e.g., I can serialize my from_prev() settings and not just my "normal" settings. Furthermore, I think this solution is also much cleaner than having atomate create a kwargs-style parameter specifically for passing into the ``from_prev()`` of a vaspinput set.

Thoughts? Is this practical to change?

shyuep:
My view is that there are both use cases. The problem with the suggested Option 2 is that the logic flow is strange. Initialization VaspInputSet requires a Structure, which presumably is being read from a previous calculation in these cases. I would also prefer not to wreak havoc on a lot of people's scripts (including a large fraction of my group). 

I would suggest a Third option that will achieve what you want and is totally backwards compatible.

```python
class VaspInputSet:

    def __init__(structure, ...):
        remains the same.

    def override_from_prev_calc(path="."):
        what you are suggesting in the new from_prev

    @classmethod
    def from_prev_calc(*still the same API and parameters*):
        uses __init__ and override_from_prev_calc to achieve exact same effect as current from_prev_calc
```

For atomate, you should use `__init__` and  `override_from_prev_calc`, which will allow you to standardize the code.

computron:
Two comments:

[...]

PR #1466, opened 2019-05-13
Add override_from_prev_calc to VASP input sets

utf:
## Summary

Based on the discussion in #1464, I have added `override_from_prev_calc` methods to the VASP input sets that already define `from_prev_calc`. In each case, the old `from_prev_calc` method has been rewritten to use `override_from_prev_calc`. This new approach has several advantages, including:

- The `from_prev_calc` and `override_from_prev_calc` parameters are now consistent for every input set.
- There are now no duplicated parameters. Previously, some defaults were defined in both `__init__` and `from_prev_calc`, now they are all defined in `__init__`.
- The implementation in atomate should be significantly cleaner.

Some additional notes:
- Previously, some from_prev functions had a `standardize` keyword that would transform the structure to the standardized primitive cell. It seemed like this would be a very useful option for every input set, regardless of whether it was continuing from a previous calculation. E.g. at the start of your workflow, you might want to standardize your structure before relaxing. I have therefore added the `standardize` keyword argument to the base `DictSet`. By default this is set to False (which also matches the previous default behavior of standardize in the from_prev methods).
- `MPHSEBSSet` has a `mode` keyword that controls the type of calculation (line, uniform or gap). The gap option was only available in the from_prev method, where it was the default. The only difference between the gap and uniform modes is that the VBM and CBM of the previous calculation are explicitly added to the k-point list. I have changed the default value of mode in `__init__` to "Gap" because this gives the best compatibility with the previous behavior (it is reproduced exactly). If the user does not continue from a previous calculation, the behavior will be exactly the same as the "uniform" mode (the previous default for the init method). However, if the user continues from a previous calculation then the default will be to add the VBM and CBM k-points unless otherwise specified (again, the default for from_prev).
- `MVLGWSet` has different default values for they `mode` kwarg in the from_prev and init functions. I think this is due to the standard workflow for GW/BSE calculations. To avoid any issues here, I left the `mode` keyword as an option to the `from_prev_calc` method (so that the default behavior is not affected). As the main use of the `override_from_prev_calc` methods will be in atomate which does not use these input sets, the slightly different behavior of `override_from_prev_calc` (it does not use the same default mode as `from_prev_calc`) should not be an issue. I am open to other suggestions for fixing this, however.
- I fixed up the documentation if I noticed it was off.

One thing to note is, I had to introduce a dummy structure in the from_prev_calc methods, so that I could initialise the input set before calling override_from_prev_calc method. This is a bit ugly, but saves reading the vasprun twice to get the structure.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -2,30 +2,6 @@
 # Copyright (c) Pymatgen Development Team.
 # Distributed under the terms of the MIT License.
 
-
-import abc
-import re
-import glob
-import shutil
-import warnings
-from itertools import chain
-from copy import deepcopy
-import numpy as np
-from pathlib import Path
-from monty.serialization import loadfn
-from monty.io import zopen
-from monty.dev import deprecated
-
-from pymatgen.core.periodic_table import Specie, Element
-from pymatgen.core.structure import Structure
-from pymatgen.io.vasp.inputs import Incar, Poscar, Potcar, Kpoints, VaspInput
-from pymatgen.io.vasp.outputs import Vasprun, Outcar
-from monty.json import MSONable
-from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
-from pymatgen.symmetry.bandstructure import HighSymmKpath
-from pymatgen.analysis.structure_matcher import StructureMatcher
-from pymatgen.core.sites import PeriodicSite
-
 """
 This module defines the VaspInputSet abstract base class and a concrete
 implementation for the parameters developed and tested by the core team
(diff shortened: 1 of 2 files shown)


mkhorton:
With regards to the ‘standardize’ option, is there a good reason to prefer the standard primitive over a reduced cell?

shyuep:
The main reason would be for defining the lines along which to plot band structure.

mkhorton:
That’s fair, but the primitive standard cells aren’t always great, eg I saw a case recently where:

`get_primitive_structure` gives:
```
Lattice
    abc : 29.205719340811687 10.269670980642069 14.881236907930036
 angles : 153.08849008764724 168.90022329343003 24.67294446652493
 volume : 358.69649610521316
```
`get_reduced_structure` (Niggli) gives:
```
Lattice
    abc : 5.7298161670563585 7.373364113749618 9.910301456455453
 angles : 71.19293240753842 73.19688199311382 67.13579991473384
 volume : 358.6964961052132
```
`get_primitive_standard_structure` gives:
```
Lattice
    abc : 7.373364129854753 7.373364129854753 10.269670980642069
 angles : 65.99141686564593 114.00858313435408 134.2715992359032
 volume : 358.69649610521344
```

This is actually not *so* bad, but I think I’d still probably prefer the reduced cell for a relaxation. I’m open to being convinced otherwise however. I can certainly see the consistency for band structure calculations being useful.

shyuep:
I agree. That is why standardize should default to False unless there is a good reason for it to be true, e.g BS calculations, which I believe is the only place where this is true.

PR #1470, opened 2019-05-22
Segmented drawing different branches of band structure

ShuaishuaiYuan:
## Summary
Segmented drawing different branches of band structure in BSDOSPlotter. Look nicer between two discontinuous k-path branches
* After fix
 
* Before fix


shyuep:
Thanks.

PR #1471, opened 2019-05-23
Make FermiDos MSONable

utf:
## Summary

Currently, FermiDos objects are not MSONable for two reasons:
1. They have a strange constructor that requires a Dos object rather than energies, densities and a Fermi level like other Dos subclasses.
2. The nelecs variable is not stored, even as a private variable.

Ideally I'd correct the constructor but I don't want to break other peoples code. Instead, I've added custom to_dict and from_dict methods and fixed point 2.

I've also tied up the class, made it pep8 compliant and added type hinting.


mkhorton:
Thanks @utf! I share your concern about breaking backwards compat, but
cases like this seem pretty clear it should be tidied up. Will merge this
fix regardless

PR #1473, opened 2019-05-29
Add Interface class and InterfaceBuilder

kylebystrom:
## Summary

Adding Interface and InterfaceBuilder classes (prototype stage) along with a few basic unit tests. Initial code written by @shyamd and @sivonxay and debugged/documented by myself.

* Store and edit interfaces in the Interface class
* Create new interfaces with the InterfaceBuilder


mkhorton:
Thanks @kylebystrom, are you happy for this to merged now while it’s still in “prototype” stage? It already looks quite substantial, though I can’t tell what the test coverage is like from a quick read.

kylebystrom:
Looking back at it, I think the test coverage is still a bit low, so Shyam and I are going to add a couple more unit tests before we merge this. Sorry for the late reply.

shyuep:
I have a more general question of how this is different from the SubstrateBuilder?

shyamd:
The substrate analyzer merely finds the matching surface lattices and does some simple analysis on that, ie elastic strain energy. This uses that same algorithm and actually builds an interface structure that can be used in DFT for instance to compute interface energies.

shyuep:
Merge the two. I do not wish for multiple modules doing roughly the same thing with minor differences. In fact, I would completely deprecated substrate analyzer and integrate its functionality into interface.

A side note that we already ahve existing code in our private code base for this purpose. See http://dx.doi.org/10.1021/acs.chemmater.7b04096 . Please coordinate with @HanmeiTang to integrate everything properly in a coherent manner.

shyamd:
They don't do the same thing. `SubstrateAnalyzer` determines matching relationships for the epitaxial constraint using the Zur and McGill algorithm. It's designed to crank through lots of surfaces planes and structure combinations quickly to evaluate which combination is most plausible. `InterfaceBuilder` constructs interface structures for evaluating using DFT **if** you already have the two structures and surface planes in mind.  You should use the `SubstrateAnalyzer` to determine which planes are appropriate to investigate and then construct that using the `InterfaceBuilder`. I  do agree the `SubstrateAnalyzer` name is a bit misleading since it can be used for more than just substrates. 

The long-term goals are:
1. Have an `Interface` class in `core`. `GrainBoundary` should inherit from this in the future
2. Keep the `ZSLGenerator` in`analysis`. This is just a geometry finding algorithm and is superior to the Stradi algorithm for coherent crystalline interfaces, but building the interface structure for DFT is more of a pain. Hence why the `InterfaceBuilder` does quite a bit. 
3. Work with @HanmeiTang to build a `StradiGenerator` so we can use that in either analysis or interface construction. 
4. Convert `SubstrateAnalyzer` into `InterfaceCoherencyAnalyzer` to be clear of what the name does. 

I want something working in pymatgen for now, since I have gotten plenty of emails asking for this functionality. Kyle is finishing up at LBL soon and moving for grad school where he will likely work on other projects, so it's better not to let this hang. I'm also not a fan of single monolith classes. This makes understanding what they do near impossible for any but the most steeled pymatgen veterans.

shyuep:
I agree they technically handle two different aspects, but I still want them merged into one module, not separate modules. I can certainly forsee a class method which says Interface.from_structures(a, b, algo="Zur").

shyamd:
This is ready to merge. I will add an issue to clean up InterfaceBuilder and assign to myself.

mkhorton:
Ok thanks @kylebystrom @shyamd, I'd like to merge now since it seems immediately useful and we can take discussion of combining these classes to #1491

@kylebystrom before I can merge could you make sure all functions and methods have docstrings? Overall this PR seems to be nicely documented but there are a few missing in interface.py

kylebystrom:
@mkhorton I documented the helper functions in interface.py. I also removed several functions that aren't currently in use. Some of that functionality might be integrated in #1491.

mkhorton:
Thanks @kylebystrom !

Issue #1477, opened 2019-05-31
Composition accepts non valid formulas


utf:
Composition objects allow initializing with an input that is clearly not a valid formula.

For example:

```python
from pymatgen import Composition

Composition("Madeupstuff")
Composition("Lonsdalite")
Composition("MyGranddadGoesToMarketEveryTuesday")  
```

None of which raise an exception.

Is this behavior expected?

shyuep:
This might be surprising, but yes, these are valid. Basically, pymatgen treats any non-element symbol starting with an upper-case as a DummySpecie, unless it violates the rule that the upper case letter itself is an element. For example, Composition("Stuff") would not work because S is an element. 

You can see this clearly when you look deeper at what your example gives. E.g.,

```python
In [11]: c = Composition("Madeupstuff")

In [12]: print(dict(c))
{DummySpecie Madeupstuff0+: 1.0}
```

Of course, we can be more restrictive about DummySpecie names. Just not sure how restrictive we want to be.

utf:
I figured out that’s what was going on but I wasn’t sure if it was expected behaviour.

I was originally trying to cast to a composition check if a string was a valid formula. 

The current behaviour of composition seems fine, but it might be nice to have an ”is_valid_formula” function if that doesn’t already exist in pymatgen.

utf:
Thanks Matt!

PR #1478, opened 2019-05-31
Add valid property to Composition and strict option to constructor

mkhorton:
For #1477


mkhorton:
Oops realized I created this branch from my NearNeighbors branch instead of from master as intended, will merge later

PR #1480, opened 2019-05-31
linear scaling method for get all neighbors in a structure

chc273:
## Summary

Implemented a linear scaling `get_all_neighbors_cell_list` method for `Structure`. This method gives identical results to `get_all_neighbors` but uses a cell list method that scales linearly with the number of atoms, i.e., O(N). The original `get_all_neighbors` has a O(N^2) scaling. 
Reference: https://en.wikipedia.org/wiki/Cell_lists

* get_all_neighbors_cell_list
* linear scaling with system size
* Identical results as `get_all_neighbors`


mkhorton:
Hi @chc273 , this looks great!

If the output is identical to `get_all_neighbors` why not combine them? You could rename `get_all_neighbors_cell_list` to `_get_all_neighbors_cell_list` and then add a new `method="cell_list"` keyword argument to `get_all_neighbors`. If someone wants the old algorithm, they could change it to `method="traditional"` instead? If the speed up is that good, it should be the default algorithm :-)

Separately, could you comment on cell lists vs e.g. something like a k-d tree? My understanding was that cell lists in particular were really useful in MD, in a situation where you had multiple time steps and minimal changes between each time step such that the cell lists would not need to be updated too often, but that for a single isolated structure the advantage was not as great (though still of course better than the traditional method!)

shyuep:
I have a few comments:

i. It would appear to me that this algo should work for a single site as well. Is there a reason why we do not want to do it for get_nn(site)? Of course the speed up is not likely to be very significant for a single site. I just prefer for things to be designed in a coherent manner.

ii. I agree with @mkhorton that we should make this the default and rename the current method to something like get_all_neighbors_old. I would like to preserve the old algo as a cross-check for now. We actually already made this modification once before.

iii. If we agree on the things above, I would actually change the signature of get_all_neighbors somewhat. Basically, 

```
def get_all_neighbors(indices=None):
```

If indices are specified, we will only get all neighbors for specified indices. This would be useful in instances where you only want to analyze certain local-env. Furthermore, the get_nn(site) becomes a trivial exercise of calling this method with a single index. We will merge all algos.

chc273:
Hi @mkhorton,

Good suggestions, I will try to merge them together. 

For k-d tree vs cell lists, my experience using k-d tree is only for high dimensional (n>3) neighbor search. My understanding is that it divides the space by iteratively subdivide each dimension and it works nicely if one wants to query the neighbors for one point, by transversing each dimension. However, distance information is not native to this method. For our problem here, we want neighbors within certain cutoff for all sites. I guess using the k-d tree, the search speed should be slower than O(N)? I don't know.

chc273:
@shyuep 

All those can be done. I will push to the PR later.

chc273:
@shyuep @mkhorton 

The `get_all_neighbors` and `get_neighbors` methods are modified. 
i. Both methods are now using the new algorithm. The previous algorithms are changed to `get_all_neighbors_old` and `get_neighbors_old`. The `get_neighbors` method calls `get_all_neighbors` by setting the `sites` argument to the specific site. `Comment by @mkhorton and Comment ii by @shyuep` 
ii. The `get_all_neighbors` methods include two more input parameters `indices` and `sites`. `indices` can be set to a list of indices, a single index or None, for computing the neighbors for a list of sites, a single site and all sites (if `sites` are not set) respectively. `Comment iii by @shyuep`. The `sites` parameters allows a list of Sites, a single Site and None for computing the neighbors for the list of sites, a single site and being not used respectively.` Comment i by @shyuep `

shyuep:
I have merged and added some fairly robust set of tests for cross-checking old algo with new algo. CircleCI will warn us if something goes awry. The performance is pretty good, with little difference at small cells, but rapidly becoming much faster beyond 100 atoms.

image

shyuep:
@chc273 There seems to be some issues with the new algo and the Magnetic transformations. See https://circleci.com/gh/materialsproject/pymatgen/8522?utm_campaign=chatroom-integration&utm_medium=referral&utm_source=slack .  Can you check?

mkhorton:
I’ll take a look at this tomorrow also since I probably wrote some of that
code.

chc273:
The error seems to be from using `get_neighbors` with a very small cutoff radius. Essentially there are no neighbors within the radius. The current `get_neighbors` and `get_neighbors_old` return an empty list `[]`, but the class expects a list of length 1.  @shyuep 

If Matt is going to take a look, I will leave this to you then @mkhorton  Thanks. 

```           neighbors = structure.get_neighbors(
                    site,
                    0.05,  # arbitrary threshold, needs to be << any bond length
                    # but >> floating point precision issues
                    include_index=True
                )
                if len(neighbors) != 1:
                    raise Exception("This shouldn't happen, found neighbors: {}"
>                                   .format(neighbors))
E                   Exception: This shouldn't happen, found neighbors: []
```

shyuep:
@chc273 I just pushed a change that removed your "sites" option. In general, I do not like duplicate mutually exclusive arguments to a function, especially since the sites and indices can be generated quite eaily. However, it has introduced a strange error when inserting a hydroxyl atom. Cna you pull and check it out? Just need to run test_structure.py. 

@mkhorton I think you should look at the MagTransformation since that is the only area where there is a specific broken test.

chc273:
@shyuep I agree . The inclusion of `sites` was to make `get_neighbors` method to call `get_all_neighbors` without much more efforts.
The added functionality of `sites` is that it can specify additional Sites that are not in the original structure. This can not be handled using `indices`.  

Here the problem seems to be from calling `self.index(site)` to a site that is not in original cell.

shyuep:
@chc273 In that case, I would suggest we keep the sites option and remove the indices option. Let me fix it now.

shyuep:
@chc273 I just pushed a fix for this particular problem.

shyuep:
@mkhorton Can you get a fix in for the MagTransformation soon? The fact that it is the only place where there is a failing test suggest it is a specific issue to magnetic structures.

mkhorton:
Will do

mkhorton:
Looks like there's also a failure in `pymatgen.alchemy.tests.test_filters.SpecieProximityFilterTest.test_filter`

shyuep:
Fixed it. That's a regression when I refactored some of the things.

mkhorton:
Ok, I have a minimal example of this bug now:

```
from pymatgen import Structure
from pymatgen.electronic_structure.core import Spin
d = {'@module': 'pymatgen.core.structure', '@class': 'Structure', 'charge': None, 'lattice': {'matrix': [[0.0, 0.0, 5.5333], [5.7461, 0.0, 3.518471486290303e-16], [-4.692662837312786e-16, 7.6637, 4.692662837312786e-16]], 'a': 5.5333, 'b': 5.7461, 'c': 7.6637, 'alpha': 90.0, 'beta': 90.0, 'gamma': 90.0, 'volume': 243.66653780778103}, 'sites': [{'species': [{'element': 'Mn', 'oxidation_state': 0, 'properties': {'spin': Spin.down}, 'occu': 1}], 'abc': [0.0, 0.5, 0.5], 'xyz': [2.8730499999999997, 3.83185, 4.1055671618015446e-16], 'label': 'Mn0+,spin=-1', 'properties': {}}, {'species': [{'element': 'Mn', 'oxidation_state': None, 'occu': 1.0}], 'abc': [1.232595164407831e-32, 0.5, 0.5], 'xyz': [2.8730499999999997, 3.83185, 4.105567161801545e-16], 'label': 'Mn', 'properties': {}}]}
struct = Structure.from_dict(d)
print("new method", struct.get_neighbors(struct[0], 0.05))
print("old method", struct.get_neighbors_old(struct[0], 0.05))
```

@chc273 could you investigate further? Note that this is a slightly pathological case with sites being co-incident with one another but it's important.

chc273:
@mkhorton Will do. Thanks

shyuep:
I think I know what is happening now. Let me fix this.

chc273:
@shyuep sure. Thanks. A distance threshold of >1e-8 was applied in the last step to filter out the centering site itself. If this is what you saw too. 
Probably we need to filter the centering site with same index and image vector (0, 0, 0) criterion.

mkhorton:
It's a little tedious, but I sometimes have structures with exactly co-incident sites as an intermediate step prior to merging two sites together (or discarding one of them). There are some edge cases where this is useful, I think it was useful for the grain boundary generator too to discard duplicate sites, so it'd be good to have this fixed. I'm not sure the best way to filter the centering site.

shyuep:
@chc273 Yes, I think that would be the way to do it. Also, pls fix the list of lists problem at the same time.

chc273:
@shyuep @mkhorton I am on it. Will push fixes for both cases.

mkhorton:
Thanks @chc273 !

PR #1484, opened 2019-06-04
Bug fix for Structure.get_all_neighbors if cutoff radius is too small

chc273:
## Bug fix for Structure.get_all_neighbors

Previously if the cutoff radius is too small and no neighbors are found, the `get_all_neighbors` method will return an empty list `[]`, while it should be a list of empty list `[[]]`. 
This PR corrects return types.


shyuep:
Actually this is still not correct. The legnth of the list of lists should be equal to number of sites.

shyuep:
Pls resubmit a PR and include a test for this case.

PR #1485, opened 2019-06-05
Bug fix for get_all_neighbors

chc273:
## Summary

1. Fixed the return type of `get_all_neighbors` when no neighbors are found. It should return a list of length n, where each item is `[]` and n is the number of atoms
2. Fixed the `magnetism` test problem. It is an issue with coincide sites. The site with DummySpecie coincides with a normal site and now the `get_neighbors` can find the correct neighbors from the Dummy site. 

## Possible issue
In the case where two identical sites exist in a structure, the `get_neighbors_old` will not find any neighbors using one of the site (small cutoff). However, if one site has a different specie, then the algorithm will be able to find neighbors. It does not seem to be physical since neighboring relationship is only spatial information and should have nothing to do with the Specie. Now I set `get_neighbors` to have the same behavior as `get_neighbors_old`. If this needs to change, it can be easily fixed. 

## Comparisons between `get_all_neighbors_old` and `get_all_neighbors`
The `get_all_neighbors_old` may give different results compared to `get_neighbors_old`, for example, in structures with coincide sites. For comparisons, always compare `get_all_neighbors` and `get_neighbors` to `get_neighbors_old`.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -1360,7 +1360,6 @@ class IStructure(SiteCollection, MSONable):
             nn_indices = list(itertools.chain(*[cube_to_indices[k] for k in ks]))
             dist = np.linalg.norm(nn_coords - i[None, :], axis=1)
             nns = []
-            is_filtered = False
             for coord, m, n, d in zip(nn_coords, nn_indices, nn_images, dist):
                 if d < r + 1e-10:
                     item = []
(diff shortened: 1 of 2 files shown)


shyuep:
I added one more test for the small distance as well as @mkhorton's Mag structure coincident site test. I also removed the unnecessary checks for empty lists. Those checks are more trouble than it is worth. Without those, the code automatically returns the correct list of empty lists when cutoff is too small.

Issue #1442, opened 2019-04-18
waveder AttributeError: can't set attribute


amyncarol:
Unable to use the Waveder class in pymatgen.io.vasp.outputs

Error message below:

```
   4400             cder_data = cder.reshape((nbands, nelect, nk, ispin, 3))
   4401             self.cder_data = cder_data
-> 4402             self.nkpoints = nk
   4403             self.ispin = ispin
   4404             self.nelect = nelect

AttributeError: can't set attribute
```

I think this because nkpoints is a property and cannot be set. 
Change nkpoints to _nkpoints and change the nkpoints method accordingly will probably fix the problem.

utf:
This is also the case for the `nbands` and `nelect` properties.

amyncarol:
I am also a little confused about cder_data. Should it be a complex vector with x,y,z components for each spin? Then it should have 6 float numbers for each (band1, band2, kpoint, spin)? But from the code, we only get 3 float numbers. It should be similar to the data in wavederf, right?

utf:
I was also having this issue, I believe the datatype for the cder should be "complex64". E.g. the data is composed of two 32 bit numbers (real and imaginary component). However, looking at the VASP code, it seems like the exact format of the data will depend on whether the wave function is complex or not. I don't know when the wave function is not complex the code seems like this could happen.

Unfortunately, even setting the dtype of cder to be complex, I couldn't get agreement between the WAVEDER and WAVEDERF file. 

Another thing to note is that `nelect` is incorrectly named. It is not actually the number of electrons. In practice I think it will just be the same as `nbands`.

mkhorton:
Yes, the wave function is not complex if it’s gamma point only.

amyncarol:
Any idea what's the unit of numbers in cder_data?

amyncarol:
Seems to be standard unit J/m, right?

shyuep:
So can someone fix this and submit a PR?

mkhorton:
@knc6, you contributed the WAVEDER class, could you comment on this or would you be willing to submit a fix as appropriate?

PR #1492, opened 2019-06-12
Fix for issue #1442

JRSuckert:
## Summary
This commit tries to fix issue #1442 in the pymatgen.io.vasp.outputs.Waveder class.

* Fix 1
The original issue is raised by the fact that the variables and member functions have the same name.
This has been fixed by prepending a underscore to the variable names.

* Fix 2
Additionally, the datatype of the records read into cder should be of type np.complex64.
This has been changed.

* Fix 3
Furthermore, the order of the reshape call for cder_data needs to be inverted, otherwise the mapping between bands, kpoints and spin won't be correct

* Feature 1
I also introduced a property function for the orbital derivative between states.


shyuep:
Thanks. Can you add unittests to ensure those bugs don't recur in future?

utf:
Based on the discussion in #1442, it seems like the datatype of the records can be float64 instead of complex64 if the wave function is generated using the gamma point version of VASP. 

Please can you also add an option in the init method called `gamma_only` which defaults to False and will toggle the datatype correctly.

JRSuckert:
I've added the option proposed by @utf .

Regarding unit tests, I have never done them before, so I might need a bit to figure out what to test.

JRSuckert:
Added a unit test. 
It would be great if somebody can look over it to see if its useful this way.

shyuep:
Great. THanks. I will merge now.

PR #1495, opened 2019-06-14
Add Outcar method to parse on-site density matrix and tests

mkhorton:
Will merge once CI passes.


mkhorton:
Thanks @clegaspi for help with the regex on this one, @mhsiron this should be ready for your workflow now if you pull from master

clegaspi:
Did you mean to change the row pattern for spin down? Shouldn't it be the same as spin up?

mkhorton:
Hmm, good catch. Yes, I’ll change that. Thanks!

PR #1498, opened 2019-06-15
Fix pip install

JaGeo:
## Summary
After using pip install, "BASIS_PBE_54.yaml" was not found. I have changed the "setup.py" to correct this.


JaGeo:
I have also corrected for another bug in the calculation of nbands... and adapted one test to avoid this in the future.

PR #1499, opened 2019-06-19
Another small update to Lobsterin class

JaGeo:
## Summary

Small BugFix in the lobsterin class to also change LWAVE from False to True when writing a new INCAR. I also adapted the tests. 

I am sorry for the very frequent bug fixes at the moment. I hope this will be one of the last ones for now.


mkhorton:
Better bug fixes than bugs!

Question regarding the io, why the write_INCAR function instead of a `LobsterSet` in io.vasp.sets? The latter may be easier for other people to discover

JaGeo:
I will probably write a LobsterSet in the future. I have to think a bit more about the design at first. I am not sure about the standard settings for the calculations yet.

JaGeo:
And, thanks for merging.

PR #1500, opened 2019-06-19
Fix serialization of Interface

sivonxay:
## Summary
-Override the as_dict() and from_dict() functions from pymatgen.core.structure.Structure for proper serialization
- Cleaned up extraneous imports
- Fixed option to create copy with site properties (Previously documented as such, with no functionality)
- Override get_sorted_structure() from Structure. Not sure if this was the right way to fix the issue; the alternative is to make an Interface.from_sites() function, although one will likely never make an interface from a list of sites. 

## TODO (if any)

- Unit tests

### DIFF
--- a/pymatgen/analysis/interface.py
+++ b/pymatgen/analysis/interface.py
@@ -294,7 +294,7 @@ class Interface(Structure):
         d["modified_film_structure"] = self.modified_film_structure
         d["strained_sub_structure"] =self.strained_sub_structure
         d["strained_film_structure"] = self.strained_film_structure
-        d['init_inplane_shift'] = self.ab_shift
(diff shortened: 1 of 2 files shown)


mkhorton:
Usually it's better not to have to write custom as_dict/from_dict methods -- what specifically about Interface means they're necessary? (Not a problem if it is necessary)

mkhorton:
Ok, verdict is that it’s because of Structure’s from_dict. Also making a note that get_sorted_structure and from_sites won’t work on subclasses as present, will need to address this in future.

mkhorton:
`test_copy` broke on this

sivonxay:
Looks like it. Looking into this now

sivonxay:
Okay, @mkhorton. Issues resolved

PR #1502, opened 2019-06-20
Fix for magnetic space groups

simonward86:
## Summary

This fixes magnetic symmetry operations in pymatgen/symmetry/maggroups.py . 

* Fix 1 : Hexagonal magnetic spacegroups start at 143 instead of 168. This can be verified by VESTA.
* Fix 2 : When reshaping data from https://stokes.byu.edu/iso/magnetic_data.txt , the data should be read in columns instead of rows.

* None

## TODO (if any)

* Verify that the `_parse_transformation(b)` is giving expected results. Before/after the fix it was giving results which are the same as reference, but in a slightly different format. I'm unsure if this is intentional.

### DIFF
--- a/pymatgen/symmetry/maggroups.py
+++ b/pymatgen/symmetry/maggroups.py
@@ -124,7 +124,7 @@ class MagneticSpaceGroup(SymmetryGroup):
 
         def _get_point_operator(idx):
             '''Retrieve information on point operator (rotation matrix and Seitz label).'''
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks very much for fixing this bug @simonward86 (the space group cut-off was my error, I was not including the trigonal int numbers within that range, so thanks for catching that also!)

I'll compare some more results to https://stokes.byu.edu/iso/magnetic_table_bns.txt via the `_write_all_magnetic_space_groups_to_file` method to double check.

PR #1504, opened 2019-06-21
Modified run_type in outputs.py.

darnoceloc:
## Summary
Updated run_type_method but phonoopy test giving error which is independent of my changes.

* Modified run_type_method in Vasprun class within outputs.py script to account for wider range of functionals/levels of theory.
   
* Added tests within test_outputs.py script in io/vasp/tests directory and new test files relevant to various functionals/levels of theory.

## TODO (if any)

to be done

* Add functionality to account for Many-Body perturbation theory (GW, GoWo, scGW, etc.), the Random Phase Approximation (ACDFT/RPA), Time-Dependent DFT (TD-DFT) and Time-Dependent Hartree-Fock equations (TDHF) via the Bethe Salpeter and Casida equations (BSE), and other higher levels of theory.


shyuep:
Thanks.

Issue #1503, opened 2019-06-20
Pymatgen function to get XAS


shyamd:
There needs to be a pymatgen function to get XAS spectra from MP:

The format to get data by MP ID seems to be:
`f"/materials/{mpid}/xas/{absorbing_element}"`

shyuep:
@yimingchen95 Pls handle.

PR #1505, opened 2019-06-21
add get_xas_data method

yimingchen95:
## Summary

This solves issue #1503 by implementing get_xas_data method in MPRester.

* Add get_xas_data method in MPRester
* Add unittest 

* none

## TODO (if any)

* none


shyuep:
The test is too basic. You cannot just test length of data. Test some of the internal structure.

PR #1508, opened 2019-06-23
add grain boundary MPRester

Tinaatucsd:
## Summary

* add grain boundary MPRester to query grain boundary structure and energy data
* add unittest

### DIFF
--- a/pymatgen/ext/matproj.py
+++ b/pymatgen/ext/matproj.py
@@ -1205,6 +1205,42 @@ class MPRester:
         millers, energies = zip(*miller_energy_map.items())
         return WulffShape(lattice, millers, energies)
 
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @Tinaatucsd! Would it be possible to rename `get_gb_data` to `get_grain_boundary_data` before we merge? Just to be more descriptive/more discoverable by people. Otherwise looks good!

shyuep:
@mkhorton I think get_gb is fine. gb is a very well-established abbreviation. We need to balance between descriptive and length of names...

mkhorton:
@shyuep Given the choice I'd prefer to err on the side of verbosity ("code is read more than it's written", tab-completion exists, etc), but don't mind being over-ruled on this one.

PR #1509, opened 2019-06-24
atat to_string method

dongsenfo:
Switched the lattice vector matrix with the scaling matrix, now in format compliant with atat


mkhorton:
Many thanks @dongsenfo!

PR #1510, opened 2019-06-25
add remove atoms option to the GrainBoundaryTransformation

Tinaatucsd:
## Summary

* Update the GrainBoundaryTransformation such that remove atoms option is available. This option is useful if atoms around the grain boundary region are too close to each other.


mkhorton:
Thanks @Tinaatucsd! Do you recommend setting "quick_gen" to True for the interactive app?

Tinaatucsd:
Yes, please. The "quick_gen" will accelerate the generation process but the generated GB maybe not the smallest possible cell. I think it's ok to do visualization using the larger cell. (The smallest cell is useful if the structure will be used to perform DFT calculation)

Issue #1513, opened 2019-06-26
Vasprun does not discriminate both Kpoint sets when LVEL flag is set


JRSuckert:
**Describe the bug**
When running VASP with the LVEL flag set, two kpoint sets are generated (probably IBZ and FBZ).
The Vasprun class will read both to the same variable overwriting the first that has been read.
Unfortunately, the set that is overwritten is the one associated to the eigenvalue array.
In the vasprun, the structure is somewhat like:
```
<modelling>
  <kpoints>
    ....
  </kpoints>
  <electronvelocities>
    <kpoints>
    ....
    </kpoints>
  </electronvelocities>
</modelling>
```

**To Reproduce**
1. Run vasp with LVEL and LOPTICS flag
2. Initialize Vasprun with resulting vasprun.xml
3. Compare size of eigenvalues in Vasprun().eigenvalues with size of Vasprun.actual_kpoints

**Expected behavior**
Separate kpoint variables, or a list of kpoint sets.

Edit: corrected xml tree

shyuep:
Thanks for reporting this. I have no idea what the LVEL tag does. If this is an issue, I would suggest you perhaps fork the repo and implement a quick fix?

JRSuckert:
The actual purpose of the LVEL flag is also unclear to me, as it is undocumented.

In my particular case, I want to use it as it will write the full set of matrix elements <u|d/dk|u> for (ikpt, nbands, nbands) into the WAVEDER file, which by default is only done for a subset.

I will attempt to fix it.

JRSuckert:
I'm not particularly experienced in parsing xml files, so maybe someone else can give me a hint.
The vasprun is parsed using the iterparse method of the ElementTree so it's not possible to determine the according parent.

A quick fix would be to check if self.kpoints has already been set and then skip later tag matches.
```python
if not hasattr(self, 'kpoints'):
  self.kpoints, self.actual_kpoints, self.actual_kpoints_weights = self._parse_kpoints(elem)
```

Right now it should not break anything, since the kpoint set used for self-consistency is quite far up in the element tree, but I'm not sure how future proof it is, maybe someone else got an idea?

PR #1514, opened 2019-06-27
Fix reading the SC K-Point set when LVEL flag is set.

JRSuckert:
## Summary
This PR tries to fix the issue of #1513.
When running vasp with the LVEL flag enabled, two kpoint sets are written to the vasprun.xml file.
One quite far up in the file, containing the kpoints in the irreducible Brillouin zone, and one after the self-consistent calculation inside the electronvelocities tag which contains points in the full Brillouin zone.
Here, I only prevent that the kpoint variables read for the first set in Vasprun are overwritten by the second set. 

Also added a test that ensures that the kpoint dimension of the eigenvalues array is equal to the number of kpoints.

Short few sentences, and summary of the major changes in bullet 
points

* Recover default behaviour, when LVEL flag in vasp calculation is set

## For a later PR

A later feature PR should introduce new variables for the second kpoint set in vasprun.xml, as well as for any quantities calculated due to the LVEL flag (if any).

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -414,9 +414,10 @@ class Vasprun(MSONable):
                     elif tag == "incar":
                         self.incar = self._parse_params(elem)
                     elif tag == "kpoints":
-                        self.kpoints, self.actual_kpoints, \
-                        self.actual_kpoints_weights = self._parse_kpoints(
-                            elem)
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @JRSuckert. I'm not familiar with the LVEL flag, I was about to ask what this is for and then saw the discussion in the issue :)

PR #1515, opened 2019-06-27
Add option to translate molecule to origin in add_adsorbate

oxana-a:
## Summary

Updated _add_adsorbate_ to work with a copy of the adsorbate molecule and added the _translate_ keyword - if _True_, the molecule is translated to the origin before being added to the surface.

* Updated _add_adsorbate_ to work with a copy of the molecule passed as an argument. Previously, if _reorient_ was _True_ (it is by default), the molecule would be rotated in place. In the case of consecutive structure generation (such as in the adsorption workflow from atomate), the adsorbate would keep rotating after each structure generation.
* Added the option to translate the molecule to the origin before adding it to the surface. Previously, the position of the adsorbate was fully dependent on the position of the input molecule - which could lead to structures with adsorbates that are positioned too close to surface, or too far from the actual adsorption site. If _translate_ is _True_, the molecule is first translated so that the center of mass of the substructure made up of the atoms that have the most negative z coordinates is placed at the origin (and hence on the adsorption site). This allows for both head-on and side-on adsorbate placement while ensuring that no atoms are placed closer to the surface than the actual adsorption site.
* Added tests for both changes.

### DIFF
--- a/pymatgen/analysis/adsorption.py
+++ b/pymatgen/analysis/adsorption.py
@@ -400,7 +400,7 @@ class AdsorbateSiteFinder:
                                   if s.coords[2] == min([s.coords[2]
                                   for s in molecule.sites])]
             x, y, z = front_atoms.center_of_mass
-            molecule.translate_sites(vector = [-x, -y, -z])
+            molecule.translate_sites(vector=[-x, -y, -z])
         if reorient:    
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @oxana-a ! We'll wait for the CI tests to pass otherwise looks good to merge

PR #1516, opened 2019-06-27
add options to query certain surface plane, and work of separation 

Tinaatucsd:
## Summary

* When query surface data, a new option to query certain surface plane is added by providing the Miller index. 
* When query grain boundary data, users can also choose to include the work of separation value (unit: J/m^2) of the corresponding grain boundary.


mkhorton:
Thanks @Tinaatucsd !

PR #1517, opened 2019-07-01
Waveder reading changed from FortranFile to directly reading binary.

JRSuckert:
## Summary
This change was made due to the fact that FortranFile can not read records with the number of
entries exceeding int32 limits (filesize >~ 2GB).


mkhorton:
Thanks @JRSuckert !

PR #1518, opened 2019-07-01
Add ShengBTE IO Functions

rees-c:
## Summary

@utf 

Initial commit for interfacing Pymatgen with ShengBTE. More features will be added later.

ShengBTE is a software package for solving the Boltzmann Transport Equation for phonons. It is capable of using 2nd and 3rd order interatomic force constants to calculate lattice thermal conductivities of bulk crystalline solids and nanowires. Full details can be found at the following link: https://doi.org/10.1016/j.cpc.2014.02.015

Added an IO for reading/writing/updating ShengBTE CONTROL files with Python dictionaries.

'f90nml' for reading FORTRAN namelists used as input for running ShengBTE.

### DIFF
--- a/pymatgen/io/shengbte.py
+++ b/pymatgen/io/shengbte.py
@@ -24,9 +24,34 @@ __date__ = "June 27, 2019"
 
 
 class Control(dict, MSONable):
-
     """
     Class for reading, updating, and writing ShengBTE CONTROL files.
+    See  https://bitbucket.org/sousaw/shengbte/src/master/ for more
+    detailed description and default values of CONTROL arguments.
+
+    Args:
+        ngrid (size 3 list): number of grid planes along each axis in
+            reciprocal space
+        lfactor (float): unit of measurement for lattice vectors (nm)
+        scalebroad (float): scale parameter for Gaussian smearing. A value
+            of 1.0 is theoretically guaranteed to work, but significant
+            speedups can sometimes be achieved by reducing it with
+            negligible loss of precision.
+        t (int or float): temperature (Kelvin)
+        **kwargs: Other ShengBTE parameters. Several parameters are required
+            for ShengBTE to run - we have listed these parameters below:
+            - nelements (int): number of different elements in the compound
+            - natoms (int): number of atoms in the unit cell
+            - lattvec (size 3x3 array): real-space lattice vectors, in units
+              of lfactor
+            - types (size natom list): a vector of natom integers, ranging
+              from 1 to nelements, assigning an element to each atom in the
+              system
+            - elements (size natom list): a vector of element names
+            - positions (size natomx3 array): atomic positions in lattice
+              coordinates
+            - scell (size 3 list): supercell sizes along each crystal axis
+              used for the 2nd-order force constant calculation
     """
 
     required_params = ["nelements", "natoms", "ngrid", "lattvec", "types",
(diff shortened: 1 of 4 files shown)


pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,275 @@
...
+"""
+
+__author__ = "Rees Chang"
+__copyright__ = "Copyright 2019, The Materials Project"
+__version__ = "0.1"
+__email__ = "rc564@cornell.edu"
+__date__ = "June 27, 2019"
+
+
+class ShengBTE_CONTROL_IO:
```
utf: Firstly, rename this to `Control`. The fact that it is in the shengbte module means you don't need the shengbte part. Also, we only capitalise the first letter of each word in classes.

I think this could be better written as a class which:
1. Takes several dictionaries in the constructor. One for each section of the file: e.g., flags, parameters, crystal, etc. The constructor then sets the defaults if they are required.
2. Has a static `from_file` method. Which just takes a file path as input.
3. Has a static `from_dict` method, which takes a fully formatted nested dictionary as input.
3. Has a (not static) `to_file` method, which writes the file and takes a file path as input and writes the settings by accessing them through `self`.

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,275 @@
...
+            del sdict['parameters']['t_min']
+        if 't_max' in sdict['parameters']:
+            sdict['parameters']['T_max'] = sdict['parameters']['t_max']
+            del sdict['parameters']['t_max']
+        if 't_step' in sdict['parameters']:
+            sdict['parameters']['T_step'] = sdict['parameters']['t_step']
+            del sdict['parameters']['t_step']
+        return sdict
+
+    def file_writer_helper_func(self, dict, filename):
```
utf: I'm a bit confused, can you not just use the f90ml writing methods instead of doing all of this?

rees-c: I tried using the f90ml writing methods but it returned a format of namelists that ShengBTE couldn't read

pymatgen/io/shengbte.py
rees-c: For required arguments like atom types, atomic positions, etc. that depend on the material, should I still set defaults or just raise an error if they are missing arguments?

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,354 @@
...
+                            self.params_dict[required_namelist_param] is None:
+                        raise AttributeError('Missing argument: {}>{}'.format(required_namelist,
+                                                                              required_namelist_param))
+                    elif required_namelist == 'flags' and \
+                            self.flags_dict[required_namelist_param] is None:
+                        raise AttributeError('Missing argument: {}>{}'.format(required_namelist,
+                                                                              required_namelist_param))
+        check_required_params()
+
+    @classmethod
```
utf: This should be a static method not class method. I.e. (`@staticmethod`).

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,354 @@
...
+        if 't_step' in sdict['parameters']:
+            sdict['parameters']['T_step'] = sdict['parameters']['t_step']
+            del sdict['parameters']['t_step']
+
+        alloc_dict = sdict['allocations']
+        crystal_dict = sdict['crystal']
+        params_dict = sdict['parameters']
+        flags_dict = sdict['flags']
+
+        return cls(alloc_dict, crystal_dict, params_dict, flags_dict)
```
utf: Change this to `return Control(...)`.

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @classmethod
```
utf: Again, make this static.

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,354 @@
...
+        try:
+            params_dict = sdict['parameters']
+        except:
+            params_dict = {}
+        try:
+            flags_dict = sdict['flags']
+        except:
+            flags_dict = {}
+
+        return cls(alloc_dict, crystal_dict, params_dict, flags_dict)
```
utf: Again, remove the reference to cls.

shyamd:
Quick note. Any factory functions should be class methods. This ensures that if you just inherit from them the factory function generates the inherited class.

utf:
Thanks for the info @shyamd 

@rees-c Can you make the `from_file` function a class method again. Also make this method call the `cls.from_dict` method.

utf:
@mkhorton This PR is no longer WIP.

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,156 @@
...
+    def to_file(self, filename):
+        """
+        Writes ShengBTE CONTROL file from 'Control' object
+        """
+
+        for param in self.required_params:
+            if param not in self.as_dict():
+                warnings.warn(
+                    "Required parameter '{}' not specified!".format(param))
+                raise AttributeError("param {} not specified".format(param))
```
utf: Remove this line. We already give a warning. If a user wants to make an invalid ShengBTE input file they should be able to.

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,156 @@
...
+
+__author__ = "Rees Chang"
+__copyright__ = "Copyright 2019, The Materials Project"
+__version__ = "0.1"
+__email__ = "rc564@cornell.edu"
+__date__ = "June 27, 2019"
+
+
+class Control(dict, MSONable):
+
```
utf: Remove this blank line

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    Class for reading, updating, and writing ShengBTE CONTROL files.
+    """
```
utf: The class docstring should contain the init argument definitions. E.g. see https://google.github.io/styleguide/pyguide.html#384-classes

Basically just move the whole of the init docstring to this docstring and don't put a docstring for init.

pymatgen/io/shengbte.py
```diff
@@ -0,0 +1,156 @@
...
+        https://bitbucket.org/sousaw/shengbte/src/master/.
+        Note some parameters are mandatory. Optional parameters
+        default here to None and will not be written to file.
+
+        Args:
+            dict: A Python dictionary of ShengBTE input parameters.
+        """
+        return cls(**sdict)
+
+    def to_file(self, filename):
```
utf: Can you set the filename default to "CONTROL"

mkhorton:
Thanks @rees-c and @utf

PR #1519, opened 2019-07-01
Added CSLD functionality

rees-c:
This branch adds functionality to ultimately be used in a high-throughput workflow implementing compressed sensing lattice dynamics for rapidly computing interatomic force tensors.

10.1103/PhysRevLett.113.185501
https://arxiv.org/abs/1805.08904
https://arxiv.org/abs/1805.08903

## Summary
@utf 

Added 'CubicSupercellTransformation' and 'PerturbedSupercellsTransformation' classes to 'advanced_transformations.py'.

'CubicSupercellTransformation' generates a transformation matrix (np.ndarray) and a supercell (Pymatgen Structure) that is as cubic as possible given a structure (Pymatgen Structure) and several parameter constraints. Constraints include a minimum number of atomic nearest neighbor distances that each dimension of the supercell must exceed, and bounds for the total number of atoms in the supercell.

'PerturbedSupercellsTransformation' generates a list of Pymatgen Structures with randomly displaced atomic coordinates given a Pymatgen Structure. Parameters are input describing the range and magnitude of the displacements as well as the number of displaced structures to generate.


pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+        self.num_nn_cutoff = num_nn_dists
+
+        self.lat_vecs = self.structure.lattice.matrix
+        bond_matrix = self.structure.distance_matrix
+        np.fill_diagonal(bond_matrix, np.Inf)
+        self.nn_dist = np.amin(bond_matrix)
+
+        self.smallest_dim = None # smallest dimension of the resulting supercell
+
+    def round_away_from_zero(self, x):
```
utf: This function should be private.

Also, what is it actually doing? Can you add some docs.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+        aX = abs(x)
+        return math.ceil(aX) * (aX / x) if x != 0 else 0
+
+    def _round_and_make_arr_singular(self, arr):
```
utf: Again, what is this doing?

You should add comments in case someone later on needs to debug your code.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
     @property
     def is_one_to_many(self):
         return False
+
+class CubicSupercellTransformation:
+    """
+    A transformation that aims to generate a nearly cubic supercell structure from a structure.
+    """
```
utf: Can you add a summary of the algorithm here.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+            zero_col_idxs = np.where(~arr_rounded.any(axis=0))[0]
+            for zero_col_idx in zero_col_idxs:
+                zero_col = arr[:, zero_col_idx]
+                row_idx_to_fix = np.where(np.absolute(zero_col) == np.amax(np.absolute(zero_col)))[0]
+                for i in row_idx_to_fix:
+                    arr_rounded[i, zero_col_idx] = self.round_away_from_zero(arr[i, zero_col_idx])
+        return arr_rounded
+
+    @staticmethod
+    def proj(b, a):
```
utf: This doesn't need to be a class method. Can move this to the bottom of advanced_transformations. It should also be private and needs a docstring.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+        # Returns vector projection of b onto a
+        return (b.T @ (a / np.linalg.norm(a))) * (a / np.linalg.norm(a))
+
+    def gen_scaling_matrix(self):
+        """
+        Returns: supercell (Structure), transformation matrix (np.ndarray)
```
utf: You should also put an outline of what the algorithm is doing here.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+                superstructure = SupercellTransformation(trans_mat).apply_transformation(self.structure)
+                num_at = superstructure.num_sites
+
+                if self.smallest_dim >= hard_sc_size_threshold \
+                        and num_at >= self.min_atoms and num_at <= self.max_atoms:
+                    return superstructure, trans_mat
+                else:
+                    # Increase threshold until proposed supercell meets requirements
+                    target_sc_size += 0.1
+                    if num_at > self.max_atoms:
```
utf: Is it possible for this while loop to run indefinitely, e.g. if max_atoms is infinite?

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+                        raise AttributeError('While trying to solve for the supercell, the max'
+                                             'number of atoms was exceeded. Try lowering the number'
+                                             'of nearest neighbor distances.')
+
+class PerturbedSupercellsTransformation:
+    """
+        Uses a supercell Structure, a transformation matrix, a min/max displacement,
```
utf: A few formatting notes:
- The line length is too long for many of the lines. Can you manually wrap lines at around 70 characters.
- This whole section is indented one tab too far.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+                    return superstructure, trans_mat
+                else:
+                    # Increase threshold until proposed supercell meets requirements
+                    target_sc_size += 0.1
+                    if num_at > self.max_atoms:
+                        raise AttributeError('While trying to solve for the supercell, the max'
+                                             'number of atoms was exceeded. Try lowering the number'
+                                             'of nearest neighbor distances.')
+
+class PerturbedSupercellsTransformation:
```
utf: Can you rename this to `PeturbSitesTransformation` as this doesn't necessarily need to be run on a supercell.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+    """
+        Uses a supercell Structure, a transformation matrix, a min/max displacement,
+        number of displacement values, and number of supercells per displacement
```
utf: Is this description correct? It doesn't seem to take a transformation matrix.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+        to generate a list of randomly perturbed supercells from a given supercell.
```
utf: Can you remove supercell from this whole description and argument names. E.g. `original_supercell` -> `structure`. As the cells might not be supercells.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+
+        Args:
+            original_supercell (Structure): supercell to displace, typically already relaxed
+            max_displacement_val (float): maximum displacement value for perturbing the supercell (in Angstroms)
```
utf: Remove "val" from argument name. See below also.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+            min_displacement_val (float): minimum displacement value for perturbing the supercell (in Angstroms)
+            num_displacements (int): number of unique displacement values to try, bounded by
+                min_displacement and max_displacement. This argument is ignored if min_displacement
+                is not passed as an argument.
+            scs_per_displacement_val (int): number of perturbed supercells to generate for each unique
```
utf: Rename to `num_structures`.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+
+        if floor_displacement is not None:
+            self.floor_disp = float(floor_displacement)
+        else:
+            self.floor_disp = None
+
+        self.disp_vals = np.linspace(min_displacement_val, max_displacement_val, num=num_displacements)
+        self.perturbed_supercells = self.generate_transformations()
+
+    def random_displacements(self, natom, rmax, rmin, dim=3):
```
utf: Make private (E.g., rename to `_random_displacements`)

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
     @property
     def is_one_to_many(self):
         return False
+
+class CubicSupercellTransformation:
```
utf: This class doesn't subclass `AbstractTransformation`.

E.g. see `class SupercellTransformation(AbstractTransformation):` Note the `(AbstractTransformation)`.

Furthermore, it doesn't implement the required functions that all abstract transformations should. See `SupercellTransformation` for an example. E.g., you should add and modify the following:

```python
    def apply_transformation(self, structure):
        return structure * self.scaling_matrix

    def __str__(self):
        return "Supercell Transformation with scaling matrix " + \
            "{}".format(self.scaling_matrix)

    def __repr__(self):
        return self.__str__()

    @property
    def inverse(self):
        raise NotImplementedError()

    @property
    def is_one_to_many(self):
        return False
```

Read the `AbstractTransformation` docs for more information.

pymatgen/transformations/advanced_transformations.py
utf: Also, see my notes for the cubic transformation:

This class doesn't subclass `AbstractTransformation`.

E.g. see `class SupercellTransformation(AbstractTransformation):` Note the `(AbstractTransformation)`.

Furthermore, it doesn't implement the required functions that all abstract transformations should. See `SupercellTransformation` for an example. E.g., you should add and modify the following:

```python
    def apply_transformation(self, structure):
        return structure * self.scaling_matrix

    def __str__(self):
        return "Supercell Transformation with scaling matrix " + \
            "{}".format(self.scaling_matrix)

    def __repr__(self):
        return self.__str__()

    @property
    def inverse(self):
        raise NotImplementedError()

    @property
    def is_one_to_many(self):
        return False
```

Read the `AbstractTransformation` docs for more information.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+            dr = self.random_displacements(na, max_disp, floor_disp) #generate random displacements
+
+            # Perturb structure with the random displacements
+            for i in range(len(structure._sites)):
+                structure.translate_sites([i], dr[i], frac_coords=False, to_unit_cell=True)
+            return structure
+        else:
+            print('Displacement entered is not a float.')
+
+    def generate_transformations(self):
```
utf: What is this for? 

I think this should instead be refactored to override `apply_transformation` superfunction

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+        dx_norms = np.linalg.norm(dx, axis=1)
+        veclen = np.full(natom, rmax) if rmin is None else np.random.uniform(rmin, rmax, natom)
+
+        if not 0 in dx_norms:
+            return dx * (veclen / dx_norms)[:, None]
+        else:
+            self.random_displacements(natom, rmax, rmin, dim)
+
+
+    def perturb_supercell(self, structure, max_disp, floor_disp):
```
utf: This should be private. Also needs documentation about how the sites are displaced and what the arguments control.

pymatgen/transformations/tests/test_advanced_transformations.py
```diff
@@ -614,6 +615,30 @@ def test_apply_transformation(self):
         self.assertDictEqual(output[-1].species.as_dict(),
                              {'Ni': 0.5, 'Ba': 0.5})
 
+class CubicSupercellTransformationTest(PymatgenTest):
+
+    def test_gen_scaling_matrix(self):
```
utf: Look at the tests for other transformers. You should be testing the `apply_transformation` function. You are also not validating that the output of the transformation is correct.

See the `assert` functions in other tests.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,213 @@ def inverse(self):
...
+        max_disp = float(max_disp)
+        if isinstance(max_disp, float):
+            dr = self.random_displacements(na, max_disp, floor_disp) #generate random displacements
+
+            # Perturb structure with the random displacements
+            for i in range(len(structure._sites)):
+                structure.translate_sites([i], dr[i], frac_coords=False, to_unit_cell=True)
+            return structure
+        else:
+            print('Displacement entered is not a float.')
```
utf: This should raise an error rather than printing something.

pymatgen/transformations/tests/test_advanced_transformations.py
rees-c: So I wasn't sure about how to test 'apply_transformation' because I don't necessarily know what the 'correct' output is. For example, if I changed the line 'target_sc_size += 0.1' to '+= 0.01', I might find a better transformation that the step size of '0.1' was too big to catch. Should I simply test that apply_transformation produces a valid Pymatgen structure? Or maybe that the supercell that is at least as cubic as what my current version yields? Even then I will need to somewhat arbitrarily define some measure of how cubic something is.

utf: The cubic transformation is deterministic, so you can just check the lattice parameters match what you expect and that the cell has the correct number of atoms.

For the perturb sites, you should check:
- That the number of structures you asked for is returned
- That the cell has been perturbed, e.g. test that some of the sites are not in their original positions.

pymatgen/transformations/advanced_transformations.py
rees-c: If 'max_atoms' is infinite, then there will always be a supercell that has at least 'min_atoms' atoms and dimensions larger than 'hard_sc_size_threshold' (i.e. some arbitrarily large supercell), in which case the loop will terminate at the return statement. I'll implement a max number of iterations anyways just in case.

pymatgen/transformations/advanced_transformations.py
rees-c: Good catch!

pymatgen/transformations/advanced_transformations.py
rees-c: Renaming 'num_displacements' to 'num_structures' might be a little misleading since the number of structures is actually 'num_displacements' times 'scs_per_displacement_val'. Would 'num_unique_displacement_vals' be more descriptive?

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+    def inverse(self):
+        return None
+
+    @property
+    def is_one_to_many(self):
+        return False
+
+class PerturbSitesTransformation(AbstractTransformation):
+    """
+    Uses a structure, a min/max displacement, a number of displacement
```
utf: This docstring is still quite confusing. How many displacements will be made?

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+                 max_displacement_val,
+                 min_displacement_val,
+                 num_displacements,
+                 structures_per_displacement_val,
+                 floor_displacement=None):
+        """
+        Args:
+            original_structure (Structure): structure to displace, typically
+                already relaxed
+            max_displacement_val (float): maximum displacement value for
```
utf: Can you remove "val" from the option names.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+            min_displacement_val (float): minimum displacement value for
+                perturbing the structure (Angstroms)
+            num_displacements (int): number of unique displacement values to
+                try, uniformly distributed between 'min_displacement' and
+                'max_displacement'. This argument is ignored if min_displacement
+                is not passed as an argument.
+            structures_per_displacement_val (int): number of perturbed structures to
+                generate for each unique displacement value.
+
+            floor_displacement (Optional float): If None (default), then for a
```
utf: Why is this a float? Should this just be a boolean? This docstring makes it sound like the value is not actually used.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+        self.structures_per_disp_val = structures_per_displacement_val
+
+        if floor_displacement is not None:
+            self.floor_disp = float(floor_displacement)
+        else:
+            self.floor_disp = None
+
+        self.disp_vals = np.linspace(min_displacement_val, max_displacement_val, num=num_displacements)
+
+    def _random_displacements(self, natom, rmax, rmin, dim=3):
```
utf: This needs a proper docstring, what do the arguments control?

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+        dx_norms = np.linalg.norm(dx, axis=1)
+        veclen = np.full(natom, rmax) if rmin is None else np.random.uniform(rmin, rmax, natom)
+
+        if not 0 in dx_norms:
+            return dx * (veclen / dx_norms)[:, None]
+        else:
+            self._random_displacements(natom, rmax, rmin, dim)
+
+    def _perturb_structure(self, structure, max_disp, floor_disp):
+        # *** Adapted from CSLD's 'polaron_main' file
```
utf: the main docstring, e.g. in the """ section should be directly after the function definition. This line can be in the docstring.

You should also give the full `args:` listing of the arguments in the docstring too.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+        First, all entries in 'arr' will be rounded to the nearest integer to
+        yield 'arr_rounded'.
+        If 'arr_rounded' has any zero rows, then one element in each zero row of 'arr_rounded'
+        corresponding to the element in 'arr' of that row with the largest absolute valued
+        magnitude will be rounded to the next integer away from zero (see the '_round_away_from_zero(x)'
+        function) rather than the nearest integer.
+        This process is then repeated for zero columns.
+        Also note that if 'arr' already has zero rows or columns, then this function
+        will not change those rows/columns.
+        :param arr: np.ndarray
```
utf: Can you make these google style docstrings.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+        # Variables to be solved for by 'apply_transformation()'
+        self.smallest_dim = None # smallest direction of the resulting supercell
+        self.trans_mat = None # transformation matrix
+
+    def _round_away_from_zero(self, x):
+        """
+        Returns 'x' rounded to the next integer away from 0.
+        If 'x' is zero, then returns zero.
+        E.g. -1.2 rounds to -2.0. 1.2 rounds to 2.0.
+        :param x: float to be rounded
```
utf: Can you make these google style docstrings.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+class PerturbSitesTransformation(AbstractTransformation):
+    """
+    Uses a structure, a min/max displacement, a number of displacement
+    values, and a number of supercells per unique displacement value
+    to generate a list of randomly perturbed supercells from a given
+    supercell.
+    """
+
+    def __init__(self,
+                 max_displacement_val,
```
utf: Can you set sensible default values for all of these options.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,295 @@ def inverse(self):
...
+                already relaxed
+            max_displacement_val (float): maximum displacement value for
+                perturbing the structure (Angstroms)
+            min_displacement_val (float): minimum displacement value for
+                perturbing the structure (Angstroms)
+            num_displacements (int): number of unique displacement values to
+                try, uniformly distributed between 'min_displacement' and
+                'max_displacement'. This argument is ignored if min_displacement
+                is not passed as an argument.
+            structures_per_displacement_val (int): number of perturbed structures to
```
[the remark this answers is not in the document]
rees-c: So the CSLD source code had it such that every atom moves a random direction but a constant distance, and you choose multiple constant distances to make different structures with. There was functionality in the code to have a mix of displacements for each structure but Junsoo said to keep what we've been using.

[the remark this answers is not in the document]
rees-c: If we want to try a mix of displacements for each structure, you can just pass a value into 'floor_displacement'

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,356 @@ def inverse(self):
...
+
+    The algorithm is as follows:
+    - Generate a list of distances between min_displacement and
+      max_displacement.
+    - If floor_displacement is None, a structure is generated for each of
+      the displacement distances where all atoms are displaced in random
+      directions from their original locations by the displacement distance.
+    - If structures_per_displacement is greater than 1, multiple structures
+      per displacement distance will be generated.
+    - If floor_displacement is a number (float less than min_displacement),
```
utf: Can you update this to the new variable name for the whole of this docstring.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,356 @@ def inverse(self):
...
+            max_displacement (float): maximum displacement value for
+                perturbing the structure (Angstroms)
+            min_displacement (float): minimum displacement value for
+                perturbing the structure (Angstroms)
+            num_displacements (int): number of unique displacement values to
+                try, uniformly distributed between 'min_displacement' and
+                'max_displacement'.
+            structures_per_displacement_distance (int): number of perturbed
+                structures to generate for each unique displacement value.
+
```
utf: Remove this blank line.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1360,3 +1361,356 @@ def inverse(self):
...
+        Args:
+            max_displacement (float): maximum displacement value for
+                perturbing the structure (Angstroms)
+            min_displacement (float): minimum displacement value for
+                perturbing the structure (Angstroms)
+            num_displacements (int): number of unique displacement values to
+                try, uniformly distributed between 'min_displacement' and
+                'max_displacement'.
+            structures_per_displacement_distance (int): number of perturbed
+                structures to generate for each unique displacement value.
```
utf: Can you change value to distance to match the docstring wording.

utf:
Hi Rees, the tests are still failing. Can you have a look and try and fix please.

utf:
@mkhorton 

This PR is no longer WIP.

mkhorton:
Thanks @utf and @rees-c, this looks really useful! I had a request for something along the lines of CubicSupercellTransformation very recently in the context of creating input cells for STEM image simulations, so this is great.

Before merging, I have a request: @rees-c in the docstrings where you mention CSLD can you expand the acronym and provide the references you mention above? Also where any reference is made to the CSLD code, a link to their code/repo would also be useful.

utf:
@mkhorton re linking to the CLSD repo: the CLSD code is not available online just yet. We have an advance version of the code in our group, so are adding features to pymatgen and atomate with the expectation that it will be released publicly in the next month or so.

mkhorton:
Ah ok, that's fine then -- explains why I couldn't find it myself :) If you could submit a quick PR to update the docstring when it's publicly released that'd be great.

Once the acronym/refs are added I'll go ahead and merge

rees-c:
@mkhorton updated the docstring :)

mkhorton:
Thanks @rees-c!

PR #1521, opened 2019-07-03
Pseudo_zval bugfix

sthartman:
Fixes a rare bug in the Outcar.read_pseudo_zval method of pymatgen.io.vasp.outputs.  The new method doesn't make any assumptions about the POSCAR's title line, since this has no meaning in VASP.  If adding a new OUTCAR test file is unwanted, I can modify the existing OUTCAR.BaTiO3.polar to check this issue instead.  Thanks to @blondegeek for the help!

* Outcars now get their zval_dict keys from more reliable part of the OUTCAR

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -2583,27 +2583,29 @@ class Outcar:
         Create pseudopotential ZVAL dictionary.
         """
         try:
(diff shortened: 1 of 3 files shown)


blondegeek:
@sthartman Thanks for catching the bug and making the fixes!

shyuep:
Thanks!

PR #1524, opened 2019-07-12
Fix reading frequency dependent dielectric constant from VASP 6.0.8 output

utf:
In VASP 6.0.8, the frequency dependent dielectric constant is written to the vasprun.xml in a slightly different format from previous versions.

This PR makes a slight modification to support parsing the new dielectric format.

I've added a test and example vasprun.xml to catch the the broken behavior.


mkhorton:
Thanks @utf

PR #1526, opened 2019-07-16
[WIP] TEM Diffraction

thefrankwan:
## Summary:
Working draft of TEM calculator.  Have noticed some problems with the intensity calculations (that seem to stem from get_atoms not working as it should).

* Feature: Calculates and displays electron diffraction patterns for any regular crystalline object in any of the seven classes of crystals.  Displays in nice "futuristic" format.

## Additional dependencies introduced:
*PrettyTable: needed to display TEM diffraction details in clean format

## TODO:
Re-examine cell_scattering_factors to make sure it's not double counting (though I believe this
should only affect calculations by a factor of two which cancels out anyway in the normalization of
cell intensity). (high priority)
Optimize runtime for patterns. Runtime is solid for simple structures like CsCl, but long (~10s) for complicated structures like rotaxanes and "molecules that chemists deal with". (best bet is to memoize certain calculations that reoccur many times) (mid-low priority)
Rearrange code so that the user decides only the structure to be tested, the Laue zone to be tested, and the point range to be tested for a given initialized calculator using the get_2d_pattern function.
Finish writing tests for the other functions. (low priority)
Make MSONable. (high priority)
Add type hints for every method. (low priority)

Start running this code on 120k-ish structures and doing cluster analyses on them.  (high priority)
Write a function that gets all unique crystal planes based on crystal symmetries (?) (high priority)


pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,788 @@
...
+class TEMCalculator(AbstractDiffractionPatternCalculator):
+    """
+    Computes the TEM pattern of a crystal structure for multiple Laue zones.
+    """
+
+    """
+    some important constants needed for calculations
+    """
+    planck = 6.62607004 * 10**-34
+    restMassElec = 9.1093835611 * 10**-31
```
shyuep: Please follow pymatgen's convention for variable names, i.e., underscores, not CamelCase.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,788 @@
...
+        # TODO: memoize this as a dict of laue zone to pointlist. prob want a system where if you input a diff range of 
+        # initial points, you edit the cache entry.
+
+        observed_points = []
+        for point in points:
+            if np.dot(self.beam_direction, np.transpose(point)) == laue_zone:
+                observed_points.append(point)
+        return observed_points
+
+    def dist_cubic(self, structure, point):
```
shyuep: The interplanar distance for any crystal system is well defined as 1/g_hkl. There is no need for specific methods for different crystal structures. See slide 19 of https://www.slideshare.net/shyuep/ucsd-nano106-03-lattice-directions-and-planes-reciprocal-lattice-and-coordinate-transformations. You simply need to compute the reciprocal lattice vector and take the norm and then the reciprocal of that, instead of relying on the lattice parameters. Alternatively, you can use the metric tensor and take the inverse to compute the reciprocal metric tensor.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -88,10 +95,10 @@ class TEMCalculator(AbstractDiffractionPatternCalculator):
...
     """
     planck = 6.62607004 * 10**-34
-    restMassElec = 9.1093835611 * 10**-31
-    chargeElec = 1.60217662 * 10**-19 
-    speedLight = 299792458
-    vacuumPerm = 8.8541878176 * 10**-12
+    rest_mass_elec = 9.1093835611 * 10**-31
+    charge_elec = 1.60217662 * 10**-19 
+    speed_light = 299792458
+    vacuum_perm = 8.8541878176 * 10**-12
```
shyuep: These should not be hard coded. Since scipy is a dependency, you can use scipy.constants for all of these. Their values are far more accurate and updated.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -20,6 +20,13 @@
...
 # Copyright (c) Pymatgen Development Team.
 # Distributed under the terms of the MIT License.
+# tempattern inherits from diffractionpattern
+#make msonable
+#run TEM on all 120k strucs
+#remove: symprec
+#voltage default set to 200 instead of 300 kv
+#beam directions: all permutations/families of 1, -1 (-1-1-1, 111, 110, 100)
+#cluster analysis of the points: min vector (1d or 2d)
 from math import *
```
shyuep: * imports are not encouraged.

PR #1527, opened 2019-07-16
Get symmetrically equivalent Miller indices

richardtran415:
## Summary

Added a new function in surface.py called get_symmetrically_equivalent_miller_indices which takes in a structure and Miller index and returns a list for the family of hkl. Useful for searching for Miller index specific entries on MPRester (surface data, GB data, Wsep) as the user does not need to enter the exact same Miller index as the one associated with an entry. Modified get_surface_data using this function to make it easier to obtain work of separation from GB data.

* Feature 1 get_equivalent_miller_indices(structure, miller_index)
* Fix 1: Can now get specific surfaces without needing the exact same Miller index

## TODO (if any):
Some overlap between get_symmetrically_equivalent_miller_indices and get_symmetrically_distinct_miller_indices.

### DIFF
--- a/pymatgen/core/surface.py
+++ b/pymatgen/core/surface.py
@@ -1542,7 +1542,6 @@ def get_d(slab):
     return slab.lattice.get_cartesian_coords([0, 0, d])[2]
 
 
-
 def get_recp_symmetry_operation(structure, symprec=0.01):
     """
     Find the symmetric operations of the reciprocal lattice,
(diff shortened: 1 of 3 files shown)


mkhorton:
Thanks @richardtran415 ! I maintain some of this would be more appropriate in core.lattice, perhaps in future

richardtran415:
Yeah I see what you mean, that would make it easier to get rid of some of overlap I mentioned.

Thanks

shyuep:
Let's move it to lattice then.

mkhorton:
Yes, if you ever decide you'd like to move it, I would support that. Although of course very important to surfaces, Miller indices and planes are fundamentally a property of a Lattice and relevant to non-surface applications too so I think it would make sense.

richardtran415:
Will do

PR #1529, opened 2019-07-17
Fix bug in bandoverlaps class

JaGeo:
Hello again, 

I am sorry. I have already found a small bug in the bandoverlaps class. It should be fixed now and I have also adapted the tests to avoid this problem in the future.


mkhorton:
No problem!

PR #1535, opened 2019-07-23
Add return_all_molecules option to BabelMolAdaptor

### LINKED ISSUE #1512: BabelMolAdaptor should be able to return all molecules found in a file
**Is your feature request related to a problem? Please describe.**
The way it's currently implemented, ``pymatgen.io.babel.BabelMolAdaptor`` will only ever return the first molecule it finds in a file and discard the rest:
https://github.com/materialsproject/pymatgen/blob/5006e2c92d669a05775fcf67dd9b1f0b972c9636/pymatgen/io/babel.py#L314-L315
It still *loads* all of them though and holds them all in memory simultaneously, so in a way it's the worst of both worlds.

**Describe the solution you'd like**
It would be cool if all molecules found within a file could be accessed.

**Describe alternatives you've considered**
- not implementing this

**Additional context**
There is precedent for this e.g. in [``pymatgen.io.xyz``](https://pymatgen.org/pymatgen.io.xyz.html), which loads and stores everything by default and lets you access either only the last frame via the ``molecule`` property or all frames via ``all_molecules``.

smheidrich:
## Summary

As suggested in https://github.com/materialsproject/pymatgen/issues/1512#issuecomment-506859409, this PR adds a ``return_all_molecules`` option to ``BabelMolAdaptor``, which, when set to ``True``, will return one instance of the adaptor for each molecule found in a file. The previous behavior was to always return just the first molecule found and discard the rest, which this PR leaves as the default behavior for backwards-compatibility.

Fixes #1512.

### DIFF
--- a/pymatgen/io/babel.py
+++ b/pymatgen/io/babel.py
@@ -145,7 +145,7 @@ class BabelMolAdaptor:
 
         Args:
             idx1: The atom index of one of the atoms participating the in bond
-            idx2: The atom index of the other atom participating in the bond 
+            idx2: The atom index of the other atom participating in the bond
         """
         for obbond in ob.OBMolBondIter(self._obmol):
             if (obbond.GetBeginAtomIdx() == idx1 and obbond.GetEndAtomIdx() == idx2) or (obbond.GetBeginAtomIdx() == idx2 and obbond.GetEndAtomIdx() == idx1):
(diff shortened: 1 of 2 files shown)


smheidrich:
The errors in CI look like they aren't my fault, so I'm marking this as ready for review.

shyuep:
Looks good. Thanks.

smheidrich:
Thanks for merging! :+1:

Issue #1536, opened 2019-07-23
CubicSupercellTransformation and PerturbSitesTransformation not MSONable


mkhorton:
@utf @rees-c 

This should be trivially fixable, just make sure the class attribute names match the keyword arg names and that the default values are JSON serializable (e.g. `None` for a default and set `np.Inf` in the body of the init), and then it should "just work." @rees-c you can check this by calling `.as_dict()`, `.from_dict()` etc

mkhorton:
I also realize there's something overlap between `PerturbStructureTransformation` and `PerturbSitesTransformation`; thoughts on merging these?

mkhorton:
Additionally, I just realized that PeturbSitesTransformation is a one-to-many but it doesn't follow the correct format of a one-to-many transformation (see the AbstractTransformation apply_transformation docstring or other one-to-many transformations for an example).

In the long term, I think we could probably clean up the AbstractTransformation base class to make the method signature more consistent between one-to-one and one-to-many transformations, but for now PerturbSites should be fixed to make it consistent with other transformations, else it will not be able to be used e.g. for TransmuterFW in workflows etc.

utf:
@rees-c Can you respond to Matts comments/fix the transformation.

rees-c:
Hey @mkhorton ! Sorry for the late reply - I'm a bit new to Git and my gmail seems to silence my Github notifications. I'll send another pull request with the JSON serialization fixes and the corrected one-to-many format for PerturbSitesTransformation.

I see now that PerturbSitesTransformation can replicate the output of PerturbStructureTransformation. On the flip side, I feel that if a user only needs a single perturbed structure, then reading through a one-to-many transformation with its corresponding variable names could be nonintuitive (i.e. if a user wanted to replicate the output of PerturbStructureTransformation using PerturbSitesTransformation, they would have to set `return_ranked_list=False` in `apply_transformation`, `min_random_distance=None` in the init, and `min_disp` to the desired distance). Do you have any thoughts about what would be the best way to merge the two while maintaining backwards compatibility if we choose to do so?

mkhorton:
No worries, and thanks for the fix.

Regarding merging them, they *should* be merged, but I'm agnostic to the specific method.

The `PerturbStructureTransformation` is basically a wrapper around the `Structure.perturb` method, so one option would be to improve that method to make it more powerful.

[...]

PR #1544, opened 2019-08-01
Update CubicSupercellTransformation and merge PerturbSitesTransformation into PerturbStructureTransformation

rees-c:
## Summary
- Added a boolean argument to CubicSupercellTransformation that can constrain the transformation matrix to be diagonal
- Made a bug fix in CubicSupercellTransformation regarding how the smallest dimension is found
- Merged PerturbSitesTransformation into PerturbStructureTransformation (as discussed in #1536)
  - Removed PerturbSitesTransformation and its unit test from advanced_transformations
  - Added an option to PerturbStructureTransformation and Structure.perturb that allows for the displacements to be randomly distributed between two values.
  - Added to unit tests for PerturbStructureTransformation and Structure.perturb() to reflect the new option
  - The rest of the advanced features of the old PerturbSitesTransformation will be implemented into my CSLD atomate workflow by calling the standard transformation multiple times
  - Changed the args of PerturbStructureTransformation to match those of Structure.perturb (i.e. changed `amplitude` to `distance`). *****NOTE THIS IS POTENTIALLY A BREAKING CHANGE FOR ANY ATOMATE WORKFLOW USES `PerturbStructureTransformation`.**


pymatgen/core/structure.py
```diff
@@ -3178,21 +3178,31 @@ def rotate_sites(self, indices=None, theta=0, axis=None, anchor=None,
...
+                'min_random_distance' and 'distance'.
+
         """
 
         def get_rand_vec():
             # deals with zero vectors.
             vector = np.random.randn(3)
             vnorm = np.linalg.norm(vector)
-            return vector / vnorm * distance if vnorm != 0 else get_rand_vec()
+            if isinstance(min_random_distance, float) or \
```
utf: You can write these two lines as just:

```python
 if isinstance(min_random_distance, (float, int)):
```

pymatgen/transformations/standard_transformations.py
```diff
@@ -680,15 +680,20 @@ class PerturbStructureTransformation(AbstractTransformation):
     Args:
         amplitude (float): Amplitude of perturbation in angstroms. All sites
             will be perturbed by exactly that amplitude in a random direction.
+        min_random_distance (None, int, or float): If None, perturb
```
utf: For the transformation, can you update this to be `min_random_amplitude` to match the other argument names.

Also edit the rest of the docstring accordingly.

pymatgen/transformations/standard_transformations.py
```diff
@@ -680,15 +680,20 @@ class PerturbStructureTransformation(AbstractTransformation):
...
-    def __init__(self, amplitude=0.01):
+    def __init__(self, amplitude=0.01, min_random_distance=None):
 
         self.amplitude = amplitude
+        self.min_random_distance = min_random_distance
 
     def apply_transformation(self, structure):
         s = structure.copy()
-        s.perturb(self.amplitude)
+        s.perturb(self.amplitude, self.min_random_distance)
```
utf: to avoid any errors arising from new variables being added to `Structure.perturb`, can you make this line:

```python
s.perturb(self.amplitude, min_random_distance=self.min_random_ampliude)
```

utf:
You should mention that your way of merging the two classes was to:
- Remove the advanced transformation.
- Add an option to `PerturbStructureTransformation` and `Structure.perturb` that allows for the displacements to be randomly distributed between two values.

The rest of the advanced features of the transformation will be implemented in your own code by calling the standard transformation multiple times.

pymatgen/core/structure.py
rees-c: Oh nice, didn't know you could do that

pymatgen/core/structure.py
```diff
@@ -3178,21 +3178,30 @@ def rotate_sites(self, indices=None, theta=0, axis=None, anchor=None,
...
+                'min_random_distance' and 'distance'.
+
         """
 
         def get_rand_vec():
             # deals with zero vectors.
             vector = np.random.randn(3)
             vnorm = np.linalg.norm(vector)
-            return vector / vnorm * distance if vnorm != 0 else get_rand_vec()
+            if isinstance(min_random_distance, (float, int)):
```
utf: In fact this whole thing could be rewritten much more nicely as:

````python
 if isinstance(min_random_distance, (float, int)):
    distance = np.random.uniform(min_random_distance, distance)

return vector / vnorm * distance if vnorm != 0 else get_rand_vec()
```

mkhorton:
Thanks for working on this, it's looking much better! Quick note that since we're modifying the `.perturb()` method there will also need to be a test in `pymatgen.core` for the method also, otherwise after @utf's comments are resolved I think this looks good.

mkhorton:
Perhaps one way to improve legibility would be to change the arg name on `perturb` to `min_amplitude` and say "if None, all displacements will be equal amplitude" (just a small note because we're changing from using the word "amplitude" to the word "random_distance" to describe essentially the same thing).

pymatgen/core/structure.py
rees-c: So for some reason when I tried that code, it treats `distance` inside `np.random.uniform(min_random_distance, distance)` as an unresolved variable local to get_rand_vec(). Would this be an ok solution:

```python
dist = distance
if isinstance(min_random_distance, (float, int)):
     dist = np.random.uniform(min_random_distance, dist)
return vector / vnorm * dist if vnorm != 0 else get_rand_vec()
```

utf:
@mkhorton currently the main perturb argument is called “distance”, should we also change this to amplitude? 

Personally, I think distance is a more obvious naming choice. However, if we were to use this throughout it would be a breaking change as amplitude is a named argument in the transformation class.

mkhorton:
Ah, apologies -- I was looking at the StandardTransformation args, not the perturb args. I'd want to rename that ideally to be consistent with the perturb argument not sure if it will break backwards compatibility somewhere though. (I have no opinion on "amplitude" vs "distance", as long as it's consistent)

rees-c:
Hey @mkhorton ! I implemented alex's comments, added to the Structure.perturb unit test, changed the doc string wording for `perturb` and `PerturbStructureTransformation` to "if None, all displacements will be equal amplitude", and refactored `amplitude` and `min_amplitude` in `PerturbStructureTransformation` to `distance` and `min_distance` (to match the arg names in `Structure.perturb`). Does this look ok? I can undo the refactoring of `amplitude` and `min_amplitude` if we think that might break something

mkhorton:
As long as tests pass I’m ok with this change, it’s not a big difference if
we need to update something in atomate etc. Thanks again!

rees-c:
Thanks for your help @mkhorton!

rees-c:
Hi @mkhorton ! The checks have passed so I think we're ready to merge :)

PR #1545, opened 2019-08-03
[WIP] Mcsqs_caller

dongsenfo:
Returns a special quasirandom structure in the pymatgen structure format given an initial structure with partial occupancies. Not quite sure how to handle the files that come out of the mcsqs caller in the tests. In addition, there is a sleep time built into the code to allows the mcsqs code to build the files before converting into a pymatgen structure. 

Dependency on mcsqs from ATAT.


mkhorton:
Thanks for working on this @dongsenfo! 

Would it be possible to add a Transformation interface to it for convenience also? (along the same lines that enumlib_caller has EnumerateStructureTransformation, mcsqs_caller could have SQSTransformation)

Few other questions:

* what’s the stopping criterion here / how do we know when we have a decent SQS?

* I saw this takes clusters as an input, is there a separate function coming to generate the clusters themselves?

dongsenfo:
@mkhorton I'll write that up! 

The stopping criterion is currently just set by finding a perfect SQS or stopping after the max time, but I should probably edit that so it still spits out the best SQS it finds after the max time.

The clusters are generated with an initial call to mcsqs so you just need to specify the cluster cutoffs your're looking for

mkhorton:
Test failure is just a few PEP8 code style issues: https://travis-ci.org/materialsproject/pymatgen/builds/580508562#L1051

Running this through [Black](https://github.com/psf/black) or similar should fix most of them, but there are a few that might need manually changing like `E711 comparison to None should be 'if cond is None:'`

mkhorton:
Thanks again @dongsenfo !

dongsenfo:
@mkhorton Lol thanks for fixing my formatting!

PR #1547, opened 2019-08-07
Small change to conversion electrodes

jmmshn:
## Small change to conversion electrodes
- added kwarg that allows unstable entry to be used in the constructor. This allows us to get a conversion voltage estimate any new cathode materials, not just ones on the convex hull.
- changed the output of all_pairs and adj_pairs to lists not iterators


shyuep:
What does Master mean?

jmmshn:
ooops I didn't see that get populated by github somehow...

PR #1548, opened 2019-08-07
Bond valence sum unordered function fix

ardunn:
## Summary

New Structure `get_neighbors` returns namedtuples which breaks usage of `calculate_bv_sum_unordered`. It appears `calculate_bv_sum` was fixed previously though.

* Fix `test_calculate_bv_sum_unordered`
* Add tests for both bv_sum functions (previously missing, hence why this bug existed)
* Cleaned up the docstrings for these two functions to match the implementation

## TODO:
There is probably a way to combine these into one function for a little less boilerplate, but that's maybe something for a future PR. This is just a minimum changes necessary fix.


shyuep:
Thanks.

PR #1550, opened 2019-08-08
Fix multi-line :param: and :return: directives in docstrings

rkingsbury:
## Summary

* Fix docstrings with multi-line :param: and :return: directives not being rendered correctly. This was due to an (undocumented) quirk in sphinx discussed [here](https://stackoverflow.com/questions/32566466/multi-line-description-of-a-parameter-description-in-python-docstring). 

* I also fixed a few issues with the display of some numbered lists.

See screenshots of the incorrect rendering attached.

@shyuep - I've tested this with ```invoke make-docs```. Do you want me to commit the built docs as well, or just the code?

image

image

### DIFF
--- a/pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
+++ b/pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
@@ -158,7 +158,7 @@ class AbstractChemenvStrategy(MSONable, metaclass=abc.ABCMeta):
         """
         Abstract constructor for the all chemenv strategies.
         :param structure_environments: StructureEnvironments object containing all the information on the
-        coordination of the sites in a structure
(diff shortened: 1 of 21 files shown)


rkingsbury:
I commited MPCompatibility.yaml by mistake; it should not be in here.

mkhorton:
Thanks @rkingsbury!

Technically both modules mentioned in this PR are using RST docstrings where (by our own convention) we should be using Google-style docstrings. This easily happens since many IDEs default to RST, and I have some older code in pymatgen where I’ve done just that myself and have been meaning to update (e.g. see magnetism module also).

A better fix might be to use a tool like [pyment](https://github.com/dadadel/pyment/) to auto-convert to the Google style.

shyuep:
Thanks!

PR #1555, opened 2019-08-20
Enhancements to ComputedEntry and MPRester

rkingsbury:
## Summary

* By analogy to the `energy_per_atom` property, it would be useful for ComputedEntry objects to have an energy per formula unit property to unambiguously return the energy associated with a single formuala unit or reduced formula. The first commit adds this property

* `get_entries_in_chemsys` currently requires a list input e.g. ['Li','O'], but the `chemical_system` attribute of Composition returns a str separated by dashes e.g. 'Li-O'. To improve interoperability, it should be possible to pass the dash-separated string into `get_entries_in_chemsys` in addition to lists. The second commit makes this possible

### DIFF
--- a/pymatgen/ext/matproj.py
+++ b/pymatgen/ext/matproj.py
@@ -663,8 +663,9 @@ class MPRester:
         creating phase diagrams of entire chemical systems.
 
         Args:
-            elements ([str]): List of element symbols, e.g., ["Li", "Fe",
-                "O"].
+            elements (str or [str]): Chemical system string comprising element
+                symbols separated by dashes, e.g., "Li-Fe-O" or List of element 
+                symbols, e.g., ["Li", "Fe", "O"].
             compatible_only (bool): Whether to return only "compatible"
                 entries. Compatible entries are entries that have been
                 processed using the MaterialsProjectCompatibility class,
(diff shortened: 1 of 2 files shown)


pymatgen/ext/matproj.py
```diff
@@ -686,6 +687,9 @@ def get_entries_in_chemsys(self, elements, compatible_only=True,
             List of ComputedEntries.
         """
         entries = []
+        if isinstance(elements,str):
```
shyuep: There needs to be a space after commas.

shyuep:
Wait, I just noticed that there are no tests for the two methods. Pls submit a second PR with the relevant tests for the two methods.

shyuep:
I have temporarily reverted this PR until the tests are added. Thanks.

rkingsbury:
@shyuep test cases are ready (see #1561 ), but are failing because the PR is on top of the reverted PR #1555 (meaning the associated methods are no longer in pymatgen). Can you reopen this PR so I can add the necessary commits here?

shyuep:
@rkingsbury I can't seem to reopen the PR. I think the easiest way is for you to use `git checkout 35f222512ee90446b2516e68b99f163eca548548 <relevant file paths>` to get back the changes you made to entries and MPRester and commit them with the test PR.

PR #1556, opened 2019-08-21
Updates to Fragmentation, Graphs, and Q-Chem IO

samblau:
## Summary

* Molecule fragmentation substantially sped up by using more efficient data structures and leveraging the faster igraph library. Additional functionality added to allow a list of previously identified fragments to be accounted for.
* Some graph functionality modified or improved in order to facilitate fragmentation upgrades. 
*  Q-Chem IO improvements given changes in the latest version of the software. Parsing more errors, warnings, and properties. Default sets also updated given latest convergence tests. 

* Now using the igraph python graph library, although like OpenBabel, it is not technically a dependency. If it cannot be imported, networkx will be used instead.


samblau:
All updates and cleanup finished. Once all the tests run (and assuming they pass), this is ready for review and / or merging.

shyuep:
Travis test has failed because of OpenBabel dependency. You need to put a skipIf for openbabel. That is not a core dependency for pmg.

shyuep:
Thanks. Merged.

shyuep:
BTW, if igraph is supposed to be used, shouldn't it be in setup.py and requirements-optional.txt? I see it as a separate install in circleci/config.yml.

mkhorton:
Why is igraph being used(?)

samblau:
It’s a lot faster than networkx since it’s core routines are written in C++
(I think) instead of python. This improved performance can be critical when
doing large numbers of isomorphism checks such as when finding all unique
fragments from many molecules. However, igraph does not provide any
additional functionality beyond networkx, just performance, thus I did not
think it merited being an actual requirement.

mkhorton:
Thanks @samblau, sounds great! I like how igraph will gracefully fall back to networkx if it's not installed

PR #1558, opened 2019-08-21
uncertainty quantification in energy corrections

awvio:
## Summary

Added uncertainty quantification for energy corrections in compatibility module.

* additional yaml file that contains uncertainties for each correction
* get_correction() methods now return (correction, uncertainty)
* process_entry() adds correction uncertainty to entry.data dictionary under key: 'correction_uncertainty'

Plotly

## TODO (if any)

WIP (needs review)


shyuep:
I think I need a longer summary to know what this is for?

awvio:
Hi, sorry I'll elaborate.

The corrections in MPCompatibility.yaml were refit and now have standard error values associated with them (from imperfect fits as well as uncertainty in experimental data). I'll add the code that calculates the corrections and their errors soon. The Correction classes which previously given an entry returned a correction now also return the error on that correction. The Compatibility classes which previously processed an entry by calculating a total correction to the energy value now also include the error on that correction.

mkhorton:
Just to make some introductions, @awvio is working with us (@shyamd, Persson group) to help improve the MP corrections.

@awvio I wasn't familiar with your GitHub handle, so I was also confused for a moment :)

shyuep:
@mkhorton @kristinpersson @shyamd  I think this needs to be discussed. I do understand of course that corrections can be changed/improved. I have no problems with putting uncertainties in. But this is a rather extensive set of changes and it is possible that there are codes out there relying on the existing correction values. Some of the corrections are not small at all.

We should have a proper review of all the correction values, as well as a possible strategy to have versioned corrections based on MP database versions. This cannot be done just by someone editing the code.

mkhorton:
Yes, agreed this cannot be done solely in code. Perhaps worth discussing in next MP dissemination meeting; it'll need to be staged with the next database release, and have accompanying docs. Agreed also we need to store the MPCompatibility version (which could be pymatgen version even) in the database, and potentially in the resulting ComputedEntries also.

shyamd:
There are plans for a docs page and a paper discussing these corrections and their effects. For the most part these don't change the old values by much. The big difference is that these corrections can be reproduced in comparison to the old numbers, which I've never been able to recreate.

pymatgen/entries/MPCompatibility.yaml
```diff
@@ -2,55 +2,46 @@ Name: MP
...
     O:
-      Mn: -1.68085015096   #Fit to MnO, Mn3O4 and MnO2 (BURP:-1.687)
-      Fe: -2.733           #Fit to FeO and Fe2O3 (Fe3O4 probably wrong)
-      Co: -1.874           #Fit to CoO, Co3O4 (BURP:-1.751)
-      Cr: -2.013           #Fit to Cr2O3 (CrO3 missing) (BURP: -2.067)
-      Mo: -3.531           #Fit to MoO3 and MoO2 (BURP: -2.668)
-      W:  -4.351           #Fit to WO2 and WO3 (BURP: -2.762)
-      V:  -1.682           #Fit to V2O3 and V2O5 (VO2 fit is way off) (BURP: -1.764)
-      Ni: -2.164           #Based on burp version as of Feb 28 2011
+      V: -1.634860584846134
```
shyamd: We probably don't need so many sig figs. The corrections are definitely not good sub 1 meV, so let's round to that for now unless someone has a different opinion

pymatgen/entries/compatibility.py
```diff
@@ -648,3 +721,277 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
              GasCorrection(fp),
              AnionCorrection(fp, correct_peroxide=correct_peroxide),
              UCorrection(fp, MPRelaxSet, compat_type), AqueousCorrection(fp)])
+
+
+class CorrectionCalculator:
```
shyamd: Let's move this to a new file. Compatability is getting too dense.

pymatgen/entries/compatibility.py
```diff
@@ -648,3 +721,277 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
+    
+    species = ['oxide', 'peroxide', 'superoxide', 'F', 'Cl', 'Br', 'I', 'N', 'S', 'Se',\
+               'Si', 'Sb', 'Te', 'V', 'Cr', 'Mn', 'Fe', 'Co','Ni', 'Cu', 'Mo'] #species that we're fitting corrections for
+
+    def __init__(self, exp_json, comp_json):
```
shyamd: Add docstrings for all the methods. Let's also add type hints.

pymatgen/entries/compatibility.py
```diff
@@ -648,3 +721,277 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
+
+        popt, pcov = curve_fit(f, self.coeff_mat, self.diffs, p0=np.ones(21), sigma=sigma, absolute_sigma=True)
+        self.corrections = popt.tolist()
+        self.corrections_std_error = np.sqrt(np.diag(pcov)).tolist()
+        for i in range(len(self.species)):
+            self.corrections_dict[self.species[i]] = (self.corrections[i], self.corrections_std_error[i])
+        return self.corrections_dict
+
+
+    def graph_residual_error(self):
```
shyamd: Switch to plotly plots so that we can have them interactive and annotated for more information.

pymatgen/entries/compatibility.py
```diff
@@ -648,3 +721,277 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
+
+    def make_yaml(self, name='MP'):
+        """
+        create the _name_Compatibility.yaml that stores corrections as well as _name_CompatibilityErrors.yaml for correction errors
+        """
+
+        if len(self.corrections) == 0:
+            self.compute_corrections()
+
+        aqueous = OrderedDict()
```
shyamd: Add a comment mentioning these come from the old YAML file and are not yet auto-generated by this class.

shyuep:
The changes are large, by ~0.5 eV/atom in some cases.

MP has gone from 2012 to 2019 with many generations of db changes along the way. I do not doubt that it would be difficult to reproduce the old values (some of which were done using proper AFM structures). The original values were actually documented in the pymatgen publication.

The point remains this is an exercise that requires careful discussion and vetting. A doc page and publication would be good. But this will not be merged until all the pieces are in place.

pymatgen/entries/MPCompatibility.yaml
```diff
@@ -2,55 +2,46 @@ Name: MP
...
-  sulfide:     -0.66346     #Sulfide Correction fit to Li2S, Na2S, K2S, Rb2S, CsS, MgS, CaS, BaS, Al2S3.
-
-# Entropic Correction term for gases. 
-# For H2O, energy is corrected such that the experimental formation energy of water of -2.46 eV is reproduced.
-# H2 is corrected such that the experimental formation energy of water is reproduced. 
-AqueousCompoundEnergies:
-  O2:  -0.316731
-  N2:  -0.295729
-  F2:  -0.313025
+  sulfide: -0.6314172089245144
```
shyamd: I'm tempted to get rid of the Sulfide correction class and just treat it as another Anion correction since we didn't find a big difference for polysulfides.

shyuep:
Just to give an example, Fe3O4 in MP seems to be ferromagnetic. The actual ground state is ferrimagnetic. The energy diffs are not small. This is a point I have made since 2013.

The original corrections are fitted to binary oxides that were properly charge-ordered and magnetized.

pymatgen/entries/MPCompatibility.yaml
rkingsbury: I agree with rounding to 1 meV

pymatgen/entries/MPCompatibility.yaml
rkingsbury: I agree. In a forthcoming commit we've consolidated OxideCorrection and SulfideCorrection into one AnionCorrections key. This will make it clearer which Class is calling the corrections as well.

pymatgen/entries/MPCompatibility.yaml
awvio: Does this also apply to the correction errors? Rounding to 3 decimal places often results in only 1 sigfig (error = 0.00x), at most 2.

rkingsbury:
I would say you could carry 1 extra digit for the uncertainty (so, round uncertainty to 0.1 meV)



Ryan Kingsbury, Ph.D., P.E.

Postdoctoral Researcher
The Material Project
C: 713-851-7231
E: RKingsbury@lbl.gov

rkingsbury:
I can't figure out why my last commit about polysulfide caused the Travis Build to fail. The test that's failing (ReactionDiagramTest.test_formed_formula) doesn't touch any of the code I changed as far as I can tell.

mkhorton:
A good start may be to modify the ReactionDiagramTest.test_formed_formula -- it doesn't look like a very good test to me because it's not clear what's changed when it fails (I appreciate you haven't modified this file/it's not your test).

Currently it's:

`self.assertTrue(formula in formed_formula)`

Maybe:

`self.assertIn(formula, formed_formula`

rkingsbury:
@mkhorton OK, I can try that. The test runs successfully on my local machine and on AppVeyor though (without modifying the assertion statement). Could it be something about the Travis environment?

mkhorton:
Yeah, it looks like the kind of test that might have machine-specific precision issues but I'm not sure, I suspect the test could be improved.

shyuep:
Yes, that test has been failing randomly. Just ignore it for now. @mkhorton Can you talk to whoever is responsible and get him to fix it.

mkhorton:
That said the test has been around for two years so I'm not sure why it would start failing now. @paulgholin it looks like you wrote this test, do you have any ideas why this may be failing?

rkingsbury:
Were the linting tests recently added to the build checks? It looks like we're failing on mypy / pycodestyle / flake8 now, but that wasn't an issue a few weeks ago.

shyuep:
Yes, linting is now enforced on all PRs.

shyuep:
Note that the linting is fairly relaxed. Codestyle in general is quite basic. mypy enforces type checks on the few classes where we actually have type annotations (they are mostly core classes that have well-defined APIs). Flake8 adds a few minor checks. Pydocstyle again just ensures that you have proper docstrings for all public modules, classes and methods - no a huge ask. We do not enforce pylint at all, though that is the most comprehensive. Pylint is to encourage you to fix the code, but if you pass the other four linters, we will not insist on passing pylint.

awvio:
One of the errors from mypy is that it can't find the plotly module:
pymatgen/entries/corrections_calc.py:3: error: Cannot find module named 'plotly'
After referring to the mypy docs ([https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports](url)), this seems to be caused by the fact that plotly doesn't have type hints. Is it ok to silence this import error?

mkhorton:
Whether plotly has type hints or not isn't relevant. The error is present because plotly is not a pymatgen dependency -- you need to add it to the requirements.txt as a dependency otherwise it will not be installed during testing. Note that while in general we're reluctant to add new dependencies, I think plotly makes a lot of sense for us so this is fine.

shyuep:
After adding plot.ly, you might need to add them to the mypy exclusions in the setup.cfg file. There are plenty of examples there. See https://github.com/materialsproject/pymatgen/blob/master/setup.cfg

mkhorton:
At least that'd be my guess -- I was looking to see if the regular tests pass but nothing in `corrections_calc.py` seems to have tests yet, so we can take another look after they've been added.

pymatgen/entries/tests/test_correction_calculator.py
```diff
@@ -0,0 +1,139 @@
...
+		calc_path = os.path.join(self.test_dir, "calc_compounds_norm.gz")
+		
+		calculator = CorrectionCalculator(exp_path, calc_path)
+		corrs = calculator.compute_corrections(allow_polyanions = True, allow_large_errors = True, allow_unstable = True)
+
+		self.assertDictEqual(corrs, self.warnings_allowed_corrections)
+
+	def test_no_uncertainties(self):
+		"""
+		Test that corrections can be calculated with no corrections.
```
rkingsbury: Typo: should be 
"Test that corrections can be calculated with no **uncertainties**."

rkingsbury:
@awvio can you add plotly under "additional dependencies introduced" in the first comment on this PR?

mkhorton:
Plotly is getting added anyway from #1685

awvio:
I created a 2020 version of MITCompatibility.yaml even though the MIT corrections have not been changed - the only difference between MITCompatibility2020.yaml and MITCompatibility.yaml is CompoundEnergies (N2, F2, Cl2, H2, SO3, SO2) have been changed to GasCorrections. However I'm not sure if it's appropriate to create a 2020 version if the value of the corrections themselves have not changed, just the formatting/the way they are applied.

mkhorton:
If they give 1:1 equivalent outputs I would omit the 2020 label.

rkingsbury:
@shyuep we are ready to begin code review on this PR. I'll explain in bullet points what it does; feel free to request more info offline if that would be helpful. @awvio and @shyamd feel free to chime in as well.

- This PR updates the existing GGA/GGA+U mixing scheme to incorporate uncertainty quantification.
- We extend the list of anions to which corrections are fit, adding Br, I, Se, Si, Sb, and Te to the existing list of corrected elements.
- The updated scheme is implemented as `MaterialsProjectCompatibility2020` to avoid conflicts with the existing `MaterialsProjectCompatibility`, which we intend to deprecate in the future
- For clarity, in the new scheme we consolidate all corrections into the `CompositionCorrection` class, because all corrections are linear functions of the elemental composition of a compound. This class replaces the `AnionCorrection`, `GasCorrection`, and `UCorrection` in the previous scheme
- Corrections are fit to experimental formation energy data as before. However, in the -2020 scheme all corrections are solved simultaneously via a linear system of equations. This approach makes it possible to include more compounds (e.g., ternaries with multiple corrected atoms like KFeO2.)
- In addition to fitting energy corrections, the linear system allows us to quantify the uncertainty in the fits arising from 1) uncertainty in the experimental data and 2) the number of observations used to fit each correction. It also generates information about cross-correlation effects between energy corrections
- `ComputedEntry` objects that are corrected now have a `correction_uncertainty` item added to their `.data` dict 
- We include a new `CorrectionCalculator` class that illustrates how to generate the .yaml file containing the corrections. We hope this new class will make our correction procedure more transparent and easier for users to understand.

We look forward to your feedback. Hope you have a nice holiday break!

pymatgen/entries/compatibility.py
```diff
@@ -437,10 +632,21 @@ def process_entry(self, entry):
...
             return None
         entry.correction = sum(corrections.values())
+        tot_error = 0
+        for e in errors.values():
+            tot_error += e ** 2
+        tot_error = sqrt(tot_error)
+        if tot_error == 0:
+            # if there are no error values available for the corrections applied,
+            # set correction_uncertainty to not a number
+            entry.data["correction_uncertainty"] = float("NaN")
```
rkingsbury: @awvio can you change `float("NaN")` to `np.nan`? The float('NaN') code causes problems with `as_dict` / `from_dict`.

pymatgen/entries/compatibility.py
```diff
@@ -239,84 +274,275 @@ def get_correction(self, entry) -> float:
...
+                err = self.cpd_errors[rform] * comp.num_atoms
+
+                correction += ufloat(corr, err)
             else:
-                correction += cpdenergies[rform] * comp.num_atoms
+                corr = cpdenergies[rform] * comp.num_atoms
+                err = self.cpd_errors[rform] * comp.num_atoms
+
+                correction += ufloat(corr, err)
         if not rform == "H2O":
```
rkingsbury: @shyamd @montoyjh does this modification to the `AqueousCorrection` make sense to you in light of our discussions?

shyamd: Yeah, this seems to be the closest to what was there before. We'll fix the whole water detection issue later on as that is much harder.

rkingsbury:
@mkhorton is `IonTest.test_as_dict()` failing in master? I don't think any changes here should have affected it.

mkhorton:
Yup, looks like it broke alongside another bug fix (https://github.com/materialsproject/pymatgen/commit/dec6ecb6a9a80dbd4bfdfb924040cdfc7fe4ebb3)

pymatgen/analysis/tests/test_structure_analyzer.py
```diff
@@ -202,15 +202,15 @@ def test_oxide_type(self):
...
+        # NaS2 -> sulfide
         latt = Lattice.tetragonal(9.59650, 11.78850)
         species = ["Na"] * 2 + ["S"] * 2
         coords = [[0.00000, 0.00000, 0.17000],
                   [0.27600, 0.25000, 0.12500],
                   [0.03400, 0.25000, 0.29600],
                   [0.14700, 0.11600, 0.40000]]
         struct = Structure.from_spacegroup(122, latt, species, coords)
-        self.assertEqual(sulfide_type(struct), "polysulfide")
+        self.assertEqual(sulfide_type(struct), "sulfide")
```
mkhorton, line 213: Was this incorrect previously?

rkingsbury:
We removed the 'polysulfide' type because it was not being used by the old corrections and it had caused a bug that in some cases caused corrections not to be applied correctly.

mkhorton:
Ok, that's fair, but NaS2 _is_ a polysulfide (i.e. has an S-S bond) so the new labelling is incorrect. Is there a better way to handle this?

Are there any other breaking changes made in this PR?

rkingsbury:
See https://github.com/materialsproject/pymatgen/pull/1587. I think I accidentally opened a separate PR for this issue at one point and then integrated it back in to the MP2020 one.

mkhorton:
I see it. Just to clarify, I have no objection to dropping polysulfide in terms of corrections, my issue is that `sulfide_type` needs to be accurate.

rkingsbury:
> I see it. Just to clarify, I have no objection to dropping polysulfide in terms of corrections, my issue is that `sulfide_type` needs to be accurate.

How about I revert the changes to `sulfide_type` so that it still can return 'polysulfide', but then treat polysulfide and sulfide the same within the`MaterialsProject2020Compatibility`?

mkhorton:
Yep, that would be great. Thanks.

rkingsbury:
Also, do you want me to squash commits?

mkhorton:
No need, I can squash on merge if necessary. Commit history is long but can be useful.

rkingsbury:
> Ok, that's fair, but NaS2 _is_ a polysulfide (i.e. has an S-S bond) so the new labelling is incorrect. Is there a better way to handle this?
> 
> Are there any other breaking changes made in this PR?

No other breaking changes that I'm aware of

rkingsbury:
@mkhorton should we be placing `DeprecationWarning` on the MIT Compatibility classes? We had the warnings in there, but I think they are left over from when we were modifying the underlying `Correction` classes that they depend on. We later moved most of the new stuff to new classes such that the behavior of `MITCompatibility` should not change at all.  There are some updates to the `AqueousCorrection` class that are utilized only by `MITAqueousCompatibility`, however.

rkingsbury:
> @mkhorton should we be placing `DeprecationWarning` on the MIT Compatibility classes? We had the warnings in there, but I think they are left over from when we were modifying the underlying `Correction` classes that they depend on. We later moved most of the new stuff to new classes such that the behavior of `MITCompatibility` should not change at all. There are some updates to the `AqueousCorrection` class that are utilized only by `MITAqueousCompatibility`, however.

I've gone ahead and removed the deprecation warnings from the MIT classes; I think the only one we really need to be warning users about is the transition from MP to MP2020. Let me know if you think we should keep them.

pymatgen/analysis/reaction_calculator.py
```diff
@@ -15,10 +15,10 @@
 from pymatgen.entries.computed_entries import ComputedEntry
 from monty.json import MontyDecoder
 from monty.fractions import gcd_float
+from uncertainties import ufloat
```
mkhorton, line 18: I like this library, but is it necessary here (considering we would have to add as a core requirement)? It looks like the functionality of `uncertainties` is not used in this class beyond book-keeping, which could be done by a named_tuple or otherwise. I'd err towards keeping it but just wanted to clarify this point.

rkingsbury: Beyond book-keeping, we use `uncertainties` to calculate the propagated uncertainty when adding/subtracting uncertain quantities. It would be possible to code this ourselves (we had it that way in a previous iteration), but I feel its more robust to let the dedicated library handle it. Also, in some of our previous discussions I thought you were in favor of adding `uncertainties` as a core requirement, because it will facilitate some other functionality we have in the pipeline? In any event, I'd prefer to keep unless you strongly oppose. I think the package is pretty lightweight.

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,433 @@
...
+
+from typing import List, Dict, Tuple
+from monty.serialization import loadfn
+
+from pymatgen import Element, Composition
+from pymatgen.analysis.reaction_calculator import ComputedReaction
+from pymatgen.analysis.structure_analyzer import sulfide_type
+
+
+def func(x, *m):
```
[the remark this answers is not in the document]
rkingsbury: Done

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,433 @@
...
+        allow_unstable: bool = False,
+    ) -> dict:
+
+        """
+        Computes the corrections and fills in correction, corrections_std_error, and corrections_dict.
+
+        Args:
+            allow_polyanions: optional variable, boolean controlling whether compounds with problematic polyanions
+                will be included in the fit
+            allow_large_errors: optional variable, boolean controlling whether compounds with large experimental
```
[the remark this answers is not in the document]
rkingsbury: You're correct that the definitions are hard-coded in the .json file. That is, they are pulled from a 'warnings' key associated with each `ComputedEntry` in the .json file. For this work I believe @awvio defined 'large error' as a relative uncertainty of 10% or greater.

It is true that we could compute the relative uncertainty (indicated by the 'large errors' flag) on the fly and therefore specify a cutoff value in `compute_corrections`, whereas for the other two warning categories - polyanions and e above hull, it's more robust to specify the warnings directly. I think we were trying to maintain consistency with the other two and hence placed the large errors warning in the .json as well.

These warnings are applied to the experimental, not the Computed, compound entries. Hence, it would be difficult to determine `allow_unstable` in code, because that would require matching each experimental entry with a `ComputedEntry` having an `e_above_hull` attribute, which we can of course assume for MP entries, but at the risk of making the calculator less general.

@shyamd @awvio, what are your thoughts about making the warnings / excluded compound filters more automatic or fine-grained?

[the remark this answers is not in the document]
shyamd: I would argue that these options should actually go into the `__init__` and the `calculate` function takes the two data sets as args.

It would also be good to encapsulate the file loading by making the main calculate take two lists of entries and having helper function that loads from GZip files/

shyuep:
Pls make sure you pull in the latest master branch changes first. There are some additional code checks that we are imposing on new code.

pymatgen/entries/corrections_calc.py
```diff
@@ -67,24 +67,22 @@ class CorrectionCalculator:
...
         Args:
-            exp_gz: name of gzip file that contains experimental data
-                    data in gzip file should be a list of dictionary objects with the following keys/values:
-                    {"formula": chemical formula, "exp energy": formation energy in eV/formula unit,
-                    "uncertainty": uncertainty in formation energy, "warnings": {"polyanion": str describing
-                    polyanions in compound or None, "large_uncertainty": str describing if compound has a large
-                    uncertainty value or None, "unstable": str describing if compound has high e above hull or None}}
-            comp_gz: name of gzip file that contains computed entries
-                    data in gzip file should be a dictionary of {chemical formula: ComputedEntry}
+            max_error: maximum tolerable relative uncertainty in experimental energy
```
rkingsbury, line 81: @awvio can make this docstring a little more explicit, e.g. "maximum tolerable relative uncertainty in experimental energy. Entries with a relative uncertainty greater than this value will be excluded from the fit."?

pymatgen/entries/corrections_calc.py
```diff
@@ -67,24 +67,22 @@ class CorrectionCalculator:
...
+            allow_unstable: whether unstable entries are to be included in the fit
```
rkingsbury, line 82: Clarify "unstable" - "whether unstable entries (i.e., e_above_hull > 0) are included in the fit"

pymatgen/entries/corrections_calc.py
```diff
@@ -67,24 +67,22 @@ class CorrectionCalculator:
...
+            allow_polyanions: whether entries with problematic are to be included in the fit
```
rkingsbury, line 83: Let's avoid saying "problematic" and instead just say "entries containing polyanions (e.g., NO3, SO4, etc.)"

rkingsbury: Actually, examining the code below, can you just include the exact list of polyanions that the code searches for? "['SO4', 'CO3', 'NO3', 'OCl3', 'SiO4', 'SeO3', 'TiO3', 'TiO4']"

rkingsbury: and perhaps a short statement to the effect of "compounds with these polyanions have been observed to contain additional sources of error that may negatively influence the quality of the fitted corrections."

pymatgen/entries/tests/test_correction_calculator.py
```diff
@@ -57,28 +58,28 @@ def setUp(self):
...
+            'V': (-1.66, 0.0765),
+            'Cr': (-1.85, 0.0684),
+            'Mn': (-1.589, 0.0695),
+            'Fe': (-2.165, 0.061),
+            'Co': (-1.563, 0.0777),
+            'Ni': (-2.516, 0.0969),
+            'W': (-4.574, 0.1209),
+            'Mo': (-3.05, 0.1094),
+            'H': (-0.137, 0.0307)
         }
```
rkingsbury, line 83: Why did the values change here? Was this just a case of the values in the test file being slightly out of date? Just want to confirm that the main .yaml file was not affected.

awvio: The values changed since the previous version actually had a mistake. This particular test case tests for whether the CorrectionCalculator works if there is no experimental uncertainty data available (in which case it should compute corrections with all compounds weighted equally). Previously, it still filtered out compounds with large uncertainties since filtering was determined by the warnings dict and not the uncertainty data, but I realized that it makes more sense if it doesn't filter by uncertainty.

rkingsbury:
@mkhorton we’re having trouble understanding the output from the new pylint tests. Everything has been linted and formatted with pycodestyle, flake8, and black, and many of the failures seem more like style suggestions than real problems. Are we expected to pass the pylint test now, and if so, can you offer any guidance on how to address all the errors in a relatively quick way?

shyuep:
Please try to address the pylint problems. If you have a message you are uncertain about, post it here and I can give guidance.

rkingsbury:
Ok @shyuep, it seems like the majority of our failures fall into 3 major categories, and I'd like to ask your guidance on each.

`consider-using-set-comprehension` is raised whenever we have a `set( [x for x in iterable ] )` construction. I'll acknowledge that a direct set comprehension would be more concise, but this feels like a style preference rather than a code quality issue. There are a lot of these constructions throughout parts of pymatgen that we haven't modified in this PR. Can we consider suppressing this warning?

`no-else-return` is raised by constructions like this one from `oxide_type`, in which the `else` is arguably not necessary (but I believe clearer to read):
```
if return_nbonds:
     return ox_obj.oxide_type, ox_obj.nbonds
else:
     return ox_obj.oxide_type
```

Again, this feels like a style preference and not a code quality problem. As with the first case, there are many instances of this in parts of the code we have not changed. Can we consider suppressing this warning?

`bad-continuation` errors are raised because `pylint` and `black` disagree on the proper way to format certain indentations (see https://github.com/psf/black/issues/48 and https://github.com/PyCQA/pylint/issues/289). Many of us are finding `black` to be a useful and time-saving way to promote code quality, so can we consider suppressing this warning at least until `black` and `pylint` implement consistent style? I feel it will be problematic if we start using `black` on other codes (see https://github.com/materialsproject/custodian/pull/149) but have to use a different style in pymatgen.

Thanks!

mkhorton:
For `set( [x for x in iterable ] )`, `{x for x in iterable}` is definitely preferred, there's no point constructing an object twice. I would not recommend suppressing this.

The pylint failures are the result of files touched in this PR, even if you haven't directly modified those specific aspects of code.

`no-else-return` is more of a style thing, I'm more agnostic about, but would just fix it. I'm ok with suppressing the `bad continuation` if `pylint` and `black` disagree, I don't think there's explicit guidance on this.

shyuep:
For indent, I can consider disabling it. But not for the rest. When it comes down to it, everything is a style preference. But I'd rather impose a consistent style.

1. I don't see why a `set([x....])` is cleaner than simply a `{x...}`. 
2. Same with the if else. The else is definitely not necessary.

I would say the above are not the most serious. pylint actually caught a lot of statements like these:

```
if x == 0:
    return True
else:
    return False
```

This is just silly. We can replace this with just return x == 0.

shyuep:
Re the indent thing, for now, just add `# pylint: disable=C0330` at the top of your file to disable this message.

Note that other error messages like for example something is unsubscriptable - this usually is actually a serious bug in the code.

mkhorton:
Just a tip regarding the order of import statements, if you're using PyCharm, "Code > Optimize Imports" is useful to fix this.

pymatgen/entries/MP2020CompatibilityUncertainties.yaml
```diff
@@ -0,0 +1,37 @@
+Name: MP2020
```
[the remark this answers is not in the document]
awvio: Yes it's for backwards compatibility. All Correction classes used by MaterialsProject2020Compatibility (with the exception of PotcarCorrection) can take either a single .yaml files containing corrections, or two .yaml files containing corrections and uncertainties.

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,504 @@
...
+    Attributes:
+        species: list of species that corrections are being calculated for
+        exp_compounds: list of dictionaries which each contain a compound's formula and experimental data
+        calc_compounds: dictionary of ComputedEntry objects
+        corrections: list of corrections in same order as species list
+        corrections_std_error: list of the variances of the corrections in same order as species list
+        corrections_dict: dictionary of format {'species': (value, uncertainty)} for easier correction lookup
+    """
+
+    species = [
```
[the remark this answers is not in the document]
awvio: Since we went through a selection process to determine which species to calculate corrections for, I expected most common use case would be to calculate corrections for that same set of species and hard-coded it in. Do we want to give the user the option to specify the species themselves, perhaps upon initialization?

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,504 @@
+"""
+This module calculates corrections for the species listed below, fitted to the experimental and computed
+entries given to the CorrectionCalculator constructor.
+"""
+# pylint: disable=C0330
+
+import warnings as w
```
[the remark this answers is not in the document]
awvio: That was a result of the previous experimental entries having a "warnings" dict and me wanting to avoid confusion between the two. Fixed!

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,504 @@
...
+                    "SO4",
+                    "CO3",
+                    "NO3",
+                    "OCl3",
+                    "SiO4",
+                    "SeO3",
+                    "TiO3",
+                    "TiO4",
+                ]:
+                    if anion in name or anion in cmpd_info["formula"]:
```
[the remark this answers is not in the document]
awvio: `name` is `Composition.reduced_formula`, which uses the helper function `reduce_formula` in the composition module. The helper function appears to check for polyanions and displays them accordingly in the formula it returns, so I think it should catch any polyanions that are properly formatted in the list (in which case I guess it technically doesn't need the second check after the `or` at all).

shyuep:
Note - I modified structure_analyzer to clean up the linting errors. Including fixing an obvious bug that was not covered by unittests. E.g.,

```python
-        t = list(zip(*voronoi_ensemble))
+        t = zip(*voronoi_ensemble)
         labels = t[0]
         val = list(t[1])
```

The original code was obvious impossible to run without error.

Please pull in the changes.

pymatgen/entries/tests/test_compatibility.py
```diff
@@ -97,62 +125,137 @@ def test_overlapping_adjustments():
...
-                        'is_hubbard': False,
-                        'pseudo_potential': {'functional': 'PBE', 'labels': ['H'], 'pot_type': 'paw'},
-                        'hubbards': {},
-                        'potcar_symbols': ['PBE H'],
-                        'oxide_type': 'None'})
+            "H8",
+            -27.1,
+            0.0,
+            parameters={
+                "run_type": "None or LDA+rVV10",
```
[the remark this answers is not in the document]
rkingsbury: @mkhorton this was an artifact of what I think was a parsing bug? in the previous database release. I see that there are actually two `self.entry4` here; will remove the redundant one.

pymatgen/analysis/structure_analyzer.py
```diff
@@ -603,7 +603,7 @@ def oxide_type(structure, relative_cutoff=1.1, return_nbonds=False):
 
 def sulfide_type(structure):
     """
-    Determines if a structure is a sulfide/polysulfide
+    Determines if a structure is a sulfide/polysulfide/sulfate
```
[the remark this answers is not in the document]
rkingsbury: If the average electronegativity of the coordinating elements around S are greater than the electronegativity of S, then yes, "sulfate" is returned (see line 642)

rkingsbury:
> Note - I modified structure_analyzer to clean up the linting errors. Including fixing an obvious bug that was not covered by unittests. E.g.,
> 
> ```python
> +        t = list(zip(*voronoi_ensemble))
> -        t = zip(*voronoi_ensemble)
>          labels = t[0]
>          val = list(t[1])
> ```
> 
> The original code was obviously impossible to run without error. pylint is valuable this way. Another egregious example of super bad code:
> 
> ```python
> +   return oxide_type(structure, relative_cutoff) == "peroxide"
> -    ox_type = oxide_type(structure, relative_cutoff)
> -    if ox_type == "peroxide":
> -        return True
> -    else:
> -        return False
> ```
> 
> Please pull in the changes.

Thanks @shyuep !

pymatgen/entries/compatibility.py
```diff
@@ -242,84 +287,291 @@ def get_correction(self, entry) -> float:
...
+                    * comp["O"]
+                )
+
+        for anion in ["Br", "I", "Se", "Si", "Sb", "Te", "H", "N", "F", "Cl"]:
+            if Element(anion) in comp and anion in self.comp_correction:
+                apply_correction = False
+                # only apply anion corrections if the element is an anion
+                # first check for a pre-populated oxidation states key
+                # the key is expected to comprise a dict corresponding to the first element output by
+                # Composition.oxi_state_guesses(), e.g. {'Al': 3.0, 'S': 2.0, 'O': -2.0} for 'Al2SO4'
```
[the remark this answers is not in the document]
rkingsbury: For consistency with how the old `MPCompatibility` works, we've built this in a way that's strictly composition-based. I recall having some conversations about a more structure-based scheme, but we decided that it would be too complicated / not worth the effort at this point. We could add an extra layer of logic to look for oxidation states when we have a  `ComputedStructureEntry` if you think that would be beneficial. That's also an enhancement we can add later as we integrate more of the oxi state information into the builders.

pymatgen/entries/compatibility.py
```diff
@@ -689,10 +989,54 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
+    """
+    This class implements the GGA/GGA+U mixing scheme, which allows mixing of
+    entries. Note that this should only be used for VASP calculations using the
+    MaterialsProject parameters (see pymatgen.io.vaspio_set.MPVaspInputSet).
+    Using this compatibility scheme on runs with different parameters is not
+    valid.
+    """
+
+    def __init__(
+        self, compat_type="Advanced", correct_peroxide=True, check_potcar_hash=False
```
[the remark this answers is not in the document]
rkingsbury: I'd advise against, considering how much confusion continues to be caused by the hashing checks we added to `InputSet`. We could certainly turn it on internally for purposes of building our database though. I haven't given this a lot of thought; we default to `False` because that's what the old MPCompatibility does.

pymatgen/entries/corrections_calc.py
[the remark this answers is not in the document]
rkingsbury: Yes @awvio I think that would be good!

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,504 @@
...
+        "Ni",
+        "W",
+        "Mo",
+        "H",
+    ]  # species that we're fitting corrections for
+
+    def __init__(
+        self,
+        max_error: float = 0.1,
+        allow_unstable: bool = False,
```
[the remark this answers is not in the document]
rkingsbury: @awvio can you make `allow_unstable` a float that specifies how high `e_above_hull` can be for compounds to be included?

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,504 @@
...
+        allow_polyanions: bool = False,
```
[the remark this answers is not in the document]
rkingsbury: @mkhorton what if we change `allow_polyanions`  to 'exclude_polyanions', with a default value of either `None` or the specific list of polyanions we excluded in our recent work? @awvio, any thoughts one way or another?

pymatgen/entries/corrections_calc.py
```diff
@@ -0,0 +1,504 @@
...
+        Args:
+            exp_gz: name of gzip file that contains experimental data
+                    data in gzip file should be a list of dictionary objects with the following keys/values:
+                    {"formula": chemical formula, "exp energy": formation energy in eV/formula unit,
+                    "uncertainty": uncertainty in formation energy}
+            comp_gz: name of gzip file that contains computed entries
+                    data in gzip file should be a dictionary of {chemical formula: ComputedEntry}
+        """
+
+        exp_entries = loadfn(exp_gz)
```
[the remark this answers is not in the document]
rkingsbury: @mkhorton we can include the structures for archival purposes if you want. The reason we didn't is partly related to the comment above that our scheme is inherently composition-based, so we weren't thinking of distributing structures originally.

pymatgen/analysis/structure_analyzer.py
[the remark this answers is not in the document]
mkhorton: That's for the inner function though, line 652 suggests None is then returned?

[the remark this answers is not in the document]
rkingsbury: Oh weird, hadn't noticed that. Can you think of any reason we shouldn't return 'sulfate' there?

pymatgen/entries/corrections_calc.py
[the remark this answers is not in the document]
awvio: Will there ever be the case that the user wants every compound to be used, regardless of its `e_above_hull`? In this case, would they just have to set the float to some arbitrarily large value?

pymatgen/entries/corrections_calc.py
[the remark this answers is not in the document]
awvio: I think a list of polyanions to exclude in place of `allow_polyanions` sounds good. If we all think that the list of polyanions that we exclude in the paper makes sense, we can set that as the default so anyone who tries to reproduce the corrections in the paper would be able to do so easily.

pymatgen/entries/corrections_calc.py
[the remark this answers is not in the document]
rkingsbury: I'd say make the type `Optional[float, bool]` so that a user can pass `True` to indicate "use all compounds" or provide a number to limit instability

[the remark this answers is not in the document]
awvio: Should it be `Optional[float, bool]`, which I believe is for defaulting to `None`, or `Union[float, bool]`, which we could default to 0.1 (which is the current energy above hull limit)?

pymatgen/entries/compatibility.py
```diff
@@ -175,22 +206,27 @@ def __init__(self, config_file):
...
+    def get_correction(self, entry) -> ufloat:
         """
         :param entry: A ComputedEntry/ComputedStructureEntry
         :return: Correction.
         """
         comp = entry.composition
 
+        correction = ufloat(0.0, 0.0)
+
+        # set error to 0 because old MPCompatibility doesn't have errors
```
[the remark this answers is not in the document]
awvio: I think originally I intended to initialize uncertainty for a given correction as 0 if there isn't uncertainty data for that correction, and then once all the corrections were added, if the total correction uncertainty is 0 (indicating that none of the corrections applied had uncertainty data), set the total correction uncertainty to np.nan. This was before the change to apply corrections through the EnergyAdjustment class, so I'm not sure if that's still how we want to handle it. @rkingsbury?

pymatgen/analysis/reaction_calculator.py
mkhorton: Just to respond to this, yes I'm in favor of `uncertainties` via being in favor of `pint`, so I don't mind it getting added here, but mostly with a view of also addressing the unit question down the road. If this were the only place we'd use it I'd be more hesitant.

mkhorton:
I think it's time to merge, thank you @awvio and @rkingsbury for the very extensive work done here.

To re-iterate changes, the major changes in this PR are a new `CorrectionCalculator` class to reproducibly generate formation enthalpy corrections for DFT energies from experimental data, including the addition of uncertainties to these corrections from the corresponding experimental uncertainties. This comes with a new MP2020 scheme containing an example fit along with numerous smaller improvements.

For any external pymatgen users reading this, note that the specific data and specific values in this MP2020 scheme are subject to change until we reach publication status, and @rkingsbury is planning further iterative improvements. This PR contains the machinery of the corrections class, and it is this machinery which has now been finalized.

The current MP correction scheme remains usable and unchanged by this PR.

Happy Friday everyone!

PR #1560, opened 2019-08-22
Minor improvements to plot_fermi_surface

ardunn:
## Summary

1. `plot_fermi_surface` did not easily work with multiple transparency factors and colors at the same time. This PR makes plotting multiple energy levels with plot_fermi_surface easier by accepting a list of colors and transparencies, one for each desired surface. The core plotting code is identical to previous versions. No features were added. 

2. Some formatting and removal of the most obvious anti-patterns (imports in functions, mutable default kwargs). Docs are changed to be a bit clearer and more PEP8 compliant.

3. Added mayavi as an optional requirement (it was used but not optionally required before)

Technically `mayavi`, but this was already implicitly "required" by the module and would just throw a boltztrap error every time if mayavi wasn't installed.

### DIFF
--- a/pymatgen/electronic_structure/plotter.py
+++ b/pymatgen/electronic_structure/plotter.py
@@ -3740,7 +3740,8 @@ def plot_fermi_surface(data, structure, cbm, energy_levels=None,
             run mlab.show().
         
     Returns:
-        a Mayavi figure and a mlab module to control the plot.
+        ((mayavi.mlab.figure, mayavi.mlab)): The mlab plotter and an interactive
+            figure to control the plot.
 
     Note: Experimental. 
           Please, double check the surface shown by using some 
--- a/pymatgen/electronic_structure/plotter.py
+++ b/pymatgen/electronic_structure/plotter.py
@@ -9,8 +9,13 @@ import warnings
 from collections import OrderedDict
 
 import numpy as np
-
 from monty.json import jsanitize
+from monty.dev import requires
+
+try:
+    from mayavi import mlab
+except ImportError:
+    mlab = None
 
 from pymatgen.core.periodic_table import Element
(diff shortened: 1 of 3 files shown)


mkhorton:
Yikes, a lot of that code looks quite old. Thanks for fixing.

Does this work with BoltzTraP 2 or is this 1 only? I seem to remember seeing a function to plot Fermi surfaces in there too. Also, is it possible to plot using any uniform band structure, or does it require a special BoltzTraP-friendly input?

ardunn:
@mkhorton yeah the code is all terribly ancient.

My understanding is that there are two separate versions. plot_fermi_surface in boltztrap.py (for BoltzTrap 1) and BztPlotter in boltztrap2.py which does not look fully implemented.

Luckily. if I am not mistaken @utf is working on (fixing? improving?) the Boltztrap2 version which will likely supercede everything current boltztrap fermi surface does.

For this Boltztrap 1 version at least, it is not possible to plot with any uniform BS. You need to do a separate Boltztrap 'FERMI' mode run beforehand. I assume this is because the fermi surf of your typical uniform BS will be too low resolution, hence why boltztrap is needed to interpolate? 

It would be great if there was instead a simple `plot_fermi_surface` method of `BSPlotter`...perhaps some more physics-minded people could weigh in on whether that would or would not be a good idea.

ardunn:
tests failing bc mayavi can't install on travis, I guess I'll remove the requirement...

ardunn:
tests now failing for unrelated reasons, as far as I can tell...`ReactionDiagramTest.test_formed_formula` is failing

mkhorton:
Seems to be a merge conflict on this now?

mkhorton:
🤦🏻‍♂️

PR #1561, opened 2019-08-22
[WIP] Add test cases for enhancements in PR #1555

rkingsbury:
Add test cases for enhancements in PR #1555


shyuep:
Tests failed because of a bug in the get_reduced_formula. Pls fix. (And also, this is why unittests are needed).

shyuep:
I decided to go with an alternative implementation. Instead of having energy_per_formula_unit, which is not very well-defined (since forumula unit can mean different things to different people). I added a normalize method to ComputedEntry. This has been pushed.

PR #1562, opened 2019-08-22
Add options and docs to periodic_table_heatmap

rkingsbury:
## Summary

Add options to adjust the range of the colorbar and the size of the label in `periodic_table_heatmap`. The ability to adjust the scale is useful if you want to generate multiple plots for comparison to one another, and make sure they all use the same scale.

Also corrected some inaccuracies in the docstring.


mkhorton:
Thanks @rkingsbury !

PR #1563, opened 2019-08-23
Remove charge from composition

kmu:
## Summary

- The method that removes charges from the composition.
- It is useful to compare the composition of the oxidized structure and the composition derived from other sources such as ICSD metadata.


pymatgen/core/composition.py
```diff
@@ -766,6 +767,15 @@ def add_charges_from_oxi_state_guesses(self,
...
         return Composition(collections.Counter(species))
 
+    def remove_charges(self):
+        d = collections.Counter()
+
+        for e, f in self.items():
+            e = re.findall(r"[A-z]+", str(e))[0]
+            d[str(e)] += f
+
+        return(Composition(d))
```
mkhorton: should be `return Composition(d)`, also missing docstring

mkhorton:
Thanks @kmu, this looks useful!

mkhorton:
Thanks @kmu!

PR #1566, opened 2019-08-26
CombinedData class and hybrid style LAMMPS data file i/o

htz1992213:
## Summary

* The CombinedData class is added to easily merge LammpsData objects or LAMMPS data files.
* The original LammpsData class is modified to handle hybrid style bond/angle/dihedral/improper coefficients merging.
* Fix: No. of significant figures to output for charges is changed to a more conservative number (4).

## TODO 

* To support i/o of the comments of coeff styles on the 'keyword' lines of LAMMPS data file


mkhorton:
Thanks @htz1992213, this looks great + very useful!

There are a few places where `for i in range(len(...)):` loops are used where `for i, ... enumerate(...)` would be more appropriate (and would remove the need for some of the "temp" variables). If you could make that change we can go ahead and merge.

htz1992213:
Hey @mkhorton! I have made commits changing the 1) `range(len())` for loop, 2) temp variable to meaningful names, 3) and from_files function. 
Thank you for reviewing my code :)! Feeling good to get my feet wet in contributing to pymatgen!

mkhorton:
No problem, and thanks for your contribution(!)

PR #1567, opened 2019-08-28
[WIP] Added SLME and spillage analysis

knc6:
## Summary 

- Analysis of solar-cell efficiency using spectroscopy limited maximum efficiency (SLME)
- Analysis of topological spillage for non-trivial materials


shyuep:
Hi, thanks for the contribution! But the test is failing due to both actual test errors as well as codestyle issues. Can you fix those? For codestyle, you can refer to http://pymatgen.org/contributing coding guidelines for how to deal with codestyle easily.

knc6:
Ok sure, I'll look into it.

mkhorton:
Thank for contributing these @knc6 !

To be more specific for some of the issues @shyuep  mentioned, we have missing docstrings for some functions (e.g. in topological), some extraneous print statements (please use a logger if you want to print debug output), the `if __name__ == "__main__"` block is fairly harmless but shouldn't be there either, and some capitalization is not PEP8 compliant (classes should be capitalized, methods should have underscores not camelCase).

On a separate note, we already have a `Wavecar` class, is there anything in the `vaspwfc` class that would make sense to add to the `Wavecar` class instead?

knc6:
Hi @mkhorton Thank you for the notes, I'll get back to this in a couple of days.

knc6:
It seems like the current pymatgen's Wavecar module is not implemented for the spin-orbit case, which is important for spillage calculation. Could you please confirm? @mkhorton

mkhorton:
@mturiansky contributed the Wavecar functionality, I think he would know best whether it correctly handles the spin-orbit case.

mturiansky:
You're correct; it does not handle spin-orbit WAVECARs nor does it handle gamma WAVECARs. It shouldn't be too difficult to modify the current code to also handle these cases; I was planning to do it once I can find some free time. #1585 references this and someone had started implementing it for the gamma case, but it was never finished.

mkhorton:
Hi @knc6, this is very timely but @mturiansky just addressed loading of SOC WAVECARs with our existing `Wavecar` class via #1861, so it should now be possible to simplify this PR by removing `vaspwfc` and get this merged in if you were interested. Enjoyed your Optimade talk today!

knc6:
Hi @mkhorton thanks for attending the talk. I'll look into the new Wavecar class. Or another option would be to make jarvis-tools an dependency/optional-dependency and just import the functionalities from there. Also, I would like to add a jarvis adaptor to pymatgen. What do you think?

mkhorton:
We have several adaptors already inside pymatgen.io and a Jarvis adaptor would be very welcome. Regarding dependencies, we try not to add dependencies unless absolutely necessary, it's important to keep this minimal. An optional dependency for the adaptor would make sense (e.g. see the [ASE adaptor](https://github.com/materialsproject/pymatgen/blob/e42ae2911757c080a2b18d8b53bacff94630fd78/pymatgen/io/ase.py#L19)), I don't think a dependency should be necessary for the spillage code however.

mturiansky:
I'm happy to help integrate with the Wavecar class. Let me know if you have any questions/need any help.

Out of curiosity: why does jarvis-tools reimplement many of the things pymatgen has instead of just making pymatgen a dependency?

I would point out that the Wavecar class here has a few extra features like constructing a PARCHG or converting to an UNK file. One thing that the jarvis-tools implementation has that is nice is that the bands are only read when called for, so the whole WAVECAR doesn't need to be stored in memory. If this is necessary, I can make some changes to make this implementation more memory efficient. There are a few places where I use lists instead of just defining an empty np.array. Second, I could add a "bands" tag to the reader so only certain bands are read in. This just adds a little extra work because then there needs to be a mapping from the input band coefficient to the array that is stored in memory; this can be accomplished with the property decorator, but as I said, just a little extra work.

knc6:
@mkhorton , I have added the SOC-Spillage module with pymatgen Wavecar. The unittests are passing on my local computer. I think some other pymatgen modules are failing. Please take a look.
New modules added: pymatgen/analysis/solar/slme and pymatgen/analysis/topological/spillage.

mkhorton:
Many thanks Kamal, I'll take a look. The first test that seems to be failing right now is the linter, which is usually some minor code style issues: https://github.com/materialsproject/pymatgen/runs/786553347#step:5:24

mkhorton:
For context, the linters that run check for code style, documentation coverage, and also some type checking, so there are typically a few iterations of getting linter tests to pass before the full test suite is run. These can also be run offline if desired (via pycodestyle, flake8, pylint, pydocstyle etc.)

knc6:
@mkhorton Thank you for the suggestions. In addition to the code style changes, fixed a small bug in core/ion.

pymatgen/analysis/solar/slme.py
```diff
@@ -0,0 +1,306 @@
...
+        This is for dealing with broadening applied in DFT absorbance
+        calculations. Probably not desired for experimental data.
+
+    3) We can calculate at different temperatures if we want to, but 25 C /
+        293.15 K is the standard temperature assumed if not specified
+
+    4) If absorbance is in cm^-1, multiply values by 100 to match units
+        assumed in code
+
+    Args:
```
mkhorton, line 160: Would it be possible to add descriptions for these args?

mkhorton:
Fantastic, thanks @knc6, and thanks for fixing the ion bug too. The only issue stopping this getting merged in I think is the missing descriptions for the args in `slme`.

There's another PEP8 issue but I can correct if necessary, this is `SOC_Spillage` should be `SOCSpillage` [(classnames don't have underscores)](https://www.python.org/dev/peps/pep-0008/#class-names) and similarly with `from pymatgen.io.vasp.outputs import Wavecar as pmg_wav`. {

knc6:
Alright @mkhorton , I have made the suggested changes. Please let me know if any additional modifications are needed.

mkhorton:
Thanks for addressing all the comments @knc6! This is a really useful PR, glad to be able to merge it :-)

PR #1568, opened 2019-08-28
Update born effective charge in outcar parsing

dongsenfo:
Vasp on nersc now produces extra text in front of the born effect charge which prevents parsing. Got rid of the line which differentiates the two old and new vasp text in front of the born effective charges


shyuep:
Can we have a unittest for the functionality pls? Thanks.

shyuep:
Thanks.

PR #1570, opened 2019-09-02
New vasp set

JaGeo:
## Summary

Hello! 

I have written a LobsterSet to create VASP inputs that can be used for lobster calculations later and I have also created tests. 
Please let me know if I have followed all conventions for VASP sets. Also, let me know if I have overlooked something. The relationships between classes in this module are quite complex due to the many class-subclass relationships.

Best,
JG


shyuep:
Thanks. This is great. I have merged it.

PR #1577, opened 2019-09-04
Incar params

richardtran415:
## Summary

* Added a JSON file containing a dictionary with INCAR flags as keys and the value indicating the object type. 380 flags scoured from the VaspWiki and vasprun.xml file. By no means is this all possible flags for the INCAR, but I will add on as I find more in future pull requests.

* Added Incar class method called check_params(self) which will show 5 different types of warnings:
- Warning if a user provided flag in the incar does not exist in our json file.
- Warning if a parameter for a flag is not a real number when a real number is needed
- Warning if an object type of the parameter for a flag does not match that in the json file
- Warning if only a specific list of strings/parameters is valid for a specific tag and the user provided parameter is not present in the list
- Warning if the parameter should be a list but is not (a future pull request will check if the elements in the list are the correct object type, hence why this is a separate warning)

* Unittest for check_params

## TODO

* Will add more flags along the way

* Add a function for recommending specific flags if a user provides a non-existent flag (e.g. if the user has NBAND in the INCAR file instead of NBANDS, the warning will recommend using NBANDS instead)

* Custodian Outcar validator for checking if all parameters in the INCAR weere properly red (see issue #100 in materialsproject/custodian)


mkhorton:
Thanks @richardtran415, this is fantastic

PR #1580, opened 2019-09-04
Fast pourbaix

montoyjh:
## Refactor of pourbaix diagram module for significant speedup to many-element systems

The pourbaix diagram module is refactored for significant speedup for many element systems.  The methodology is outlined [here](https://arxiv.org/abs/1909.00035).

* Many-element pourbaix diagrams pre-compute valid "MultiEntries" using a convex hull from a grand-canonical phase diagram which includes electrons and protons as free species.
* The `MPRester.get_entries_in_chemsys` method is refactored to use a MongoDB `$in` operation on the chemsys, rather than iteratively querying the rest endpoint.  This is significantly faster for many-element systems, and similarly performant for small-element systems.
* The `PourbaixDiagram.get_decomposition_energy` (i. e. the "pourbaix hull" method) now returns eV/atom, rather than eV/formula unit.
* `PourbaixDiagram.get_decomposition_energy` is now constrained to use the same composition as the pourbaix diagram, rather than attempting to make an inference about which phase the user intends to compute in pourbaix space.  I'd like to reintroduce this functionality at some point with a more concrete use case and example that has a more rigorous physical interpretation, but as it stood was costly and likely misleading.  This may introduce some hiccups to existing data pipelines.


mkhorton:
Very happy to see this :) will merge when tests pass

PR #1589, opened 2019-09-12
Fixed some minor bugs in Trajectory and more features

sivonxay:
## Summary

- Fixed issue with slicing, wherein only fully specified slices work. example: traj[0 : -1 : 10] worked, but traj[ : : 10] and traj[ : -1 : 10] did not work.
- Added property to keep track of information of each time step such as Temperature, Pressure, and Energy. This variable is called 'frame_property', but if a better name is proposed, I have no problem changing it.
- Added more tests for slicing, site properties, and frame properties

### DIFF
--- a/pymatgen/core/tests/test_trajectory.py
+++ b/pymatgen/core/tests/test_trajectory.py
@@ -1,3 +1,4 @@
+
 from pymatgen.util.testing import PymatgenTest
 from pymatgen.io.vasp.outputs import Xdatcar
 from pymatgen.io.vasp.inputs import Poscar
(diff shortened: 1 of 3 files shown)


shyuep:
Thanks. While you are fixing the errors (travis fails immediately with codestyle issues), can you add some basic type annotations for the main classes in Trajectory? Thanks.

mkhorton:
Regarding "frame_property," this is really a problem of associating metadata with a structure. We have specie properties, site properties, but no structure properties. This tends to lead to structure properties being created in an ad-hoc way, like charge, or here, with the energy/temperature/pressure.

I think the solution proposed here with "frame_property" is fine, I'm just noting this as something we may want to think about since there are many cases where this might be useful.

On the other side of the issue, a problem with site_properties etc. at the moment is that it's a free-form dict and can't be type-hinted (or, rather, it's `Dict[str, List[Any]]`, which doesn't tell us much). This makes it powerful and versatile, but also makes it more difficult to reason about and prone to error, and would also be an issue here with frame_property.

sivonxay:
@shyuep Fixed! :D

@mkhorton I do see where the issue is with site_properties and "frame_property" being free-form. Perhaps there is a way to let the user know which keys are supported by default/ used by other bits of code in pymatgen, but still allow users to have the flexibility of using other properties? As for structure properties, I would be in support of creating structure properties.

shyuep:
Many thanks!

PR #1594, opened 2019-09-17
Unify lattice.get_points_in_sphere and structure.get_all_neighbors

### LINKED ISSUE #1592: Inconsistent Structure.get_distance and Lattice.get_distance methods.
Recently, @chc273 implemented a much better distance computation algo that scales linearly with # of atoms Structure.get_neighbors. But some of the other methods seem to still use Lattice.get_points_in_sphere.

I would like to unify all these treatments. Basically port the general concept of get_neighbors to Lattice.get_points_in_sphere and then have all relevant methods just call that. This is reduce code duplication and ensure we have one consistent algo.

@chc273 can do the honors?

chc273:
## Summary

This fixes #1592 

1. Moved neighbor computations from `structure.py` to `lattice.py`
2. Added a function `lattice.get_points_in_spheres` that computes, for each `center_coords`,  the neighboring points from `all_coords` within a cutoff `r`. All coordinates are cartesian. This function takes optionally boolean `pbc` to indicates periodic boundary for each dimension. When any direction is periodic, a periodic `lattice` should be supplied. The function returns `List[List[Tuple[coord, distance, index, image]]]`
3. `Lattice.get_points_in_sphere` calls `get_points_in_spheres` by supplying only one `center_points`
4. Previous `Lattice.get_points_in_sphere` was moved to `Lattice.get_points_in_sphere_old` with deprecation decorator. 
5. `Structure.get_all_neighbors` calls `get_points_in_spheres` and constructs the neighbor list with site properties.


shyuep:
Thanks. Do you mind closing this PR and resubmititng another one? I am testing out Github Action workflows for PRs. I want to make sure that the Actions get run. You can just copy and paste the same PR text.

chc273:
@shyuep  ok

PR #1595, opened 2019-09-17
Unify lattice.get_points_in_sphere and structure.get_all_neighbors 

### LINKED ISSUE #1592: Inconsistent Structure.get_distance and Lattice.get_distance methods.

chc273:
This fixes #1592

Moved neighbor computations from structure.py to lattice.py
Added a function lattice.get_points_in_spheres that computes, for each center_coords, the neighboring points from all_coords within a cutoff r. All coordinates are cartesian. This function takes optionally boolean pbc to indicates periodic boundary for each dimension. When any direction is periodic, a periodic lattice should be supplied. The function returns List[List[Tuple[coord, distance, index, image]]]
Lattice.get_points_in_sphere calls get_points_in_spheres by supplying only one center_points
Previous Lattice.get_points_in_sphere was moved to Lattice.get_points_in_sphere_old with deprecation decorator.
Structure.get_all_neighbors calls get_points_in_spheres and constructs the neighbor list with site properties.


shyuep:
Ok, there are some errors related to mypy that you need to fix. 

@mkhorton For some reason, the Github actions are not running on PRs. Can you check?

mkhorton:
Looks like this may be a limitation during the beta on PRs from forked repos.

mkhorton:
Actually https://help.github.com/en/articles/events-that-trigger-workflows#pull-request-events-for-forked-repositories

shyuep:
@mkhorton The article doesn't say how to get it to work.

mkhorton:
Yes, I'm finding conflicting information online -- it sounds like this was only recently added, e.g. see this thread https://github.com/sdras/awesome-actions/issues/98. I'm not seeing the answer in the documentation (aside to use push events, which we already do).

shyuep:
@chc273 There are still failing tests. These should be fixed.

chc273:
@shyuep  i am on it.

pymatgen/core/lattice.py
```diff
@@ -1052,6 +1052,8 @@ def get_points_in_sphere(
     ) -> Union[
         List[Tuple[np.ndarray, float, int, np.ndarray]],
         Tuple[List[np.ndarray], List[float], List[int], List[np.ndarray]],
+        List[Any],
```
mkhorton: Why [Any] here?

chc273: Just to accommodate the case where no neighbors are found, in which I simply return an empty list or a tuple of empty lists. Otherwise, I will need to design an empty return that suits the return format.

mkhorton: Would `Optional` not be appropriate here?

chc273: Let me check. Thanks

shyuep:
Thanks.

PR #1597, opened 2019-09-19
[Needs test] Add vacuum mode to Bader caller

mkhorton:
Needs test before merging.

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -106,7 +106,8 @@ class BaderAnalysis:
               " Please download the library at http://theory.cm.utexas"
               ".edu/vasp/bader/ and compile the executable.")
     def __init__(self, chgcar_filename, potcar_filename=None,
-                 chgref_filename=None, parse_atomic_densities=False):
+                 chgref_filename=None, parse_atomic_densities=False,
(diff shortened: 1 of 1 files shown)


pymatgen/command_line/bader_caller.py
```diff
@@ -106,7 +106,8 @@ class BaderAnalysis:
               " Please download the library at http://theory.cm.utexas"
               ".edu/vasp/bader/ and compile the executable.")
     def __init__(self, chgcar_filename, potcar_filename=None,
-                 chgref_filename=None, parse_atomic_densities=False):
+                 chgref_filename=None, parse_atomic_densities=False,
+                 vacuum_mode="auto"):
```
shyamd: Can we make this an enum to provide some unambiguity? I think in general we should do that for anything that is a selection of choices.

mkhorton:
Even better, we have new Literal type hints now!

We have added MSONable support for Enums in the latest monty but I still
find it verbose/a little clumsy to have to init the Enum before you can
init the class. You can also create the Enum from a string kwarg inside the
class initializer and raise an error that way if the option you choose in
unsupported but again a lot of boilerplate.

mkhorton:
Will re-open this pending changes to Bader from #1804

Issue #1212, opened 2018-07-20
Cythonized linear assignment algorithm very slow ONLY with negative costs


emanuele:
Unexpectedly, in `pymatgen.optimization`, the cythonized linear assignment algorithm LAPJV, inside `linear_assignment.pyx`, becomes extremely slow when the cost matrix has some negative values. Adding a constant to the cost matrix does not change the solution and when the constant is high enough, the equivalent matrix becomes non-negative. In that case the algorithm finds the (same) solution in a much shorter time (10x - 100x). This issue does not affect the NumPy implementation of LAPJV. Example code:
```python
import numpy as np
import time
try:
    from pymatgen.optimization.linear_assignment import LinearAssignment
except ImportError:
    print("WARNING: Cythonized LAPJV not available. Falling back to Python/NumPy.")
    from pymatgen.optimization.linear_assignment_numpy import LinearAssignment

# matrix with some negative values
matrix = np.random.normal(size=(1000, 3000))
# matrix with all non-negative values
matrix_nonnegative = matrix - matrix.min()

t0 = time.time()
assignment = LinearAssignment(matrix).solution
print("Linear Assignment with some negative values: %s seconds" %(time.time()-t0))
t0 = time.time()
assignment_nonnegative = LinearAssignment(matrix_nonnegative).solution
print("Linear Assignment with non-negative values: %s seconds" %(time.time()-t0))
print("Are the solutions the same? %s" %(assignment == assignment_nonnegative).all())
```

Output with cythonized LAPJV, showing a 50x slower solution when negative costs are present:
```
Linear Assignment with some negative values: 5.35161304474 seconds
Linear Assignment with non-negative values: 0.124813079834 seconds
Are the solutions the same? True
```

Output with NumPy LAPJV, showing no difference between the timing of non-negative vs some-negative costs:
```
WARNING: Cythonized LAPJV not available. Falling back to Python/NumPy.
Linear Assignment with some negative values: 213.514488935 seconds
Linear Assignment with non-negative values: 209.583215952 seconds
Are the solutions the same? True
```

I suspect that there is a bug in the Cython implementation of LAPJV. A short-term solution is to force the cost matrix to be non-negative, or to make it positive adding a constant and then subtracting its effect from the final optimized cost.

PR #1610, opened 2019-09-30
Speed up finding neighbor points with cython

chc273:
## Summary

Previously, for finding neighbors from in a structure with given centering points, I need to use `Structure.get_all_neighbors` which has a python loop of constructing `PeriodicSite`. In cases where the sites are not used, this is very slow. Therefore, a general function that find points in different sphere centers directly will be of value. 

Here I am isolating a `find_neighbors_in_spheres` function that takes a two sets of cartesian coordinates `all_coords` and `center_coords`, as well as a `lattice` matrix and a cutoff radius `r`. The function will return all neighbors within `r` for `center_coords` that are in a lattice constructed by `lattice` matrix and `all_coords` point coordinates.  This is written in `cython` syntax, and compiled to cpp. 

The return data structure for the `find_neighbors_in_spheres` are four `np.ndarray`s: 
`index1`, `index2`, `offsets`, `d`, where `index1` is a 1d array (size n) of `center_coords` indices, `index2` is a 1d array (size n) of `all_coords` indices, `offsets` is a 2d array (size nx3) that stores the image offset vectors for each `all_coords[index2]`, and `d` is a 1d array (size n) measuring the distance between center points and the neighbor points. 

The new `find_points_in_spheres` should be extremely fast for specific purposes if accessed directly.  However, it does not accelerate `Structure.get_all_neighbors` by much due to the slow speed of constructing sites and append to lists etc. Test results show that `find_points_in_spheres` takes less than 1% of the total time in `Structure.get_all_neighbors` when the system is large (> 1000 atoms), and is even less when the system size is even larger. 

I also changed the definition of image offset vectors. In previous implementations, all site coordinates were first wrapped inside the periodic boundary and the image offset vectors are defined w.r.t. the wrapped sites. It is fine if the initial sites are all within the unit cell, but is problematic when we have initial coordinates of the structure that are outside the unit cell. 
For example, if we have a structure `s=Structure(Lattice.cube(1), ['H', 'H'], [[1.1, 0, 0], [0.05, 0, 0]])`, previous implementation will find unexpected `image` offsets for neighbors of the second atom. Comparing `s.get_all_neighbors(0.06)[1]` and `s.get_all_neighbors_old(0.06, include_image=True)[1]`, the `old` version will give a neighbor `H(0.1, 0, 0)` with `image=(0, 0, 0)`. Apparently, this neighbor is a `(-1, 0, 0)` shifted `H[1.1, 0, 0]`.  
 

* `find_points_in_spheres` cython implementation
*  Changed the image offset vectors definition


mkhorton:
I really like this code and these optimizations, but can we discuss how we expose this to the user?

Some users have trouble compiling C extensions, so it would be nice to have a pure Python fallback for those users.

I see that currently you have an `@deprecated` `get_all_neighbors_py`. We already have `.pyx` files inside `pymatgen.optimization` -- you could put the new neighbors `.pyx` file inside that folder. You could then add `try: ... except ImportError: ...` inside `pymatgen.core.structure` and if it fails to import, then fall back to the old pure python version?

I'm also open to other ideas.

This would work if the Python and Cython functions are functionally equivalent.

chc273:
@mkhorton  Thanks for the suggestion.. The code also used c++ and it seems that c++ is even more troublesome for distribution. I will go back and revise it. Thanks again.

shyuep:
Wait first... 
 personally, I don't think we should avoid C-extensions. With the conda-forge distributions, c extensions are fine. In any case, we already have them for coord_pbc and spglib.

mkhorton:
I would say a good 50% of first-time-user queries we have are from broken C extensions.

I think pymatgen should be *fast*, so I'm on-board with including C extensions, but if we already have a pure Python implementation why not fall back to that gracefully if the C extensions don't compile?

mkhorton:
The spglib extensions aren't an issue because a lot of pymatgen can be used without them, the coord_pbc extensions are the ones that cause users problems (e.g. if they're just installing pymatgen to do some basic manipulations, or use the rester, etc.)

shyuep:
Do they cause issues in conda installs? If not, we should just point them there....

shyuep:
FYI, numpy is basically in all of pymatgen, and it is largely c....

mkhorton:
Yeah, I don't know -- maybe numpy has more wheels available (e.g. see https://pypi.org/project/numpy/1.17.2/#files) or it's just pre-installed on more systems. In either case, we definitely do get queries about import errors with coord_cython.

mkhorton:
Maybe something like this will help our wheel coverage https://github.com/joerick/cibuildwheel -- I'll investigate further. I still think having a graceful Python fallbacks makes sense if we already have a working, consistent pure Python version though.

shyuep:
I know we get queries. I am saying we should just point them to a conda install and tell them to go debug themselves if they do not want to use conda. Check out https://anaconda.org/conda-forge/pymatgen/files . As far as we are concerned, we have complete coverage of Win-OSX-Linux here. Since the uploading to conda-forge channel, we have gotten ~5k downloads in a matter of 1 month (it took us three years to get to 20k in the matsci channel). 

My problem with maintaining a separate python version is the biggest no-no of coding - duplication. If I write the same routine in two different ways in two places, I have to remember to change two places when I make any changes to the API.

mkhorton:
Yes, that's true, and arguably something we've encountered already.

The flip side of this discussion is that by adding a second program
language (C) we're reducing the pool of people that can read and understand
the algorithm, including making improvements to it. It'll in practice
become static/dead code. At least having the Python version gives us a
"reference" implementation that's more readable, and easier to compare the
output of the optimized version to and see when they disagree.

As an example of both points we have a bug in the linear assignment code
with a fix proposed (see
https://github.com/materialsproject/pymatgen/issues/1212) that is still
open because I'm not sure that anyone has a good understanding of that code
(e.g. if there's a bug in
https://github.com/materialsproject/pymatgen/blob/67e4fbc78890f93a8db0b38166eca3015c6cc4ca/pymatgen/optimization/linear_assignment.pyx#L80
I'm not sure I'd ever find it).

chc273:
> so I'm on-board with including C extensions, but if we already have a pure Python implementation why not fall back to that gracefully if the C extensions don

If the C extension does not compile then user would not be able to install in the first place right? It seems that the `linear_assignment` fallback to python was deleted.

mkhorton:
> If the C extension does not compile then user would not be able to install in the first place right? 

I don't know, but I don't think so; I've seen installations where pymatgen is apparently installed but the Cython code doesn't work. So you can attempt to import pymatgen but you get an ImportError when the specific Cython submodule tries to get imported, e.g. `ImportError: cannot import name coord_cython`.

mturiansky:
Just to add something to this discussion: I see you're keen on keeping the readability of python code with the speed of c. Have you put any thought into using [numba](https://numba.pydata.org/)? I've had success with applying numba to speed up my code. It allows you to get speed ups in performance critical sections while maintaining the readability of python. You literally only need to add a decorator to the function that you want to speed up. This also removes the problem of having to compile c code when distributing pymatgen.

mkhorton:
I think numba’s a great suggestion, I know it plays nicely with numpy too.

Might be worth comparing the speed to Cython; if it’s similar it will be a
lot easier for us to optimize other parts of pymatgen too.

mturiansky:
Yes, numba plays nicely with numpy, and can even be faster than the numpy implementation in certain cases. My experience has been that numba is comparable to cython (sometimes faster because they use aggressive optimizations), e.g., [here's a comparison](https://jakevdp.github.io/blog/2013/06/15/numba-vs-cython-take-2/).

chc273:
> Just to add something to this discussion: I see you're keen on keeping the readability of python code with the speed of c. Have you put any thought into using [numba](https://numba.pydata.org/)? I've had success with applying numba to speed up my code. It allows you to get speed ups in performance critical sections while maintaining the readability of python. You literally only need to add a decorator to the function that you want to speed up. This also removes the problem of having to compile c code when distributing pymatgen.

Good suggestion! @mturiansky I played with numba long time ago and my experience was that you have to enable `nopython` to speed it up, and in that case some python object does not work. I prefer cython for this since it is statically typed and there will be fewer unexpected behaviors except for some memory handling. If you can try use numba to optimize some `pymatgen` code and show the results, that should be extremely helpful.

chc273:
> I really like this code and these optimizations, but can we discuss how we expose this to the user?
> 
> Some users have trouble compiling C extensions, so it would be nice to have a pure Python fallback for those users.
> 
> I see that currently you have an `@deprecated` `get_all_neighbors_py`. We already have `.pyx` files inside `pymatgen.optimization` -- you could put the new neighbors `.pyx` file inside that folder. You could then add `try: ... except ImportError: ...` inside `pymatgen.core.structure` and if it fails to import, then fall back to the old pure python version?
> 
> I'm also open to other ideas.
> 
> This would work if the Python and Cython functions are functionally equivalent.

The `try: ... except ImportError` fall back to python is implemented in the latest commit. Thanks for the great suggestion.

mkhorton:
My understanding of numba's "nopython" mode is just that you have to be strict with your types going in (which is fine) and that you have to avoid allocating new Python objects inside your loop (which you're not doing here, since you mentioned you're avoiding creating new Site objects).

This blog post makes a very compelling case for numba, https://jakevdp.github.io/blog/2013/06/15/numba-vs-cython-take-2/

I too would be very interested in seeing a proof-of-concept -- for now this current solution looks good (@chc273, make sure you catch the `except ImportError` immediately after the `from ... import ...` line and put the remainder of that code in an `else` clause).

mkhorton:
A choice quote:

> On top of [numba] being much easier to use (i.e. automatic type inference by`autojit`) it's now about 50% faster, and is even a few percent faster than the Cython option.

> And though I've seen similar things for months, I'm still incredibly impressed by the results enabled by Numba: a single function decorator results in a 1300x speedup of simple Python code. I'm becoming more and more convinced that Numba is the future of fast scientific computing in Python

mturiansky:
> Good suggestion! @mturiansky I played with numba long time ago and my experience was that you have to enable `nopython` mode to speed it up, and in that case some python object does not work. I prefer cython for this since it is statically typed and there will be fewer unexpected behaviors except for some memory handling. If you can try use numba to optimize some `pymatgen` code and show the results, that should be extremely helpful.

@chc273 You're right, nopython mode is necessary to get the speedup. The best way to approach the problem is essentially how you would with cython. You wrote a separate function `find_neighbors_in_sphere` which takes and returns numpy arrays and handles no other python objects. This is something numba handles quite well. You would essentially do the same with numba, write a separate python function that only handles numpy arrays and slap the numba `njit` decorator on it.

I don't really have time at the moment to play around with optimizing some pymatgen, but I have a smallish example on my forked repository that speeds up wavecar processing (not much though because it's IO limited). You can see it [here](https://github.com/mturiansky/pymatgen/blob/hacks/pymatgen/io/vasp/outputs.py#L4058-L4090). All I did was move the "_generate_G_points" function outside the class and slap the `njit` on it. This function only takes numpy arrays as input and output so it's a prime candidate for numba optimization. I even made numba optional with [this try,except statement](https://github.com/mturiansky/pymatgen/blob/hacks/pymatgen/io/vasp/outputs.py#L45-L49).

mkhorton:
@mturiansky was moving it outside the class necessary?

mturiansky:
@mkhorton I believe so because if you have the `self` python object, then numba will complain. Maybe there's a way to do it with static methods or something but moving it outside was an easy solution.

They're working on support for classes, but it's in [beta](https://numba.pydata.org/numba-doc/dev/user/jitclass.html).

shyuep:
I am merging this now. If someone else want to take up the mantle of numba-ing it, ifine.

shyuep:
@chc273 I have sped up the get_all_neighbors function in Structure further by avoiding PeriodicSite construction. The main reason why PeriodicSite init is so slow is because of all the checks implemented. These checks are unnecessary when we are just creating a neighbor. 

Now the get_all_neighbors return a PeriodicNeighbor, which is much faster. Profiling suggests it shaves ~1s off a 10x10x10 Li2O cell with r = 5. Pls test it out. It is in the nnspeedup branch for now to make sure I haven't broken anything.

chc273:
ok @shyuep

chc273:
> @chc273 I have sped up the get_all_neighbors function in Structure further by avoiding PeriodicSite construction. The main reason why PeriodicSite init is so slow is because of all the checks implemented. These checks are unnecessary when we are just creating a neighbor.
> 
> Now the get_all_neighbors return a PeriodicNeighbor, which is much faster. Profiling suggests it shaves ~1s off a 10x10x10 Li2O cell with r = 5. Pls test it out. It is in the nnspeedup branch for now to make sure I haven't broken anything.

@shyuep It does seem to increase the speed by 30% amount a few cases I tested with atom number > 1000. The speed is still throttled by the loop step, particularly, the comparison of site (line 1243) and the construction of neighbor object. 

image

shyuep:
I sped it up further a litte by using a trick with checking the if statements. This is probably as fast as it is going to get. The PeriodicNeighor constructor is barebones and requires very few ops. So it is at the limit of what we can do.

chc273:
> I sped it up further a litte by using a trick with checking the if statements. This is probably as fast as it is going to get. The PeriodicNeighor constructor is barebones and requires very few ops. So it is at the limit of what we can do.

I agree, there does not seem to be other things to optimize in python. The issue seems to be also in the use of `dict` and `list` append, and for those we cannot optimize much in python. 

The users can always use `Lattice.get_points_in_sphere` or just the cython func to gain much more speed if they know what they are doing. Thanks for the efforts.

mkhorton:
Fantastic speed-up :-)

PR #1613, opened 2019-10-01
Include additional check to avoid problems for summed cohps

JaGeo:
Hello!

I included an additional test to make sure that two lists that should have the same length actually have the same lengths. These lists are parameters for "get_summed_icohp_by_label_and_orbital_list". I have also included a test to make sure an error will be raised if they have different lengths. 

Best, 
JG


shyuep:
Thanks.

PR #1623, opened 2019-10-04
robust pbc considerations for get neighbors

chc273:
## Summary

Small changes to revise details of how to treat mixed periodic boundary and non-periodic boundary in the general `find_points_in_spheres` function. 

If the structure has periodic and non-periodic boundary mixed case (2d material, 1d material etc) the find points in spheres function will not augment the non-periodic dimension. Previous one was not robust and needs to supply a large enough lattice to make it correct. This is fixed in this PR.

### DIFF
--- a/pymatgen/core/lattice.py
+++ b/pymatgen/core/lattice.py
@@ -1510,6 +1510,7 @@ def get_points_in_spheres(all_coords: np.ndarray, center_coords: np.ndarray, r: 
     """
     if isinstance(pbc, bool):
         pbc = [pbc] * 3
+    pbc = np.array(pbc, dtype=bool)
     if return_fcoords and lattice is None:
         raise ValueError("Lattice needs to be supplied to compute fractional coordinates")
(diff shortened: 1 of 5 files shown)


mkhorton:
@chc273 do you have any thoughts on wider support in pymatgen for mixed periodic boundary conditions? e.g. like ASE

chc273:
> @chc273 do you have any thoughts on wider support in pymatgen for mixed periodic boundary conditions? e.g. like ASE

I think it's more of a choice to make. VASP works fine by putting molecules and surfaces in a periodic lattice. Some other softwares, for example, lammps do have nonperiodic boundaries by setting 'boundary' keyword. For me I want to have structure and molecule somewhat unified in our attempts to build unified models. I do like ASE's approach of using "Atoms", but its not handy when dealing with complex crystal-based computations. Those computations are also not compatible with molecules. Are we interested in pursuing the unification which may sacrifice certain functionality or make those difficult to use?

PR #1630, opened 2019-10-14
Use caching in MinimumVIRENN class

utf:
## Summary

The `MinimumVIRENN` near neighbors class relies on the `ValenceIonicRadiusEvaluator` which is extremely slow.

If using `MinimumVIRENN` to get the coordination of all sites in a structure, the `ValenceIonicRadiusEvaluator` class will be initialized once per site. This is very inefficient as `ValenceIonicRadiusEvaluator` calculates ionic radii for the whole structure simultaneously.

I've therefore enabled caching of the `ValenceIonicRadiusEvaluator` class using the `lru_cache` functionality in `functools`. In my testing, for some structures this result in a 10x speedup.


utf:
Failing test is for `CrystalAIResterTest.test_query` nothing to do with this PR.

PR #1650, opened 2019-10-31
fix previous destructive usage of input LammpsData

htz1992213:
* Fix: Fixed the previous behavior that the CombinedData initializes by destructively using the input LammpsData objects, which causes errors when reusing those LammpsData.


mkhorton:
Hi @htz1992213, this looks good but missing a few docstrings: https://travis-ci.org/materialsproject/pymatgen/builds/605287251#L1275

shyuep:
I would also suggest adding tests to make sure that the destructive usage problem will not happen again.

shyuep:
Thanks.

PR #1655, opened 2019-11-08
TemCalculator

welltemperedpaprika:
## Summary

- Edits made to Frank Wan's previous TEM calculator, a project under Shyam's direction:
      - Changed method get_interplanar_spacing to call lattice.d_hkl rather than separately calculating for different crystal systems.
      - Vectorized method generate_points to be using numpy arrays rather than nested for loops.
      - Vectorized methods to reduce for loops when looping over dictionaries, if possible.
      - Changed original implementation of prettytable to pandas dataframe.
      - Removed TEM Dot object to a more lightweight named tuple.
      - Will be used to generate dataset analyzing material differences in TEM space.
      - Updated structural factors. Matches international table of crystallography.

Request submitted by Jason Liang, student working with Shyam.


shyamd:
I think during your merge of the master branch, some of your code got deleted. Take a look at the diff for that commit.

welltemperedpaprika:
> I think during your merge of the master branch, some of your code got deleted. Take a look at the diff for that commit.
> 
> There may also be issues where your code isn't correct. For instance: `is_perm([1,1,1],[1,-1,1])`
> 
> Maybe something like this?
> 
> ```
> from collections import defaultdict
> 
> def is_perm(hkl1 : Union[List,Tuple],hkl2 : Union[List,Tuple]) -> bool:
>     """
>     Determines if two HKL indicies are permutations of each other.
>     This does not consider crystallographic symmetry, just index symmetry
>     """
>     if len(hkl1) != len(hkl2):
>         return False
> 
>     counts = defaultdict(int)
>     for i in hkl1:
>         counts[abs(i)] += 1
> 
>     for j in hkl2:
>         counts[abs(j)] -= 1
> 
>     return sum(counts.values()) == 0
> ```
> 
> Also, can you make some tests for `get_unique_families` so that we catch errors?

is_perm and get_unique_families are part of the core code, I will make the change but it will cause conflict when I merge. Is that ok? I will take a look at the deleted codes and put them back if necessary.

shyamd:
That's fine. You're improving code that is already there and that can include "core" code. When you update now there shouldn't be a conflict anymore. Just make sure its a new commit and not reverting an old commit.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+from IPython.display import set_matplotlib_formats  # type: ignore
+from pymatgen import Structure  # type: ignore
+from pymatgen.analysis.diffraction.core import AbstractDiffractionPatternCalculator  # type: ignore
+
+with open(os.path.join(os.path.dirname(__file__),
+                       "atomic_scattering_params.json")) as f:
+    ATOMIC_SCATTERING_PARAMS = json.load(f)
+set_matplotlib_formats('retina')
+poff.init_notebook_mode(connected=True)
+# coding: utf-8
```
mkhorton: Remove duplicate lines

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+__email__ = "fwan@berkeley.edu, yhljason@berkeley.edu"
+__date__ = "06/19/2019, updated 01/2020"
+
+
+class TEMCalculator(AbstractDiffractionPatternCalculator):
+    """
+    Computes the TEM pattern of a crystal structure for multiple Laue zones.
+    """
+
+    def __init__(self, wavelength_cache: Dict[float, float] = {}, symprec: float = None, voltage: float = 200,
```
mkhorton: `wavelength_cache` not in docstring, dangerous {} default

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+                 beam_direction: List[int] = [0, 0, 1], camera_length: int = 160,
+                 debye_waller_factors: Dict[str, float] = None, cs: float = 1) -> None:
```
mkhorton: I'm not sure the return type is accurate(?)

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+__date__ = "06/19/2019, updated 01/2020"
+
+
+class TEMCalculator(AbstractDiffractionPatternCalculator):
+    """
+    Computes the TEM pattern of a crystal structure for multiple Laue zones.
+    """
+
+    def __init__(self, wavelength_cache: Dict[float, float] = {}, symprec: float = None, voltage: float = 200,
+                 beam_direction: List[int] = [0, 0, 1], camera_length: int = 160,
```
mkhorton: Dangerous list default, should be a tuple `(0, 0, 1)`, see https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+            list of 3-tuples
+        """
+        if any(isinstance(n, tuple) for n in points):
+            return points
+        if len(points) == 0:
+            return []
+        filtered = np.where(np.dot(np.array(self.beam_direction), np.transpose(points)) == laue_zone)
+        result = points[filtered]
+        result_tuples = [tuple(x) for x in result.tolist()]
+        return result_tuples  # type: ignore
```
mkhorton: Why `# type: ignore`?

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+            dict of hkl plane to cell intensity
+        """
+        csf = self.cell_scattering_factors(structure, bragg_angles)
+        plane = bragg_angles.keys()
+        csf_val = np.array(list(csf.values()))
+        cell_intensity_val = (csf_val * csf_val.conjugate()).real
+        cell_intensity = dict(zip(plane, cell_intensity_val))
+        return cell_intensity
+
+    def get_pattern(self, structure: Structure, scaled: bool = True, two_theta_range: tuple = (0, 90)) \
```
mkhorton: type should be `Tuple[float, float]`?

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+            pos = np.array([coeffs[0] * positions[first_point][0] + coeffs[1] * positions[second_point][0],
+                            coeffs[0] * positions[first_point][1] + coeffs[1] * positions[second_point][1]])
+            positions[plane] = pos
+        points.append((0, 0, 0))
+        points.append(first_point)
+        points.append(second_point)
+
+        return positions
+
+    def TEM_dots(self, structure: Structure, points: list) -> list:
```
mkhorton: capitalization in function name, PEP8 violation

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,520 @@
...
+            position = positions[plane]
+            hkl = plane
+            intensity = cell_intensity[plane]
+            film_radius = 0.91 * (10 ** -3 * self.cs * self.wavelength_rel() ** 3) ** Fraction('1/4')
+            d_spacing = interplanar_spacings[plane]
+            TEM_dot = dot(position, hkl, intensity, film_radius, d_spacing)
+            dots.append(TEM_dot)
+        return dots
+
+    def show_plot_2d(self, structure: Structure):
```
mkhorton: Method is a lot more useful if it returns the `fig` object directly

requirements-optional.txt
```diff
@@ -16,4 +16,5 @@ pycodestyle==2.5.0
 mypy==0.761
 pydocstyle==5.0.1
 flake8==3.7.9
-pylint==2.4.4
\ No newline at end of file
+pylint==2.4.4
+plotly==3.10.0
```
mkhorton: This is an out-of-date plotly version, please bump to latest

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: > This looks great, there are a few code style issues and other comments. When these are addressed, I'll take another look + test it for correctness, and then we can merge.

Thanks!

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: Fixed.

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: I googled around and according to https://www.python.org/dev/peps/pep-0484/ the return type for init method is none.

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: Thanks!

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: I am using hkl plane to a float here, so I think it is Tuple[int, int, int]?

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: Thanks!

requirements-optional.txt
welltemperedpaprika: Updated

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: Fixed!

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: When I run mypy, it keeps giving me an error that I have Tuple[any, any...] even though I checked that they are int inside the tuple. If I change the type hint to Tuple[any, any..] instead, it creates several more errors because other methods are specified as Tuple[int, int, int]... So to keep the type hints as meaningful as possible I added it

pymatgen/analysis/diffraction/tem.py
mkhorton: Sorry, I was referring to the two_theta range specifically in get_pattern

pymatgen/analysis/diffraction/tem.py
mkhorton: Ah, yes. I suspect that's because it doesn't know the length of `result`. You may be able to fix by explicitly declaring `result_tuples: Tuple[int, int, int] = ...`, I'm not sure. You could also try `Tuple[int, ...]` maybe. Annoying!

welltemperedpaprika: I will take another look today ASAP! Thanks

mkhorton: Thanks, and not the end of the world if you can't fix. Type hints are definitely preferred but they are optional for non-core classes, and the method signature is the main thing the matters.

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: Ah! Sorry, those arguments were left from previous implementation of the method, which were relevant to powder diffraction, and as I work on the code, I found them to be unnecessary. They should not have an effect on the code.

pymatgen/analysis/diffraction/tem.py
welltemperedpaprika: Thank you! I tried different combinations of type hints for multiple methods to make them work... but mypy still won't recognize it. I am now casting the result to be List[Tuple[int, int, int]] (even though when I run the debugger, it is already List[Tuple[int, int, int]] without casting...).

mkhorton:
Thanks @welltemperedpaprika, I'll take another look

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,523 @@
...
+            Args:
+                structure (Structure): The input structure.
+                scaled (boolean): Required value for inheritance, does nothing in TEM pattern
+                two_theta_range (Tuple): Required value for inheritance, does nothing in TEM pattern
+            Returns:
+                PandasDataFrame
+        """
+        points = self.generate_points(-10, 11)
+        TEM_dots = self.tem_dots(structure, points)
+        field_names = ["Pos", "(hkl)", "Intnsty (norm)", "Film rad", "Interplanar Spacing"]
```
mkhorton: I'd write out these in full, e.g. "Intensity", "Position", "Film radius" ... abbreviating doesn't help much.

Also `TEM_dots` (capitalized) is a PEP8 violation.

Below, why the line `dict1 = {}`?

requirements-optional.txt
```diff
@@ -16,4 +16,5 @@ pycodestyle==2.5.0
 mypy==0.761
 pydocstyle==5.0.1
 flake8==3.7.9
-pylint==2.4.4
\ No newline at end of file
+pylint==2.4.4
+plotly==4.5.0
```
mkhorton: plotly should be moved into requirements.txt, and make sure it's included in setup.py too (with a >= version number)

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,523 @@
...
+            position = positions[plane]
+            hkl = plane
+            intensity = cell_intensity[plane]
+            film_radius = 0.91 * (10 ** -3 * self.cs * self.wavelength_rel() ** 3) ** Fraction('1/4')
+            d_spacing = interplanar_spacings[plane]
+            TEM_dot = dot(position, hkl, intensity, film_radius, d_spacing)
+            dots.append(TEM_dot)
+        return dots
+
+    def show_plot_2d(self, structure: Structure) -> go.Figure:
```
mkhorton: Now that this returns the go.Figure, the method name should be "get_plot_2d"

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,523 @@
...
+                showticklabels=False,
+            ),
+            width=600,
+            height=600,
+            paper_bgcolor='rgba(100,110,110,0.5)',
+            plot_bgcolor='black'
+        )
+
+        fig = go.Figure(data=data, layout=layout)
+        poff.iplot(fig, filename='stuff')
```
mkhorton: Remove this line(?)

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,523 @@
...
+class TEMCalculator(AbstractDiffractionPatternCalculator):
+    """
+    Computes the TEM pattern of a crystal structure for multiple Laue zones.
+    """
+
+    def __init__(self, wavelength_cache: Dict[float, float] = None, symprec: float = None, voltage: float = 200,
+                 beam_direction: Tuple[int, int, int] = (0, 0, 1), camera_length: int = 160,
+                 debye_waller_factors: Dict[str, float] = None, cs: float = 1) -> None:
+        """
+        Initializes the TEM calculator with a given radiation.
```
mkhorton: This docstring needs cleaning, it's not clear to me (as a user) what I need to input... e.g. what wavelength_cache is, what kind of structure refinement is done (does this put it into a conventional setting?), also the "later on" comment at the end should be put into a "#TODO:" comment in the code itself rather than the docstring.

mkhorton:
Added a few more comments -- this looks very close to being ready to merge, thanks for your patience waiting for the review.

welltemperedpaprika:
> Added a few more comments -- this looks very close to being ready to merge, thanks for your patience waiting for the review.

Thanks! I have addressed the changes and updated the code.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,608 @@
...
+            "twin direction" for certain materials, "sample thickness", and "excitation error s"
+    """
+
+    def __init__(self, wavelength_cache: Dict[float, float] = None, symprec: float = None, voltage: float = 200,
+                 beam_direction: Tuple[int, int, int] = (0, 0, 1), camera_length: int = 160,
+                 debye_waller_factors: Dict[str, float] = None, cs: float = 1) -> None:
+        """
+        Initializes the TEM calculator with given voltage,
+        Args:
+            wavelength_cache (dict): Reference of voltage to relativistic wavelength. Value dependent on
```
shyamd: Let's actually get rid of this wavelength cache and use the `lru_cache` decorator from `functools` to take care of that caching.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,608 @@
...
+        conversion based on the International Table of Crystallography.
+    #TODO: Could add "number of iterations", "magnification", "critical value of beam",
+            "twin direction" for certain materials, "sample thickness", and "excitation error s"
+    """
+
+    def __init__(self, wavelength_cache: Dict[float, float] = None, symprec: float = None, voltage: float = 200,
+                 beam_direction: Tuple[int, int, int] = (0, 0, 1), camera_length: int = 160,
+                 debye_waller_factors: Dict[str, float] = None, cs: float = 1) -> None:
+        """
+        Initializes the TEM calculator with given voltage,
```
shyamd: This should just be the Argument list description since the docstring for the module actually explains what this calculator does.

mkhorton:
Thanks, just had a chance to take another look at this, have got it working now. I have a few style tweaks to ask before we merge.

image

Could you remove the "trace 1" legend (it doesn't tell us anything) so that the colorbar is better aligned, maybe slightly increase the zoom to be more like the above (contrast on the higher order dots is so low you can't see them, and default zoom makes points overlap, see below). Finally, could you add a plot title (the "2D Diffraction Pattern, Beam Direction: 001") instead of the current label and just leave it as a default font. Also, for beam directions that are negative can you make sure to use the overbar (utils to convert to unicode in pymatgen.util.string might be helpful here).

Finally, what are your thoughts on making the colorbar optional? I think intensity is fairly clear + it doesn't tell us much from a quantitative perspective. If the colorbar was non-grey I think this would be a different matter.

image

mkhorton:
This also now has some minor merge conflicts, should be easy to resolve though

welltemperedpaprika:
> This also now has some minor merge conflicts, should be easy to resolve though

Thanks! I will address the suggestions and make changes accordingly in the next few days.

shyamd:
Should be good to go now.

welltemperedpaprika:
> Thanks, just had a chance to take another look at this, have got it working now. I have a few style tweaks to ask before we merge.
> 
> image
> 
> Could you remove the "trace 1" legend (it doesn't tell us anything) so that the colorbar is better aligned, maybe slightly increase the zoom to be more like the above (contrast on the higher order dots is so low you can't see them, and default zoom makes points overlap, see below). Finally, could you add a plot title (the "2D Diffraction Pattern, Beam Direction: 001") instead of the current label and just leave it as a default font. Also, for beam directions that are negative can you make sure to use the overbar (utils to convert to unicode in pymatgen.util.string might be helpful here).
> 
> Finally, what are your thoughts on making the colorbar optional? I think intensity is fairly clear + it doesn't tell us much from a quantitative perspective. If the colorbar was non-grey I think this would be a different matter.
> 
> image

Thanks, I've addressed the changes. I agree that the colorbar isn't necessary, so I took it out. I also zoomed in a little bit on the x and y axis. It helps on some structures.

mkhorton:
Ok, I'm still having some issues with the title formatting and some other small issues, but I'm happy to take a look at it myself and/or we can iterate on it in an additional PR.

I'm going to merge this in now. Thanks again @welltemperedpaprika, @thefrankwan and @shyamd for all your work on this PR!

shyuep:
@mkhorton This shouldn't be merged. The TEM calculator uses IPython.display for some reason. It is failing on Appveyor. IPYthon is not a dependency of pymatgen. Pls revoke.

shyamd:
It looks like all the tests are passing. Where is the Appyveyor test?

shyuep:
appveyor.yml. In any case, no Ipython.

mkhorton:
Ok, will either fix or revert the PR before next release. This shouldn't use IPython.display.

shyuep:
I am surprised the tests pass. Ipython is definitely not anywhere in the req list. Maybe one of the other dependencies installed it? It is certainly not on appveyor.

pymatgen/analysis/diffraction/tem.py
```diff
@@ -0,0 +1,588 @@
...
+from collections import namedtuple
+from fractions import Fraction
+from typing import List, Dict, Tuple, cast
+from functools import lru_cache
+import numpy as np  # type: ignore
+import scipy.constants as sc  # type: ignore
+import pandas as pd  # type: ignore
+import plotly.graph_objs as go  # type: ignore
+import plotly.offline as poff  # type: ignore
+from IPython.display import set_matplotlib_formats  # type: ignore
```
mkhorton: This is the offending line; the lines before with `set_matplotlib_formats` and `poff` are entirely unnecessary too, I think this is remnants of debug code

shyuep:
There are other issues with this. 

- `from __future__` imports should never be used anymore.
- what is with all the #type: ignore. We didn't set up type checking just so people can ignore them. Unless there is a good justification for ignoring types for something, the default should not be to pepper the entire code with type ignores.

mkhorton:
* Agreed about `__future__`
* Given that our type hint policy is currently that type hints are optional, and this is a non-core module, I thought this was fine, and at least was a starting point for someone to fully add type hints at a later date.

shyuep:
@mkhorton Type hints are optional if they are writing a NEW module and using type hints on the new module. Here, the type hints are being introduced willy nilly on imports that are property type-hinted (e.g., Structure). The whole point is for non-core classes to rely on the type hints to ensure correct behavior.

I have taken it on myself to remove all the extraneous code and to remove the #type:ignores. In fact, they did nothing at all. The unittest class name was XRDCalculatorTest, instead of TEMCalculatorTest.

mkhorton:
Ok, thanks for removing them and fixing the name.

welltemperedpaprika:
Sorry about the issues ... Thanks for the help!

welltemperedpaprika:
@mkhorton Sorry for the troubles. I recently noticed that if the beam direction is set to 100, or -100, the pattern fails to generate, due to a division by zero error in one of the calculations. I have rewritten the calculation process to avoid that. It is now working. Furthermore, I added a new method to get the interplanar angle between planes, since the previous implementation in the code did not consider crystal systems. Sorry! And Thank you!

mkhorton:
No problem, this functionality has not been formally released yet. Please make sure the previous issue receives a new test when you fix it. Thanks!

mkhorton:
You'll have to create a new pull request too

PR #1667, opened 2019-11-14
Sign flip

mhsiron:
Correcting sign in BaderAnalysis' get_charge_transfer method

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -265,7 +265,7 @@ class BaderAnalysis:
         for i, v in enumerate(self.natoms):
             potcar_indices += [i] * v
         nelect = self.potcar[potcar_indices[atom_index]].nelectrons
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @mhsiron !

mkhorton:
Test will likely need to be updated also. What system did you verify this against?

mhsiron:
For ZnTe

But currently running on 50 other materials, so maybe I’ll wait!

This is the charge transfer data for Bader vs. DDEC:


For DDEC - the raw output is simply the charge transfer, while for Bader, the raw output is the total amount of charge for the atom. It would be weird for both to have a sign flip?

Also here fields 0-3 in Zn, and 4-7 is Te. I think it would make more sense for Te to be the negative transfer since its more electronegative than Zn?

Martin

mhsiron:
Actually - upon further investigation - I'm not entirely sure this is the right conclusion.

Issue #1668, opened 2019-11-15
Fail to parse OUTCAR file when two numbers are connected in spin components


yuuukuma:
**Describe the bug**
In the OUTCAR file, when there is a part such that, the Outcar parse fails, because the two float numbers are connected.
--------------------------------------- Iteration      1(   1)  ---------------------------------------

POTLOK:  cpu time    0.0040: real time    0.0065
SETDIJ:  cpu time    0.1360: real time    0.1339
atom =   2  type =  2  l = 3
 
 onsite density matrix
 
spin component  1
 
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000  0.0000
 
spin component  2
 
  0.0000  0.0000  0.0000  0.0000 19.2672  **1.6427-28.9219**
  0.0000  0.0000  0.0000  0.0000 19.0213 -1.8641-31.3417
  0.0000  0.0000  0.0000  0.0000 18.2881 -5.4857-33.4018
  0.0000  0.0000  0.0000  0.0000 17.0809 -9.1595-35.0760
  0.0000  0.0000  0.0000  0.0000 15.4217-12.8227-36.3455
  0.0000  0.0000  0.0000  0.0000 13.3405-16.4136-37.1994
  0.0000  0.0000  0.0000  0.0000 10.8748-19.8728-37.6345

**Expected behavior**
Could you divide these two numbers?

**Desktop (please complete the following information):**
 - Version  '2019.11.11'

shyuep:
Pls provide a sample OUTCAR with this problem. I would think you can probably debug this by modifying one of the regex and submitting a PR.

yuuukuma:
Thanks for the prompt reply.
Please find the attached. I will also send the pull request soon.
[OUTCAR.txt](https://github.com/materialsproject/pymatgen/files/3850578/OUTCAR.txt)

PR #1670, opened 2019-11-15
Fix regular expression in read_onsite_density_matrices in Outcar

yuuukuma:
To parse the connected numbers, modify the regular expression.

See Issue #1668


mkhorton:
Thanks @yuuukuma !

mkhorton:
I saw you posted an example Outcar to the issue thread, would it be possible to add this to `test_files` and add a test too?

Here would be an appropriate place to add it: https://github.com/materialsproject/pymatgen/blob/ff1a1239811dcd175259d4e77cd9d3daec8d6b05/pymatgen/io/vasp/tests/test_outputs.py#L974

shyuep:
I have added the test.

PR #1678, opened 2019-11-20
Respect KSPACING setting in the INCAR

rkingsbury:
## Summary

VASP now [recommends](https://www.vasp.at/wiki/index.php/KSPACING) setting KSPACING in the INCAR file rather than supplying KPOINTS manually. This PR updates the teterror handler to accommodate calculations that use KSPACING instead of an explicit KPOINTS file.

Currently, if a user adds KSPACING to user_incar_settings when initializing a VaspInputSet, the setting will be ignored and a KPOINTS file generated anyway. This PR ensures that the KSPACING setting passed to VASP as expected.

See also Custodian [#139](https://github.com/materialsproject/custodian/pull/139)

See also a [similar issue](https://gitlab.com/ase/ase/merge_requests/226/commits) recently fixed in ASE.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -517,7 +517,7 @@ class DictSet(VaspInputSet):
         
         # Return None if KSPACING is present in the INCAR, because this will
         # cause VASP to generate the KPOINTS file automatically
-        if incar.get("KSPACING",None) is not None:
+        if self.incar.get("KSPACING",None) is not None:
             return None
 
         # If grid_density is in the kpoints_settings use
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -1110,6 +1110,11 @@ class MPNonSCFSet(MPRelaxSet):
         """
         :return: Kpoints
         """
+        # override pymatgen kpoints if provided
+        user_kpoints = self.kwargs.get("user_kpoints_settings", None)
+        if isinstance(user_kpoints, Kpoints):
+            return user_kpoints
+
         if self.mode.lower() == "line":
             kpath = HighSymmKpath(self.structure)
             frac_k_points, k_points_labels = kpath.get_kpoints(
@@ -1140,12 +1145,7 @@ class MPNonSCFSet(MPRelaxSet):
         else:
             self._config_dict["KPOINTS"]["reciprocal_density"] = \
                 self.reciprocal_density
-            kpoints = super().kpoints
-
-        # override pymatgen kpoints if provided
-        user_kpoints = self.kwargs.get("user_kpoints_settings", None)
-        if isinstance(user_kpoints, Kpoints):
-            kpoints = user_kpoints
+            return super().kpoints
 
         return kpoints
 
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -341,6 +341,15 @@ class DictSet(VaspInputSet):
         self.sym_prec = sym_prec
         self.international_monoclinic = international_monoclinic
 
+        if self.user_incar_settings.get("KSPACING") and user_kpoints_settings is not None:
+            warnings.warn(
+                "You have specified KSPACING and also supplied kpoints "
+                "settings. KSPACING only has effect when there is no "
+                "KPOINTS file. Since both settings were given, pymatgen"
+                "will generate a KPOINTS file and ignore KSPACING."
+                "Remove the `user_kpoints_settings` argument to enable KSPACING.",
+                BadInputSetWarning)
+
         if self.vdw:
             vdw_par = loadfn(str(MODULE_DIR / "vdW_parameters.yaml"))
             try:
@@ -385,6 +394,8 @@ class DictSet(VaspInputSet):
                     del settings[k]
                 except KeyError:
                     settings[k] = v
+            elif k == "KSPACING" and self.user_kpoints_settings != {}:
+                pass  # Ignore KSPACING if user_kpoints_settings are given
             else:
                 settings[k] = v
         structure = self.structure
@@ -465,7 +476,8 @@ class DictSet(VaspInputSet):
         # An error handler in Custodian is available to
         # correct overly large KSPACING values (small number of kpoints)
         # if necessary.
-        if "KSPACING" not in self.user_incar_settings.keys():
+        # if "KSPACING" not in self.user_incar_settings.keys():
+        if self.kpoints is not None:
             if np.product(self.kpoints.kpts) < 4 and incar.get("ISMEAR", 0) == -5:
                 incar["ISMEAR"] = 0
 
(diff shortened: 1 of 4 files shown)


pymatgen/io/vasp/sets.py
```diff
@@ -458,8 +458,21 @@ def incar(self) -> Incar:
...
-            incar["ISMEAR"] = 0
+        # Ensure adequate number of KPOINTS are present for the tetrahedron
+        # method (ISMEAR=-5). If KSPACING is in the INCAR file the number
+        # of kpoints is not known before calling VASP, but a warning is raised
+        # when the KSPACING value is > 0.5 (2 reciprocal Angstrom).
+        # An error handler in Custodian is available to
+        # correct overly large KSPACING values (small number of kpoints)
+        # if necessary. 
+        if "KSPACING" not in self.user_incar_settings.keys():
+            if np.product(self.kpoints.kpts) < 4 and incar.get("ISMEAR", 0) == -5:
```
mkhorton: Curious, why is this `np.product(self.kpoints.kpts)` and not `len(...)`?

pymatgen/io/vasp/sets.py
```diff
@@ -498,7 +511,7 @@ def nelect(self) -> float:
             return nelect
 
     @property
-    def kpoints(self) -> Kpoints:
+    def kpoints(self) -> Kpoints or None:
```
mkhorton: This is not a valid type annotation, it should be `Optional[Kpoints]`

pymatgen/io/vasp/sets.py
utf: If the user is relying on automatic k-point generation then `kpoints.kpts` will just be the mesh, e.g.., `[[4, 4, 4]]`.

This should probably be updated to handle the case where the list of k-points is supplied explicitly?

rkingsbury: I'm not sure @mkhorton ; that's what was there previously

pymatgen/io/vasp/sets.py
rkingsbury: changing to `Optional[Kpoints] is triggering a `NameError`

mkhorton: You may need to import `Optional` if it's not already imported, the `or` is definitely incorrect though (to my knowledge). The `Optional` syntax is equivalent to `Union[None, Kpoints]`.

pymatgen/io/vasp/sets.py
mkhorton: Ah, hence the product previously ... yes, agreed. To handle this robustly I think we'd have to both explicitly test number of k-points from the mesh if the user is using a mesh (testing `np.product(...) < 4` doesn't seem like it'd cover all cases? doesn't spglib have a function to give k-points from a mesh?) and then also for the length of explicit k-point lists as you mention.

In either case, just changing ISMEAR=-5 to ISMEAR=0 may not be the right call here, it may be better to warn the user and/or up the k-point density, it seems like the right action would be context dependent.

pymatgen/io/vasp/sets.py
rkingsbury: Any objection to using `Union[Kpoints,None]` instead? That's more explicit and `Union` is already imported

pymatgen/io/vasp/sets.py
rkingsbury: I'd be happy to refine this scheme but would prefer to do so under a separate PR, since it's not directly related to KSPACING. If you guys tell me what you think the logic should be I'll be glad to implement it.

rkingsbury:
@mkhorton and I realized that we need to somehow handle the case when a user specifies KSPACING but there is ALSO a KPOINTS file or settings present (in `user_kpoints_settings`).

In VASP, if KSPACING is set and a KPOINTS file exists, the KPOINTS file overrides KSPACING and there is apprently no error or warning generated. We now preserve this behavior so that if a user has specified KSPACING but also provided `user_kpoints_settings`, a warning is generated and KSPACING is ignored.

rkingsbury:
I cannot for the life of me figure out a good way to resolve the mypy type hinting error at line 1143 in `sets.py`.

rkingsbury:
I can't for the life of me figure out a good way to resolve the mypy TypeHinting error at 1143 in `sets.py`. It's caused by the fact that the internal variable `kpoints` may take on either `Kpoints` or `None` types, depending on the situation. I tried instantiating it as a TypeVar or adding a check, e.g.
```
if super().kpoints is not None:
    kpoints = super().kpoints
else:
    return None
```

without any luck. Both the superclass (DictSet) and the local kpoints method are hinted with `Union[Kpoints,None]`

pymatgen/io/vasp/sets.py
utf: `Optional[Kpoints]` is definitely the more pythonic.

rkingsbury: @utf my understanding is that `Optional` is really intended for the case of optional arguments that default to None. We don't have an optional argument here - None has the specific meaning that kpoints are being determine by VASP.

mkhorton: Ah, actually that's not the case. "Optional" has a specific meaning in type theory, https://en.wikipedia.org/wiki/Option_type It's not intended for "optional arguments" in the colloquial sense of the word

rkingsbury: Guess I shouldn't believe everything [I read on the internet](https://realpython.com/python-type-checking/#the-optional-type).

Sounds like Optional is preferred then?

rkingsbury:
Also a note for code review - there are a couple of input sets that implement their own `kpoints` property methods, and I have not touched these. The sets are MPHSEBSSet, MPNonSCFSet, MVLGWSet, MVLSlabSet, MVLGBSet, MITMDSet, and MPMDSet. Most of them inherit from DictSet.kpoints, but not all. As such, some may still behave unexpectedly when KSPACING is supplied.

I don't feel I know enough about the intended use of these input sets to modify the kpoints methods. Thoughts?

pymatgen/io/vasp/sets.py
mkhorton: Yeah, let's go with Optional. If it's helpful, consider that a function might *return* an Optional type. In this case there was no "optional argument," nothing was chosen by the user, it's the type itself that is optional. In terms of resources for this, PEP 484 is the canonical source of information, and that doesn't take an explicit stance.

pymatgen/io/vasp/tests/test_sets.py
```diff
@@ -389,6 +389,22 @@ def test_init(self):
...
         self.assertTrue(lcalcpol_vis.incar["LCALCPOL"])
 
+    def test_user_incar_kspacing(self):
+        # Make sure user KSPACING settings properly overrides KPOINTS.
+        si = self.get_structure('Si')
+        vis = MPRelaxSet(si, user_incar_settings={"KSPACING": 0.22})
+        self.assertEqual(vis.incar["KSPACING"], 0.22)
+        self.assertEqual(vis.kpoints, None)
+
+    def test_kspacing_overridce(self):
```
mkhorton: Type here

mkhorton:
This looks good. As far as I can tell, this is an additive feature, you're not changing defaults and not responsible for the default settings or behavior of other sets (which are overriding kpoints for good reason). As long as existing tests pass without modification that's fine. The mypy failures will have to be fixed however, we can investigate those further.

PR #1680, opened 2019-11-22
Fixing bugs with run_type

shyamd:
## Summary

GGA `run_type`s were not being properly detected. This fixes that bug.

* Properly detect GGA `run_type`s
* add tests for various `run_type`
* Moved BPARAM for rVV10 enforcement to MVL set


mkhorton:
Thanks @shyamd

mkhorton:
Can we discuss this more? This line:

```
        elif self.potcar_symbols[0].split()[0] == 'PAW':
```

is a little problematic because VASP has distributed (and continues to distribute) POTCARs that are actually PBE but have the "PAW" string (instead of the "PAW_PBE" string), meaning that it's possible we'll identify a PBE calculation as LDA with the current logic. This is very much not our fault, but we might need to handle it ...

PR #1695, opened 2019-12-03
Add pymatgen.Molecule support to AseAtomsAdaptor

smheidrich:
## Summary

* Added ``pymatgen.Molecule`` support to ``AseAtomsAdaptor``:
  * ``AseAtomsAdaptor.get_atoms()``now also accepts ``Molecule``s (previously, it would only accept ``Structure``s)
  * New (static) method ``AseAtomsAdaptor.get_molecule()`` (analogous to the existing ``get_structure()``method, but returns ``Molecule``s instead of ``Structure``s)
* Slightly extended tests

## Checklist

      Run [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/) and [flake8](http://flake8.pycqa.org/en/latest/) on your local machine.
  - [x] There weren't any before in the module this is about, so it should be fine if my stuff doesn't have any either. Probably better for a separate PR.
- [ ] All existing tests pass.
  - Only one way to find out for sure...

### DIFF
--- a/pymatgen/io/ase.py
+++ b/pymatgen/io/ase.py
@@ -14,7 +14,7 @@ __maintainer__ = "Shyue Ping Ong"
 __email__ = "shyuep@gmail.com"
 __date__ = "Mar 8, 2012"
 
-from pymatgen.core.structure import Structure
+from pymatgen.core.structure import Molecule, Structure
 
 try:
     from ase import Atoms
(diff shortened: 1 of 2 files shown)


shyuep:
Many thanks.

smheidrich:
Thanks for merging! :+1:

PR #1698, opened 2019-12-07
Check POTCAR hashes when creating Vasp Input Sets

rkingsbury:
When a VaspInput set is instantiated, check that the POTCAR hashes match the specified functional and POTCAR symbol (i.e., check that you are really using the POTCARs you intend). @mkhorton was consulted on this.

The check was implemented in `sets.py` rather than `input.py` because the `InputSets` specify their POTCARs by symbol (really filename) and functional, which uniquely identify the hashes they should have. There may be cases in which a user wants to instantiate a `Poscar` object directly from an arbitrary file on disk, in which case it wouldn't make sense to check its hash against anything.

md5 hashes were computed for the pseudopotentials distributed with VASP 5.4.4. We hashed both the entire POTCAR file and also the data itself (via the pymatgen `get_potcar_hash()` method. Both sets of hashes are provided as `.json` files. 

No new PSPs were added to the test_files. I copied an existing PSP for Fe that already existed in the test_files directory and made some small modifications to test the validation. However, I need some help to get these tests to work properly.


mkhorton:
Thanks @rkingsbury! This looks basically good, there may be a pycodestyle error in there. Only comment is that I'd recommend leaving the `symbol` property in the class, but have it return `self._symbol`. Then, in your `__init__` set `self._symbol` instead of `self.symbol`, this way there's still a docstring and the user is still informed that the "symbol" attribute exists and what it is.

mkhorton:
A code comment to explain how the .json files were generated where they're loaded would also be useful, since it may be non-obvious what they are and what the distinction is between a "file hash" and a "pymatgen hash".

rkingsbury:
> Thanks @rkingsbury! This looks basically good, there may be a pycodestyle error in there. Only comment is that I'd recommend leaving the `symbol` property in the class, but have it return `self._symbol`. Then, in your `__init__` set `self._symbol` instead of `self.symbol`, this way there's still a docstring and the user is still informed that the "symbol" attribute exists and what it is.

Thanks! See changes; I believe I implemented what you were asking in my latest batch of commits.

As for formatting, I took the somewhat extreme step of just formatting the entirety of these three files with black. I've become a fan of it recently, but unfortunately it doesn't allow you to only format a selection of text. The formatting changes are contained in a single commit, so let me know if it's a problem and I can revert.

> A code comment to explain how the .json files were generated where they're loaded would also be useful, since it may be non-obvious what they are and what the distinction is between a "file hash" and a "pymatgen hash".

See edits to the docstring of `validate_potcar_hash`

rkingsbury:
Looks like `analysis/gb/grain.py` is triggering a linting failure. I'm not sure why that's happening as I haven't touched that file.

mkhorton:
I should have fixed that, you might just need to merge changes from master
back into your branch. There were a few trivial missing docstrings in the
grain boundary module.

rkingsbury:
What is the proper way to fix `No library stub file for module 'pytest'`?

rkingsbury:
I'm also unsure how to interpret this error 

```
$ export MODIFIED_FILES=`git diff --diff-filter=ACMRTUXB --name-only HEAD v2019.10.16 | grep -E 'pymatgen.*\.(py)$' | sed '/test_/d' | tr '\n' ' '`

fatal: ambiguous argument 'v2019.10.16': unknown revision or path not in the working tree.
```

mkhorton:
I believe it's related to the previous issue, are you working on an outdated branch? Merge in changes from master?

rkingsbury:
> I believe it's related to the previous issue, are you working on an outdated branch? Merge in changes from master?

I did that prior to the last few commits, but just did a pull/rebase again and found more changes. Fingers crossed...

rkingsbury:
Same error, `No library stub file for module 'pytest'`

rkingsbury:
Moved definition of the `potcar_functional` into the .yaml file that defines each `InputSet` and renamed `potcar_functional` argument to `user_potcar_functional`. See [discussion](https://github.com/materialsproject/pymatgen/pull/1708#issuecomment-564779903)

rkingsbury:
I was able to silence the `No library stub file for module 'pytest'` error from mypy by adding `# type: ignore` to the import statement. I'm not sure whether this is an acceptable long-term fix or not. See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports

mkhorton:
The hashes .json files are still empty in this PR

rkingsbury:
> The hashes .json files are still empty in this PR

Are you sure @mkhorton ? I see the hash data if I click on Files Changed.

mkhorton:
image

Showing as empty to me ...

rkingsbury:
> image
> 
> Showing as empty to me ...

I still see the hashes when I open this PR in a private window (without logging in). I'm not sure what's going on. Can you try clearing cache, etc.?

image

mkhorton:
Thanks @rkingsbury !

PR #1705, opened 2019-12-11
Fix bug of missing sp2 N-N bond data

kwaters4:
Added data for an sp2 N-N bonds in bond_lengths.json
Use value of 1.25 from the following source
http://www.wiredchemist.com/chemistry/data/bond_energies_lengths.html

### DIFF
--- a/pymatgen/core/bond_lengths.json
+++ b/pymatgen/core/bond_lengths.json
@@ -162,7 +162,7 @@
     "bond_order": 1.0
   },
   {
-    "ref": "",
+    "ref": "http://www.wiredchemist.com/chemistry/data/bond_energies_lengths.html",
     "length": 1.25,
     "elements": [
       "N",
--- a/pymatgen/core/bond_lengths.json
+++ b/pymatgen/core/bond_lengths.json
@@ -162,7 +162,7 @@
     "bond_order": 1.0
   },
   {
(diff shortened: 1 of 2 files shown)


shyuep:
Can you add a small unittest in the existing unittests to make sure this additional data works as intended? There are already tests there - so it is just copying and pasting one or two lines and changing it to the N-N bond. Thanks.

pymatgen/core/bond_lengths.json
```diff
@@ -161,6 +161,15 @@
     ],
     "bond_order": 1.0
   },
+  {
+    "ref": "",
```
mkhorton: @kwaters4 Could you add the reference here? Reference is at the footer of the page you linked. Thanks!

kwaters4: Looking for it right now,one sec, looking for the right one. Added the wrong one in the last commit.

shyuep:
THanks!

mkhorton:
Hi @kwaters4, please feel welcome to add your name to [this form](https://forms.gle/JnisFb38QDR8QTFTA) so we can credit you appropriately for your contribution in the [pymatgen docs](https://pymatgen.org/team.html). We want to make sure we credit all contributors, even for small additions or bug fixes.

PR #1708, opened 2019-12-11
MP SCAN input set

rkingsbury:
## Summary

This is a new VASP input set for the SCAN metaGGA functional. Compared to previous input sets, we have adjusted a number of INCAR tags based on internal benchmarking of energy convergence, and utilized some modernized features of VASP such as KSPACING.

This input set is utilized in a forthcoming atomate workflow (see [PR](https://github.com/hackingmaterials/atomate/pull/345))

@mkhorton and @shyamd have been involved in the benchmarking and decision making regarding parameters.


shyuep:
Hmm... why don't you just edit the current input set rather than having yet another one?

rkingsbury:
Failing with `No library stub file for module 'pytest'

rkingsbury:
> Hmm... why don't you just edit the current input set rather than having yet another one?

@shyuep my thought (informed by my experience as I've learned pymatgen) was that it's more transparent to users to have a single .yaml file to look at that defines the key INCAR parameters, rather than having to follow multiple steps of inheritance that each modifies a few parameters. This can make it hard to understand what INCAR settings you are ultimately getting. 

While we could inherit from `MPRelaxSet` or `MVLScanRelaxSet`, we are changing enough of the parameters that we thought it more transparent to define a new set.

mkhorton:
This could also replace the current `MVLScanRelaxSet`

shyamd:
@shyuep  it's also important to have separation of concerns. Things related to MP, and data that will be on the website should be explicitly `MP blah blah`, with the expectation that they only get modified by group consent. This doesn't seem to be the case for the MIT or MVL parameters which have changed in the past and caused headaches for MP.

shyuep:
I would say we have no reason to have two different SCAN input sets unless there is a goal that they are substantially different. Just integrate the new parameters into the MVLScanSet for now. As far as I know, changes to either MIT or MVL sets are rare. In fact, changes to MP sets occur more frequently. And Matt and I are always gatekeepers to ensure no strange parameters are added.

mkhorton:
In the rare cases it happens, we should have a procedure in place. Unfortunately we cannot protect individual files on GitHub. We did have a serious issue with one of the MD sets a year or so ago (Prec set to Low and the relevant individuals were not notified/did not notice the change, resulting in a lot of lost compute time).

I'm not sure how best to avoid this -- we've discussed hashing the .yaml files, the key issue seems to be notifying the end user if the input set has changed. One idea could be to store the input set hashes in .pmgrc.yaml, and then if a set is initialized and the hash doesn't match the one previously stored, the user would be notified that the set has changed. This perhaps seems over-complex.

shyuep:
I think hashing is overly complex and rigid. There are parameters that we care about, and there are those that we don't or are more lenient on. E.g., we are fine with mixing ISMEAR=-5 and ISMEAR=0 runs for the most part depending on the kpoints. Custodian also modifies some inputs based on errors. Hashing would mean that even if say, VASP decides to add a new parameter to tune convergence and we add it to an input set, the hashes would be different.

I personally think the best way is to expand the compatibility engine. Instead of just PSP that is being checked right now, we will add parameters that we care about. Right now, I can't really think of what I would add. Maybe PREC, ENCUT and ISPIN for starters since those obviously will have an effect on comparability. U is treated as a special case since we are modifying energy values based on that.

shyuep:
Also, I am not aware of the PREC issue on the MD sets, but I would say that if this is a rare occurrence, we shouldn't be designing systems to handle it. I think the original MITMD set was indeed set to prec=low to save computing costs.

zbwang:
Hi all, the yaml file for MVLScanRelaxSet was constructed based on the PSP52 recommendations in vasp online manual, which is now no longer updated. Later the VASP developer removed `W_pv` "because of s ghoststate above E-fermi" and add `W_sv`. 

I hope you have noticed that there is no `W_sv` in PBE52. If preset `W_sv` in yaml for SCAN, there will be a problem when using PBE52 for W.  

For the MPSCAN default settings, I am not sure if you have been tested these two params
`ALGO = All ; LMIXTAU = True`. In my experience, these two params (suggested by SCAN developers) work much better in energy convergence than "ALGO = Fast/Normal". I would also suggest `ICHARG = 2` instead of 0/1 because I have seen convergence issues when using `ICHARG = 0/1`

shyuep:
Thanks @zbwang . 

@rkingsbury @mkhorton @shyamd I suggest we try to solve this problem once and for all. The input sets should define not just the symbol for each element, but also the PSP set it comes from. I.e., we add something in the yaml that says this is from PBE and that one is from PBE_52. This should apply to all prior input sets as well.

For the purposes of SCAN, I suggest we just move to PBE_54 and stick with that, unless we have a reason to keep PBE_52? You'd probably want to look into the parameters suggested by @zbwang as well. It is a new enough functional that I am not too worried about backward compatibility.

rkingsbury:
Thank you for those suggestions, @zbwang ! We will investigate.

I agree that defining the functional within the InputSets alongside the PSPs and INCAR parameters makes sense. For SCAN I think we're comfortable defaulting to PBE_54. If so, I can drop the `potcar_functional` argument so that the 54's are always used, and then we don't need to worry about toggling W_pv and W_sv.

shyuep:
Ok with me. I think you should keep the potcar_functional argument (but set it to default to None). The idea being that sometimes I just want to run something with an input set but I want to change the potcar functional. That should be allowed. But the default should be the one specified in the yaml itself.

rkingsbury:
> Ok with me. I think you should keep the potcar_functional argument (but set it to default to None). The idea being that sometimes I just want to run something with an input set but I want to change the potcar functional. That should be allowed. But the default should be the one specified in the yaml itself.

But if we allow changing of the potcar_functional, there is the potential for potcar symbols in the .yaml file not to match (e.g., W_pv vs. W_sv). Are you comfortable relying on the user to override the potcar symbols in those cases?

shyuep:
Users can do whatever they want.... The rule is that we specify defaults that we think makes the most sense for 99% of the population. For the 1% who wants to mess around with the potcar functional as well as the specific symbols, they can do so....

rkingsbury:
> Users can do whatever they want.... The rule is that we specify defaults that we think makes the most sense for 99% of the population. For the 1% who wants to mess around with the potcar functional as well as the specific symbols, they can do so....

OK, so to clarify what is being proposed (@shyuep correct me if I'm misunderstanding): 

- add a `POTCAR_FUNCTIONAL` key to each .yaml file
- retain `potcar_functional` as a kwarg in `DictSet` but default to `None`
- [Ryan's proposal]: rename `potcar_functional` to `user_potcar_functional` for consistency with other args that override defaults, like `user_kpoints_settings`.
- InputSets will read the value from the .yaml files unless the user specifies something else.

@shyuep since this would affect all InputSets I'd prefer to handle via a separate PR / branch. Ok?

shyuep:
Ok. Thanks!

shyamd:
@shyuep The issue with parameters changing might be a rare occurrence, but it can have significant consequences. The PREC change was part of a larger PR for other reasons and as a result, was not caught. The consequence was 13M CPU hours being lost. We're also putting SCAN on the MP Website. Thus there should be an MP input set for this. This is a simple sociological fix rather than a technical fix.

Everything else I like.

shyuep:
@shyamd The proposed solution does not solve the "rare problem" at all. It really doesn't matter whether something is called MVLScanSet or MPScanSet. In either of these, someone can make a change as part of a bigger PR and there is a non-zero probability that it does not get caught. If you really want to solve this problem, by all means fix the compatibility checks.

shyamd:
Compatability checks come after the calculations are finished, so they won't fix the issue of just wasting CPU time. Having an MP set gives @mkhorton something to watch out for. :>

rkingsbury:
@zbwang I'm incorporating your suggestions about `ALGO=ALL` and `ICHARG=2`. We'll likely add `LMIXTAU=TRUE` as well, but want to run a couple of benchmarks first (in progress now). 

We originally set `ICHARG=1` because our SCAN workflow employs a PBE preconditioning calculation, so we always start SCAN from a CHGCAR, but you make a good point that this set should be designed to facilitate standalone SCAN calculations. We can override ICHARG in our atomate workflow.

shyuep:
I am somewhat confused as to why you'd want to use PBE as a preconditioner. Is the PBE doing relaxation or just a static charge calculation?

rkingsbury:
> I am somewhat confused as to why you'd want to use PBE as a preconditioner. Is the PBE doing relaxation or just a static charge calculation?

@shyuep We've tried workflows with both a PBE static and a PBE relaxation prior to SCAN, but have settled on using a PBE relaxation as our first step. It's main purpose is to accelerate convergence by initializing the wavefunctions.

shyuep:
I would argue you should not use a PBE relaxation unless your results show that it is indeed faster. SCAN gets lattice parameters close to the expt. PBE does not.

In fact, our experience with HSE is that if you have expt lattice parameters, directly computing from the expt structure is usually faster than relaxing with PBE to wrong lattice parameters and then re-relaxing it with HSE.

Of course, this assumes you are starting from expt structures. I would actually argue for a PBE static -> SCAN low -> SCAN high sequence of runs.

mkhorton:
@shyuep I'm not commenting on the workflow here, but just a note on SCAN itself: convergence is nasty, and as currently formulated it's numerically badly conditioned. I'm told that *internally* inside VASP they generate a PBE wavefunction first before generating the SCAN one, and this is at every SCF step resulting in a ~30% performance penalty. But overall convergence is also bad enough that pre-conditioning with a self-consistent PBE WAVECAR is itself necessary (whether PBE *specifically* is important I couldn't say).

rkingsbury:
> I would argue you should not use a PBE relaxation unless your results show that it is indeed faster. SCAN gets lattice parameters close to the expt. PBE does not.
> 
> In fact, our experience with HSE is that if you have expt lattice parameters, directly computing from the expt structure is usually faster than relaxing with PBE to wrong lattice parameters and then re-relaxing it with HSE.
> 
> Of course, this assumes you are starting from expt structures. I would actually argue for a PBE static -> SCAN low -> SCAN high sequence of runs.

We thought so too. We did a series of benchmarks comparing PBE > SCAN > SCAN workflows, and found that the second SCAN relaxation changed the output very little, hence the two step PBE>SCAN that we plan to adopt. That said, we are generally starting from PBE relaxed structures from MP rather than experimental ones.

rkingsbury:
> > Users can do whatever they want.... The rule is that we specify defaults that we think makes the most sense for 99% of the population. For the 1% who wants to mess around with the potcar functional as well as the specific symbols, they can do so....
> 
> OK, so to clarify what is being proposed (@shyuep correct me if I'm misunderstanding):
> 
>     * add a `POTCAR_FUNCTIONAL` key to each .yaml file
> 
>     * retain `potcar_functional` as a kwarg in `DictSet` but default to `None`
> 
>     * [Ryan's proposal]: rename `potcar_functional` to `user_potcar_functional` for consistency with other args that override defaults, like `user_kpoints_settings`.
> 
>     * InputSets will read the value from the .yaml files unless the user specifies something else.
> 
> 
> @shyuep since this would affect all InputSets I'd prefer to handle via a separate PR / branch. Ok?


> Ok. Thanks!

@shyuep I've implemented this as part of #1698

rkingsbury:
@shyuep starting from some code contributed by @mkhorton , I added a unittest that checks the hashes of all `InputSet` .yaml files to address some of the concerns raised above about changing parameters. The test will fail if anything in the .yaml changes, and raise a warning that the user needs to contact you or @mkhorton about the change.

mkhorton:
Just to clarify, this new hash is *not* something that runs at run-time, it's purely for CI -- I wanted an explicit test that would fail if someone changes an input set .yaml file, to make it more likely we would catch a change if it was part of a large PR.

rkingsbury:
Well, I can confirm that the new `SetChangeCheckTest` works! This is a good safeguard.

rkingsbury:
@shyuep @mkhorton we've compelted our internal discussions on this and it is ready for review.

mkhorton:
This looks good to me, thanks @rkingsbury

PR #1715, opened 2019-12-16
Add fast `Structure.get_neighbor_list` method

chc273:
## Summary

Added a Structure method to construct neighbor list without using Neighbor class. Previously this is achieved by using `get_all_neighbors` method, which is much slower than the proposed methods.  It may become a problem when the structure is very large, e.g., >10,000 atoms.

Here I am adding a new method `get_neighbor_list` to `Structure`, that returns the full information about neighbors without the use of many python native objects. This new method returns a tuple of numpy array `(center_indices, points_indices, offset_vectors, distances)` as the neighbors. Here atom `center_indices[i]` has neighbor atom index `points_indices[i]` that is translated by `offset_vectors[i]` lattice vectors, and the distance between them is `distances[i]`.  

For a Mo structure with atom numbers of `1000` ([10, 10, 10] supercell), `10,000` ([10, 10, 100] supercell) and `100,000 ` ([10, 10, 1000] supercells), constructing the neighbor list of radius 4 in this way takes about `0.04`, `0.24` and `2.37` seconds, respectively using 2.7 GHz Dual-Core Intel Core i5.  `get_all_neighbors` (not including the time it takes to convert neighbors to numpy format) is about five times slower.  I think this method is highly needed for high-performance materials computations, for example, in the construction of force fields, machine learning, etc. 

- [Y] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
      on your local machine.
- [Y] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [Y ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) 
      to type check your code.
- [ Y] Tests have been added for any new functionality or bug fixes.
- [ Y] All existing tests pass.

allowing commits.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -1230,8 +1230,8 @@ class IStructure(SiteCollection, MSONable):
                               numerical_tol: float = 1e-8,
                               exclude_self: bool = True) -> Tuple[np.ndarray, ...]:
         """
-        A python version of getting neighbor_list. The returned values are a list of
-        numpy arrays [center_indices, points_indices, offset_vectors, distances].
+        A python version of getting neighbor_list. The returned values are a tuple of
+        numpy arrays (center_indices, points_indices, offset_vectors, distances).
         Atom `center_indices[i]` has neighbor atom `points_indices[i]` that is
         translated by `offset_vectors[i]` lattice vectors, and the distance is
         `distances[i]`.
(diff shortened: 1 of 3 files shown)


mkhorton:
Fantastic! Thanks @chc273 

Could you comment a little on the change in the C++ compile args? Is this going to be any easier/more difficult to compile than previously?

chc273:
@mkhorton `c++0x` is an older standard than `c++11`. Some supercomputer platforms (like tscc we are using) are still using the old gcc version that is not compatible with `c++11`. The code I wrote actually does not use new c++ features, and hence the compilation flags can be set to older c++ standards without affecting the results.

mkhorton:
Ok, sounds good

PR #1716, opened 2019-12-18
determine kspacing and smearing from bandgap in MPScanRelaxSet

rkingsbury:
This morning @mkhorton @shyamd @kristinpersson and I decided to compute the KSPACING value based on the bandgap of the input material in our SCAN workflow. At the time I thought this functionality would be part of atomate, but later learned it needs to be in MPScanRelaxSet.

I also identified and corrected a potential bug / unintended behavior when a user tries to override KSPACING, ISMEAR, or SIGMA.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -803,6 +803,9 @@ class MPScanRelaxSet(DictSet):
         else:
             rmin = 25.22 - 1.87 * bandgap  # Eq. 25
             kspacing = 2 * np.pi * 1.0265 / (rmin - 1.0183)  # Eq. 29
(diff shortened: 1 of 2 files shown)


rkingsbury:
@mkhorton @shyamd this is how E_gap maps to KSPACING. The VASP default value is 0.5. Should we cap the value at 0.5, or allow it to increase beyond that?  (Would change the k-point density of materials with Egap > 6 eV)

image

shyamd:
I think we'd bound it. Anything over say 0.5 eV, we pin it to 0.44 all the way down to 0.22 for metallic?

mkhorton:
Thanks @rkingsbury !

Issue #1717, opened 2019-12-18
Reaction Calculator -- better stoichiometry coefficient balancing


mattmcdermott:
The current Reaction class’s stoichiometry/balancing code allows for the automatic shuffling of reactants/products to opposite sides -- or even deleting the reactants/products entirely. While this is occasionally a nice feature for underdetermined systems, it can lead to undesirable/unexpected solutions. For example, when we run:

```
reactants = ["Mn(HO)2", "Li2CO3", "Ni(HO)2", "O2"]
products = ["Li2MnO3", "Li2MnNiO4", "CO2", "H2O"]
reactants = [Composition(r) for r in reactants]
products = [Composition(p) for p in products]

rxn = Reaction(reactants, products)
print(rxn) 
```

We get the following reaction:

_2 Mn(HO)2 + 2 Li2CO3 + Ni(HO)2 + O2 -> Li2MnO3 + Li2MnNiO4 + 2 CO2 + 3 H2O_

However, when we simply reverse the reaction via `Reaction(products, reactants)`, we get:

_Li2MnNiO4 + CO2 + 2 H2O -> Mn(HO)2 + Li2CO3 + Ni(HO)2 + 0.5 O2_

In the above example, Li2MnO3 is unexpectedly removed, just by simply changing the direction of the reaction. Albeit not that common, I feel like this is very counterintuitive result to a user.

To fix this issue, I suggest we change the starting constraints and iterate until we find the “best” solution, i.e. the solution which causes the fewest number of swaps/deletions. I am submitting a PR with a proposed fix!

mattmcdermott:
Closed with commit #1718.

PR #1718, opened 2019-12-18
Reaction Calculator -- better stoichiometry/balancing

mattmcdermott:
This PR references the issue described in #1717: i.e. that the stoichiometric coefficient solver in the  _Reaction_ class often finds solutions which are unexpected/undesirable.

The fix adds tracking for the `num_errors`, which counts how many components flip sides or are deleted entirely. The `_balance_coeffs()` method loops through different combinations of constraints and keeps track of the best solution (that which minimizes the number of errors).

Note that I use _pinv_ from numpy instead of _lstsq_, however the code functions similarly and is a little easier to follow I think. 

Also note that I altered a few tests that were enforcing less than optimal solutions, and added the solutions found by the new method.

Regarding possible increase in computational time -- I am already using this code for high-throughput generation of hundreds of thousands of reactions and it seems to be roughly equivalent timing wise for reactions which have 1,2 reactants going to 1,2 products (this might not hold for much larger reactions).


mkhorton:
This looks excellent, thanks @mattmcdermott !

Before merging could you comment a little on the test changes, e.g. is `test_singular_case` still testing a singular case? It's not clear to me how the old test was correct in any case.

mattmcdermott:
@mkhorton Yeah! All changes I made to the tests were because the previous solutions were either not optimal, or in some cases, not even given (i.e. components were changing sides or being deleted more than necessary).

It does seems weird in general to create a test which enforces a specific solution when many exist, as in the case of underdetermined systems. At least now though we know it is looking for the best solution, instead of just what is found first when iterating through possible constraints.

As for the singular case -- I think the original composition matrix is singular, but as far as I know, this isn't really that special of a case. The Moore-Penrose pseudoinverse will always be defined and unique so it shouldn't really matter.

mkhorton:
Ok, that makes sense. Thanks for the additional info!

mkhorton:
Hey @mattmcdermott, not sure why this wasn't caught in CI but there are some missing docstrings that should be added to the ReactionCalculator: https://travis-ci.org/materialsproject/pymatgen/builds/627925624#L2734

I understand these are not methods you've added, but whenever you touch a file that causes the doc checks for that file to be run.

mattmcdermott:
@mkhorton Haha I was wondering about that... thought it was weird that those old functions had none. Should I add docstrings to all the functions? And should I open a new PR or can we tag it into this?

mkhorton:
You should open a PR since this has already been merged now

mkhorton:
Hi @mattmcdermott I noticed you were missing from our developer list, please fill out [this form](https://forms.gle/JnisFb38QDR8QTFTA) so we can credit you appropriately in the [pymatgen docs](https://pymatgen.org/team.html).

mattmcdermott:
@mkhorton Thanks so much -- done! Also I haven't forgotten about the docstrings, it is on my to-do :)

PR #1731, opened 2020-01-04
add potcar_spec option to write_input

rkingsbury:
## Summary

Adds a boolean `potcar_spec` option to the `write_input` method of a `VaspInputSet`. When this option is enabled, `write_input` generates a POTCAR.spec file rather than a POTCAR file (similar to the existing `write_spec` method, which outputs a .zip file containing all the inputs.) The purpose of this option is to facilitate sharing of inputs and automated testing in environments where POTCAR files are not available due to licensing. 

In this particular case, I developed some tests for a new atomate workflow in which the input set utilized POTCAR files that are not distributed with pymatgen (see https://github.com/hackingmaterials/atomate/pull/345). It was not possible for the CI testing to call `write_input` on my inputset because of the missing POTCARs. A similar issue affects some of the Ferroelectric workflow tests in atomate.

With this new option in pymatgen, a powerup can be added to atomate to enable the `potcar_spec` option in unittests and avoid problems with missing POTCARs. The existing `write_spec` method is less desirable for this because it outputs a zip file rather than just writing the files into a directory.

@utf and @computron were consulted on this solution.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -177,7 +177,7 @@ class VaspInputSet(MSONable, metaclass=abc.ABCMeta):
                 not have a license to specific Potcar files. Given a "POTCAR.spec",
                 the specific POTCAR file can be re-generated using pymatgen with the
                 "generate_potcar" function in the pymatgen CLI.
-            zip_output (bool): If True, output will be zipped into a file with the 
+            zip_output (bool): If True, output will be zipped into a file with the
                 same name as the InputSet (e.g., MPStaticSet.zip)
         """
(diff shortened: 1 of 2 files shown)


mkhorton:
How about you remove the existing `write_spec` method (which hasn't been there long in any case) and add a `zip` flag to `write_input` also?

The `write_spec` was added to formalize some functionality that previously existed in Materials Project, but was only present in the website code, and since pymatgen has an option to reconstruct from a spec (via its command-line interface) it seemed to make sense to have it here.

mkhorton:
Otherwise this looks good.

mkhorton:
Great, thanks!

PR #1741, opened 2020-01-22
Isym uniform bug fix

fraricci:
As discussed in issue #1453, ISYM has to be set to 1 in case of uniform mode in MPNonSCFSet.
This pull request should fix the problem.


utf:
Thanks @fraricci! Can we use ISYM = 2 as this should be more efficient. See https://www.vasp.at/wiki/index.php/ISYM

shyuep:
I am fine with this, but please implement tests.

fraricci:
@utf , thanks for the suggestion, that's preferable indeed.
@shyuep test on isym added as well.

mkhorton:
Thanks @fraricci !

PR #1742, opened 2020-01-24
Small fix in connected components

davidwaroquiers:
* Fixing a small error in serialization of connected components.


mkhorton:
Thanks @davidwaroquiers

PR #1743, opened 2020-01-26
Update xas

yimingchen95:
## Summary

Update XAS module with EXAFS module and stitching code
* Add EXAFS module 
* Add stitching method for XANES and EXAFS
* Add stitching method for L2 and L3 edge XANES

## TODO (if any)

None

## Checklist

in the pull request title.

- [ ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
      on your local machine.
      to type check your code.
- [ ] All existing tests pass.

allowing commits.


mkhorton:
Thanks @yimingchen95 !

@shyamd is there anything else you need from this for the builders/api?

shyuep:
I don't think we need separate classes for XANES and EXAFS, since both have the same init. I think you can add a name arg to the XAS to allow you to specify whether it is K,L XANES or EXAFS.

Then, what you need is as XAS.stitch(other) method that does the stitching, rather than two individual methods that violate OOP for no reason. I do not quite understand why the stiching is different for XANES+EXAFS vs L2 + L3. Is it simply because one is an addition? If so, Spectrum already implements linear addition? Surely we can also enable interpolated addition?

yimingchen95:
I agree to merge the XANES and EXAFS module, which makes things more concise, and also move stitching methods inside class. 

Yes, while Spectrum has the linear addition and interpolated addition is doable, things are bit more complicated for stitching L2+L3. This is because fill value needs to be added to prevent sudden change in mu. This fill value when x-axis is out of range is more tricky and less common. Hence I would like to have another method for stitching L2 and L3.

For stitching XANES+EXAFS, extra math are required and this is unique for this process.

pymatgen/analysis/xas/spectrum.py
```diff
@@ -2,57 +2,178 @@
...
-            self.absorption_specie, self.edge,
+        return "%s %s Edge %s for %s: %s" % (
+            self.absorption_element, self.edge, self.spectrum_type,
             self.structure.composition.reduced_formula,
-            super().__str__()
+            super(XAS, self).__str__()
         )
+
+    @classmethod
+    def stitch_xanes_exafs(cls, xanes, exafs, num_samples=500):
```
shyuep: This should not be a class method. It should simply be XAS.stitch(self, other, num_samples=500). I would prefer we don't have two different stitch methods. You can have a parameter called mode that defines whether it is the XANES/EXAFS or the L23.

shyamd: You should be able to detect what to do based on edge, absorption element, and type automatically correct? If the user provides a set of spectra that are not valid, then throw an error.

shyuep: Technically yes, but I can see different requirements. E.g., some people may want to stitch in different ways, etc. Right now, there is no difference in XANES and EXAFS in terms of class. So other than the spectrum_type, there is nothing to distinguish them. I would hesitate to try to determine the stitching automatically based on the spectrum_type and let the user be explicit, with a sensible default.

shyuep:
Also, pls have type hints for all the methods.

pymatgen/analysis/xas/spectrum.py
shyamd: I'm pretty sure we're the only ones that have the data to perform stitching, so it might be useful to have `stitch` be automatic and then have `stitch_` variants for the different methods. `stitch` should also do some sanity checking that the user isn't trying to stitch a K-edge XANES with a M-edge EXAFS for instance.

yimingchen95: I agree that we might be the only ones to have data for stitching, but whole XAS class is not only about stitching. I believe a more general usage is to simply obtain those already implemented attributes/methods while automatic stitching could delay such processes and provide data more than enough. 
Thus, I would keep `stitch` a separate method inside XAS class with modes for XANES+EXAFS and L23 as well as more sanity check.

pymatgen/analysis/xas/spectrum.py
```diff
@@ -2,57 +2,195 @@
...
+                 spectrum_type="XANES"):
+        """
+        Initializes a spectrum object.
+        """
+
+        super(XAS, self).__init__(x, y, structure, absorption_element,
+                                  edge)
         self.structure = structure
-        self.absorption_specie = absorption_specie
+        self.absorption_element = absorption_element
```
shyamd: Can we change this to `absorbing_element`?

yimingchen95: I already changed absorption_specie to absorbing_element because usually oxidation states are not known so absorbing_element is a more accurate way to describe the x-ray absorption process.

shyamd: You have absor**ption**_element right now, and I'd prefer it if it was absor**bing**_element: switching from "ption" to "bing" which is just a bit clearer. It is a bit pedantic, but might as well take care of it now.

yimingchen95: Thanks for suggestion. Just renamed it.

shyuep:
Thanks.

Issue #1746, opened 2020-01-28
POTCAR file hashes appear to be incorrect


bjmorgan:
The `vasp_potcar_file_hashes.json` ([here](https://github.com/materialsproject/pymatgen/blob/master/pymatgen/io/vasp/vasp_potcar_file_hashes.json)) is supposed to contain md5 hashes of POTCAR files (if I have this understood correctly). But the hashes in the file do not match the hashes generated from `md5sum POTCAR` from the command line.

e.g. using the PBE_54 Ac POTCAR
```python
from pymatgen.io.vasp.inputs import PotcarSingle

psingle = PotcarSingle.from_file(filename='POTCAR_PBE54_Ac')

print(psingle.hash)
# note: this hash is included in the vasp_potcar_pymatgen_hashes.json file

from hashlib import md5
print(md5(psingle.data.encode("utf-8")).hexdigest())
# This is the same hash generated using `md5sum POTCAR` from the command line.
# it is *not* included in the vasp_potcar_file_hashes.json file
```
produces
```
4175a0a39300f5c90c84142814f8cc72
cb0e0b98d520a284cf11d1161f25e6a5
```
The first hash is included in the `vasp_potcar_pymatgen_hashes.json` file:
```
grep -c 4175a0a39300f5c90c84142814f8cc72 vasp_potcar_pymatgen_hashes.json 
1
```
The second hash is *not* included in the `vasp_potcar_file_hashes.json` file:
```
grep -c cb0e0b98d520a284cf11d1161f25e6a5 vasp_potcar_file_hashes.json 
0
```

Note: this is using the "original unvie release version" PBE_54 pseudopotential files.

Using the updated PBE_54 pseudopotentials gives:
```
md5 POTCAR 
MD5 (POTCAR) = 2837362d49736e188edb5cc8d2a39c71
grep -c 2837362d49736e188edb5cc8d2a39c71 vasp_potcar_file_hashes.json 
0
```

shyuep:
Thanks for reporting this.
@mkhorton @shyamd I honestly dislike the file hash approach to dealing with POTCARs. Should we just do away with this? I think basic checks such as the functional, date and parameters would suffice?

shyamd:
These aren't supposed to be file hashes, but rather data hashes, so this is the expected behavior.

bjmorgan:
[...]

PR #1749, opened 2020-01-28
update POTCAR hashes

rkingsbury:
Addresses #1746 

Hashes have been recomputed using the complete set of POTCARs distributed with VASP 5.4.4 on Cori.


mkhorton:
thanks @rkingsbury, can you double-check that *all* the potcar hash in the .json file I provided way back are now included in this? because that should have been almost complete

Issue #1665, opened 2019-11-13
non symmetric structure matching


jmmshn:
**Describe the bug**
structure_A == structure_B does not mean structure_B == structure_A
This also results in the `group_structure` function returning different results depending on the order.

```python3
from pymatgen.analysis.structure_matcher import StructureMatcher, ElementComparator
from pymatgen import Structure
sm = StructureMatcher(comparator=ElementComparator(), #ignored_species=['Li'],
    ltol=0.6,
    stol=0.6,
    angle_tol=6,)
[...]

PR #1763, opened 2020-02-10
Minor fixes to structure mactcher and volumetric data

jmmshn:
- Allow the fit function in structure matcher to be symmetric
- Updated the from_hdf5 function so that inherited classes can use it more readily.
- Changed the way Chgcar objects are initialized so work with the above change


mkhorton:
References #1665

PR #1769, opened 2020-02-18
Replace deprecated modules

richardtran415:
Replace all deprecated use of scipy for linear algebra with numpy in wulff.py.

@mkhorton I think this should fix the noisy py3test issue on CircleCi, as I didn't see the previous messages from my tests flooding CircleCi anymore, but I did see some errors in things like test_molecule_matcher.py, though I don't think that is related to this pull right?


mkhorton:
Thanks!

Yes, I think they're unrelated -- I was trying to track those down too but the noisy tests were making it difficult, so thanks for fixing :-)

PR #1771, opened 2020-02-20
[WIP] Implement units using pint

ayushsgupta:
The goal of this project is to replace the current pymatgen unit system with that of pint, which will allow for more comprehensive and robust unit handling

* Replace all instances of FloatWithUnit with the Quantity class from pint
* Calculations in pymatgen use Quantity.magnitude
* TODO: Lots of tests still fail, will be fixed

* Pint - whole project based on the Quantity class from pint


mkhorton:
Thanks @ayushsgupta, this looks like a good start but there's a lot of items that will need to be addressed:

* you should move the unit registry out of the root pymatgen init, this should live in pymatgen.core.units

* this will be *much* easier to merge if it's two separate pull requests: the first pull request simply replaces current usage of units in pymatgen, the second pull request adds unit functionality to new modules -- this second case will require more discussion

* in the second case, care should be taken to make sure this does not adversely affect performance

* it's not clear to me why ".magnitude" has to be used in so many places, since math can be performed on pint Quantity objects directly

* I understand this is a "work in progress / WIP" PR so it's not a problem right now, but there are a lot of print statements and commented out blocks here -- as a guideline, these should never be committed to start with, it's easier to keep commits clean than to go back and tidy them later

Thanks for your work on this!

rkingsbury:
Thanks @mkhorton , we know this still has a ways to go. I asked @ayushsgupta to go ahead and open a WIP PR to facilitate feedback as his work progresses. 

Re: bullet 2 - This was our initial goal, but due to some idiosyncratic differences between pymatgen's `XXWithUnit` classes and `Quantity` objects, it wasn't as straightforward where to draw the boundary as we expected. We were trying to completely eliminate, e.g., `FloatWithUnit` rather than have to write a wrapper around `Quantity`, which is how we wound up with a bit deeper integration than we first expected. But I agree it makes sense to have separate PRs for the initial migration to pint vs. a more comprehensive integration. What exactly do you have in mind when you say "replaces current usage in pymatgen"?

Re: point 3 - Ayush has done some performance testing and so far it's a mixed bag - certain methods are slower but a few are somehow faster with Quantity!

Re: point 4 - you are right that math can be done on `Quantity` objects directly, but don't we need to be concerned about breaking existing scripts that are expecting floats? We consider using .magnitude the shortest, safest path to getting the initial `Quantity` support working.

pymatgen/__init__.py
```diff
@@ -42,6 +41,12 @@ def _load_pmg_settings():
 
 SETTINGS = _load_pmg_settings()
 
+from pint import UnitRegistry
+ureg = UnitRegistry()
+ureg.define('@alias angstrom = ang') # Need a more permanent solution - create own registry.txt
```
mkhorton: +1 for creating our own registry definition file

mkhorton:
> I asked @ayushsgupta to go ahead and open a WIP PR to facilitate feedback as his work progresses.

Excellent -- definitely appreciate the open PR, I agree it makes feedback a lot easier!

> What exactly do you have in mind when you say "replaces current usage in pymatgen"?

I meant one PR where it replaces current `FloatWithUnit` usage (etc.), and a second PR if you want to add it to any new classes.

> certain methods are slower but a few are somehow faster with Quantity!

Really interested in what the faster ones might be! Could reveal some easy performance optimizations we could make.

> We consider using .magnitude the shortest, safest path to getting the initial Quantity support working.

Possibly -- I think it just makes the code a lot more difficult to parse. If you have to get the magnitude conversion, you could convert relevant quantities at the top of the function/method/wherever rather than inline. Also make sure to check [this page](https://pint.readthedocs.io/en/0.11/wrapping.html) in the pint docs which might be useful, as well as the second section [here](https://pint.readthedocs.io/en/0.11/performance.html).

>  but don't we need to be concerned about breaking existing scripts that are expecting floats? 

Definitely concerned, I'm not entirely sure though. Maybe there's a clean way to do it -- I'm interested to see how this PR turns out. It could definitely break type hints too, depending on implementation. I'm optimistic we can find a good solution though.

ayushsgupta:
@rkingsbury feel free to correct me on anything I say here or if you have something else to add. @mkhorton thanks for the feedback. 
> I meant one PR where it replaces current `FloatWithUnit` usage (etc.), and a second PR if you want to add it to any new classes.

There seem to be 3 steps to integrating Quantity into pymatgen:
1. Replacing all instances of FloatWithUnit with Quantity.magnitude to preserve float return type
2. Have functions return Quantity objects instead of FloatWithUnit (or float)
3. Using Quantity for all calculations in pymatgen, beyond where FloatWithUnit was used

To my knowledge I've completed (1), and have done (2) on files that Ryan and I identified as important, but I haven't done (3). To clarify, do you mean that it's a good idea to separate (1) and (2) into their own pull requests? The reason I went ahead and started on (2) right after doing (1) is because (1) is simply replacing a FloatWithUnit instance with a Quantity, and whenever a calculation was done on that FloatWithUnit, use Quantity.magnitude. That would defeat the purpose of the Quantity class, so classes in pymatgen/core (see Element) return Quantity now. Quantity.magnitude is called in other places because we didn't want to touch the return types of other modules because of the problems that would cause. However, I recognize that my whole approach here was inconsistent, so I'll fix it.

> Also make sure to check [this page](https://pint.readthedocs.io/en/0.11/wrapping.html) in the pint docs which might be useful, as well as the second section [here](https://pint.readthedocs.io/en/0.11/performance.html).

I will check it out, thanks.

> Really interested in what the faster ones might be! Could reveal some easy performance optimizations we could make.

File | Time_Q | Time_O | Std_Dev_Q | Std_Dev_O
-- | -- | -- | -- | --
analysis/elastic | 89 | 90 | 4.35 | 2.37
analysis/local_env | 112 | 119 | 9.98 | 8.18
analysis/nmr | 2.71 | 2.25 | 1.45E-04 | 5.35E-05
analysis/quasiharmonic | 3.16 | 2.64 | 1.69E-04 | 7.66E-05
core/lattice | 6.53 | 5.7 | 7.18E-05 | 1.14E-04
core/periodic_table | 6.66 | 6.12 | 1.31E-04 | 2.66E-04
core/structure | 120 | 121 | 3.94E+00 | 5.75E+00
core/surface | 84 | 102 | 2.71 | 6.07
entries/compatibility | 3.29 | 2.45 | 1.45E-04 | 4.39E-05
entries/computed_entry | 3.86 | 3.8 | 4.23E-04 | 1.10E-04

This is the timing data for 10 files I chose because I had modified them to use Quantity instead of FloatWithUnit. I used %timeit from jupyter on the test files, 7 trials each. Q = Quantity, O = original, everything's in seconds.

shyuep:
Just a note that breaking existing scripts is NOT an option. I would rather have whatever suboptimal implementation in pymatgen currently that is completely compatible with normal Float use, than to try to replace it with pint and break everything expecting a float. Note that most people do not even care that the things are properly unitized and they just want the float value and they can deal with units separately themselves.

mkhorton:
Yes, agreed, requiring a type cast to float after every functional call is clearly not an acceptable outcome. But I'm optimistic we can find a good solution, this PR is intended to explore the idea of what it could look like.

> Note that most people do not even care that the things are properly unitized and they just want the float value and they can deal with units separately themselves.

I don't really agree with this, I think we can be *much* better as a community with unit handling. Angstroms and eVs are fine, but anything more complex than that is how mistakes are made (and I've seen eV vs eV/atom mistakes myself, that would be caught by a more rigorous unit system).

mkhorton:
Or, I think you're right that most people probably don't care, but I think we're in a position we can push to raise the bar a little on this.

shyuep:
Just to clarify, it is not that I don't think units are useful. They are. I cannot recall how many times I have to do the eV/atom to J/mol conversion. A proper unit system makes this a lot more painless.

What I meant is that most people just take the float value and run with it. Because most people work with whatever existing unit system is from the output. The only time units matter is when you have to compute the capacity in mAh/g or something that an experimentalist uses. So an unacceptable outcome is that all outputs from pymatgen require someone to do `.magnitude` or `float(x)` to use. 

If you can solve this, I am fine with moving to pint. Otherwise, there is no value proposition from pint that would convince me it is a good idea.

rkingsbury:
Thanks @ayushsgupta for digging into the details of this; we're eager to hear everyone's thoughts on the best way forward. Here is a synopsis of the challenge we face:

#### Difference between `Quantity` and `FloatWithUnit`

Part of the reason for so many internal calls to `.magnitude` is that pint `Quantity` has stricter units handling than `FloatWithUnit`. For example, a `FloatWithUnit` will allow you to add a regular float (which drops the unit and changes the return type to a regular float), whereas `Quantity` will not. 

`FloatWithUnit` allows arithmetic with heterogeneous units (regular floats) by dropping units information:

```
>>> from pymatgen.core.units import FloatWithUnit
>>> e = FloatWithUnit(1.1, "eV")
>>> print(e)
1.1 eV
>>> e + 2
3.1
```

`Quantity` only allows math in consistent units:
```
>>> from pint import UnitRegistry
>>> ureg = UnitRegistry()
>>> Q=ureg.Quantity
>>> e = Q(1, 'eV')
>>> print(e)
1 electron_volt
>>> e+2
---------------------------------------------------------------------------
DimensionalityError                       Traceback (most recent call last)
~/pymatgen/pymatgen/entries/compatibility.py in 
      4 e = Q(1, 'eV')
      5 print(e)
----> 6 e+2

~/miniconda3/envs/mp/lib/python3.6/site-packages/pint/quantity.py in __add__(self, other)
    752             return self.to_timedelta() + other
    753         else:
--> 754             return self._add_sub(other, operator.add)
    755 
    756     __radd__ = __add__

~/miniconda3/envs/mp/lib/python3.6/site-packages/pint/quantity.py in wrapped(self, *args, **kwargs)
     73         elif isinstance(other, list) and isinstance(other[0], type(self)):
     74             return NotImplemented
---> 75         result = f(self, *args, **kwargs)
     76         return result
     77     return wrapped

~/miniconda3/envs/mp/lib/python3.6/site-packages/pint/quantity.py in _add_sub(self, other, op)
    663                                _to_magnitude(other, self.force_ndarray))
    664             else:
--> 665                 raise DimensionalityError(self._units, 'dimensionless')
    666             return self.__class__(magnitude, units)
    667 

DimensionalityError: Cannot convert from 'electron_volt' to 'dimensionless'
```

I think the more robust approach taken by `pint` will definitely benefit pymatgen in the long run, but we'll have to think carefully about the best way to make the transition without breaking too many existing scripts. That said, in my opinion we *should* be breaking existing scripts that only work because they ignore units. For example, Ayush discovered that pymatgen's `CostEntry` class actually uses the `.energy` attribute to store cost information, so you are doing computations with a "cost" in energy units even though it's supposed to be $/mol. These types of inconsistencies will only become visible when we have something like pint tracking all the calculations and those inconsistent things start to break. We can of course use deprecation warnings and other means to transition gradually though.

#### Units in pymatgen

In the pymatgen codebase, most of the methods that return `FloatWithUnit` are in the `periodic_table` module, which is heavily utilized throughout the rest of pymatgen. So clearly these methods are intended to pass units information to downstream methods. However, almost nothing calls the `._unit` attribute of `FloatWithUnit`, so the returned values are more or less behaving as floats. 

Given the above difference between `Quantity` and `FloatWithUnit`, we have a couple of options for how to migrate to pint:

1. Change the return type of everything in `periodic_table` to `Quantity` (as done here), and use pint's wrapper to enforce specific units on downstream methods without having to modify the downstream methods themselves. We have the option here of either requiring input arguments to be `Quantity` or allowing floats as inputs (and assuming they have the correct units). We also have the option to return either `Quantity` or Float.

2. Use `Quantity` internally in `periodic_table` but return `.magnitude` (losing any units information in downstream code, e.g. the `._unit` attribute)

3. Keep `FloatWithUnit` as a lightweight wrapper around `Quantity` that mimics existing behavior (which would not enforce the more rigorous units handling)

There are some more complicated approaches we could use to make this more transparent to users. For example, we could try to 4. add a `unit=` kwarg to pymatgen methods, with an expected default, and then construct `Quantity` internally, e.g.:

```def some_func(float, unit='eV'):
    self.q = Q(float, unit)
    <do math with self.q>
    return self.q OR self.q.magnitude
```

Alternatively, 5. we could use type checking so that the method call signatures would either be Quantity -> Quantity or Float -> Float.

Personally I am in favor of Option 1, where we pass `Quantity` within pymatgen but also allow floats (and assume correct units) for compatibility reasons. We might introduce a deprecation warning at some point and eventually require `Quantity` inputs to those methods. This way, methods in pymatgen outside of `periodic_table` will still work with floats as inputs, and it will be easy for us to move to more robust handling in the future.

shyuep:
Thanks @rkingsbury . I am not in favor of stricter checking. I agree of course that if we fully adopt pint, it will help avoid certain errors. But it is *not* the role for pymatgen to play. Most people do not want someone else to be babysitting units for them.

For example, a fairly accomplished users knows that VASP returns energies in eV. When I am lazy and I want to convert that to joule, I just do X * 1.6e-19. I do not want it such that I have to do X.to(“J”) or X.magnitude * 1.6e-19. That violates what people mostly think of as a float construct.

Personally, I’d rather have pymatgen return no unitized quantity whatsoever, and when people actually want to use units, they can pass those quantities to pint to manipulate them.

shyuep:
Just a note that something like e+2 throwing an error is precisely why I do not want to enforce type checking.

rkingsbury:
Thanks @shyuep. I understand your concerns about being able to do basic arithmetic such as `e+2` without errors. I'm confident we can find some kind of workaround or compromise there.

> I agree of course that if we fully adopt pint, it will help avoid certain errors. But it is _not_ the role for pymatgen to play. Most people do not want someone else to be babysitting units for them.

I see this differently. It makes sense to me that the entire DFT community is "standardized" on eV and angstroms, but from my perspective as someone with 1) an experimental background and 2) visions of using pymatgen for analysis beyond DFT results, I absolutely do want someone else checking units for me. I have seen (and committed) many errors due to wrong units, and I think avoiding those kinds of mistakes is something that codes such as pymatgen are well-suited for.

I'm optimistic we can find a solution that's flexible enough to satisfy both use cases. Stay tuned!

rkingsbury:
@shyuep @mkhorton can you give us some feedback on a creative solution @ayushsgupta came up with? This work is not complete yet, but before we continue in this direction we wanted to get your opinion.

- Ayush created a custom class from `pint.Quantity` called `CompatibleQuantity` that allows one to add floats to Quantity (similar to the original `FloatWithUnit`. However, if you try to add a float to a Quantity, the float is assumed to have the same units as the Quantity. Therefore (unlike `FloatWithUnit`), units information is preserved. For example:

```
from pymatgen.core.units import CompatibleQuantity
a=CompatibleQuantity(1, 'eV')
print(a)
1 electron_volt
b=2
print(a+b)
1 electron_volt
```

- We propose to replace `FloatWithUnit` with `CompatibleQuantity` in all classes in `pymatgen.core`. This means that many attributes of `Element`, `Structure,` etc. will now return `CompatibleQuantity` instead of `FloatWithUnit`.
- Outside of `pymatgen.core`, anytime a method calls something from a core method, we would use the returned `CompatibleQuantity` to convert the value to an appropriate unit, and then pass the magnitude to the rest of the method. You can see a nice example of this in my next comment
- Methods outside of `pymatgen.core` will continue to return floats for the time being

Please let us know what you think.

pymatgen/analysis/elasticity/elastic.py
```diff
@@ -670,8 +674,10 @@ def omega(self, structure, n, u):
         """
         l0 = np.dot(np.sum(structure.lattice.matrix, axis=0), n)
         l0 *= 1e-10  # in A
-        weight = float(structure.composition.weight) * 1.66054e-27  # in kg
-        vol = structure.volume * 1e-30  # in m^3
+        # weight = float(structure.composition.weight) * 1.66054e-27  # in kg
+        weight = structure.composition.weight.to('kg').magnitude
+        # From Ayush: structure.volume is apparently not a Quantity, should it be?
+        vol = structure.volume * 1e-30  # in m^3
```
rkingsbury: This is an example of how `CompatibleQuantity` can add value to pymatgen. Clearly there was a need here to do calcs in a specific set of units, and we currently rely on hard-coded conversions. Using `CompatibleQuantity`, this conversion is done automatically

pymatgen/analysis/elasticity/elastic.py
```diff
@@ -670,8 +674,10 @@ def omega(self, structure, n, u):
...
         l0 *= 1e-10  # in A
-        weight = float(structure.composition.weight) * 1.66054e-27  # in kg
-        vol = structure.volume * 1e-30  # in m^3
+        # weight = float(structure.composition.weight) * 1.66054e-27  # in kg
+        weight = structure.composition.weight.to('kg').magnitude
+        # From Ayush: structure.volume is apparently not a Quantity, should it be?
+        vol = structure.volume * 1e-30  # in m^3 
         vel = (1e9 * self[0].einsum_sequence([n, u, n, u])
                / (weight / vol)) ** 0.5
         return vel / l0
```
rkingsbury: This is an example of how `CompatibleQuantity` can add value to pymatgen. Clearly there was a need here to do calcs in a specific set of units, and we currently rely on hard-coded conversions. Using `CompatibleQuantity`, this conversion is done automatically

rkingsbury:
> What happens if you try to use CompatibleQuantity with numpy operations?

@mkhorton it should be the same behavior as `Quantity`, which means most np operations should work as expected; although I know there are a few that aren't supported by `Quantity` Did you have specific cases in mind?

mkhorton:
Hi @ayushsgupta, I'm not sure if @rkingsbury relayed the message on this, but basically I think `CompatibleQuantity` (as a concept) is a good direction to try, but the implementation (over-riding __add__, __mul__ etc.) is going to be brittle and likely won't play nicely with numpy.

One approach that may be worth attempting is to over-ride `__getattr__`, check if units exist then (1) if they do exist, apply the operations with Quantities as normal, (2) if units do not exist, apply the operations on the underlying `.magnitude`, and cast the result back to the appropriate units, and issue a warning.

If this approach is viable depends, ultimately, on performance and robustness. Note that we would also need a global option for unit checking (e.g. that we can put in the .pmgrc.yaml), which can be set to `STRICT`, `WARN` or `IGNORE` respectively. This could open the door for us to be strict when doing internal testing, but ignore issues for end users who want to use existing scripts as normal.

mkhorton:
Closing this since work seems to have stalled. Please feel welcome to re-open if you want to take another go at this. Thanks for helping to explore this idea!

PR #1780, opened 2020-03-01
Refactor ComputedEntry and PDEntry with new Entry class

ayushsgupta:
* Creates a new Entry class from which ComputedEntry and PDEntry now inherit
* If some form of pull request #1771 (units with pint) is implemented, this will simplify the integration of units into pymatgen
* In general, will make pymatgen easier to maintain by reducing redundancy in the various entry classes 
existing_entry_classes
* Currently only covers ComputedEntry and PDEntry, but can be expanded in the future
* No new dependencies

@rkingsbury please add any other comments/corrections as necessary


rkingsbury:
Thanks @ayushsgupta !  I'll just add some additional context, as I think this has value to pymatgen users regardless of what we decide about the units.

In addition to facilitating the units implementation, this PR was also partly motivated by my experience learning pymatgen as a new user over the last year. I found the large number of `__Entry` classes confusing, and I also found that they did not always behave consistently or have the same method names.

While I understand the need to have multiple types of `__Entry`, and this is not going to change anytime soon, I think minimizing the number of different classes and making their member methods as consistent as possible will really help new users while also facilitating code maintenance.

This PR is a first step in that direction that we hope to continue building on later. By inheriting as much functionality as possible from the new `Entry` class, all of the existing `__Entry` classes will become easier to maintain and more consistent. For example, the `.normalize()` method that we added to `ComputedEntry` a few months ago was never added to `PDEntry`. Now it will be available there automatically.

rkingsbury:
@shyuep @mkhorton , any idea why the Travis build (pydocstyle) is failing? The failures are coming from files that weren't changed in this PR. Has the pydocstyle configuration changed recently or something?

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,105 @@
+# coding: utf-8
+# Copyright (c) Pymatgen Development Team.
+# Distributed under the terms of the MIT License.
+
+"""
+This module implements equivalents of the basic ComputedEntry objects, which
+is the basic entity that can be used to perform many analyses. ComputedEntries
+contain calculated information, typically from VASP or other electronic
+structure codes. For example, ComputedEntries can be used as inputs for phase
+diagram analysis.
```
rkingsbury: @ayushsgupta , can you update the header for the new `entry.py` file to more accurately reflect what's in here?

Something like "This module implements the base Entry class, which is ....   Other Entry classes such as ComputedEntry and PDEntry should inherit from this class."

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,105 @@
...
+__maintainer__ = "Shyue Ping Ong"
+__email__ = "shyuep@gmail.com"
+__status__ = "Production"
+__date__ = "Apr 30, 2012"
+
+
+class Entry(MSONable):
+    """
+    An lightweight Entry object containing key computed data
+    for many purposes.
```
rkingsbury: @ayushsgupta let's not say "computed data" b/c we also have an `ExpEntry` class for experimental data. Maybe say something like "...containing various types of data associated with a specific chemical composition'

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        Initializes an Entry.
+
+        Args:
+            composition (Composition): Composition of the entry. For
+                flexibility, this can take the form of all the typical input
+                taken by a Composition, including a {symbol: amt} dict,
+                a string formula, and others.
+            energy (float): Energy of the entry. Usually the final calculated
+                energy from VASP or other electronic structure codes.
```
rkingsbury: Again, reword to make less specific to calculations. I would just delete the part after "Energy of the entry".

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        return self.energy / self.composition.num_atoms
+
+    def __str__(self):
+        return self.__repr__()
+
+    def normalize(self, mode: str = "formula_unit") -> None:
+        """
+        Normalize the entry's composition, energy and any corrections.
+        Generally, this would not have effect on any
```
rkingsbury: Delete the 2nd line of the docstring here. It seems like it wasn't finished (I know this came from code you copied)

pymatgen/analysis/phase_diagram.py
```diff
@@ -74,42 +75,22 @@ def __init__(self, composition: Composition, energy: float,
                 analysis and plotting purposes. An attribute can be anything
                 but must be MSONable.
```
rkingsbury: It looks like this part of the docstring is supposed to go underneath the `attribute` attribute and not in the init

rkingsbury: I think you can delete the "analysis and plotting purposes"

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,105 @@
...
+from monty.json import MSONable
+
+
+__author__ = "Shyue Ping Ong, Anubhav Jain"
+__copyright__ = "Copyright 2011, The Materials Project"
+__version__ = "1.1"
+__maintainer__ = "Shyue Ping Ong"
+__email__ = "shyuep@gmail.com"
+__status__ = "Production"
+__date__ = "Apr 30, 2012"
```
rkingsbury: @ayushsgupta you can add yourself to the author list after the other two, and update the copyright and date fields

shyuep:
While I think this is a good initiative, I think perhaps you misunderstood the original organization hierarchy. The idea has always been that the PDEntry is the basic unit, and ComputedEntry (and its subclasses) inherit from that. Subsequently, the inheritance got removed but since Python works on duck typing, no functionality was broken. 

For PD computations, all that is needed is composition + energy - structure and everything else are not required. Extra things like "corrections" etc. only apply to ComputedEntries. E.g., I can generate a phase diagram simply by generating three PDEntries with the following args: ("Fe", 0.0), ("O", 0) and ("Fe2O3", -0.1).

So I actually think that having a separate Entry base object serves to confuse more than help. Suddenly things like corrections become part of all entries, even when they are completely irrelevant. In fact, all the new `Entry` base class does is to take all the entire PDEntry infrastructure and adding corrections in it. That is basically a `ComputedEntry`!

In short, I am not in favor of adding complexity to the fundamental base object (whether we call it Entry or PDEntry). Corrections are an artificial construct due to the way MP chose to do its scheme. 

I would be fine, however, if you want to create an `Entry` abstract base class from which both PDEntry and ComputedEntry inherits from. That class would define the core methods and properties that need to be implemented for any type of entry and also implement all common functionality like energies per atom and normalize. As far as I can see, the main abstract properties needed are just energy and composition. Then the subclasses can define energy however they see fit.

pymatgen/entries/computed_entries.py
```diff
@@ -28,7 +27,7 @@
 __date__ = "Apr 30, 2012"
 
 
-class ComputedEntry(MSONable):
+class ComputedEntry(Entry):
     """
     An lightweight ComputedEntry object containing key computed data
     for many purposes. Extends a PDEntry so that it can be used for phase
```
rkingsbury: @ayushsgupta let's change the docstring to:

'Lightweight Entry object for computed data. Contains facilities for applying corrections to the .energy attribute and for storing calculation parameters`

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Args:
+            composition (Composition): Composition of the entry. For
+                flexibility, this can take the form of all the typical input
+                taken by a Composition, including a {symbol: amt} dict,
+                a string formula, and others.
+            energy (float): Energy of the entry.
+            correction (float): A correction to be applied to the energy.
+                This is used to modify the energy for certain analyses.
+                Defaults to 0.0.
```
rkingsbury: @ayushsgupta delete the correction attribute from the docstring

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,101 @@
...
+__maintainer__ = "Shyue Ping Ong"
+__email__ = "shyuep@gmail.com"
+__status__ = "Production"
+__date__ = "Mar 02, 2020"
+
+
+class Entry(MSONable, metaclass=ABCMeta):
+    """
+    A lightweight Entry object containing various types 
+    of data associated with a specific chemical composition.
```
rkingsbury: @ayushsgupta let's change this docstring to `A lightweight object containing the energy associated with a specific chemical composition. This base class is not intended to be instantiated directly. Note that classes which inherit from Entry must define a .energy property.`

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        :return: Whether composition of entry is an element.
+        """
+        return self.composition.is_element
+
+    @property
+    @abstractmethod
+    def energy(self) -> float:
+        """
+        :return: the energy of the entry. Each subclass must define its own energy property.
```
rkingsbury: @ayushsgupta Given the changes to the main docstring, we can now delete 'Each subclass must define its own energy property.'

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+
+    @property
+    def energy_per_atom(self) -> float:
+        """
+        :return: the *corrected* energy per atom of the entry.
```
rkingsbury: remove '*corrected*'

pymatgen/entries/entry.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        return self.energy / self.composition.num_atoms
+
+    def __str__(self):
+        return self.__repr__()
+
+    def normalize(self, mode: str = "formula_unit") -> None:
+        """
+        Normalize the entry's composition, energy and any corrections.
```
rkingsbury: remove 'and any corrections'

rkingsbury:
@ayushsgupta it looks like you will also need to copy the docstring for `.normalize` into `ComputedEntry`. Here is one of the failures from the Travis build:

```
pymatgen/entries/computed_entries.py:81 in public method `normalize`:
        D102: Missing docstring in public method
```

shyuep:
Pls move the Entry abc to pymatgen.entries.__init__.py. I prefer less typing during imports.

rkingsbury:
Thanks for your feedback, @shyuep ; we're ready for you to review. Please let us know if you have any other comments. 

The only reason the Travis CI is failing is because of the docstring issues introduced by a different commit; nothing in our changes is causing problems.

shyuep:
Great. Thanks.

PR #1782, opened 2020-03-03
automatic box size overwrite

htz1992213:
## Summary

*The box size values in the CombinedData object is now automatically updated by the min/max values of the coordinates.

### DIFF
--- a/pymatgen/io/lammps/data.py
+++ b/pymatgen/io/lammps/data.py
@@ -1239,7 +1239,9 @@ class CombinedData(LammpsData):
 
         """
 
-        self.box = list_of_molecules[0].box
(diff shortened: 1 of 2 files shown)


htz1992213:
Am I supposed to fix those docstring??? Why those are merged in the first place...

mkhorton:
The shengbte, structure_vtk and advanced_transformations are the responsibility of other contributors. Any other errors you would have to correct, this is the result of stricter checking for docstrings on commit (if you touch a file that has bad docstrings, it will flag an error).

htz1992213:
Thanks Matt! So it's good to merge?

shyuep:
Thanks! It has been merged.

PR #1785, opened 2020-03-04
Pourbaix fix

montoyjh:
## Fix to two bugs in Pourbaix Diagram code

* Fixes issue where Pourbaix diagram was considering MultiEntries where reaction calculator finds a balanced reaction with no product.
* Fixes issue with MPRester PourbaixEntry processing due to Entry refactor.  Specifically, `_energy`/`uncorrected_energy` mix-up.

## TODO (if any)

The regression test that I've added is heavier than I'd like (on an already too-heavy set of tests for the pourbaix_diagram module), I can create a more minimal test that encapsulates the issue more neatly, but will need a bit more time to do so, so figured I'd issue the PR in case the bug fixes are a priority.


shyuep:
Thanks!

PR #1798, opened 2020-03-11
Fix the notes for python version dependent of the ASE package.

hongyi-zhao:
## Summary

See here for more info:

https://wiki.fysik.dtu.dk/ase/install.html

Which told the following:
Note that the ASE package is compatible with Python 3.5+ at the moment.


shyuep:
thanks

PR #1753, opened 2020-01-30
Update OpenBabel; Fix NN classes, graph classes

espottesmith:
## Summary

This PR addresses two major issues:

* The recent OpenBabel release 3.0.0 breaks the previous API in several ways, most notably in how the Python interface is imported
* A bug was found in which certain NearNeighbors classes failed to approrpriately detect bonds in molecules

With respect to the second change, all NearNeighbor classes now have properties describing if they can be used with Molecules, Structures, or both. MoleculeGraph and StructureGraph now prevent the user from using with_local_env_strategy with an inappropriate strategy.

Additionally, the bug which caused an automatic re-indexing of Molecules has been identified and fixed.

## Additional dependencies introduced (if any)

* openbabel 3.0.0

## Checklist

- [ ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
      Run [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/) and [flake8](http://flake8.pycqa.org/en/latest/)
      on your local machine.
- [X] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). (exceptions for some properties)
      Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [ ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) 
      to type check your code.
- [X] Tests have been added for any new functionality or bug fixes.
- [x] All existing tests pass.

PR #1799, opened 2020-03-12
Fix QChem API

utf:
## Summary

A recent pull request ~~(#1753)~~ (efd386b) made the `check_for_structure_changes` method private. Unfortunately, custodian relies on this method so this PR broke both custodian and atomate.

Furthermore, #1753 changed the arguments for `MoleculeGraph.with_local_env_strategy` which also broke atomate.

Can I suggest that when breaking API changes are made to the qchem module, that the authors also test their changes on atomate and custodian. @espottesmith @samblau

Thanks!

### DIFF
--- a/pymatgen/io/qchem/outputs.py
+++ b/pymatgen/io/qchem/outputs.py
@@ -834,7 +834,7 @@ class QCOutput(MSONable):
             self.data["energy_trajectory"] = real_energy_trajectory
             self._read_geometries()
             if have_babel:
-                self.data["structure_change"] = _check_for_structure_changes(
+                self.data["structure_change"] = check_for_structure_changes(
                     self.data["initial_molecule"],
                     self.data["molecule_from_last_geometry"])
             self._read_gradients()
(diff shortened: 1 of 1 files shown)


utf:
Wow, actually. I take back my accusation. I'm sorry @espottesmith @samblau 

This bug was actually introduced by @shyuep  in  efd386bcdbcc5541aa01c3bb4116999ae8924f68

mkhorton:
The changes made in #1753 at least were at my request -- it fixed a long-standing bug and cleaned up some of our neighbor detection, but it did also therefore make some kwargs superfluous which is why the method call changed. Apologies for breaking atomate with this one!

shyuep:
Basic principle - is something does not have a doc, I treat it as something private and I can modify as I well please unless the tests fail. So, write tests and doc?

utf:
I've added a docstring curtsey of @espottesmith so that linting passes.

I have nothing to do with the qchem module so I will not add a test now. I suggest @samblau or @espottesmith add one in a future pull request.

This should be ready to merge.

mkhorton:
Thanks @utf!

shyuep:
Thanks. But I will wait until a test is added to that method before merging.

utf:
Ok, I will leave this up to someone else to fix.

I was not involved in the original code and am only trying to revert changes made in pymatgen to resolve upstream breakages in atomate and custodian.

PR #1800, opened 2020-03-12
[WIP] New KPath and HighSymmKpath classes

munrojm:
## Summary

This pull request introduces support for generating high-symmetry k-paths using two additional conventions. Namely, the SeeK-path convention by Hinuma and coworkers [Y. Hinuma, et al. Band structure diagram paths based on crystallography, Comp. Mat. Sci. 128, 140 (2017)], and a new "on-the-fly" convention out of the Persson group that is soon to be published. 

Below is a summary of the code changes:

* New KPath classes added to handle high-symmetry k-path generation.
* Existing HighSymmKpath code to generate paths using the convention by Setyawan and Curtarolo converted to a KPath class.
* New HighSymmKpath class added which calls k-path generation classes, and provides methods for amalgamating and comparing path data across conventions.

It should be noted that the default HighSymmKpath behavior has not been altered.

## Additional dependencies 

The SeeK-path package is required for the convention by Hinuma and coworkers, and has been added to the optional requirements list.

### DIFF
--- a/pymatgen/symmetry/bandstructure.py
+++ b/pymatgen/symmetry/bandstructure.py
@@ -178,8 +178,13 @@ class HighSymmKpath(KPathBase):
         Returns:
         Setyawan and Curtarolo k-path with labels.
         """
+        kpath = KPathSetyawanCurtarolo(self._structure, symprec, angle_tolerance, atol)
 
-        return KPathSetyawanCurtarolo(self._structure, symprec, angle_tolerance, atol)
+        self.prim = kpath.prim
+        self.conventional = kpath.conventional
+        self.prim_rec = kpath.prim_rec
+
+        return kpath
 
     def _get_hin_kpath(self, symprec, angle_tolerance, atol, tri):
         """
--- a/pymatgen/symmetry/bandstructure.py
+++ b/pymatgen/symmetry/bandstructure.py
@@ -1,779 +1,305 @@
-# coding: utf-8
-# Copyright (c) Pymatgen Development Team.
-# Distributed under the terms of the MIT License.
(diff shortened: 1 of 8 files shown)


mkhorton:
Just to double-check, have you verified the original class still works as before when SeeK optional dependency is not installed?

Otherwise this pull request looks essentially good. Since it's a large one, I'll take some more time to read over it, and can merge once all linting/tests pass. Happy to see it finally coming out!

munrojm:
>Just to double-check, have you verified the original class still works as before when SeeK optional dependency is not installed?

Yes, the original class and its default behavior work even if SeeK-path is not installed. It is only when you ask HighSymmKpath to use the convention by Hinuma and coworkers, or simply ask it to use all of the conventions that an error is issued.

>Otherwise this pull request looks essentially good. Since it's a large one, I'll take some more time to read over it, and can merge once all linting/tests pass. Happy to see it finally coming out!

Great! That sounds good.

mkhorton:
Hi @munrojm, I'm going to merge this -- thanks for your very diligent efforts to make sure backwards compatibility wasn't broken!

mkhorton:
Actually, one request first -- can you make `KPathBase` explicitly an abstract base class? Just to prevent confusion should one try to initialize it. Making it an abstract base class makes it clear to the user it shouldn't be used directly.

munrojm:
> Actually, one request first -- can you make `KPathBase` explicitly an abstract base class? Just to prevent confusion should one try to initialize it. Making it an abstract base class makes it clear to the user it shouldn't be used directly.

Sure! No problem. That makes sense.

PR #1804, opened 2020-03-14
[WIP] CP2K Module

nwinner:
Work in progress for integration of CP2K in pymatgen. Currently in working condition, but outputs.py and sets.py need to be fleshed out with more features. Also some small changes to existing files have been made that could be useful for this integration.

**Features**
1. Creation of pymatgen.io.cp2k module with the following contents:

- inputs.py: support for the the input file structure of cp2k executable
- sets.py: input sets for running default cp2k calculations using predefined input structure
- outputs.py: cp2k output file parsing
- utils.py: a handful of functions to help with input sets, more to be inserted
- test_files: just a few test files, not exhaustive

2. Added skip_step to the xyz parser. For large MD runs, this can be hugely beneficial. While reading in the raw xyz text as frames is very fast, creating a pymatgen structure object takes a while, so this makes dealing with large xyz files easier.

3. In bader_caller.py in the command line module, it assumed a VASP structure. For Cp2k, the more general .cube files are used. These require a simpler version of the bader caller which I'm currently working on.

**Fixes**
1. pymatgen.io.xyz had a tiny issue in how it was playing with cp2k. It seemed that zopen was not using the 'rt' argument, so when gzipped .xyz files were read in, they couldn't be manipulated like strings.

**TODO**

1. Make the output parsing a little more robust. At the moment it makes some assumptions that will be fine if you know how its written, but for general use it should be a lot better at dealing with variables in the output file structure.

2. Create more input sets.

3. Fix the broken symmetry initialization of the input sets using a util that maps between structure objects and the &BS section of the cp2k input file.

4. Bader caller needs to integrate the "charge transfer" that the VASP one does.


mkhorton:
Hi @nwinner, this looks great.

Regarding the Bader analysis, could you explore the idea of merging implementations -- I'd prefer not to have two separate classes each for Gaussian Cube / VASP Chgcar given that they're basically equivalent. This might also be good time to add `user_bader_settings` as an option to the Bader caller, e.g. for vacuum cut-off and other optional parameters that it takes.

Regarding xyz parsing and large runs, I just wanted to highlight the `Trajectory` class in `pymatgen.core.trajectory` which is a more light-weight version of storing outputs from large runs: it stores it as a single Structure object + a list of displacements.

Issue #1801, opened 2020-03-13
SpaceGroup.from_int_number(hexagonal=False) ineffective


jacksund:
**Describe the bug**
When initiating a SpaceGroup class using the SpaceGroup.from_int_number(), the hexagonal option is ineffective for the following spacegroups: 146,148,155,160,161,166,167

NOTE: This is similar to issue #1203. The big difference is that here, it's claimed functionality that is ineffective. While a workaround exists, the functionality of hexagonal=False needs to be corrected.

**To Reproduce**
Below is an example with spacegroup 166 that illustrates the bug:

```
from pymatgen.symmetry.groups import SpaceGroup

SpaceGroup('R-3m') # correctly returns hexagonal info (per hexagonal is set as default)
SpaceGroup('R-3m:H') # correctly returns hexagonal info
SpaceGroup('R-3m:R') # correctly returns rhombohedral info

SpaceGroup.from_int_number(166, hexagonal=True) # correctly returns hexagonal info
SpaceGroup.from_int_number(166, hexagonal=False) # incorrectly returns hexagonal info
```

**Expected behavior**
```
sg = SpaceGroup.from_int_number(166, hexagonal=False) 
sg.symbol # should be 'R-3mR', but instead give 'R-3mH' which is the hexagonal form
sg.order # should be 12, but instead gives 36 which is the hexagonal form
```

**Desktop (please complete the following information):**
 - OS: Linux
 - Version: 2019.12.22 (though code looks to be identical in the up-to-date github version)

**Additional context**
In the from_int_number() code ([located here on github](https://github.com/materialsproject/pymatgen/blob/5a3284fd2dce70ee27e8291e6558e73beaba5164/pymatgen/symmetry/groups.py#L470)), the method really is just calling sg_symbol_from_int_number() and running that result back through SpaceGroup(). The bug exists in sg_symbol_from_int_number(). More specifically, it is a logical error in the following line:

`syms = list(filter(lambda s: not s.endswith("H"), syms))`

This line is run when hexagonal=False. The issue is that this line will return ‘R-3m’, which then calls SpaceGroup(‘R-3m’) – this still gives the hexagonal object! We need that line to return ‘R-3m:R’, not ‘R-3m’. We could add ':R' in multiple places:

OPTION 1: within SpaceGroup.from_int_number():
```
##### original code ######
return SpaceGroup(sg_symbol_from_int_number(int_number,hexagonal=hexagonal))

##### fixed code ######
[...]

Issue #1802, opened 2020-03-13
Structure.from_spacegroup() inaccurate for trigonal crystal systems


jacksund:
I apologize in advance for the long post, but I think this is a confusing topic. I wanted to give a full justification for the changes I'm suggesting!

**Summary:**
Structure.from_spacegroup() will inaccurately produce structures for the following spacegroups:
143, 144, 145, 147, 149, 150, 151, 152,153, 154, 156, 157, 158, 159, 162, 163,164, 165.
This occurs because SpaceGroup.is_compatible() requires these spacegroups to be hexagonal lattices, when the [symm_op.yaml](https://github.com/materialsproject/pymatgen/blob/master/pymatgen/symmetry/symm_ops.yaml) only provides operations for the trigonal system. This is a bug that is introduced by issue #454, where discussion is misled by rhombohedral/trigonal/hexagonal confusion.

**Issue #454 inaccuracies:**
In the posted issue (to summarize), [blondegeek](https://github.com/blondegeek) argues that trigonal spacegroups (specifically spacegroup 165) should have hexagonal lattices. Also, [wenhaosun](https://github.com/wenhaosun) states that spacegroup 165 is trigonal. [blondegeek](https://github.com/blondegeek) then makes the final statement that spacegroup 165 is indeed tigonal crystal system but possesses a hexagonal lattice system. The result of this discussion is that all trigonal spacegroups (see list in summary above) are checked to be hexagonal in the [SpaceGroup.is_compatible()](https://github.com/materialsproject/pymatgen/blob/5a3284fd2dce70ee27e8291e6558e73beaba5164/pymatgen/symmetry/groups.py#L362) code. 
Their discussion comes close, but it is still incorrect -- leading to a logical error in the code. The confusion can actually be summarized with a quote from wikipedia: 

> Crystal systems, crystal families and lattice systems are similar but slightly different, and there is widespread confusion between them: in particular the trigonal crystal system is often confused with the rhombohedral lattice system, and the term "crystal system" is sometimes used to mean "lattice system" or "crystal family".

[...]

PR #1805, opened 2020-03-16
Hexagonal Sym Issues

jacksund:
This merge request addresses issues #1801 and #1802. I decided to add the #1801 fix to the SpaceGroup.from_int_number() method, which is not the lowest function with the hexagonal=True input. The respective unittests for each fix were also added.


jacksund:
I limited the merge request to only fix #1801, as I found a mistake in my post for #1802.

shyuep:
Thanks

PR #1807, opened 2020-03-17
Reorder check for metal

mfherbst:
## Summary

* Reorder logic in BSPlotter, such that check for metal is only executed when needed.
* The current implementation causes trouble in my downstream usage of pymatgen, so I'd appreciate a merge.


mfherbst:
Thanks!

PR #1809, opened 2020-03-26
Rewrite EconNN implementation

utf:
## Summary

The EconNN class was rewritten to properly reproduce the algorithm outlined in:

https://www.degruyter.com/view/j/zkri.1979.150.issue-1-4/zkri.1979.150.14.23/zkri.1979.150.14.23.xml

This improves the performance of the method, particularly when only considering bonding between sites of opposing charge (controlled using the new `cation_anion` option).

The method has a tolerance parameter that determines when to consider a site bonded. After trialling several values, it seems like the results are rather insensitive in the range 0.1–0.6. I've set it to the optimal value of 0.2 that I found for my [test set](https://github.com/hackingmaterials/materialscoord).

### DIFF
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -3394,7 +3394,7 @@ class EconNN(NearNeighbors):
 
     def __init__(
         self,
-        tol: float = 0.5,
+        tol: float = 0.2,
         cutoff: float = 10.0,
         cation_anion: bool = False
     ):
--- a/pymatgen/analysis/local_env.py
+++ b/pymatgen/analysis/local_env.py
@@ -10,20 +10,34 @@ of single sites in molecules and structures.
 
 import math
 import warnings
-from collections import namedtuple, defaultdict
-from functools import lru_cache
-from typing import Union
-
-import ruamel.yaml as yaml
 import os
 import json
+
+import ruamel.yaml as yaml
+import numpy as np
+
 from copy import deepcopy
+from math import pow, pi, asin, sqrt, exp, sin, cos, acos, fabs, atan2
+from collections import namedtuple, defaultdict
+from functools import lru_cache
+from typing import Union, List, Optional
+from bisect import bisect_left
 
+from scipy.spatial import Voronoi
 from monty.dev import deprecated
+from monty.dev import requires
+from monty.serialization import loadfn
 
+from pymatgen import Element, Structure, IStructure
+from pymatgen.analysis.bond_valence import BV_PARAMS, BVAnalyzer
+from pymatgen.core.structure import PeriodicNeighbor
 from pymatgen.analysis.molecule_structure_comparator import CovalentRadius
 from pymatgen.core.sites import PeriodicSite, Site
 
+try:
+    from openbabel import openbabel as ob
+except Exception:
+    ob = None
 
 __author__ = "Shyue Ping Ong, Geoffroy Hautier, Sai Jayaraman," + \
              " Nils E. R. Zimmermann, Bharat Medasani, Evan Spotte-Smith"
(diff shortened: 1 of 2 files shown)


shyuep:
I think the tests are a bit too rudimentary. Diamond, NaCl and CsCl are hardly difficult structures. Can we include at least one more difficult structure?

utf:
I added LiFePO4 and tidied the tests a little.

I didn't add the LiFePO4 test for MinimumVIRENN as it is very slow.

mkhorton:
> I didn't add the LiFePO4 test for MinimumVIRENN as it is very slow.

Is still slow if you run it on a structure that already has oxidation states annotated?

utf:
Adding oxidation states beforehand didn't change the timing but it only takes an additional 4 seconds seconds so I added in.

mkhorton:
These are so slow :(

CutoffDictNNs have some obvious optimizations, I really need to re-visit at some point. The NN time is going to be a big factor in driving the visualizations on the web too.

utf:
image

mkhorton:
@sivonxay I remember we discussed optimizations to JmolNN (and other CutoffDictNNs), I remember you had a suggestion but that there was a roadblock with the implementation, could you remind me what the status was with this?

@utf given that JmolNN is already on the Pareto front, it's probably not urgent that it's improved for this analysis, but I'm going to take a quick look at it

sivonxay:
@mkhorton  Ahhh yes. I had initially thought the CutoffDictNN was poorly optimized, since it was taking a while to run. After a bit of digging, I found out that the slowdown was not due to the neighbor finding code, but due to the NearestNeighbor._get_image() function. With my structures of > 100 atoms, this was a large bottleneck.

mkhorton:
@utf with the additional LiFePO4 test added, is this ready to merge?

@sivonxay thanks for the reminder! I'll take a look at optimizing that, will bring discussion of `_get_image` to a new issue/PR

utf:
Ok, once tests pass this should be ready to merge

utf:
I also ran into the bottleneck of `get_image` for some large structures. I think a major reason is that frac_coords is continually called. 

I made an optimised version of MinimumDistanceNN by only calculating frac_coords once and moving the get_image function inside get_nn_info. This isn't the best modular implementation but might be useful if you are doing some optimisation:

```
from pymatgen.analysis.local_env import MinimumDistanceNN

class FastMinimumDistanceNN(MinimumDistanceNN):
    """
    Optimized version of MinimumDistanceNN
    """
    
    def get_all_nn_info(self, structure):
        """Get a listing of all neighbors for all sites in a structure

        Args:
            structure (Structure): Input structure
        Return:
            List of NN site information for each site in the structure. Each
                entry has the same format as `get_nn_info`
        """
        from tqdm.cli import tqdm
        all_neighbors = structure.get_all_neighbors(self.cutoff)
        
        return [self.get_nn_info(structure, n, all_neighbors=all_neighbors)
                for n in tqdm(range(len(structure)))]

    def get_nn_info(self, structure, n, all_neighbors=None):
        """
        Get all near-neighbor sites as well as the associated image locations
        and weights of the site with index n using the closest neighbor
        distance-based method.

        Args:
            structure (Structure): input structure.
            n (integer): index of site for which to determine near
                neighbors.

        Returns:
            siw (list of tuples (Site, array, float)): tuples, each one
                of which represents a neighbor site, its image location,
                and its weight.
        """

        site = structure[n]
        if not all_neighbors:
            neighs_dists = structure.get_neighbors(site, self.cutoff)
        else:
            neighs_dists = all_neighbors[n]
            
        from pymatgen.util.coord import pbc_diff
        frac_coords = structure.frac_coords

        def get_original_site(site):
            """Private convenience method for get_nn_info,
            gives original site index from ProvidedPeriodicSite."""
            diff = pbc_diff(site.frac_coords, frac_coords)
            matches = np.where(np.all(np.abs(diff < 1e-8), axis=-1))[0]
            if len(matches) > 0:
                return matches[0]
            else:
                raise Exception('Site not found!')

        def get_image(site):
            original_site_n = get_original_site(site)
            original_site = structure[original_site_n]
            image = np.around(np.subtract(site.frac_coords, original_site.frac_coords))
            image = tuple(image.astype(int))
            return original_site_n, image

        siw = []
        if self.get_all_sites:
            for nn in neighs_dists:
                w = nn.nn_distance
                original_site, image = get_image(site)
                siw.append({'site': nn,
                            'image': image,
                            'weight': w,
                            'site_index': original_site})
        else:
            min_dist = min([nn.nn_distance for nn in neighs_dists])
            for nn in neighs_dists:
                dist = nn.nn_distance
                if dist < (1.0 + self.tol) * min_dist:
                    w = min_dist / dist
                    original_site, image = get_image(site)
                    siw.append({'site': nn,
                                'image': image,
                                'weight': w,
                                'site_index': original_site})
        return siw
```

PR #1810, opened 2020-03-26
Improvements to mcsqs integration

rwoodsrobinson:
## Summary

* Improved tests
* Return objective function
* Simplified method signature
* Improved docs
* Fixed issue with `run_mcsqs` output

## TODO (if any)

* Further improvements in a future pull request (e.g. include MC temperature)

## Checklist

      on your local machine.
      to type check your code.


mkhorton:
Thanks @rwoodsrobinson, will merge once tests pass

@dongsenfo be aware that the keyword args for `run_mcsqs` have changed here to merge `supercell` and `numatoms` options into a single `scaling` option (since both cannot be used simultaneously); I think the improvement is worthwhile to make the change, but since you're the primary user of this code I wanted to make sure you were notified

dongsenfo:
@mkhorton That sounds good to me. 

Thanks @rwoodsrobinson!

mkhorton:
@dongsenfo there was also an issue with the old caller returning structures that were parsed incorrectly (which we discussed in the office), this caller now returns guaranteed correct structures however the underlying parsing bug still has not been fixed yet. Just flagging this in case you've encountered this in your own work. Maybe @rwoodsrobinson can comment further or supply examples

rwoodsrobinson:
You're welcome @dongsenfo, and thanks @mkhorton for the assistance with these changes.

To elaborate, I was looking at cif files for some structure outputs from the old code for what should have been a 128 atom sqs derived from mp-1014244 (ZnZrN2 in the "wurtsalt" structure) and they looked like this:
ZZN-wurtsalt-old

With the updated code using the "str2cif" function in atat, it looks more reasonable:
ZZN-wurtsalt-new

I also added the objective function to the output of run_mcsqs so we can test convergence for future calculations, though this is a work in progress (I plan to add T, w_p, w_n for test). Happy to discuss other changes.

PR #1814, opened 2020-03-29
Lammpsdata fix

### LINKED ISSUE #1812: Unexpected structure.get_sorted_structure in structure to lammps data
https://github.com/materialsproject/pymatgen/blob/47c8223dfe2ef4af6ebc273c950a72ccf038c5c6/pymatgen/io/lammps/data.py#L805

I found this behavior unexpected for some applications. 

If I want to calculate the site descriptors for a given site index in a structure using lammps, I can get wrong results since this line of code reordered the sites in the structure before outputting to lammps data. 

For example, if my structure has two sites ['S', 'Mo'], and I want to calculate local environment descriptors using lammps on the first site (S), I would always get the properties of the 'Mo' site due to the structure sort.

I had to sort my structures before using this code every time to avoid error and I wonder if this behavior can be changed.

chc273:
## Summary

This PR fixes #1812 

Previously Structure is reordered before writing to lammps data file. This could result in unexpected site index mismatch between structure and the final outcome from lammps.


shyuep:
Pls add a test to ensure this does not occur again. I would also add a new kwarg to allow people to sort if they want.

chc273:
@shyuep Sure.

shyuep:
Thanks.

PR #1817, opened 2020-04-01
Tem Calculator

welltemperedpaprika:

Edits made to Frank Wan's previous TEM calculator, a project under Shyam's direction:
- Changed method get_interplanar_spacing to call lattice.d_hkl rather than separately calculating for different crystal systems.
- Vectorized method generate_points to be using numpy arrays rather than nested for loops.
- Vectorized methods to reduce for loops when looping over dictionaries, if possible.
- Changed original implementation of prettytable to pandas dataframe.
- Removed TEM Dot object to a more lightweight named tuple.
- Updated structural factors. Matches international table of crystallography.
---------------------------------------
Updated March 2020:
- Fixed a bug where beam direction of 110 and -1-10 are not generating any patterns.
- Added a new method to more accurately calculate interplanar angles, accounting for different crystal systems.
- Recalculated diffraction pattern using Moore-Penrose inverse.

Request submitted by Jason Liang, student working with Shyam.


mkhorton:
Thanks for the quick turnaround on this :-)

Before merging, have you also tested a few random structures with random directions? I'm wondering what the test coverage is like here -- I only tested myself along some high symmetry directions with some common structures.

welltemperedpaprika:
> Thanks for the quick turnaround on this :-)
> 
> Before merging, have you also tested a few random structures with random directions? I'm wondering what the test coverage is like here -- I only tested myself along some high symmetry directions with some common structures.

Thank you!
Calculations are based on the following indexing procedure:
https://www.doitpoms.ac.uk/tlplib/diffraction-patterns/maths.php
I have tested individual formulas of calculation according to the international table of crystallography. I have also generated 600 + patterns with random beam directions of different iron oxides obtained from the materials project, and all of them was able to generate a pattern. There are some directions where the pattern only appears to be a line of dots, however. 
I would assume the results for high symmetry system and directions are most accurate (which I was able to compare to some reference pattern online).

mkhorton:
Woo, doitpoms 🎉 I worked on that website myself as an undergrad many years ago.

>  I have also generated 600 + patterns with random beam directions of different iron oxides obtained from the materials project, and all of them was able to generate a pattern. 

Great!

> There are some directions where the pattern only appears to be a line of dots, however.

This would be expected too.

There's commercial software out there that's very good called SingleCrystal from the makers of CrystalMaker, if you wanted to do further testing I think they offer free trials, otherwise this sounds great.

PR #1818, opened 2020-04-01
Optimize CubicSupercellTransformation for diagonal transformations

utf:
## Summary

`CubicSupercellTransformation` had some unnecessary complexity for diagonal only transformations. This PR implements a more optimised method.

### DIFF
--- a/pymatgen/transformations/advanced_transformations.py
+++ b/pymatgen/transformations/advanced_transformations.py
@@ -1626,6 +1626,7 @@ class CubicSupercellTransformation(AbstractTransformation):
     """
     A transformation that aims to generate a nearly cubic supercell structure
     from a structure.
+
     The algorithm solves for a transformation matrix that makes the supercell
     cubic. The matrix must have integer entries, so entries are rounded (in such
     a way that forces the matrix to be nonsingular). From the supercell
@@ -1636,6 +1637,7 @@ class CubicSupercellTransformation(AbstractTransformation):
     and the number of atoms in the supercell falls in the range
     ``min_atoms < n < max_atoms``.
     """
+
     def __init__(
         self,
         min_atoms: Optional[int] = None,
@@ -1669,9 +1671,11 @@ class CubicSupercellTransformation(AbstractTransformation):
         side length is at least 'num_nn_dists' times the nearest neighbor
         distance and the number of atoms in the supercell falls in the range
         defined by min_atoms and max_atoms.
+
         Returns:
             supercell: Transformed supercell.
         """
+
         lat_vecs = structure.lattice.matrix
 
         # boolean for if a sufficiently large supercell has been created
@@ -1697,10 +1701,12 @@ class CubicSupercellTransformation(AbstractTransformation):
             )
 
             proposed_sc_lat_vecs = self.transformation_matrix @ lat_vecs
+
             # Find the shortest dimension length and direction
             a = proposed_sc_lat_vecs[0]
             b = proposed_sc_lat_vecs[1]
             c = proposed_sc_lat_vecs[2]
+
             length1_vec = c - _proj(c, a)  # a-c plane
             length2_vec = a - _proj(a, c)
             length3_vec = b - _proj(b, a)  # b-a plane
--- a/pymatgen/transformations/advanced_transformations.py
+++ b/pymatgen/transformations/advanced_transformations.py
@@ -1626,7 +1626,6 @@ class CubicSupercellTransformation(AbstractTransformation):
     """
     A transformation that aims to generate a nearly cubic supercell structure
     from a structure.
-
     The algorithm solves for a transformation matrix that makes the supercell
     cubic. The matrix must have integer entries, so entries are rounded (in such
     a way that forces the matrix to be nonsingular). From the supercell
@@ -1637,7 +1636,6 @@ class CubicSupercellTransformation(AbstractTransformation):
     and the number of atoms in the supercell falls in the range
     ``min_atoms < n < max_atoms``.
     """
-
     def __init__(
         self,
         min_atoms: Optional[int] = None,
@@ -1658,7 +1656,6 @@ class CubicSupercellTransformation(AbstractTransformation):
         self.min_length = min_length
         self.force_diagonal = force_diagonal
         self.transformation_matrix = None
-
     def apply_transformation(self, structure: Structure) -> Structure:
         """
         The algorithm solves for a transformation matrix that makes the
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @utf.

Out of interest, is there an easy way to modify Cubic to give you a square slab instead? e.g. square in x, y but a small z -- this would be useful for some folks doing (S)TEM image simulation, square slabs are useful as multi slice inputs

utf:
I think that should be possible. I wonder if @rees-c (original author) has any thoughts on how to do that?

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1682,31 +1682,23 @@ def apply_transformation(self, structure: Structure) -> Structure:
...
-            else:
-                # Update trans_mat (without diagonal constraint)
-                target_sc_lat_vecs = np.eye(3, 3) * target_sc_size
-
-                self.transformation_matrix = (
-                        np.linalg.inv(lat_vecs) @ target_sc_lat_vecs
-                )
+            target_sc_lat_vecs = np.eye(3, 3) * target_sc_size
+            self.transformation_matrix = (
+                    np.linalg.inv(lat_vecs) @ target_sc_lat_vecs
```
rees-c, line 1695:
```suggestion
                     target_sc_lat_vecs @ np.linalg.inv(lat_vecs)
```

@utf I believe the order here might be wrong - sorry! Looking at `SupercellTransformation` within `pymatgen.transformations.standard_transformations` shows the relationship is new_lattice = transformation_matrix * old_lattice, which is also how it's written a few lines down with `proposed_sc_lat_vecs = self.transformation_matrix * lat_vecs`.

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1682,31 +1682,23 @@ def apply_transformation(self, structure: Structure) -> Structure:
...
-                self.transformation_matrix = np.diag(trans_mat_diagonal)
-
-            else:
-                # Update trans_mat (without diagonal constraint)
-                target_sc_lat_vecs = np.eye(3, 3) * target_sc_size
-
-                self.transformation_matrix = (
-                        np.linalg.inv(lat_vecs) @ target_sc_lat_vecs
-                )
+            target_sc_lat_vecs = np.eye(3, 3) * target_sc_size
```
rees-c, line 1693: @utf if you want a square supercell where the _c_ vector is left untouched, then you need the last/third row of the transformation matrix to be fixed to [0, 0, 1]. So in that case, I would try adding something along the lines of `if want_square_supercell: target_sc_lat_vecs[-1,-1] = 1` right after `target_sc_lat_vecs = np.eye(3, 3) * target_sc_size`. Hope that helps.

mkhorton: Thanks @rees-c, I'd be happy to add this in another PR (so as not to create merge conflicts with this one)

mkhorton:
Thanks @utf and @rees-c

pymatgen/transformations/advanced_transformations.py
mkhorton: @rees-c, making this change I have issues with the transformation matrix being singular, do you have any suggestions?

mkhorton: To share an example, I'm getting a matrix like:

```
[[-2.83564613  2.83564613  2.83564613]
 [ 2.83564613 -2.83564613  2.83564613]
 [ 0.18413287  0.18413287 -0.18413287]]
```

which is

```
[[-3,  3,  3],
[ 3, -3,  3],
[ 0,  0, -1]]
```

after `_round_and_make_arr_singular`, but is still singular.

rees-c: Ah yikes, sorry about that. It looks like I only considered zero rows and columns when writing `_round_and_make_arr_singular`. Let me think about it and get back to you.

rees-c: I've figured out what's happening and how to fix it. For documentation's sake, I'll explain here and then add a suggestion later with the code fix. So the goal of `_round_and_make_arr_singular` is to round a matrix's entries to the nearest integers, and then make as small changes as possible to make the array singular. So in your example, my algorithm would have rounded the original matrix to
`[[-3, 3, 3],`  
`[3, -3, 3],`  
`[0, 0, 0]]`,
identified the third row as zero, and then chosen the entry in that row corresponding with the largest magnitude entry of the original row, `[0.18413287  0.18413287 -0.18413287]`. Since every entry of this original row have the same magnitude, my code broke ties randomly and in your case, just so happened to pick the one out of the three choices that doesn't make the matrix singular. I'll add a suggestion that fixes this by trying every entry instead of just randomly picking one. I guess it might be possible that trying every entry would still not work, but it should be incredibly unlikely (singular square matrices apparently are an infinitely thin subset of all square matrices). If by divine intervention the matrix is still not singular, I would then try all pairs of entries in the zero row, then all the entries in the zero row, then repeat this process (of changing singlets, pairs, triplets, etc.) for zero entries in the full matrix, and then just start picking random entries in the full matrix.

mkhorton: Thanks for this exposition @rees-c, much appreciated in clearing it up!

To clarify, I'm assuming you're talking about this line:

https://github.com/materialsproject/pymatgen/blob/ce35198c83d4541cab36bfe4b12ef0631c8ec89c/pymatgen/transformations/advanced_transformations.py#L1915

And suggesting we try all values there along with an explicit check that the matrix isn't singular.

It seems like this bit of logic is only implemented for the rows, not for the columns, why is that?

mkhorton: I have a test implementation up here: https://github.com/materialsproject/pymatgen/compare/square-transformation

It basically works, in that `|a| == |b| != |c|`, but angles are 45º, 90º, 90º -- the transformation I was aiming for had angles 90º, 90º, 90º with `|a| == |b| != |c|`.

rees-c: Hi @mkhorton! I see now that my code rounded every entry of each zero column instead of picking a random entry... if there was a good reason for why I would've done that, it has since escaped my brain haha.

With regard to the angles, it's my understanding that's it's not always possible to get perfect 90 degree angles (due to the requirement for all entries of the transformation to be integers for preserving periodicity). Not sure if that's why you're getting a 45 degree angle but I'll take a closer look at your test implementation and let you know if I spot anything.

mkhorton:
Ok, thanks @rees-c, no worries about the 45deg angles etc, this will
already be useful as is.

PR #1819, opened 2020-04-01
Fix BrunnerNN methods

utf:
## Summary

Two of the BrunnerNN methods were implemented incorrectly leading to poor performance. This PR fixes their implementation to be in line with: https://scripts.iucr.org/cgi-bin/paper?S0567739477000461


mkhorton:
Thanks @utf!

PR #1822, opened 2020-04-08
Gau

gVallverdu:
## Summary

Improve parsing of structures if multiple optimization, for example in the case of relaxed scan.

*  In case of scan the attribute `opt_structures` store optimized structure contrary to the old `structures` attribute who store all structures
* Fix a bug in read_scan() function linked to a change between g09 and g16.


gVallverdu:
It seems there is an error because of a whitespace after an open `[` .
Let me know if you want me to fix this.

shyuep:
Yes pls. All linting has to pass as well as the tests.

shyuep:
Many thanks.

PR #1826, opened 2020-04-14
Abstract Compatibility interface + self-contained Aqueous Compatibility class

### LINKED ISSUE #1823: Pourbaix diagrams: AqueousCorrection unexpected behavior re: hydrogen energy
**Describe the bug**
@JosephMontoya-TRI and I recently discovered a bug in the `AqueousCorrection` class (used by `MaterialsProjectAqueousCompatibility` and `MITProjectAqueousCompatibility`) that causes unexpected behavior when correcting the energy of aqueous H2. As a result, the hydrogen energy may vary depending on whether or not the `ComputedEntry` for H2 already has a correction applied (e.g., when downloading from MPRester).

**To Reproduce**
Below are three different ways of generating entries for Pourbaix diagram construction. These should all result in the same energy for H2, but they do not. Note that H2 should always be assigned an energy of -3.6019 eV/atom (the value in the `MPCompatibility.yaml` file.

**Method 1: Manually process MPRester entries**

```
compat = MaterialsProjectAqueousCompatibility()

with MPRester() as m:
    ion_ref_entries = m.get_entries_in_chemsys(["O", "H"])
    ion_ref_entries = compat.process_entries(ion_ref_entries)
```
(print statements omitted for clarity)
```
Entry ID 	 Correction 	 E_per_atom	Corrections
mp-632291	   -0.2112	   -3.6019	{'MP Gas Correction': 0.30189724724000033}
mp-1188177	   -0.2798	   -3.6019	{'MP Gas Correction': 1.3179495979200055}
mp-850274	   -0.2139	   -3.6019	{'MP Gas Correction': 0.2964079872400003}
```

**Method 2: Remove corrections, then manually reprocess MPRester entries**

```
compat = MaterialsProjectAqueousCompatibility()

with MPRester() as m:
    ion_ref_entries = m.get_entries_in_chemsys(["O", "H"])
    for e in ion_ref_entries:
        e.correction=0
    ion_ref_entries = compat.process_entries(ion_ref_entries)
```

```
Entry ID 	 Correction 	 E_per_atom	Corrections
mp-632291	   -0.0603	   -3.4509	{'MP Gas Correction': 0.30189724724000033, 'MP Aqueous Correction': -0.30189724724000033}
mp-1188177	   -0.1974	   -3.5195	{'MP Gas Correction': 1.3179495979200055, 'MP Aqueous Correction': -1.3179495979200055}
mp-850274	   -0.0657	   -3.4537	{'MP Gas Correction': 0.2964079872400003, 'MP Aqueous Correction': -0.2964079872400003}
```

**Method 3: compatible_only=False, then manually reprocess MPRester entries**

```
compat = MaterialsProjectAqueousCompatibility()

with MPRester() as m:
    ion_ref_entries = m.get_entries_in_chemsys(["O", "H"], compatible_only=False)
    for e in ion_ref_entries:
        e.correction=0
    ion_ref_entries = compat.process_entries(ion_ref_entries)
```

```
Entry ID 	 Correction 	 E_per_atom	Corrections
mp-632291	   -0.0603	   -3.4509	{'MP Gas Correction': 0.30189724724000033, 'MP Aqueous Correction': -0.30189724724000033}
mp-1188177	   -0.1974	   -3.5195	{'MP Gas Correction': 1.3179495979200055, 'MP Aqueous Correction': -1.3179495979200055}
mp-850274	   -0.0657	   -3.4537	{'MP Gas Correction': 0.2964079872400003, 'MP Aqueous Correction': -0.2964079872400003}
```

Method 1 produces the correct energy, while Method 2 and 3 fail. Method 3 (which parallels the implementation in `get_pourbaix_entries` is equivalent to Method 2. 

The reason Method 1 works is that the `ComputedEntry` returned by MPRester already has a non-zero energy correction. `AqueousCorrection` requires the energy of H2 to be already corrected in order to compute the correct final energy, but this is not guaranteed if you process entries "from scratch", as shown in Method 2 or set `compatible_only=False` as in Method 3.

Because of this bug, Pourbaix diagrams generated via `.get_pourbaix_entries` slightly understabilize hydrogen, which can result in hydrides disappearing from the diagrams. For example:

Method 2 /3: 

image

vs. the more correct Method 1:

image

**Additional context**
We are developing a fix to correct this and also to ensure that `MaterialsProjectAqueousCompatibility` is fully compatible with a forthcoming new set of energy corrections (See #1558 ).

rkingsbury:
## Abstract `Compatibility` interface

This pull request refactors `Compatibility` to define a common interface for pymatgen energy correction and referencing schemes. The goal of doing this is to enable more flexible correction schemes that do not necessarily require a list of `Correction` classes, and also to increase the robustness and transparency of the `Compatibility` classes by preventing double-counting and by documenting energy adjustments within the `ComputedEntry` instead of relying on `get_corrections_dict` or `explain`. 

This functionality is important for future enhancements and bugfixes to the Pourbaix scheme (see #1823 ), and to ongoing work with other energy referencing classes.

### Major changes

 * `Compatibility` is now an abstract base class, with required methods `process_entries` and `get_adjustments`.

 * `process_entries` gains a `clean` kwarg that optionally removes all previously-applied corrections / adjustment before executing

 * The old / existing Compatibility base class, which relies on an internal list of `Correction` classes, is renamed `CorrectionsList` and now inherits from the abstract class.

 * `ComputedEntry` gains an attribute `energy_adjustments` that stores information about all energy adjustments / corrections that have been applied to the `Entry` by means of lightweight `EnegyAdjustment` objects. 

 * `ComputedEntry.correction` is now calculated dynamically from the contents of `energy_adjustments`, enforcing transparency regarding the provenance of energy adjustments.

 * Manual setting of `ComputedEntry.correction` is still supported, and causes a `ManualEnergyAdjustment` object to replace the contents of `parameters.energy_adjustments`.

 * `ComputedEntry` `repr` method has been updated to list uncorrected energy, total corrections, and final energy in both eV and eV/atom. Each `EnergyAdjustment` is also listed.

## Self-contained Aqueous Compatibility class

This PR also replaces `MaterialsProjectAqueousCompatibility` with a new implementation to serve as an example, concrete implementation of the new `Compatibility`. This new class fixes several bugs (see #1823 ) and allows the energy referencing scheme to be applied to any set of DFT entries, regardless of the underlying solid DFT energy corrections.

### Major changes / features

 * `MaterialsProjectAqueousCompatibility` is now a standalone `Compatibility` whose sole purpose is to adjust DFT energies to free energies. It does NOT apply any of the solid energy corrections that are applied by `MaterialsProjectCompatibility`. Hence, it is meant to be used after the entries have been processed by another solid DFT energy correction scheme. Another `Compatibility` class can be passed as a kwarg to perform this pre-processing automatically.
 * No longer uses `AqueousCorrection`. However, this is still used by `MITAqueousCompatibility`
 * Energy adjustments are computed on the fly without reference to hard-coded values in `MPCompatibility.yaml`. To do this, the class requires three DFT energies (the ground-state energies of O2, H2O, and the total energy adjustments / corrections applied to H2O). These can be passed as kwargs to `__init__` or inferred from the entries passed to `process_entries`.
 * As a result of recalculating / refitting the energy adjustments, the H2 free energy has changed from a previous value of -3.6019 eV/atom to -4.05 eV/atom
 * Fitted DFT H2O free energy (not used by Pourbaix) changed from a previous value of -4.972 eV/atom to -5.26 eV/atom
 * The energy correction applied to hydrate compounds is now calculated in a more accurate way. This may slightly change the free energies of some compounds containing OH.
 * Added unittests for the complete `MaterialsProjectAqueousCompatibility` class. Previous tests only covered the underlying `AqueousCorrection` class and not the whole compatibility class.
 * Fixes #1823 .


mkhorton:
Hi @rkingsbury, can you talk a little bit about the logic here?

The current Compatability (now named MITMPCompatibility in this PR) class is almost an abstract base class itself, it just takes a list of `corrections` to apply, e.g. its `process_entry`, `get_explanation_dict` etc. methods are pretty general.

Explicitly making this an abstract base class seems fine, but default implementations for these methods seem sensible, and including e.g. default implementations for `process_entry` in the abstract class also seem sensible.

rkingsbury:
@mkhorton the idea here is to define the high-level interface for all `Compatibility` classes. The `MITMPCompatibility` works well as a base class for our existing approach to corrections, which assumes that we want to apply corrections all at once and that they can be computed independently. In that case, it makes sense to have default functionality like a list of corrections and a `get_corrections_dict` method in the base class.  

But the existing approach is cumbersome and error-prone when we want to apply energy adjustments sequentially (e.g., take corrected solid DFT energies and convert them to Gibbs free energies for Pourbaix diagrams). What I'm trying to do with this PR is facilitate a way to write Compatibility classes that don't have an underlying list of `Correction` classes. They will still have `process_entries`, `explain`, and `process_entry`, but the logic within the class is more flexible - it does not have to depend on a list of `Correction` classes. I should mention that @shyamd was consulted on the initial implementation as well. 

As I went about implementing one such `Compatibility` class to fix #1823 , I added some extra structure to this abstract implementation (see latest commits). The main change is to populate `Entry.data["Energy Adjustments"]` with the contents of the `corrections_dict` (or equivalent). This allows us to tell by inspection whether 

1. an Entry has already been corrected
2. what `Compatibility` classes have processed it, and 
3. whether the documentation in `Entry.data` accounts for all the corrections in `Entry.correction`.

Knowing these things is important for future energy referencing schemes in which we want to process `ComputedEntry` that already have energy corrections applied (as is the case in the AqueousCorrection / Pourbaix scheme).

I also added a `validate_corrections` method to ensure that all corrections are documented in `Entry.data` (and raise a warning if this is not the case). Finally, I now provide a new default `explain` method that performs this validation and lists corrections in both eV and eV/atom units.

I've included default implementations for `process_entries` and `process_entry` in `Compatibility`. However, the `process_entry` implementation is still marked as an abstractclass because it is intended to be overridden. The default code is just an example to illustrate that `.process_entry` needs to both validate the corrections before doing anything and also populate `Entry.data["Energy Adjustments"] before exiting.

This all works without changing anything about how the existing `Compatibility` subclasses behave. The only difference is that `ComputedEntry` will have their `.data` attribute populated with a dict documenting all the corrections that are applied.

Based on my experience the last few months, I believe these changes will lay a foundation for making our energy corrections code substantially more transparent and robust. It will also facilitate schemes like Pourbaix that require knowledge of previously-applied corrections

Let me know what you think. If this seems like the right direction to go, I'll add some tests for the new `Compatibility` class.

shyamd:
Abstract methods shouldn't have an implementation.

mkhorton:
> Abstract methods shouldn't have an implementation.

Not true... abstract classes certainly can have default methods, it is only the methods specifically marked as abstract that must be overridden .

mkhorton:
> I've included default implementations for process_entries and process_entry in Compatibility. However, the process_entry implementation is still marked as an abstractclass because it is intended to be overridden. [...] The default code is just an example to illustrate that .process_entry needs to both validate the corrections before doing anything and also populate `Entry.data["Energy Adjustments"]` before exiting.

If it's marked as abstract it's required to be overridden and should not have a default implementation.

If, however, you want to require validation in `process_entry`, that has to be enforced by the base class itself somehow.

Also, regarding naming, from your suggested changes it seems like the existing Compatability class is more a CompatabilityList class or similar, there seems to be nothing MP/MIT-specific in it that suggests it should be named `MITMPCompatibility`.

In general, I think improving the compatibility provenance is essential, and structural improvements are welcome, I just want to make sure this makes sense.

rkingsbury:
An example using a concrete implementation of `Compatibility` (not part of the PR):

The entry returned by `MPRester` has already been processed by `MaterialsProjectCompatibility`
```
>>> with MPRester() as m:
             my_entry = m.get_entry_by_material_id("mp-697111", compatible_only=True)
>>> my_entry.data["Energy Adjustments"]
{'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916}}
```

I now use a new class to apply Aqueous corrections on top of the existing ones:

```
>>> compat = MaterialsProjectAqueousCompatibility2020()
>>> compat.process_entries([my_entry])
[ComputedEntry mp-697111 - H8 O4
 Energy = -59.5408
 Correction = -6.3096
 Parameters:
 run_type = GGA
 is_hubbard = False
 pseudo_potential = {'functional': 'PBE', 'labels': ['H', 'O'], 'pot_type': 'paw'}
 hubbards = {}
 potcar_symbols = ['PBE H', 'PBE O']
 oxide_type = hydroxide
 Data:
 oxide_type = hydroxide
 correction_uncertainty = nan
 Energy Adjustments = {'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916}, 'MaterialsProjectAqueousCompatibility2020': {'Aqueous free energy correction': -3.500397110000003}}]
```
```
>>> compat.explain(my_entry)
The uncorrected energy of H8 O4 is -59.541 eV (-4.962 eV/atom).
The following energy adjustments have been applied to this entry:
	MaterialsProjectCompatibility:
		MP Anion Correction: -2.809 eV (-0.234 eV/atom)
	MaterialsProjectAqueousCompatibility2020:
		Aqueous free energy correction: -3.500 eV (-0.292 eV/atom)
The final energy after adjustments is -59.541 eV (-4.962 eV/atom).

```



This new Aqueous class currently has a bug in which reprocessing the entries will keep applying the correction on top of the old ones. This is where the `validate_corrections` method is valuable. If I process `my_entry` again, we get a warning:

```
>>> compat.process_entries([my_entry])
/Users/ryan/pymatgen/pymatgen/entries/compatibility.py:719: UserWarning: The provenance of the energy correction of -9.810 eV (-0.817 eV/atom) for entry mp-697111 is unknown! The contents of entry.data['Energy Adjustments'] contain total corrections of -6.310 eV (-0.526 eV/atom). Consider re-applying all energy corrections and proceed with caution.
  warnings.warn("The provenance of the energy correction of {:.3f} "

[ComputedEntry mp-697111 - H8 O4
 Energy = -59.5408
 Correction = -13.3104
 Parameters:
 run_type = GGA
 is_hubbard = False
 pseudo_potential = {'functional': 'PBE', 'labels': ['H', 'O'], 'pot_type': 'paw'}
 hubbards = {}
 potcar_symbols = ['PBE H', 'PBE O']
 oxide_type = hydroxide
 Data:
 oxide_type = hydroxide
 correction_uncertainty = nan
 Energy Adjustments = {'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916}, 'MaterialsProjectAqueousCompatibility2020': {'Aqueous free energy correction': -3.5003971099999998}}]
```

mkhorton:
For your new class, can you share the `.as_dict()` representation of the corrected ComputedEntry?

rkingsbury:
> For your new class, can you share the `.as_dict()` representation of the corrected ComputedEntry?

```
>>> my_entry.as_dict()
{'@module': 'pymatgen.entries.computed_entries',
 '@class': 'ComputedEntry',
 'energy': -59.54084689,
 'composition': defaultdict(float, {'H': 8.0, 'O': 4.0}),
 'parameters': {'run_type': 'GGA',
  'is_hubbard': False,
  'pseudo_potential': {'functional': 'PBE',
   'labels': ['H', 'O'],
   'pot_type': 'paw'},
  'hubbards': {},
  'potcar_symbols': ['PBE H', 'PBE O'],
  'oxide_type': 'hydroxide'},
 'data': {'oxide_type': 'hydroxide',
  'correction_uncertainty': nan,
  'Energy Adjustments': {'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916},
   'MaterialsProjectAqueousCompatibility2020': {'Aqueous free energy correction': -3.500397110000003}}},
 'entry_id': 'mp-697111',
 'correction': -6.309557110000003}
```

shyamd:
I suggest rather that "Energy Adjustments" and "correction_uncertainty" become root level keys and that rather than putting the list of corrections within the Compatability scope, they become root level
`'Energy Adjustments': {'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916}`
becomes
`'Energy Adjustments': {'MP Anion Correction': -2.80916}`

mkhorton:
Why not dump the compatibility class as MSON? There's not enough information in the dict otherwise. Also `energy_adjustments` not `Energy Adjustments`. This is essentially constructing something similar to a transformation pipeline, so you want a list of the actual compatibility classes themselves (e.g. what if they were run with non-default args).

rkingsbury:
> Why not dump the compatibility class as MSON? There's not enough information in the dict otherwise. Also `energy_adjustments` not `Energy Adjustments`. This is essentially constructing something similar to a transformation pipeline, so you want a list of the actual compatibility classes themselves (e.g. what if they were run with non-default args).

My main concern here is for the corrections to be human-readable and transparent about what corrections have been applied, rather than to create a complete tranformation pipeline. Wouldn't dumping the whole class be kind of messy? I'm not opposed to adding it to a separate .data dict key though

rkingsbury:
Regarding the abstract class, the only way I can think of to enforce the `validate_correction` and populating the dict is to separate the actual correction into a separate method (e.g., a class-level `get_correction`. I can do that, but I was trying to avoid it for the sake of simplicity. One of the things that I find challenging about the existing implementation is tracing the logic (and bugs!) through `process_entries` -> `process_entry` -> `get_corrections_dict` -> `get_correction`. 

Can y'all think of any other ways to enforce the overall structure of `process_entry` without specifying the logic for calculating `correction`?

rkingsbury:
> I suggest rather that "Energy Adjustments" and "correction_uncertainty" become root level keys and that rather than putting the list of corrections within the Compatability scope, they become root level
> `'Energy Adjustments': {'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916}`
> becomes
> `'Energy Adjustments': {'MP Anion Correction': -2.80916}`

@shyamd can I ask why?

rkingsbury:
@mkhorton since this is growing in scope, do you want to see a concrete implementation (e.g, the new AqueousCompatibility in my example) as part of this PR, or in a separate one?

mkhorton:
> My main concern here is for the corrections to be human-readable and transparent about what corrections have been applied, rather than to create a complete tranformation pipeline

You can choose how this is printed using `__str__`, the dict format is not intended to be human readable. If you want to improve provenance it needs to contain necessary information to re-run the compatibility logic.

>  since this is growing in scope, do you want to see a concrete implementation (e.g, the new AqueousCompatibility in my example) as part of this PR, or in a separate one?

Let's keep it in this PR since they affect each other and it will be easier to prototype.

shyamd:
@rkingsbury we're no longer treating them as arbitrary data that comes along with the `Entry` but rather actual pieces of data that are critical to the definition of the `Entry`. The data dict is really a place for us to throw things that we want to bring along but nothing in Compatability / PhaseDiagrams directly interacts with it.

shyamd:
I'm a bit opposed to dumping the Correction class into the Entry. For the transformation pipeline, the dump of the transformation serves two purposes: documentation of provenance and a means of reproducing that structure in a data minimal way. Our alternative would be to store each structure along the way with some minimal documentation provenance, which would quickly blow up in terms of data structure size.

For corrections, storing the name, maybe date, and the value is sufficient provenance. We can always take off the corrections and return back to the original energy. By breaking them out into which correction contributes to what we have reproducibility of each step. This doesn't guarantee reproducibility of the correction value, but neither does a dump of the Correction class. It might also make sense to make `correction` a dynamic property of `Entry` that relies on the `energy_adjustments` lists to make the dependence on the list of corrections a stronger.

rkingsbury:
> I'm a bit opposed to dumping the Correction class into the Entry. For the transformation pipeline, the dump of the transformation serves two purposes: documentation of provenance and a means of reproducing that structure in a data minimal way. Our alternative would be to store each structure along the way with some minimal documentation provenance, which would quickly blow up in terms of data structure size.
> 
> For corrections, storing the name, maybe date, and the value is sufficient provenance. We can always take off the corrections and return back to the original energy. By breaking them out into which correction contributes to what we have reproducibility of each step. This doesn't guarantee reproducibility of the correction value, but neither does a dump of the Correction class. It might also make sense to make `correction` a dynamic property of `Entry` that relies on the `energy_adjustments` lists to make the dependence on the list of corrections a stronger.

I like the idea of dynamically calculating `correction` from the dict; that will simplify the code quite a bit and improve robustness.

For the corrections dict, @shyamd am I understanding that what you're proposing would be 

`Entry.energy_adjustments = {{"MP Anion Correction: -4.2}, {"MP Aqueous Corrections: -3}}`

etc.?

shyamd:
Yeah, that is definitely the right idea, although a set of Dicts isn't allowed. Rather just have it as Dict[str,float]:
```
Entry.energy_adjustments = {"MP Anion Correction: -4.2,  "MP Aqueous Corrections: -3}
```
I don't see a reason why a correction should ever be applied twice, which the Dict will enforce with only one entry per key. 

We could add the PMG version but that maybe overkill. It's up to each correction class as to how it stores the individual components. The `CorrectionList` could just store one for each correction it has, while the Pourbaix correction might lump it all into one. Either way the point is to be able to separate out the components of the correction value. 

As for what is shown in `__repr__`, that is up to you. There is nothing wrong with being detailed and having both the corrected and uncorrected values. There is a diff between `__str__` and `__repr__`. I just forget what.

rkingsbury:
Separate question - for writing new tests for this class, should I:

1. Try to directly test the abstract class using something like `mock` OR
2. Test a subclass, e.g., `MaterialsProjectCompatibilty` OR
2. write a dummy subclass in the test file ?

and 

1. use `unittests' OR
2. use 'pytest' ?

I'm inclined toward No. 2 in both cases, but let me know what you would like to see.

rkingsbury:
> > ```
> > Entry.energy_adjustments = {"MP Anion Correction: -4.2,  "MP Aqueous Corrections: -3}
> > ```
> 
> 👍 makes sense
> 
> > We could add the PMG version but that maybe overkill. It's up to each correction class as to how it stores the individual components. The `CorrectionList` could just store one for each correction it has, while the Pourbaix correction might lump it all into one. Either way the point is to be able to separate out the components of the correction value.
> 
> So, if a set of dicts isn't allowed, how would a class store additional data in this structure? Wouldn't each class need its own dict if we wanted to store pmg version, etc. in addition to the correction itself? e.g.
> 
> {'MP Anion Correction': {"value": -4.2, "pmg_version": 2020-xx-xx, ...}}

It's also occurred to me that I think it would be good to have a record of whether an Entry has been processed by a specific class, even if the resulting correction is zero. I guess we can accomplish that by populating the dict with zero values when appropriate?

shyamd:
You're asking if there should be unit tests or integrations tests.
There should be unit tests that ensure a dummy class with specific functionality creates certain responses for the built-in methods in the Abstract class.
There should be integration tests that ensure these methods also return the right answer when in actual use in the MPCompatability class.

I will leave it up to @mkhorton  as per pytest vs unittest. I find it easier to write pytest tests.

If the entry is processed by a class but it is not a valid correction for that entry, it should probably be set to None. If the correction is valid and its just 0, then it should get set to 0. We want to differentiate between an actual 0, vs an effective 0.

mkhorton:
Catching up on the above ...

> For corrections, storing the name, maybe date, and the value is sufficient provenance. We can always take off the corrections and return back to the original energy. By breaking them out into which correction contributes to what we have reproducibility of each step. This doesn't guarantee reproducibility of the correction value, but neither does a dump of the Correction class. 

I agree there's definitely an argument to be made that the numbers alone are sufficient. The MSON dump does include `@version` though, so it makes it very easy to find out exactly what bit of code generated that correction number, and that's what I'm seeing as most valuable by dumping the class (the `@version` at the root level is presumably sufficient for this).

I see the argument that it makes the dict a bit heavy, I think an argument can be made either way.

>  I guess we can accomplish that by populating the dict with zero values when appropriate?

This would also be the benefit of explicitly storing the list of corrections as dicts, even if the value is zero.

> I will leave it up to @mkhorton as per pytest vs unittest. I find it easier to write pytest tests.

I prefer pytest, I stick to unittest in pymatgen for historical consistency, but perhaps it's time we switch. No strong opinions.

mkhorton:
Yes, I'm going to call it -- let's see a prototype with `energy_adjustments` as a list of the compatibility classes themselves (so that, when serialized, it's going to store these as their dict representations).

>   'correction_uncertainty': nan,

Separately, how is uncertainty going to be propagated when multiple corrections are applied? (and why `nan` not `None`?)

shyamd:
That breaks the definition of compatibility which is not defined on a per-entry basis.

mkhorton:
> That breaks the definition of compatibility which is not defined on a per-entry basis.

To be more specific, what I'm talking about here is e.g. the dict pasted above:

```
{'@module': 'pymatgen.entries.computed_entries',
 '@class': 'ComputedEntry',
 'energy': -59.54084689,
 'composition': defaultdict(float, {'H': 8.0, 'O': 4.0}),
 'parameters': {'run_type': 'GGA',
  'is_hubbard': False,
  'pseudo_potential': {'functional': 'PBE',
   'labels': ['H', 'O'],
   'pot_type': 'paw'},
  'hubbards': {},
  'potcar_symbols': ['PBE H', 'PBE O'],
  'oxide_type': 'hydroxide'},
 'data': {'oxide_type': 'hydroxide',
  'correction_uncertainty': nan,
  'Energy Adjustments': {'MaterialsProjectCompatibility': {'MP Anion Correction': -2.80916},
   'MaterialsProjectAqueousCompatibility2020': {'Aqueous free energy correction': -3.500397110000003}}},
 'entry_id': 'mp-697111',
 'correction': -6.309557110000003}
```

Keys like `MaterialsProjectAqueousCompatibility2020` are not sufficient to point to a class, dumping the `.as_dict()` gives us pointers (`@module`, `@class`, any other args/kwargs).

This is all I'm asking for -- this might not be sufficient to re-produce the number but it at least gives us a pointer to what was applied rather than just a bare string.

mkhorton:
The other advantage of this is that in a construction of a `PhaseDiagram` it'll make it easier to check that a consistent set of corrections were applied to all `entries` used.

shyamd:
Honestly, the provenance of the corrections has never been issue.  How complicated they are has been one, but never which is applied, since we can just reply what ever we want to enforce. I don't want to solve the provenance problem and make them even more complicated.

I also think you're mixing `Compatability` with `Correction`.  All of the parameters you want to track are in the `Compatability` classes which get set from code. These are predefined, so there is no need for Args or Kwargs. In the case of the Scan referencing scheme which will require the other entries to parametrized, we simply can't store all of this in one entry. So again, there is not much of a reason to store the `Compatability` class. 

I do agree we should store the corrections and there is a much better way of doing this. I'd rather see `Correction` be converted into something that is just to store that info and not as a means of computing it. For now, there would be two different kinds of correction classes: `CompositionCorrection` and `ConstantCorrection` that store the correction value or multiplier and a description. We can make the entry `correction_list` just those and when dumped via MSONable they serialize. This is far more descriptive and actually provide some traceability of the critical parameters. The `Compatability` classes will build `Correction` objects for each entry and store them.

rkingsbury:
> I do agree we should store the corrections and there is a much better way of doing this. I'd rather see `Correction` be converted into something that is just to store that info and not as a means of computing it. For now, there would be two different kinds of correction classes: `CompositionCorrection` and `ConstantCorrection` that store the correction value or multiplier and a description. We can make the entry `correction_list` just those and when dumped via MSONable they serialize. This is far more descriptive and actually provide some traceability of the critical parameters. The `Compatability` classes will build `Correction` objects for each entry and store them.

I like this idea. I didn't want to cause confusion by duplicating the name of the `Corrections` class, so I made a lightweight `EnergyAdjustment` class that stores a name, value, class name, and pymatgen version. So now the `repr` for an entry shows

```
ComputedEntry mp-632291 - H2
Energy = -6.7814
Correction = -0.6171
Energy Adjustments:
	MP Gas Correction: 0.302 eV (0.151 eV/atom)
	Fit H2 and H2O energy to experiment: -0.919 eV (-0.459 eV/atom)
	Compound entropy at room temperature: 0.000 eV (0.000 eV/atom)
	Hydrate energy adjustment: 0.000 eV (0.000 eV/atom)
Parameters:
	run_type = GGA
	is_hubbard = False
	pseudo_potential = {'functional': 'PBE', 'labels': ['H'], 'pot_type': 'paw'}
	hubbards = {}
	potcar_symbols = ['PBE H']
	oxide_type = None
Data:
	oxide_type = None
```

mkhorton:
Can I ask for the `as_dict` again? (Thanks!)

rkingsbury:
> Can I ask for the `as_dict` again? (Thanks!)

Here's an example for FeHO2 processed with `MaterialsProjectCompatibility`

```
{'@module': 'pymatgen.entries.computed_entries',
 '@class': 'ComputedEntry',
 'energy': -375.85317445,
 'composition': defaultdict(float, {'Fe': 16.0, 'H': 16.0, 'O': 32.0}),
 'energy_adjustments': [{'@module': 'pymatgen.entries.computed_entries',
   '@class': 'EnergyAdjustment',
   '@version': '2020.4.2',
   'name': 'MP Anion Correction',
   'value': -22.47328,
   'cls': 'MaterialsProjectCompatibility',
   'version': '2020.4.2'},
  {'@module': 'pymatgen.entries.computed_entries',
   '@class': 'EnergyAdjustment',
   '@version': '2020.4.2',
   'name': 'MP Advanced Correction',
   'value': -43.728,
   'cls': 'MaterialsProjectCompatibility',
   'version': '2020.4.2'}],
 'parameters': {'run_type': 'GGA+U',
  'is_hubbard': True,
  'pseudo_potential': {'functional': 'PBE',
   'labels': ['Fe_pv', 'H', 'O'],
   'pot_type': 'paw'},
  'hubbards': {'Fe': 5.3, 'H': 0.0, 'O': 0.0},
  'potcar_symbols': ['PBE Fe_pv', 'PBE H', 'PBE O'],
  'oxide_type': 'hydroxide'},
 'data': {'oxide_type': 'hydroxide'},
 'entry_id': 'mp-626570',
 'correction': -66.20128}


```

rkingsbury:
So arguably, the `.version` in `EnergyAdjustment` is redundant b/c it will be provided by serializing anyway

mkhorton:
Yeah, I agree `.version` probably redundant.

Ok, shall we leave it here for the weekend and have a call next week to discuss and hack on this a little? I'm concerned I might be missing some things that are obvious under my shelter-in-place cabin fever and think it might just be easier to talk it out :-)

rkingsbury:
@mkhorton sure. Would be good to get consensus among you, me, and @shyamd before I invest a lot of time into adding tests, etc. I'll set up a call.

The last commit was just some extra polish on `repr` for `ComputedEntry` and `EnergyAdjustment`.

pymatgen/entries/compatibility.py
```diff
@@ -407,58 +408,82 @@ def __str__(self):
...
+                                                 .format(entry.entry_id,
+                                                         ea.name,
+                                                         ea.value
+                                                         )
+                                                 )
+                else:
+                    # Add the correction to the energy_adjustments list
+                    entry.energy_adjustments.append(ea)
+
+            return entry
```
rkingsbury, line 483: @shyamd if we make `process_entry` an abstract method, how can we enforce that corrections aren't applied twice? See the logic here.

In principle this could go into `process_entries`, but then we could have a situation where `process_entry` gives you a different result than `process_entries`.

shyamd: I think you're confusing how `CorrectionList` works with `Compatibility`. Things that are specifically an issue in `CorrectionList` get taken care of there. This also becomes less of a problem if `Compatibility`'s don't become user composable: Apply one then any other and have them add up.

rkingsbury: @shyamd not sure I follow completely. The issue of corrections being applied twice is not specific to `CorrectionList`. I want to enforce, for any `Compatibility`, that once an entry is processed, it will not be processed again (or it will only be processed again if the previous correction is removed). You are right that the new infrastructure will handle adding corrections from multiple `Compatibility`, but there is nothing in that paradigm to prevent the SAME correction being added multiple times.

For example, (in pseudo code), I want the abstract class to guarantee that this situation won't happen:
```
>>> compat = Compatibility()
>>> entry = ComputedEntry()
>>> entry.correction
0
>>> compat.process_entry(entry)
>>> entry.correction
2
>>> compat.process_entry(entry)
>>> entry.correction
4
>>> compat.process_entry(entry)
>>> entry.correction
6
```

Related to this, I want to enforce that any `Compatibility` can accept a `clean` argument to specify whether or not previously-applied corrections are removed before processing. This would apply to ALL corrections on the entry, not just those added by the `Compatibility` class that is doing the processing.

For example (pseudo code):
```
>>> entry = ComputedEntry(... correction=2)
>>> entry.energy_adjustments
Manual adjustment: 2
>>> compat = Compatibility(clean=False)
>>> compat.process_entry(entry)
>>> entry.energy_adjustments
Manual adjustment: 2
Example energy adjustment: 5
>>> entry.correction
7
```
vs

```
>>> entry = ComputedEntry(... correction=2)
>>> entry.energy_adjustments
Manual adjustment: 2
>>> compat = Compatibility(clean=True)
>>> compat.process_entry(entry)
>>> entry.energy_adjustments
Example energy adjustment: 5
>>> entry.correction
5
```

If there is a way we can enforce these behaviors while eliminating `get_corrections_dict`, I'm all for it. I just haven't been able to figure out how to do that. Open to ideas!

shyamd: I think you're trying too hard to enforce something that is expected by whoever writes `process_entry`. You have to expect that whoever writes that knows what they are doing and in all-reality, might have a good reason for having two of the same correction. 

One thing you can enforce is that all corrections, except `ManualAdjustment` get stripped before going into `process_entry`

rkingsbury: Fair point, future developers will know what they're doing so to some extent we can trust them to implement these things. But doesn't this compromise the objective of defining a consistent interface for users (who may NOT know what they are doing)? 

If we don't enforce this, then the behavior of `process_entries` will be ambiguous. In some cases, it could be something that happens one time (sort of a boolean, like is this entry processed or not), but in others, it might be something that I can do over and over and keep changing the energy.

I think it's really important that we enforce one behavior or another in order to prevent unexpected behavior / bugs (this has tripped me up multiple times)

rkingsbury: > One thing you can enforce is that all corrections, except `ManualAdjustment` get stripped before going into `process_entry`
> 
> Edit: This would be in `process_entries`.

True, but if I do this, then `process_entry` could give me a different answer than `process_entries`. That's going to violate users expectations. If I process an entry with `process_entry` or `process_entries`, I need to wind up with the same energy.

rkingsbury:
@mkhorton It looks like we have some new pymatgen `pull_request` tests? I'm getting a failure that isn't related to the code I changed. (in `LarsenDimensionalityTest.test_calculate_dimensionality_of_site`)

mkhorton:
Yup, fixing them now -- we weren't testing on Python 3.8 in CI before this morning, and some tests depend on arbitrary orders of lists which now change.

mkhorton:
You will have to merge changes to master into your branch after tests in master pass again.

pymatgen/entries/compatibility.py
shyamd: Would that be the case for a `Scan` compatibility? I think what you're getting at is that `process_entry` is public so a user might use it; and I'm suggesting that maybe that's not the case. `process_entry` is an implementation method that gets used by `process_entries`.  `process_entries` is used in the MPRester and I think most of the example notebooks.

shyamd: You can even rename it to `_process_entry` if you want so that it's naming is that it is for internal purposes. @mkhorton  how do you feel about that?

rkingsbury: FYI, `process_entry` is used in three places in `emmet-old`: https://github.com/materialsproject/emmet/search?q=process_entry&unscoped_q=process_entry

rkingsbury: I'm ok with an explicitly private `_process_entry` being the abstract method and `process_entries` being public. I still think there is some risk of confusion if `_process_entry` doesn't do exactly the same thing that `process_entries` does, but we can manage that with docstrings, etc. 

The one caveat is that I want to allow `process_entries` to take in either a list of entries or a single entry, so that users can work on single entries without having to pass `[entry]`.

If this sounds good to you @shyamd and @mkhorton I'll proceed with that implementation.

shyamd: You can also call it something else from `process_entry`, but basically there should be one interface at the `Compatability` level that users use to get entries adjusted and one interface on the implementation side on where people put their code. Since we want to keep this as integrated with existing pymatgen as possible, it's important to keep `process_entries`. 

Having `process_entries` deal with both single and list of entries is perfectly reasonable.

shyamd: It's important that `process_entries` return be consistent though. It should always return a list of adjusted entries.

rkingsbury: Good morning guys. After sleeping on it I want to make one more counterproposal, because I'm still concerned about inconsistency between `_process_entry` and `process_entries` causing confusion for users and problems for maintenance / debugging.

What about the following, in which `adjust_energy` is the abstract method?
```
@abstractmethod
adjust_energy(Entry) -> [EnergyAdjustment]

_process_entry(Entry) -> Entry
    calls adjust_energy 

process_entries(Entry or [Entry]) -> [Entry]
    calls _process_entry 
```

This way we have only two public interfaces as @shyamd suggests, we enforce de-duplication and the `clean` argument in `_process_entry`, and we maintain complete consistency between `_process_entry` and `process_entries`.

Compared to what's shown now, I'm proposing to replace `get_corrections_dict` with `adjust_energy` that actually returns `EnergyAdjustment` instead of just a dict, and then making `process_entry` private.

Thoughts @shyamd and @mkhorton ?

shyamd: I'm mostly fine with this. There is no point in having `process_entry` be private then. It's a fully functioning public method. The point of making something private is not to hide it, but to ensure it only gets used in specified situations or manners.

rkingsbury: As I look at it, I suppose we could also eliminate `_process_entries` entirely in this implementation. I think my main concern is avoiding having two methods (private or not) that look like they do the same thing but have subtle differences. `_process_entry` and `process_entries` "look" like they should behave exactly the same way, just operate on different numbers of `Entry`. But we could just have `get_adjustments` and `process_entries`. What about that?

rkingsbury: That way, in addition to the semantic difference in the method names, they also clearly have different functions. One returns a list of adjustments, the other modifies the entry.

shyamd: That's fine with me.

pymatgen/entries/computed_entries.py
```diff
@@ -12,25 +12,78 @@
...
+    """
+    pass
+
+
+class TempEnergyAdjustment(EnergyAdjustment):
+    """
+    An energy adjustment applied to a ComputedEntry based on the temperature.
+    Used, for example, to add entropy to DFT energies.
+    """
+    pass
```
rkingsbury, line 80: @shyamd I've put a layer of abstraction in place for `EnergyAdjustmetnt` classes, but currently the child classes don't do anything. What did you have in mind for the different types of `EnergyAdjustment`?

I'm thinking we could make `.value` an abstract method/property, then define it as follows:

`ConstantEnergyAdjustment(value)` -> value = value (no change from base class)

`CompositionEnergyAdjustment(adj_per_atom, comp)` -> value = adjustment per atom * no. of atoms

`TempEnergyAdjustment(adj_per_deg, temp)` -> value = adjustment per degree * temperature in K

What are your thoughts?

shyamd: Yeah, that was definitely my idea.

I would also add `ManualAdjustment` as just a child of `ConstantAdjustment` with an updated docstring. This was you can filter it out.

pymatgen/entries/computed_entries.py
```diff
@@ -11,26 +11,187 @@
...
+                                                                                        self.temp,
+                                                                                        self.n_atoms,
+                                                                                        )
+
+    @property
+    def value(self):
+        """
+        Return the value of the energy correction in eV.
+        """
+        return self._value * self.temp * self.n_atoms
```
rkingsbury, line 188: @shyamd any thoughts on these `EnergyAdjustment` classes? I'm not real happy with this implementation yet. I don't see how to make a clear distinction between a `ConstantCorrection` and a `CompositionCorrection`, for example, because ANY kind of energy correction we apply will be scaled based on the number of atoms. (For example, if we normalize an entry with composition Fe4O6 to Fe2O3, all corrections, no matter what kind, should get cut in half). So in that sense, everything is composition dependent.

pymatgen/entries/computed_entries.py
```diff
@@ -11,26 +11,187 @@
...
+        """
+        Args:
+            value: float, value of the energy adjustment in eV
+            name: str, human-readable name of the energy adjustment.
+                (Default: Manual adjustment)
+            cls: str, the name of the Compatibility class used to generate the
+                energy adjustment. (Default: None)
+            description: str, human-readable explanation of the energy adjustment.
+        """
+        self._value = value
```
shyamd, line 46: No need to take in a value here.

shyamd, line 46: No need to take in a value here. `EnergyAdjustmet` just cares about name, cls, description and that any implementation provides a value prop of some type.

pymatgen/entries/computed_entries.py
```diff
@@ -11,26 +11,187 @@
...
+        self.name = name
+        self.cls = cls
+        self.description = description
+
+        @abc.abstractmethod
+        def value(self):
+            """
+            Return the value of the energy correction in eV
+            """
+            return
```
shyamd, line 56: I believe you can remove this `return` and it should be fine.

pymatgen/entries/computed_entries.py
```diff
@@ -11,26 +11,187 @@
...
+
+    @property
+    def value(self):
+        """
+        Return the value of the energy correction in eV.
+        """
+        return self._value
+
+    @value.setter
+    def value(self, x):
```
shyamd, line 93: There should be no setter here since this is constant adjustment.

pymatgen/entries/computed_entries.py
```diff
@@ -11,26 +11,187 @@
...
+        super().__init__(value, name, cls="None", description=description)
+
+    @property
+    def value(self):
+        """
+        Return the value of the energy correction in eV.
+        """
+        return self._value
+
+    @value.setter
```
shyamd, line 118: If you make ManualEnergyAdjustment inherit from Constant and add the setter, that will make the separation of capabilities a bit clearer.

pymatgen/entries/computed_entries.py
shyamd: In the current Entry, the corrections are absolute. The idea with these EnergyAdjustments is to add metadata on where and how they were computed. So in that sense, an EnergyAdjustment is tightly linked to the ComputedEntry it is contained in. It only has to represent that specific full composition, not any scaling. If you want to scale the composition the EnergyAdjustment has to be recomputed for the new composition as is the case right now.

pymatgen/entries/computed_entries.py
```diff
@@ -52,9 +214,9 @@ def __init__(self,
                 a string formula, and others.
             energy (float): Energy of the entry. Usually the final calculated
                 energy from VASP or other electronic structure codes.
-            correction (float): A correction to be applied to the energy.
-                This is used to modify the energy for certain analyses.
-                Defaults to 0.0.
+            energy_adjustments (dict): An optional list of EnergyAdjustment to
```
shyamd, line 217: no need for type annotations in docstrings if using them in the arg list.

pymatgen/entries/computed_entries.py
```diff
@@ -63,7 +225,16 @@ def __init__(self,
         """
         super().__init__(composition, energy)
         self.uncorrected_energy = self._energy
-        self.correction = correction
+        self.energy_adjustments = energy_adjustments if energy_adjustments else []
+
+        if correction:
```
shyamd, line 230: I suggest this instead.
``` python
        if correction > 0.0 and energy_adjustments is not None:
```

pymatgen/entries/computed_entries.py
```diff
@@ -86,21 +271,45 @@ def normalize(self, mode: str = "formula_unit") -> None:
                 normalizes such that the composition amounts sum to 1.
         """
         factor = self._normalization_factor(mode)
-        self.correction /= factor
         self.uncorrected_energy /= factor
+        for e in self.energy_adjustments:
```
shyamd, line 275: Is this new? It seems like an odd function.

pymatgen/entries/computed_entries.py
rkingsbury: But then I would miss the second part of the logic: if the correction >0 and energy_adjustments IS none, we have to set the correction:

```
if correction >0.0:
            if energy_adjustments is not None:
                raise ValueError("Argument conflict! Setting correction = {:.3f} conflicts "
                                 "with setting energy_adjustments. Specify one or the "
                                 "other.".format(correction))

            self.correction = correction
```

pymatgen/entries/computed_entries.py
rkingsbury: I shouldn't need a setter for `ManualAdjustment` either, right? Once the value is set it is what it is (other than scaling with composition)

pymatgen/entries/computed_entries.py
rkingsbury: OK, that makes sense. So we need to make sure that `normalize` can scale the values of every type of `EnergyAdjustment` (see comment below).

pymatgen/entries/computed_entries.py
rkingsbury: Yes it is new. As you noted above, whenever we scale the composition (as `normalize` does), we need to rescale all the `EnergyAdjustment` too. This is why I had added the setters. I'm having trouble figuring out how to make a combination of regular attrs and @property work here.

In order to make `value` and abstract method, it has to be a @property, right? In that case, I have to have setters for every `EnergyAdjustment` class to enable scaling. I've tried with a mixture of @property and regular attributes for the different classes but can't seem to make it work.

rkingsbury:
@shyamd @mkhorton I've just added a revamped `MaterialsProjectAqueousCompatibility` that is dynamic and self contained: all the corrections are computed on the fly, independently of the underlying solid DFT correction scheme. The class is parameterized with three energies that are needed to reference the solid energies to aqueous energies, and these parameters can either be passed as arguments to `__init__` or inferred from entries passed to `process_entries`. I've also written a more robust set of tests (there were actually never any tests for the Aqueous Compatibility class before, just the `AqueousCorrection`).

The implementation I have causes mypy to complain because some of these parameters are initially set to `None` Interested in your thoughts on how to work around that issue, and on the overall approach to this new style Compatibility implementation.

pymatgen/entries/compatibility.py
```diff
@@ -640,34 +759,224 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
+                   ), 6
+        )
+
+        # Free energy of H2O, fitted for consistency with the O2 and H2 energies.
+        self.fit_h2o_energy = round((2 * self.h2_energy +
+                                    (self.o2_energy - self.cpd_entropies["O2"]) +
+                                    self.MU_H2O
+                                     ) / 3,
+                                    6
+                                    )
```
rkingsbury, line 871: This is why mypy is unhappy. `get_adjustments` performs a check to make sure that none of the required parameters are `None` before executing the rest of the code (which would required floats), but mypy picks up the fact that these parameters could initially be equal to `None` and says you can't multiple or add `None` with floats. 

I have a hard time understanding mypy's logic in cases like this: there is no way that, for example, line 859 will execute if `self.h2o_energy` is `None`, but I still get the error.

mkhorton:
Really happy you had the chance to take a look at the aqueous corrections @rkingsbury !

Mypy errors can be rather cryptic. You have to make sure you have appropriate guard statements in place (checking something is not actually None before you try to perform a multiplication with it), you can also explicitly declare types of different variables. I'd have to take a longer look to make specific suggestions. 

Are there some circular imports happening between this and Pourbaix?

pymatgen/entries/compatibility.py
mkhorton: > I have a hard time understanding mypy's logic in cases like this: there is no way that, for example, line 859 will execute if self.h2o_energy is None, but I still get the error.

It is possible though? I can set `self.h2o_energy = None` and then call `.get_adjustments` ? This isn't using the class as intended, but it is possible -- so I think mypy is correct.

mkhorton: Oh, I see the `all` statement ...

mkhorton: It's possible mypy doesn't understand the `all` maybe (?) ... docs seems to suggest they use `assert x is not None` or similar (`if x is None: raise ...`). If these are required, why are they optional kwargs with None defaults anyway?

rkingsbury: > It's possible mypy doesn't understand the `all` maybe (?) ... docs seems to suggest they use `assert x is not None` or similar (`if x is None: raise ...`). If these values are required anyway, why do they optional kwargs with None defaults?

Those kwargs are optional to preserve flexibility, so that if a user prefers to just populate automatically by including entries for O2 and H2O when they call `process_entries`, they can do that instead of having to manually set the entries up-front.

rkingsbury:
> 
> Are there some circular imports happening between this and Pourbaix?

Not to my knowledge; although I would like to add something to `get_pourbaix_entries` that pre-populates the required kwargs (rather than getting them from entries) in order to avoid raising warnings

mkhorton:
Ok, was just wondering why the `from pymatgen.analysis.pourbaix_diagram import MU_H2O` import was where it was (and why it's getting assigned as an instance variable when it's a constant)

rkingsbury:
> Ok, was just wondering why the `from pymatgen.analysis.pourbaix_diagram import MU_H2O` import was where it was (and why it's getting assigned as an instance variable when it's a constant)

Yeah I just noticed that; it might make more sense to move MU_H2O into `compatibility.py` and import it into pourbaix just to keep the imports going in the same direction.

I initially tried importing it as a regular constant but I wasn't able to access it within the class. I didn't try very hard to troubleshoot, but thought it made sense to carry that value as an instance variable along with the other energies that are used to compute the corrections.

pymatgen/entries/compatibility.py
rkingsbury: An alternative would be to always determine the parameters in `process_entries` from the contents of the entries, but I figured some users may have scripts or workflows in which it would not always be convenient to include H2O and O2 entries every time.

rkingsbury: Also, because `get_adjustments` is public, there needs to be way of handling the situation where a user instantiates the class, then tries to call `get_adjustments` before having run `process_entries`. I think there is an argument for making `get_adjustments` a private method though, to discourage that type of use. In that case, always relying on `process_entries` to populate the kwargs would make more sense.

rkingsbury:
> Mypy errors can be rather cryptic. You have to make sure you have appropriate guard statements in place (checking something is not actually None before you try to perform a multiplication with it), you can also explicitly declare types of different variables. I'd have to take a longer look to make specific suggestions.

I was able to solve this by adding an else: statement.

pymatgen/entries/compatibility.py
```diff
@@ -407,9 +411,127 @@ def __str__(self):
...
+        if clean:
+            for e in entries:
+                # remove any previously documented corrections from the entry
+                e.energy_adjustments = []
+
+        # apply the corrections
+        for entry in entries:
+            try:
+                adjustments = self.get_adjustments(entry)
+                for ea in adjustments:
```
shyamd, line 471: I think you're doing too much checking here. There is some expectation that the implenter knows what they're doing.

pymatgen/entries/compatibility.py
```diff
@@ -407,9 +411,127 @@ def __str__(self):
...
+                    # Has this correction already been applied?
+                    if ea.name in [e.name for e in entry.energy_adjustments]:
+                        if ea.value in [e.value for e in entry.energy_adjustments]:
```
shyamd, line 474: What if you have two of the same correction with different values, but one of the values also is from another correction? That would fail this check.

pymatgen/entries/compatibility.py
```diff
@@ -407,9 +411,127 @@ def __str__(self):
...
+                            # we already applied this exact correction. Do nothing.
+                            pass
+                        else:
```
shyamd, line 477: I think you're best of with a simple check that just looks for duplicate corrections
Something like:
```python
adjustment_names = [ea.name for ea in adjustments]
if len(adjustment_names) != len(set(adjustment_names)):
    raise ComptabilityError(....)
else:
    entry.energy_adjustments = adjustments
```

pymatgen/entries/compatibility.py
```diff
@@ -407,9 +411,127 @@ def __str__(self):
...
+        processed_entry_list = []
+
+        if clean:
+            for e in entries:
+                # remove any previously documented corrections from the entry
+                e.energy_adjustments = []
+
+        # apply the corrections
+        for entry in entries:
+            try:
```
shyamd, line 469: This try should except should just be around getting the adjustments from the child implementation. Everything else after should be outside the try/catch.

pymatgen/entries/compatibility.py
```diff
@@ -407,9 +411,127 @@ def __str__(self):
...
+            are not compatible are excluded.
+        """
+        # convert input arg to a list if not already
+        if isinstance(entries, ComputedEntry):
+            entries = [entries]
+
+        processed_entry_list = []
+
+        if clean:
+            for e in entries:
```
shyamd, line 463: We should keep `ManualAdjustments` here.

pymatgen/entries/compatibility.py
```diff
@@ -640,34 +761,228 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
-                corrections are to be applied or not.
-            check_potcar_hash (bool): Use potcar hash to verify potcars are correct.
+            o2_energy: The ground-state DFT energy of oxygen gas, including any adjustments or corrections, in eV/atom.
+                Default: None
+            h2o_energy: The ground-state DFT energy of water, including any adjstments or corrections, in eV/atom.
+                Default: None
+            h2o_adjustments: Total energy adjustments applied to one water molecule, in eV/atom.
+                Default: None
+        """
+        from pymatgen.analysis.pourbaix_diagram import MU_H2O
```
shyamd, line 809: This should probably be defined here and imported into `pourbaix_diagram`

pymatgen/entries/computed_entries.py
shyamd: I would put this into the energy adjustment as an abstract method then. But in general, I don't think it's necessary. Any adjustment should only be valid within the context of the composition it applies to. If we want an adjustment for a similar composition, we should go through compatibility.

pymatgen/entries/compatibility.py
```diff
@@ -640,34 +761,228 @@ def __init__(self, compat_type="Advanced", correct_peroxide=True,
...
+
+        Returns:
+            [EnergyAdjustment]: Energy adjustments to be applied to entry.
+
+        Raises:
+            CompatibilityError if the required O2 and H2O energies have not been provided to
+            MaterialsProjectAqueousCompatibility during init or in the list of entries passed to process_entries.
+        """
+        adjustments = []
+        if not all([self.o2_energy, self.h2o_energy, self.h2o_adjustments]):
```
shyamd: Mypy doesn't understand this check. You should do explicit checks for float. Also since the rest of the code is in the else, you can just omit that:
``` python
if not all([isinstance(v,float) for v in [self.o2_energy, self.h2o_energy, self.h2o_adjustments]]):
   raise CompatabilityError(...)

self.h2_energy = ...
```

pymatgen/ext/matproj.py
```diff
@@ -492,8 +492,26 @@ def get_pourbaix_entries(self, chemsys):
...
+
+        h2o_entries = [e for e in ion_ref_entries if e.composition.reduced_formula == 'H2O']
+        h2o_entries = sorted(h2o_entries, key=lambda e: e.energy_per_atom)
+        h2o_energy = h2o_entries[0].energy_per_atom
+        h2o_adjustments = h2o_entries[0].correction / h2o_entries[0].composition.num_atoms
+
+        compat2 = MaterialsProjectAqueousCompatibility(o2_energy=o2_energy,
+                                                       h2o_energy=h2o_energy,
+                                                       h2o_adjustments=h2o_adjustments)
+        ion_ref_entries = compat2.process_entries(ion_ref_entries)
```
rkingsbury, line 514: @montoyjh I want to call your attention to this PR, which contains a new implementation of the Aqueous Correction scheme (see the first comment / summary). In particular, I wanted you to be aware of this change to `get_pourbaix_entries`. Let us know if you have any feedback!

rkingsbury:
@shyamd @mkhorton I updated the first comment with a detailed summary of the major changes so you can see them all in one place.

shyamd:
I still don't like that `MaterialsProjectAqueousCompatibility ` layers on corrections without knowing the underlying solid-correction scheme. I think that scheme should be an argument to `MaterialsProjectAqueousCompatibility `, that it then ensures is properly applied by running that schemes's `process_entries` with `clean=True`.

pymatgen/entries/compatibility.py
rkingsbury: The concern here is not that `get_adjustments` will return duplicate adjustment names (in fact, that's allowed b/c we assume that the implementer knows what they're doing), but rather that the adjustments returned by `get_adjustments` already exist in `ComputedEntry.parameters.energy_adjustments`. 

I want this block to ensure two behaviors when a user runs `process_entries`

1. If the exact same correction has already been applied, don't do anything. There's no need to raise a `CompatibiliityError` , as this might just have resulted from calling `process_entries` twice (e.g., user gets processed entries from MPRester and then tries to reprocess themself).

2. Raise a `CompatibilityError` if an adjustment with the same (name, cls) already exists but the values don't match. This indicates a definite problem indicative of unintended behavior or incorrect use of `Compatibility`.

pymatgen/entries/compatibility.py
rkingsbury: Per the abstract class, `process_entries` is supposed to catch all `CompatibilityError`. Since we want to raise a `CompatibilityError` when we find a duplicate correction, are you saying that duplicate checking should be in a separate try/except block?

In other words, we might get `CompatibilityError` from the child method because an entry is incompatible, or we might get a `CompatibilityError` within `process_entries` from duplicate checking.

pymatgen/entries/compatibility.py
rkingsbury: The checking is more for the user than the implementer. To avoid unexpected behavior, I think it's really important that we enforce `EnergyAdjustment` only being added once by any given class. I should be able to run `Compatibility.process_entries` 10 times over the same set of entries and get the same result as if I only ran it once. I know running 10 times would not be intended use, but it could easily happen inadvertently with different pieces of code talking to each other (e.g., user gets processed entries from MPRester and then tries to reprocess themselves)

shyamd: You're using one tool to accomplish a complex thing. Protecting users beyond sensible defaults is not easy. If the user wants to break your logic they will. The harder they have to fight your checks, the worse the break will be. It's best to have safe sensible defaults that will encourage the appropriate behavior, and if the user decides to do something stupid, that is on them.

In this case, you have another way to enforce good practices via `clean`. Set `clean` to be default `True` and document this in the docstrings. If the user or implementer decides to set `clean` to `False`, then they are on their own.

pymatgen/entries/compatibility.py
rkingsbury: See revised code based on comparing tuples of (name, cls, value) and (name, cls).

pymatgen/entries/compatibility.py
shyamd: Simple code = easy to follow and test. Complex code will breed bugs and even with your checks as I mentioned above you have a bug. 

You're working too hard to "fix" issues caused by implementations. If the implementor wants a correction twice, they can use enumerated names or something. 

If you insist on checking the name and value use a `tuple`:
```python
adjustment_tuples = [(ea.name,ea.value) for ea in adjustments]
if len(adjustment_tuples) != len(set(adjustment_tuples)):
    raise ComptabilityError(....)
else:

You can put the logic for finding which ones are duplicate within the `if` which encapsulates it away. 

If you don't want process_entry to raise an error, then don't. Purposefully throwing an error to just catch it and then remove it's `Exception` quality is awkward code. Use the logger and print out an error message instead.

File: pymatgen/entries/compatibility.py (hunk 15 above)
Comment (shyamd) [reply pymatgen/entries/compatibility.py:473]: If you don't want process_entry to raise an error, then don't. Purposefully throwing an error to just catch it and then remove it's `Exception` quality is awkward code. Use the logger and print out an error message instead.

File: pymatgen/entries/compatibility.py (hunk 12 above)
Comment (rkingsbury) [reply pymatgen/entries/compatibility.py:481]: I agree we can only put in so many guardrails for users doing 'stupid' things. But 'stupid' is different from 'not obvious and unintentionally wrong'. I want to provide enough of a safety net to prevent at least a few 'not obvious' unintended behaviors that have personally tripped me up. What do you think is the harm in too much checking?

Enforcing this via `clean=True` is going to conflict with your comment to preserve `ManualEnergyAdjustment`. To maintain backward compatibility, if I instantiate a `ComputedEntry` with `correction = XXX`, that creates a `ManualEnergyAdjustment`. If we preserve those when `clean=True`, then we're going to have problems with legacy entries.

**Comment by rkingsbury** [comment]
@mkhorton are the pylint tests new?

**Comment by rkingsbury** [comment]
> I still don't like that `MaterialsProjectAqueousCompatibility ` layers on corrections without knowing the underlying solid-correction scheme. I think that scheme should be an argument to `MaterialsProjectAqueousCompatibility `, that it then ensures is properly applied by running that schemes's `process_entries` with `clean=True`.

The reason I took that approach is that the Aqueous energy referencing framework is independent of the underlying solid DFT corrections, and I thought the structure of the class should reflect that. I also want to make sure it's flexible enough to accommodate other correction schemes in the future (like MP2020) , so that we don't have to duplicate the code for every scheme. But we can accomplish that with a `solid_compat` kwarg.

**Comment by mkhorton** [comment]
@rkingsbury not new, but new version released today so behavior might have changed

File: pymatgen/entries/compatibility.py (hunk 12 above)
Comment (shyamd) [reply pymatgen/entries/compatibility.py:481]: That sounds like a problem with `from_dict`. It should treat the `dict` correction as a ConstantCorrection. Only those coming in via `setter` should create a `ManualEnergyAdjustment`

**Comment by rkingsbury** [comment]
> @rkingsbury not new, but new version released today so behavior might have changed

There are a ton of errors being triggered from all over pymatgen, including lots of code I haven't touched.

File: pymatgen/entries/compatibility.py (hunk 15 above)
Comment (rkingsbury) [reply pymatgen/entries/compatibility.py:473]: That makes sense, but I'm having trouble finding an alternative. The reason for raising and catching the exception within `process_entries` was to ensure that, if we found a collision between two energy adjustments, that `Entry` would not be returned. One of the key features of our interface here is that  "if anything goes wrong when correcting an Entry, you get a `CompatibilityError`. That's why I originally put all of that logic into a single try/except block.

File: pymatgen/entries/compatibility.py (hunk 12 above)
Comment (rkingsbury) [reply pymatgen/entries/compatibility.py:481]: > That sounds like a problem with `from_dict`. It should treat the `dict` correction as a ConstantCorrection. Only those coming in via `setter` should create a `ManualEnergyAdjustment`

I thought one of the fundamental tenets of `as_dict` / `from_dict` was that you should get back exactly the same object when you call ` from_dict` ? Wouldn't this break that principle?

Comment (shyamd) [reply pymatgen/entries/compatibility.py:481]: The objects should be equivalent, but in this case we're updating an old representation to a new one, so its allowed. Backward compatibility is important when possible.

Comment (rkingsbury) [reply pymatgen/entries/compatibility.py:481]: Yes, definitely agree that we need to preserve backward compatibility here. I feel like a safer way around all this though is to have `clean=True` strip out `ManualEnergyAdjustment` along with any other. That's more straightforward to understand, and it avoids all this ambiguity about what "manual" really means. It seems to me that anytime as_dict / from_dict results in ANY change to the object, there is a huge potential for confusion and bugs.

I'll note that the current behavior of `MaterialsProjectCompatibility` is effectively the same as `clean=True`. So users should be accustomed to manually re-applying manual corrections if they reprocess entries.

File: pymatgen/entries/computed_entries.py [hunk 20]
```diff
@@ -11,26 +11,189 @@
...
+        """
+        Return the value of the energy adjustment in eV.
+        """
+        return self._value * self.n_atoms
+
+    def _normalize(self, factor):
+        self.n_atoms /= factor
+
+
+class TempEnergyAdjustment(EnergyAdjustment):
```
mkhorton, line 155: Sorry to nitpick, but can you change this to `TemperatureEnergyAdjustment`?

pymatgen/entries/tests/test_compatibility.py
```diff
@@ -72,8 +170,8 @@ def test_process_entry(self):
                                          'hash': '994537de5c4122b7f1b77fb604476db4'},
                                         {'titel': 'PAW_PBE O 08Apr2002',
                                          'hash': '7a25bc5b9a5393f46600a4939d357982'}]})
-        self.assertIsNone(self.compat.process_entry(entry))
-        self.assertIsNotNone(self.ggacompat.process_entry(entry))
+        self.assertIsNone(self.compat._process_entry(entry))
```
mkhorton, line 173: Why are we testing private methods? Is it necessary for this to become private?

pymatgen/entries/tests/test_compatibility.py
```diff
@@ -632,10 +731,102 @@ def test_ozonide(self):
...
+    -x- check compound entropies appropriately added
+    -x- check hydrate adjustment appropriately applied
+
+    Notes:
+        Argument values from MaterialsProjectCompatibility as of April 2020:
+            corrected DFT energy of H2O = -15.5875 eV/H2O (mp-697111) or -5.195 eV/atom
+            corrected DFT energy of O2 = -4.9276 eV/atom (mp-12957)
+            total energy corrections applied to H2O (eV/H2O) -0.70229 eV/H2O or -0.234 eV/atom
+    """
+    def test_MPAqeous_H_H2O_energy_with_args(self):
```
mkhorton, line 757: typo (also pep violation).. should've started a review rather than individual comments, apologies

rkingsbury: What exactly is the PEP violation in `test_MPAqeous_H_H2O_energy_with_args`? are function names not supposed to have capitals?

pymatgen/entries/tests/test_compatibility.py
rkingsbury: It's not strictly necessary, but I think it's wise to make `process_entry` private going forward, for two reasons. First, we've made `process_entries` flexible enough to take in either a single `Entry` or a list. That way people don't have to change their scripts based on the number of entries they're processing. Second, some of the more sophisticated energy adjustment classes (including the new Aqueous) really have to analyze groups of entries at a time, it isn't really feasible to calculate the correction for just one entry.

I'm happy to remove the tests, but it seemed to me that more tests are always better than fewer, and I wanted them there to make sure we didn't break backward compatibility (although I suppose we already are since we've made `process_entry` private).

mkhorton: Definitely don't remove the tests!

pymatgen/entries/tests/test_compatibility.py
rkingsbury: how about `test_h_h2o_energy_with_args`?

pymatgen/entries/computed_entries.py
rkingsbury: 👍

pymatgen/entries/tests/test_compatibility.py
rkingsbury: I could add a public method to the `CorrectionsList` abstract class, so that all the legacy `Compatibility` classes will still have it available. But I'd really prefer `process_entry` not be a part of the base `Compatibility` interface.

mkhorton: > First, we've made process_entries flexible enough to take in either a single Entry or a list. 

So should the tests not be equivalent to `.process_entries([entry])` ? And, in that case, should not `.process_entry` just do the same?

In general, not good form to have tests for private methods -- usually a sign the method should actually be public and documented.

> Second, some of the more sophisticated energy adjustment classes (including the new Aqueous) really have to analyze groups of entries at a time, it isn't really feasible to calculate the correction for just one entry.

Yes, makes sense.

rkingsbury: Yes, the tests are equivalent to `process_entries(entry)`, and that's exactly what `._process_entry` does "under the hood." Should I change the tests to use `process_entries`?

rkingsbury: Actually, the slight difference is that `process_entry` returns a single entry, whereas `process_entries` always returns a list.

rkingsbury: OK, I had forgotten that `.process_entry` was already a part of `CorrectionsList`. In that case I agree it makes sense to keep it public.

mkhorton: Ok, I think it makes sense to put `process_entry()` on the same class as `process_entries()` , i.e. the abstract base class, and have it call `.process_entries([entry])` by default (if this needs to be overridden by a subclass that's fine, and is as it should be).

I hear you that it doesn't make sense in certain cases, might need to be overridden elsewhere, etc., but the alternatives (making it private, or not having it in the same class as `process_entries`) doesn't make much sense to me. That way the tests can proceed as previously.

rkingsbury: OK, done.

mkhorton:
Thanks @rkingsbury !

mkhorton:
@rkingsbury looks like this PR might have broken Pourbaix on the MP website, tagging @tschaume too -- the current public version is fine, but we're holding off upgrading until we resolve it (haven't had a chance to look into yet but I'm hoping this is a fairly minimal fix, possibly not requiring any change to pymatgen)

pymatgen/ext/matproj.py
```diff
@@ -465,20 +466,20 @@ def get_entries(self, chemsys_formula_id_criteria, compatible_only=True,
             entries = sorted(entries, key=lambda entry: entry.data["e_above_hull"])
         return entries
 
-    def get_pourbaix_entries(self, chemsys):
+    def get_pourbaix_entries(self, chemsys, solid_compat=MaterialsProjectCompatibility()):
```
mkhorton, line 469: @tschaume note the method signature changed, but the default should be fine

tschaume: @mkhorton yes that's part of the two minor changes I made in pymatpro to make the Pourbaix diagram at least show up again :P The other one is `energy` -> `_energy`. See https://github.com/materialsproject/pymatpro/commit/8aa0bac0779beb4d5504de4b80054add48dda447

mkhorton: Thanks @tschaume! Thought this PR was mostly backwards-compatible, so my main interest (aside from fixing) is getting an appropriate test added to pymatgen so we know if we break anything the website depends on

rkingsbury: @tschaume @mkhorton I can look into this tonight / tomorrow morning. Can you give me more specifics about what the problem is and/or open an issue in django?

I agree the PR should have been backwards-compatible, but I know a lot of the stuff in pymatpro (esp. related to Pourbaix) is quite archaic. What version of pymatgen is the current version of the website running?

shyamd: No, this is fine. There's no way to make a test in pymatgen for the behavior in pymatpro. That is a dumpster of a code-base that I don't think we should be pinning behavior in pymatgen to. Rather pymatpro needs to be updated with respect to pymatgen. If there is a test for pymatpro working properly, that has to be in pymatpro.

pymatgen/ext/matproj.py
```diff
@@ -492,7 +493,12 @@ def get_pourbaix_entries(self, chemsys):
         ion_ref_entries = self.get_entries_in_chemsys(
             list(set([str(e) for e in ion_ref_elts] + ['O', 'H'])),
             property_data=['e_above_hull'], compatible_only=False)
-        compat = MaterialsProjectAqueousCompatibility("Advanced")
+
```
rkingsbury, line 496: I suspect we need to add this logic to `DbPourbaix`.

rkingsbury: Actually, I think the quickest fix is to pass `solid_compat=MaterialsProjectCompatibility()` as an arg to `MaterialsProjectAqueousCompatibility` when calling `DbPourbaix`. I'll push that change

rkingsbury: https://github.com/materialsproject/pymatpro/commit/e8577128a42d6b780c7a3b7aaac4b81367fc05f0

rkingsbury: @tschaume let me know if my last commit fixes the issues. There's a lot of duplicated code between pymatgen, pymatpro, and django re: construction of Pourbaix Diagrams, so I think this was just a case in which we updated MPRester but didn't duplicate the updates in pymatpro. I'll look into consolidating this so that future releases won't have these issues like these.

tschaume: @rkingsbury thank you! Unfortunately, the Fe-Cr Pourbaix Diagram still looks off:

Materials_Project

rkingsbury: @tschaume can you try again with the latest? I think there was a problem with how `DbPourbaix` was constructing a temporary entry

tschaume: @rkingsbury yay! This looks sensible :) Only had to fix the `ComputedEntry` import from `pymatgen.entries.computed_entries`. I'll build a new image and we'll try to deploy again. I'll get back to you if any other issues pop up. Thanks!!

Materials_Project

mkhorton: Thanks both for addressing this + fixing so quickly!

PR #1832, opened 2020-04-20
Get neighbors save memory

chc273:
## Summary

This is a minor change to the previous `optimization.neighbors.find_points_in_sphere` cython method to save memory. 

The critical step for finding neighbors is the generation of candidate atoms that are potentially within the cutoff radius of some atoms. This is achieved by repeating the unit cell in 3d and generating image cells. 

In the previous approach, the memory of the maximum number of atoms `(n_atom_in_unit_cell * n_cell_images)` is allocated first and then those atoms that are beyond the cutoff are filtered out. This introduces a potentially large memory consumption in the beginning.

In the revised methods, I looped through the lattice offset vectors and gradually increase the number of candidate atoms and hence the memory of the calculation is smaller.


mkhorton:
Thanks @chc273 !

PR #1834, opened 2020-04-21
KPathLM labels bug fix

munrojm:
## Summary

This PR fixes a small bug in the `KPathLM` class where the code would hang when assigning labels to each of the points in a k-path for some complex cases.

### DIFF
--- a/pymatgen/symmetry/kpath.py
+++ b/pymatgen/symmetry/kpath.py
@@ -2195,6 +2195,8 @@ class KPathLatimerMunro(KPathBase):
                         if max_cosine_orbits_copy[grouped_ind][j][0] not in initial_max_cosine_label_inds:
                             next_choices.append(max_cosine_orbits_copy[grouped_ind][j][1])
                             break
+                        else:
+                            j += 1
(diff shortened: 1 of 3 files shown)


mkhorton:
Thanks @munrojm. Is there an existing KPath labels test this case could be added to?

munrojm:
> Thanks @munrojm. Is there an existing KPath labels test this case could be added to?

Good point! I will make one and update the existing test.

mkhorton:
Should be fixed now if you merge latest changes from master. Seems to be a _crash_ in pylint for something that seems correct to me -- will revisit in a future version of pylint assuming it's a bug, for now have excluded this specific version from our CI. Thanks for your patience with this.

munrojm:
> Should be fixed now if you merge latest changes from master. Seems to be a _crash_ in pylint for something that seems correct to me -- will revisit in a future version of pylint assuming it's a bug, for now have excluded this specific version from our CI. Thanks for your patience with this.

Not a problem, that sounds good. Thanks @mkhorton!

mkhorton:
Thanks @munrojm !

Issue #1841, opened 2020-05-02
Comment in KPOINTS generator


arosen93:
**Describe the bug**
When using [`pymatgen.io.inputs.vasp.Kpoints.automatic_density()`](https://github.com/materialsproject/pymatgen/blob/v2020.4.29/pymatgen/io/vasp/inputs.py#L1237) (and likely related functions), the [comment line](https://github.com/materialsproject/pymatgen/blob/v2020.4.29/pymatgen/io/vasp/inputs.py#L1256) is

```python
comment = "pymatgen v%s with grid density = %.0f / atom" % (__version__, kppa)
```

I am having some self-doubt on whether this is _strictly speaking_ the best way to write this, specifically the "/ atom". For instance, if I set `kppa=1000`, it writes out "... grid density = 1000 / atom". Is that better described 1000/number of atoms, as written out in the [Materials Project documentation](https://materialsproject.org/docs/calculations)? Should it instead be written as 1000 per reciprocal atom, as in [the JHU module](https://pymatgen.org/pymatgen.ext.jhu.html)? 1000/per atom always confuses me, but I see that kind of phrasing in [the VASP documentation](https://cms.mpi.univie.ac.at/vasp/vasp/Number_k_points_method_smearing.html). Regardless of the best way to do this, my question really boils down to; "is 1000/atom" the ideal way to phrase this, and is it strictly accurate? A user may see that and think "so, if I increase the number of atoms, the number of k-points goes up," which of course is the opposite.

I know this is me just being pedantic and isn't really a bug, but I thought it'd be worth bringing up here.

mkhorton:
Hi Andrew; agreed, I think this could be better phrased. A pull request is welcome if you'd like to change it!

As a general comment, a lot of the default MP sets are kept as they are for historical, backwards compatibility reasons, including how we determine k-points. For our newer sets (e.g. SCAN) we're looking at KSPACING or one of the generalized schemes (e.g. the Mueller/JHU scheme) anyway. Regardless, the comments for the kppa scheme should be accurate and unambiguous, and I agree the current comment string is misleading.

arosen93:
Alright, I'll make a PR. I just wanted to make sure I wasn't misunderstanding it!

[...]

PR #1842, opened 2020-05-02
Clearer description in generated KPOINTS comment line

arosen93:
Change k-points / atom to k-points / number of atoms

## Summary
As mentioned in https://github.com/materialsproject/pymatgen/issues/1841, the comment line in the Pymatgen-generated VASP KPOINTS file is somewhat misleading. This PR changes it from being # kpoints/atom to # kpoints/number of atoms in the comment line, which is also consistent with what's on the [Materials Project documentation](https://materialsproject.org/docs/calculations). This should make it clearer that the number of k-points would be higher for smaller systems.

Perhaps worth noting: there is still the very minor ambiguity in that [pymatgen.ext.jhu](https://pymatgen.org/pymatgen.ext.jhu.html) takes `kppra` as an input argument, whereas [`pymatgen.io.vasp.inputs.Kpoints.automatic_density()`](https://github.com/materialsproject/pymatgen/blob/v2020.4.29/pymatgen/io/vasp/inputs.py#L1237) takes `kppa` as an input argument, but obviously the keywords should not be adjusted, and I think the documentation for the latter (which simply states `kppa` = grid density) is accurate without being confusing. Therefore, I did not make any changes to the documentation.

### DIFF
--- a/pymatgen/io/vasp/inputs.py
+++ b/pymatgen/io/vasp/inputs.py
@@ -1253,7 +1253,7 @@ class Kpoints(MSONable):
         Returns:
             Kpoints
         """
-        comment = "pymatgen v%s with grid density = %.0f / atom" % (__version__, kppa)
+        comment = "pymatgen v%s with grid density = %.0f / number of atoms" % (__version__, kppa)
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks!

PR #1847, opened 2020-05-12
File update

rajeshsharma98:
###  Summary
Depending on the element, the electronic configuration may be very long. Because of this, scientists have developed a shorthand notation that involves using a noble gas to represent electrons that are not valence electrons. This function does the same, it simplifies the electronic configuration for better understanding.

Function Name: nobelgas_electronic_structure()

Example-  for 'Fe' this function will output:  ['Ar', (3, 'd', 6), (4, 's', 2)]

### DIFF
--- a/pymatgen/core/periodic_table.py
+++ b/pymatgen/core/periodic_table.py
@@ -280,7 +280,7 @@ class Element(Enum):
 
             Mendeleev number from definition given by Pettifor, D. G. (1984).
             A chemical scale for crystal-structure maps. Solid State Communications,
-            51 (1), 31-34 
+            51 (1), 31-34
 
         .. attribute:: electrical_resistivity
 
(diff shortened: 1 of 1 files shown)


shyuep:
I don't understand the difference between this and Element.Si.electronic_structure?

mkhorton:
It looks like `Element.Si.electronic_structure` contains HTML tags, so is not as versatile?

rajeshsharma98:
Yes @mkhorton, this was the only reason I created this function.
As for 'Fe'; _Element.Si.electronic_structure_  output contains HTML tags  whereas my function(noblegas_electronic_structure) gives output: ['Ar', (3, 'd', 6), (4, 's', 2)].

**Drawbacks** of using - _Element.Si.electronic_structure_
1. its output is not easily understandable because of tags.
2. its output is of string type that means it needs some time and few lines of code to fetch a particular part of the output.
3. output format is not useful.

**Improvements I made** with the development of noblegas_electronic_structure function:
1. one can easily understand the output.
2. output formatting of full_electronic_structure function and my function is same.
3. output is of list type, which makes its implementation easier.
4. its(noblegas_electronic_structure) name also suggests that it gives a similar or comparable output as compared to _full_electronic_structure_ function.

shyuep:
We are not creating a function just to remove html tags. By all means remove the html tags and make the the output of electronic structure.

mkhorton:
Yes, I'd agree -- I think it'd be easier to address this just by changing the underlying data in the periodic_table.json file, e.g. see https://github.com/materialsproject/pymatgen/blob/4e138b9f0d1caf040aabc7799875809a4596a05a/pymatgen/core/periodic_table.json#L36

rajeshsharma98:
I thought creating a function would be an easy task compared to changing the values, as function is just manipulating the output of full electronic structure function. Means it’s implementation and use will be easy. 
This was the only reason I created the function , as I am working on a project that requires noble gas configuration, mainly the electrons that are not valanced.

shyuep:
Actually, modifying the file is not necessary at all. What you want can be achieved with a single line of code:
```python
re.sub("<[^>]*>", "", Element.Si.electronic_structure)
```
gives
```
'[Ne].3s2.3p2'
```

PR #1854, opened 2020-05-26
[WIP] Adding interpolation functionality to VolumetricData class

ayushsgupta:
## Summary
* Adds the ability to interpolate between discrete data points
* Adds the `val_at` method, which accepts three coordinates and returns the data value at the associated point (possibly interpolated)
* Adds the `linear_slice` method, which gives a linear slice of the volumetric data between two specified points, with a specified resolution

* Uses RegularGridInterpolator from scipy.interpolate

## TODO
* Should there be some mechanism to raise a warning whenever a data point is interpolated as opposed to directly taken from data? There's a little bit of that (3222-3224, 3241) but not sure about the best implementation
* Is there a better way of checking list length than the assert statements on lines 3258-3259?
* If no warnings are necessary, then `val_at` and `_interpolate` can be combined
* Need to add testing!
* Add type checking for method arguments, specify return types
* Remove extra comments once everything's done


mkhorton:
Hi @ayushsgupta , thanks for this, this looks useful.

Do you think you could add a method to take a slice of a plane, returning a 2D numpy array? This could make use of the functionality in `Lattice` to provide a plane as a `(h, k, l)` tuple. If this feature would not be useful for your work however, please disregard this request.

ayushsgupta:
Hi @mkhorton, I think that would be a very good complement to what's here now. I did do some plane slicing, however it was limited to getting data on the `(x, 0, 0)`, `(0, x, 0)`, and `(0, 0, x)` planes only, not something like `(1, 1, 1)` or `(0, 2, 1)`. The first case is easier to add, but not quite sure how to do the second, as it would produce a possibly non-rectangular 2d array (?). I can look into it some more - what methods in `Lattice` would be helpful? I saw `get_miller_index_from_coords`, is there an inverse?

ayushsgupta:
Also the CircleCI tests are failing, but it seems to be unrelated to the code I edited. The problem comes from some imports from sympy to `pymatgen.alchemy` and `pymatgen.analysis`

mkhorton:
Hi @ayushsgupta, apologies for letting this PR sit here, everything looks great, thanks!

mkhorton:
Can I confirm from you that this is no longer a [WIP] before I merge?

ayushsgupta:
Hi @mkhorton no worries. I want to hold off for just a sec because I know @rkingsbury had some suggestions (change method name) and I want to make sure I covered it all. Also if you think that no warning needs to be raised when data is interpolated then I can combine two methods.

mkhorton:
I wouldn't raise a warning, I think that's fine and clear from the docstring that interpolation is used.

ayushsgupta:
@mkhorton ok, I've made all changes that I can think of. Build errors seem to come from the `Ion` class in `pymatgen.core`

mkhorton:
Thanks Ayush, looks good! The IonEntry failures are known.

mkhorton:
Could we have a test here too?

ayushsgupta:
Ok. There don't seem to be any current tests for the VolumetricData class, so do you want me to create one or test the interpolation through a different class (elfcar, chgcar)?

mkhorton:
Testing on one of the inherited classes is fine, thanks

ayushsgupta:
@mkhorton done!

mkhorton:
Thanks @ayushsgupta!

Issue #1858, opened 2020-05-29
Elfcar.as_dict() from pymatgen.io.vasp.outputs not working


ayushsgupta:
**Describe the bug**
I could be doing something wrong, but the `as_dict()` method of the `Elfcar` class doesn't appear to work. `Elfcar` ultimately inherits `as_dict()` from `monty.MSONable`, which requires that every argument passed to the constructor be an attribute of that class. However, one argument for the `Elfcar` constructor is `poscar`, but in the super call, `Elfcar` passes up `poscar.structure` to the `VolumetricData` constructor and does nothing further with `poscar`, so there isn't any `poscar` attribute associated with `Elfcar`. So `as_dict()` tries looking for a `poscar` attribute that doesn't exist, and a `NotImplementedError` is [raised](https://github.com/materialsvirtuallab/monty/blob/952ea44164d22767809374d04e8b9d6c41e13cd7/monty/json.py#L142). A simple fix could be to include `self.poscar = poscar` in the `Elfcar` constructor.

**To Reproduce**
`any_elfcar_object.as_dict()`
In my case, I created the object by calling `Elfcar.from_file()` on a gz file.

**Expected behavior**
Expected that a dictionary object would be returned.

**Screenshots**
![elfcar_asdict_error](https://user-images.githubusercontent.com/16586331/83303014-31805a00-a1b1-11ea-8c47-32a2da961184.PNG)

**Desktop (please complete the following information):**
 - Windows 10

PR #1859, opened 2020-05-29
Add poscar attribute to Elfcar modeled after Chgcar

ayushsgupta:
## Summary
* Add a `poscar` attribute to the `Elfcar` class patterned after the `Chgcar` constructor
* Fixes bug described in #1858 
@rkingsbury @mkhorton

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -3364,7 +3364,7 @@ class Chgcar(VolumetricData):
     def __init__(self, poscar, data, data_aug=None):
         """
         Args:
-            poscar (Poscar): Poscar object containing structure, or Structure object.
+            poscar (Poscar or Structure): Object containing structure.
             data: Actual data.
             data_aug: Augmentation charge data
         """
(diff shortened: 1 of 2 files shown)


pymatgen/io/vasp/outputs.py
```diff
@@ -3419,7 +3419,15 @@ def __init__(self, poscar, data):
             poscar (Poscar): Poscar object containing structure.
```
rkingsbury, line 3419: Can you modify the docstring to reflect the fact that you can also pass a `Structure` ?

rkingsbury:
Thanks for identifying the bug and for the fix, @ayushsgupta ! Can you add a test to verify that `elfcar = elfcar.from_dict(elfcar.as_dict())` ?

ayushsgupta:
@rkingsbury ok I added tests & updated the docstring. Did you want the new tests in a different method or are they fine as is?

rkingsbury:
> 
> 
> @rkingsbury ok I added tests & updated the docstring. Did you want the new tests in a different method or are they fine as is?

Hi @ayushsgupta , I think the test is great where it is. For the docstring, we typically include all types in parentheses (unless we're using type hinting, which we are not in this module, yet). So the docstring should read

```
poscar (Poscar or Structure): Object containing structure.
```

rkingsbury:
Other than that, I think this is ready for review by @mkhorton . You might want to do another git pull in case that resolves the issues with the CircleCI test failure (not sure if that is fixed yet in master). Then you can go ahead and remove the [WIP].

mkhorton:
Thanks @ayushsgupta and @rkingsbury!

CircleCI has been failing recently, going to look into it today.

PR #1866, opened 2020-06-08
Fix appveyor

mturiansky:
I assume it has the same problem as CircleCI with conda sympy.


mkhorton:
We're probably overdue removing AppVeyor now anyway, we used it because it offered Windows testing, but we can do that within Github Actions now and reduce our CI overload.

mturiansky:
Yeah, probably worth it. I was just annoyed that the appveyor badge said the build was failing :-)

mkhorton:
Haha, I admire your diligence

PR #1871, opened 2020-06-11
Add "unit" to plot_compare in PhononBSPlotter

ab5424:
The frequency unit can only be changed for a single bandstructure plot: https://github.com/materialsproject/pymatgen/blob/604d59c93f21c4a180663bdc6cd64bf1329c5239/pymatgen/phonon/plotter.py#L329 This is missing for plot_compare. I also provided a test (although it is not very useful to compare a bandstructure object with itself).


JaGeo:
Hello, 

is there anything else we would need to do before this can be merged? I think, only py3test is currently failing but this has nothing to do with this change.

Best, 
JG

mkhorton:
Thanks @ab5424! And apologies both @ab5424 and @JaGeo, been a little slow reviewing PRs this week. We currently have a test failing on master that needs to be addressed, it is unrelated to this PR.

JaGeo:
Thank you very much for merging, @mkhorton !

Issue #1875, opened 2020-06-22
XYZ File Parser can incorrectly handle comment string, miscounting the number of sites


stevetorr:
**Describe the bug**
For some XYZ files, the parser will incorrectly use the end of the comment line as the number of sites present on the molecule, which can result in an `IndexError` if this exceeds the number of sites on the Molecule. (This could presumably also result in fewer atoms being loaded in then should be, but the 'too many sites' case is how I caught the bug).

**To Reproduce**
A minimal example is as follows (I included a case which works, if it helps with debugging):

```
xyz_string_good = """17
gdb 73000       2.52453 1.59492 1.2866  1.942   70.04   -0.2468 -0.0322 0.2146  1026.4232       0.133427        -458.930556     -458.92199      -458.921046     -458.963286     32.347
C        0.0518622525    1.4677402164    0.1677182445   -0.508727
C        0.149931291    -0.0198512527    0.1549183945    0.387141
O       -1.0420748387   -0.7755524938   -0.1098982654   -0.333377
C       -0.6310443779   -1.7822217642    0.8717895672   -0.126641
C        0.6995752769   -1.0823398662    1.0716407643   -0.218953
C        1.3176925774   -0.8886070534   -0.3223080253    0.261566
O        1.126586626    -1.8073146011   -1.3481691631   -0.482546
C        2.6622014749   -0.2647569257   -0.3439137992    0.098299
O        3.1033556604    0.4506871538    0.5272840719   -0.280588
H        1.022706895     1.903540259     0.4101533166    0.164412
H       -0.2702279642    1.8391291673   -0.8111459282    0.133684
H       -0.6788048518    1.7939332661    0.9141342159    0.133712
H       -1.290366036    -1.7421645256    1.743319629     0.126989
H       -0.6147407007   -2.7905570041    0.4474086287    0.123682
H        1.2938641569   -0.9969950588    1.9709215363    0.126881
H        0.2057168389   -1.7109885321   -1.6301219799    0.297897
H        3.2419713095   -0.515832035    -1.256903708     0.096568
[...]

PR #1876, opened 2020-06-23
Add support for *^ scientific notation for XYZ parser (Useful for QM9 Database)

stevetorr:
## Summary

The QM9 database uses a special format for scientific notation in their .xyz files, wherein e.g. `3.14e-6` would be represented `3.14*^-6`. This addresses issue #1875 (which was not a bug per se, but simply brushing up against unimplemented behavior).

- The XYZ file parser can now handle this formatting for coordinates.
* This means changing the regular expressions which detect coordinate lines, as well as a substitution from the '*^' notation to 'e' notation.
- This capability is unit tested.

## Checklist

      on your local machine.
- [N/A] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [N/A] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) 
      to type check your code.
- [X] All existing tests pass.

allowing commits.

### DIFF
--- a/pymatgen/io/tests/test_xyz.py
+++ b/pymatgen/io/tests/test_xyz.py
@@ -83,6 +83,16 @@ C 1.16730636786 -1.38166622735 -2.77112970359e-06
         self.assertTrue(abs(mol[0].z) < 1e-5)
         self.assertTrue(abs(mol[1].z) < 1e-5)
 
+        mol_str = """2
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @stevetorr!

PR #1877, opened 2020-06-23
change setup.py

JaGeo:
## Summary

Hi, 

I forget to change the setup.py to reflect the correct folder structure of the new lobster module. Currently, the basis functions are not found when pymatgen is installed... (Is it maybe possible to check for this within an automatic test in the future?)

Best, 
JG

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -128,6 +128,7 @@ setup(
         "pymatgen.io": ["*.yaml"],
         "pymatgen.io.vasp": ["*.yaml", "*.json"],
         "pymatgen.io.lammps": ["templates/*.*", "*.yaml"],
+        "pymatgen.io.lobster": ["lobster_basis/*.*"],
         "pymatgen.io.feff": ["*.yaml"],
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks -- the only way to really test this would be a test that does a pip install + then runs the test suite, which is perhaps not a bad idea. The other thing is that we could change the setup.py to just include all .json/.yaml it finds, which might also be sensible (rather than the current method which just includes those in specific submodules).

JaGeo:
Both are fine with me. It's just that I forgot to include the yaml file in the setup.py the last time I added one...

JaGeo:
Let me know what you decide (I am mainly making these changes to finally include the Lobster workflow in atomate)

mkhorton:
I've merged this now to get it working, I think we should do both: modify the setup.py to include all `.json` and `.yaml`, and run the test suite on pip installs, I'll add it to the list of things to do :) If you want to modify the setup.py in a future PR, that would also be welcome.

Issue #1815, opened 2020-03-30
Boltztrap2.py unit problem


adam5353:
I was using boltztrap2 module in pymatgen. I discoverd unit problem in electrical conductivity. When I checked the source, I have not found unit conversion of kS/m. I think the unit should be S/m.

fraricci:
Let me check that.
I have a couple of edits done but not pushed yet.
Thank for reporting. Do not hesitate if you find something else.

adam5353:
It is great to hear that. Looking forward to your edits. Thanks.

fraricci:
I pushed a new version. 
Please have a look.
Any suggestion is welcome.

PR #1879, opened 2020-06-24
Boltztrap2 spin

fraricci:
Here an upgrade of the boltztrap2 interface.
Major points are:
- transport properties, dos, and bands can be computed for spin polarized DFT computations
- PMGLoader should be used to load a vasprun/bandstructure pymatgen object, instead of the two previous loaders.
- load and save methods of BztInterpolator and BztTransportProperties allows to store/read in/from file the data they contain
- it includes the fix to the typo in the plotter raised in issue #1815 
- bands can be interpolated along a custom path
- carrier concentration for different doping levels and temperatures is now available

I did some tests. But those of you that use it, please let me know any issue. Thanks.


pymatgen/electronic_structure/boltztrap2.py
```diff
@@ -56,16 +56,162 @@
 __date__ = "August 2018"
 
 
-class BandstructureLoader:
-    """Loader for Bandsstrcture object"""
+class PMGLoader:
```
shyuep, line 59: This name is too uninformative. it can literally mean loading anything in pymatgen.

pymatgen/electronic_structure/boltztrap2.py
```diff
@@ -56,16 +56,162 @@
...
+                 structure=None,
+                 nelect=None,
+                 mommat=None,
+                 magmom=None):
         """
         :param bs_obj:
         :param structure:
         :param nelect:
         :param spin:
+        :param momat:
```
shyuep, line 209: I need proper documentation for every parameter.

pymatgen/electronic_structure/boltztrap2.py
fraricci: what about VasprunBSLoader ;-)

pymatgen/electronic_structure/boltztrap2.py
fraricci: right, done!

shyuep:
Thanks!

PR #1880, opened 2020-06-24
Fix suggest_larger_supercells bug

kavanase:
## Summary

There's a small bug in the `DefectPhaseDiagram.suggest_larger_supercells()` method (line 390 in `pymatgen/analysis/defects/thermodynamics.py`, where the current code is:
```python
 for tl, chgset in self.transition_level_map.items():
```
which gives the wrong variables of Defect-Name, {Transition-Level: Charge-set}. It should be:
```python
 for tl, chgset in self.transition_level_map[def_type].items():
```
(it's within the `for def_type in self.defect_types:` loop), which gives the correct variables of which gives the wrong variables of Transition-Level, Charge-set.

See below for a quick demonstration of the bug, and the result when fixed:
image

### DIFF
--- a/pymatgen/analysis/defects/thermodynamics.py
+++ b/pymatgen/analysis/defects/thermodynamics.py
@@ -387,7 +387,7 @@ class DefectPhaseDiagram(MSONable):
                     # consider if transition level is within
                     # tolerance of band edges
                     suggest_bigger_supercell = True
-                    for tl, chgset in self.transition_level_map.items():
+                    for tl, chgset in self.transition_level_map[def_type].items():
                         sorted_chgset = list(chgset)
                         sorted_chgset.sort(reverse=True)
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks for the bug fix @kavanase, my apologies for letting this sit for a while. I've merged this in now as-is since it's a fairly minimal change, but for future bug fixes including a minimal extra test that demonstrates the buggy behavior is always appreciated.

Please make sure to include your details in [this form](https://forms.gle/JnisFb38QDR8QTFTA) so that you can be credited appropriately in the [pymatgen documentation](https://pymatgen.org/team.html) too.

kavanase:
Great, thanks @mkhorton !!
I'll make sure to include extra tests for future bug fixes!

I've added my details to the developer attribution form now.

PR #1885, opened 2020-06-27
parse scientific notation in OUTCAR (possibly without spaces in betwe…

computron:
## Summary

This PR is very similar to #1836 , except it now supports the dielectric tensor section of the OUTCAR as well (instead of just plasma frequencies)

The unit test from #1836 should also cover this one


computron:
Closing for now, since two features got intermingled in this PR

PR #1886, opened 2020-06-27
two changes: parse scientific notation dielectric data, update cost information

computron:
## Summary

This PR includes two different changes (poor style, I know - but the Github UI makes it difficult to separate them into different PRs after the fact):

### Change 1:
This update for scientific notation is very similar to #1836 , except it now supports the dielectric tensor section of the OUTCAR as well (instead of just plasma frequencies)

The unit test from #1836 should also cover this one

### Change 2:
Added cost data to CSV file for He,H,Ar,Ne,Kr,Tc; remove duplicate O row.

There is no new code here, just a data file update.


mkhorton:
Thanks @computron :)

PR #1891, opened 2020-07-02
Fix zpsp to have element names as its keys

### LINKED ISSUE #1890: Error of Critic2Caller
[test.zip](https://github.com/materialsproject/pymatgen/files/4850861/test.zip)
**Describe the bug**
When using the critic2 with Critic2Caller, the result shows 
```
%% load int.CHGCAR id chg_int zpsp Al 3 As 5 O 6

!error! id chg_int zpsp Al 3 As 5 O 6
ERROR(setfield): Unknown extra keyword
ERROR(setfield): Unknown extra keyword
```
and fails.
This result does not depend on the system.
However, when using Pymatgen ver.2020.1.28, Critic2Caller works.
So, the update between ver.2020.1.28 and ver.2020.6.8 may break the code related to Critic2Caller.

The version of critic2 is the latest stable version (+ critic2 (stable), commit (config) 3081176) downloaded last Saturday.

**To Reproduce**
1. Copy the attached files to a directory.

2. Run the following command with the attached files. First argument is the path to the directory.

```
from pymatgen.command_line.critic2_caller import Critic2Caller

def create_critic2_output(path: str):
    critic2_caller = Critic2Caller.from_path(path)

if __name__ == "__main__":
    create_critic2_output(sys.argv[1])
```

yuuukuma:
##  Summary

This fix is needed as symbol shows the POTCAR label e.g. "Ca_pv" while element its name "Ca",

Fix #1890

### DIFF
--- a/pymatgen/command_line/critic2_caller.py
+++ b/pymatgen/command_line/critic2_caller.py
@@ -297,7 +297,7 @@ class Critic2Caller:
 
             if potcar_path:
                 potcar = Potcar.from_file(potcar_path)
-                zpsp = {p.symbol: p.zval for p in potcar}
+                zpsp = {p.element: p.zval for p in potcar}
 
         if not zpsp:
 


mkhorton:
Thanks!

PR #1897, opened 2020-07-08
Make EwaldSummation MSONable

lbluque:
## Summary
Computing the Ewald matrices for large structures can be very time consuming. By making the `EwaldSummation` class `MSONable` we can at least save time by reloading instances without recreating the matrices.

* `as_dict` method 
* `from_dict` method
* To implement the above I changed the constructor to initialize private ewald parts as None and lazy compute when requested.

## Checklist

- [X] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
      on your local machine.
- [X] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [X] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) 
      to type check your code.
- [X] All existing tests pass.


lbluque:
If anything in the implementation looks fishy please comment...

shyuep:
Thanks but I am unclear about the value of this. The EwaldSummation is an analysis. Typically people execute it once for a particular structure. They may want to store the final results of it in some form (son or otherwise) but I fail to see when I would want to reinitialize a EwaldSummation object. There is no additional analysis that can be done?

lbluque:
We are using the ewald matrices as part of Monte Carlo sampling. Most often just the total energy matrix, but having the separate parts allows finer grained control in sampling, having access to the properties defined in the class is quite useful for this.

For this use case I could definitely just save all the matrices separately, and then write functions to get the total matrix, energies, etc, but this would essentially be re-implementing the class properties in the EwaldSummation. I guess it comes down to this value extending further than our particular use case...

shyuep:
Thanks for the explanation. But I guess my question is how you are using the ewald matrices. You are using them unchanged right? What you really need is an output method. as_dict can be implemented. What I don't see is why we need a from_dict method since the dict (which can be json serialized) can be read in with a two-line code and used quite easily.

lbluque:
Thanks for the quick reply!

Yes, we use them unchanged. The reason for the from_dict is to recreate without needing to recompute a previously constructed set of matrices, for future sampling runs.

lbluque:
Ah ok, I see what you mean. I put it in there to handle the two cases where the matrices have been and have not been computed.

Here, I am out of my understanding comfort zone. Can this be automatically handled by proper serialization?

shyuep:
I see. Ok. I will merge.

lbluque:
Thanks @shyuep 
I tried using `functools.cached_property` to clean this up, but since some of the properties (i.e. real & point) are computed at the same time, I could not find a way around it.

PR #1900, opened 2020-07-09
Changed tolerance for lattice match when creating defect objects 

mbkumar:
## Summary

Stringent atol of 1e-8 used in numpy.allclose function for lattice match is causing problems for users. The tolerance for lattice match check for defect instantiation is now made looser. This fixes many failures experienced by users of pycdt and  atomate defect workflows 

* Instead of using default lattice match function of pymatgen, direct call to np.allclose function is made. A looser absolute tolerance of 1e-5 is supplied inplace of the default 1e-8 in the call to np.allclose.


shyuep:
Can we have a unittest pls, since the claim is that the lattice match is causing functionality failure.

mbkumar:
Unit cell added demonstrating the failure with default lattice match and success with looser absolute tolerance.

shyuep:
Thanks.

PR #1904, opened 2020-07-10
Fix serialisation of slabs with list scale_factor

utf:
## Summary

A recent bug fix (1a3ce601c932ed80db14930da6e2c11a8a246771) introduced an incompatibility of the `Slab` class with the atomate `TransformerFW` and therefore broke surface workflows. Specifically, the `Slab` class now assumes the `scale_factor` input is a numpy array, whereas the `TransformerFW` supplies a list `scale_factor`. This caused an error during serialisation of the Slab when `tolist()` was called on a list.

This PR casts the `scale_factor` to a numpy array during initialisation of the Slab.

I've added a test.


mkhorton:
Thanks @utf

Issue #1905, opened 2020-07-10
QChem Frequency Calculation Import Fails


asmith97:
I'm unable to use QCOutput to import data from a QChem frequency calculation. An error is thrown while loading the eigenvectors of the vibrational modes. I get an `IndexError: list index out of range` from `shape=(len(freqs), len(temp_freq_mode_vecs[0]), 3))` in line 558 of _read_frequency_data because temp_freq_mode_vecs is an array with no elements. I printed out the contents of temp_freq_mode_vecs to confirm this. It seems like either there's an issue with the QChem output file that I have, which comes directly from a QChem frequency calculation, or the regex pattern given to read_table_pattern isn't working. I'm not sure if other people have ran into this issue or if this is the right place for such a message.

mkhorton:
Thanks @asmith97 for the report, would it be possible to share your output file to diagnose? @samblau could you assign someone to this, or know someone who could take a look?

samblau:
I'm happy to take a look and diagnose. @asmith97, please provide your output, and I'll get right on it!

asmith97:
Thank you! I redid a frequency calculation with benzene to see if I got the same error, and it threw the same error. I was using `QCOutput.multiple_outputs_from_file(QCOutput, 'benzene.txt')`
[benzene.txt](https://github.com/materialsproject/pymatgen/files/4906403/benzene.txt)

samblau:
Hi @asmith97 - I see the problem. We never calculate Raman intensities, so the regex parsing the frequency modes assumes that they will be immediately preceded by the "Raman Active: ...." line rather than the "Depolar: ..." line like they are once you set doraman = true. I can work on a fix on Monday that will resolve the issue you're seeing and also parse the Raman intensities and Depolar values.

asmith97:
Thanks!

PR #1908, opened 2020-07-14
Fixing QChem frequency parsing when calculating Raman intensities 

samblau:
* Fixes issue #1905 
* All frequency information can now be correctly parsed even when Raman intensities are being calculated


shyuep:
Thanks.

PR #1911, opened 2020-07-17
Add validation and extrapolation for XAS stitching

yimingchen95:
## Summary

* Add check for empty spectrum and negative intensities
* Fix the jump in intensities at the L2 and L3-edge junction by extrapolating L2-edge XANES
* Add tests


shyuep:
There are failing pylint tests. Pls fix before I merge.

PR #1912, opened 2020-07-19
Adding JARVIS-Tools adaptor module.

knc6:
## Summary

JARVIS-Tools module integration in Pymatgen.

* Creation of pymatgen.io.jarvis
* Fixed setup.cfg for jarvis modules
* Fixed requirements-optional.txt


shyuep:
Thanks. 

1. Can you rename jarvisio to just Jarvis. The idea is that it is pymatgen.io.jarvis. For example, pymatgen.io.vasp, not pymatgen.io.vaspio (this was what we started with, but since IO is already in the earlier package, we don't need io again).
2. Pls pin jarvis-tools to a specific version in requirements-optional.txt.

Issue #1868, opened 2020-06-10
StructureMatcher.get_rms_dist illegal instruction on release pymatgen==2020.6.8.


mortengjerding:
**Describe the bug**
On the most recent release of pymatgen (2020.6.8) I am now getting "Illegal instruction" when running the `pymatgen.structure_matcher.StructureMatcher.get_rms_dist`. This did not happen on the previous release (2020.4.29).

**To Reproduce**
The following script
```
from ase.build import bulk
from pymatgen.io.ase import AseAtomsAdaptor
from pymatgen.analysis.structure_matcher import StructureMatcher
al = bulk('Al')
adaptor = AseAtomsAdaptor()
structure = adaptor.get_structure(al)
StructureMatcher().get_rms_dist(structure, structure)
```
results in "Illegal instruction (core dumped)".

**Expected behavior**
The result of the get_rms_dist should be zero.

**Desktop (please complete the following information):**
```
morten@hp ~ $ python3
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0] on linux
>>> import pymatgen
>>> pymatgen.__version__
'2020.6.8'

```

shyuep:
If you installed pymatgen via conda, pls try to do pip install --upgrade pymatgen. This error is usually a symptom that something is wrong with the compilation of the C extensions and happens quite frequently on Linux based systems.

mortengjerding:
I have not installed pymatgen through conda. I have simply pip installed it. This is also making my pipelines fail on gitlab so it's not just my laptap. I just ran the above script in a virtual environment and it still fails. I set the venv up like in the following:

```
$ python3 -m venv venv
$ . venv/bin/activate
$ pip install pymatgen
$ python test_script.py
Illegal instruction (core dumped)
```
So I don't see that it should be something with my setup. Since you closed the issue I was hoping you could help me figure out what is wrong. As I wrote: It worked in the previous release.

mortengjerding:
Sorry I forgot that I also had to install ASE in the venv to make the script work. The magic incantation to set up the environment is:

```
$ python3 -m venv venv
$ . venv/bin/activate
$ pip install pymatgen ase
$ python test_script.py
Illegal instruction (core dumped)
```

shyuep:
Does it fail for all scripts, or is it just the specific code you showed that is an issue? As far as I know, nothing has changed since v2020.4 that would cause this. There were some issues with sympy...

[...]

PR #1916, opened 2020-07-30
Remove CPP and use only C in get neighbors optmization

chc273:
## Summary

Recently, we have seen some platform-dependent errors caused by this cython extension. 

The known error is  "illegal instructions (core dumped)" which cause crash of python. This error was found on conda installations. In particular, I have encountered issues on `ubuntu 18.04.4 LTS` (Matt raised this issue to me), and the github action VM `macOS-latest, python3.7`. Such errors however do not show up if pymatgen is installed using `pip` or from source. Interestingly, it also does not show up on github action VM `macOS-latest, python3.8`. 

This PR is an attempt to fix the issue. My suspicion is that this is related to C++ compilation since it requires the platform-dependent flags. In this PR, I have removed all CPP-dependent codes and replace them with C only codes.


mkhorton:
Many thanks for addressing this @chc273.

For documentary purposes, I'm just noting we also noticed this issue on conda installs via Binder (which runs Ubuntu), and previously we also had some issues with Linux wheels via PyPI on #1868 that I think might be related.

chc273:
> Many thanks for addressing this @chc273.
> 
> For documentary purposes, I'm just noting we also noticed this issue on conda installs via Binder (which runs Ubuntu), and previously we also had some issues with Linux wheels via PyPI on #1868 that I think might be related.

Ah, I missed that issue. Thanks for bringing that to my attention... I hope this fix will do the job, otherwise, it could take much more time, since it is not very obvious where the problem is and the problem seems to be very dependent on the installation methods..

mkhorton:
I'm pretty confident that a switch to pure-C will fix this.

In that other issue, we were trying a new method of building PyPI wheels, so I'd assumed there was an issue with the build process rather than an issue in the code itself. However, since we're seeing it on conda too, it's clear there must just be a compatibility issue across Linux distributions with the current code.

PR #1918, opened 2020-08-03
gaussian std orientation bug fix

eimrek:
Hi pymatgen community!

In gaussian log, there are geometries outputted either in "standard" or "input" orientation or both together. The current implementation assumed that "input orientation" is always outputted but that is not the case (try e.g. [g.txt](https://github.com/materialsproject/pymatgen/files/5019019/g.txt)) and I was getting an error.

I modified the code such that by default, if it's there, the `opt_structure` will be of "standard orientation" and otherwise "input orientation". This is reflected by the `standard_orientation` boolean attribute and the same convention was already used for the `structures` attribute.

Previous version was mainly modified by @gVallverdu, so ping to you in case you're curious.

### DIFF
--- a/pymatgen/io/gaussian.py
+++ b/pymatgen/io/gaussian.py
@@ -537,7 +537,8 @@ class GaussianOutput:
 
     .. attribute:: opt_structures
 
-        All optimized structures from the calculation in the input orientation
+        All optimized structures from the calculation in the standard orientation,
+        if the attribute 'standard_orientation' is True, otherwise in the input
         or the Z-matrix orientation.
 
     .. attribute:: energies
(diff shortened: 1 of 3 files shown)


shyuep:
Thank you. Can you pls add a unit test for the new functionality pls? Thanks.

eimrek:
hi @shyuep, there is no new functionality, just a small bug fix

mkhorton:
The test would be to catch the bug you've encountered (e.g. with g.txt) so that future code modifications don't re-introduce the bug.

shyuep:
Yes, I understand that. but the fact is that the bug fix is fixing something that is not caught by any tests at the moment (otherwise, the tests would be failing since the bug exists). The idea is to write a simple test that demonstrates that the fix actually solves the bug,

shyuep:
Also, the new code seems to cause the existing tests to fail. So that needs to be fixed too.

eimrek:
Ah yep, I just realized this as well. The new convention (that standard orientation is preferred by default) is incompatible with one of the test, which compares against the "input orientation". Should I change that test as well? Other option would be to prefer the "input orientation" for the opt_structure but that would be weird as the standard orientation is already preferred for the `structures` attribute

eimrek:
Hi @shyuep, @mkhorton 

I added the test and modified an existing one. But now I'm failing `pybtex` import in some completely unrelated file: `pymatgen\util\provenance.py`. Any ideas why this could be?

shyuep:
Ignore that for now. I have merged it.

gVallverdu:
Thanks @eimrek for this bug fix.

PR #1920, opened 2020-08-07
Fix JMolNN tolerance

utf:
## Summary

Changed the JMolNN tolerance parameter to match the that used by the JMol source code.


utf:
The actual change of this PR was changing a single value (0876de1) but to get linting to pass I had to make many changes in the local_env module.

shyuep:
Thanks for fixing everything.

PR #1921, opened 2020-08-11
Added GibbsComputedStructureEntry for predicting ΔGf(T)

mattmcdermott:
## Summary

- Added a new entry class called **GibbsComputedStructureEntry**. This is an extension to ComputedStructureEntry which
allows for the estimation of Gibbs free energy of formation, ΔGf(T), via a machine-learned model. This can be used to estimate temperature-dependent phase diagrams using MP data.
- Currently incorporates the Gibbs SISSO descriptor for solid compounds as created and described by [Bartel et al. (2018) ](https://www.nature.com/articles/s41467-018-06682-4). This class should be extendable to future models based on pymatgen structural properties if/when they are created. 
- As explained in the paper, the calculation of ΔGf(T) requires the use of the temperature-dependent chemical potentials of the elements, which are derived from FactSage and supplied in the _g_els.json_ file.
- Since the SISSO model only applies to inorganic solid compounds, functionality has also been added to use NIST-JANAF data for gases where possible. This data is supplied in the _nist_gas_gf.json_ file.
- The `from_pd()` method constructs and returns a list of GibbsComputedStructureEntry objects given an existing MP phase diagram -- this is typically the easiest way to use the class.

## Checklist

      on your local machine.
      to type check your code.


rkingsbury:
Thanks @mattmcdermott! Is there any reason this functionality couldn't just be added to the existing `ComputedStructureEntry` rather than made its own class?  In general I feel we have too many `Entry` classes already. Interested to hear other's thoughts on this.

Separately, I'm working on creating an MPContribs project for experimental thermochemical data that might be a logical home for the NIST gas data you've included. I can upload the data using the file here, but we'll have to discuss whether it makes sense to modify your code to talk to the MPContribs project now, or leave it as-is until we streamline that project more. (Talking to MPContribs requires another API key that not everyone has)

rkingsbury:
I've also been thinking that separate attributes for `enthalpy`, `entropy`, and `free_energy` in `ComputedEntry` might be a good idea to better clarify how some of our analysis works, and your new class takes a step in that direction.

mkhorton:
Sub-classing is appropriate here, having many classes is not a problem in itself as long as the inheritance makes sense and it's well documented. This Gibbs model, though good, is still its own thing with its own set of caveats and shouldn't be added onto `ComputedStructureEntry` which is more general. Separately, making the distinction between energy/enthalpy might be a good conversation to have, we often interchange these terms inappropriately.

@mattmcdermott once tests are added this looks about ready to merge, I am thinking a little about `from_pd` though and if this is the best place for it.

mattmcdermott:
@rkingsbury I am always ambivalent about furthering an already long chain of subclasses, but I agree with @mkhorton 's reasoning that there is enough distinct functionality here to not integrate it into existing classes... at least not yet.

I will also add that Chris's model we are using here can only predict this "G_δ" term, aka the difference between ΔHf (0K) and ΔGf (desired temp), so it can't quite be separated into enthalpy and entropy terms, although we know from his paper that enthalpy changes much less with temperature than entropy does.

@mkhorton I always thought that having a method which constructs a list of objects was a bit strange (not sure if this is common practice), but the class is basically useless to the average user without this method, since you have to make a 0 K phase diagram to get MP formation energies in the first place. Were you thinking about putting it into the phase diagram class?

mkhorton:
@mattmcdermott Yeah, I actually think this is fine after thinking about it more. I was thinking PhaseDiagram, was also thinking Compatibility, but it's a small method and only there for convenience, and it doesn't preclude additional integration elsewhere later.

mattmcdermott:
@mkhorton @rkingsbury Just added tests! I also added the `from_entries()` method (another convenient initialization) and cleaned up a problem with saving/loading dicts.

mkhorton:
Thanks @mattmcdermott !

PR #1926, opened 2020-08-12
fix valence bug, add test

rkurchin:
Previously, asking for the valence property on an Element with full shells that wasn't noble (e.g. column 2 or 12) would cause an error. This resolves that.

## Summary

* Added a check for if the last orbital listed is fully-occupied and to choose that number as the valency
* Added a case to the tests to trigger this

none

## Checklist

- [ x ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
      on your local machine.
- [ x ] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [ x ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) 
      to type check your code.
- [ x ] Tests have been added for any new functionality or bug fixes.
- [ ? ] All existing tests pass. --> I ran the test_periodic_table.py manually, but nosetests continually hangs after 1089 tests. I installed all dependencies (including optional and dev ones) in my environment so I'm not sure what the issue is with that, but I can't imagine what this would break that's not covered by the test I checked.

### DIFF
--- a/pymatgen/core/periodic_table.py
+++ b/pymatgen/core/periodic_table.py
@@ -653,9 +653,13 @@ class Element(Enum):
         L_symbols = 'SPDFGHIKLMNOQRTUVWXYZ'
         valence = []
         full_electron_config = self.full_electronic_structure
-        for _, l_symbol, ne in full_electron_config[::-1]:
+        last_orbital = full_electron_config[-1]
+        for n, l_symbol, ne in full_electron_config:
             l = L_symbols.lower().index(l_symbol)
             if ne < (2 * l + 1) * 2:
+                valence.append((l, ne))
+            # check for full last shell (e.g. column 2)
+            elif (n, l_symbol, ne) == last_orbital and ne == (2 * l + 1) * 2 and len(valence) == 0:
                 valence.append((l, ne))
         if len(valence) > 1:
             raise ValueError("Ambiguous valence")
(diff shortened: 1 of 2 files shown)


mkhorton:
Many thanks @rkurchin !

mkhorton:
On the test note, I agree this is unlikely to fail but I'll let the tests pass on CI first before merging. Your issue might be related to using `nose` which is now deprecated, `pytest` should be sufficient to run tests.

We're also asking new contributors to [fill out this form](https://forms.gle/JnisFb38QDR8QTFTA) so that everyone can be acknowledged appropriately in the [pymatgen documentation](https://pymatgen.org/team.html).

rkurchin:
Got it – might want to update the instructions [here](https://pymatgen.org/contributing) then, that's why I used `nose`, even though I saw on its docs page that it was deprecated.

For my own stuff (since I need this to work properly for something I'm doing), I imagine once this is merged in I can expect it in the version I get from `pip install` within a couple days? In the meantime I can just use my local version obviously.

mkhorton:
Thanks for the catch! Yes that page is definitely out of date, I've pushed some minor edits to the documentation for now but we're over-due for a more thorough documentation update.

We don't currently have a fixed schedule for pymatgen releases, we usually make one or two per month as pull requests come in or on an as-needed basis. I'll see if I can get this released by the end of the week.

mkhorton:
FYI, this was released as part of 2020.8.13

PR #1933, opened 2020-08-17
MP2020 Compatibility refinements

rkingsbury:
## Summary

* Refactor `MaterialsProject2020Compatibility` to utilize the new abstract compatibility interface (see #1826 ) instead of the legacy `CorrectionsList` class. Note that the `PotcarCorrection` class is still utilized, but not any other `Correction` classes.
* `MaterialsProject2020Compatibility()` now exposes all keyword arguments that were previously passed to `Correction` classes
* Consolidate .yaml files into a single file and change the keys to more closely match the way each correction type is described in the forthcoming manuscript. Update `make_yaml` to generate this format.
* Refit energy corrections against the `v2020.08.21` database release, using experimental ground-state structures instead of DFT ground states for selected compounds where the DFT ground state is not in the ICSD
* update contents of `calc_compounds.json.gz` to include `ComputedStructureEntry` instead of `ComputedEntry`
* Update docstrings


rkingsbury:
FYI @awvio. When you are ready to update the .json.gz, the easiest way would be to open a pull request against [my branch](https://github.com/rkingsbury/pymatgen/tree/mp2020-refactor), which I can then merge so that it shows up here. Thanks!

pymatgen/entries/compatibility.py
```diff
@@ -991,45 +827,247 @@ def __init__(
...
     valid.
     """
 
     def __init__(
-        self, compat_type="Advanced", correct_peroxide=True, check_potcar_hash=False
+        self,
+        compat_type="Advanced",
+        correct_peroxide=True,
+        check_potcar_hash=False,
+        input_set=MPRelaxSet,
```
rkingsbury, line 844: @mkhorton @shyuep can you please advise on how to resolve a serialization problem I'm having? I want to pass `MPRelaxSet` as a kwarg to this class so that it can be used by `PotcarCorrection`, which only needs to access the class attribute `CONFIG`. I don't want to instantiate the class because that requires an input structure, which wouldn't make sense to provide here. However, if the class isn't instantiated, it won't serialize, and I fail the `as_dict()` test.

mkhorton: There are a number of different ways to do this... you could pass the kwarg as a string and access the class on demand, for example (e.g. `{subcls.__name__: subcls for subcls in DictSet.__subclasses__()}`). The larger question is if you should do it at all, I guess. The `CONFIG` attribute itself isn't enforced by `DictSet`, so in principle it's possible to write a input set for which you couldn't use the `PotcarCorrection`. Not really a problem in practice but feels a little messy.

rkingsbury: Thanks, I'm leaning towards "we should not." The MP2020 compatibility class is fitted to and therefore inextricably linked to MPRelaxSet calculations, so it probably doesn't make sense to ever check the POTCARs against any other `InputSet`. If I hard-code `MPRelaxSet` into the class instead of making it an attribute, this serialization problem will go away. Thoughts?

rkingsbury:
@mkhorton @shyamd @shyuep I've reformatted the .yaml file to consolidate uncertainties and corrections into one place, and also to make the keys more descriptive and consistent with the way we describe the corrections in the forthcoming manuscript. Please let me know if you have any feedback.

rkingsbury:
Ready for review per recent discussions @mkhorton. Let me know any further suggestions. Thanks!

PR #1936, opened 2020-08-18
Overhaul of PhaseDiagram plotting: adding Plotly backend

mattmcdermott:
## Summary

- Added `backend="plotly"` as a default argument to `PDPlotter`. Option `backend="matplotlib"` returns to existing plotting functionality.
- New features includes 2D scatter/line plot for binaries, and 3D scatter/mesh plots for ternaries and quaternaries. The latter include convex hull shading via Plotly's mesh plotting capabilities.
- Benefits of a Plotly backend: **highly increased user interactivity** -- ability to see and rotate around hull in three dimensions, hide traces, see hover annotations for phase info, etc.
- Plotting seems to be pretty fast: ternaries take about ~ 100 ms on my computer, especially with added caching of `pd_plot_data`

### EDITS:

- Improved color scheme with added colorbars
- Default layouts now in json file.
- Added uncertainty error bars for ComputedEntry's new uncertainty property. 
- Added uncertainty window shading along the hull for binaries.
- Separated stable labels into a separate trace (to allow hiding).
- Added some functionality for smarter placement of labels (should make this even better in the future).
- Improved some camera angle / rotation capability (should also make this even better in the future).

### Binaries: 
binary
### Ternaries:
ternary
### Quaternaries:
quaternary

## TODO (if any)

- Would be nice to add Plotly backend for chemical potential range map plots (i.e. predominance diagrams) **EDIT**: This has been done, but will be added in a future PR.
- Would be nice to add rotation support for ternaries; no option for choosing which element is plotted on top
- Phase diagram annotations are not always sensibly placed -- I can't seem to find a good fix for this, other than hiding annotations in general or writing some pretty arbitrary geometrical shifting functions **EDIT**: A simple first pass of this has been implemented, and will likely be edited in the future.
- A couple minor bugs that I've noticed but had trouble fixing: 1) Color of "Stable" trace in legend is not always green (as I would like) which also becomes an annoyance when working with ternary/quaternary hovertext, and 2) Shading of convex hull in quaternaries is obscured by the faces of tetrahedron which are also colored **EDIT**: Both problems fixed. Deleted shading of quaternary hull (too strange).

## Checklist

      on your local machine.
      to type check your code.

## Projects
This addresses Project #1288.


pymatgen/util/plotly_pd_layouts.py
```diff
@@ -0,0 +1,157 @@
+# coding: utf-8
```
mkhorton, line 1: Looks like this file could equally be a .json file, might be a better fit? Should be easy to dump the dictionaries that are in here, and then you load them directly from the JSON in phase_diagram.py.

mkhorton:
Looks fantastic, very exciting. Will need tests of course but otherwise will seems well structured.

pymatgen/util/plotly_pd_layouts.py
mattmcdermott: Oh yes that's totally do-able and I can/will change. Is there a reason why this would be preferred though?

mkhorton: Not really. More just that if it looks like a datafile, we should store it as a datafile. Could be pretty-printed JSON to make it easier to edit. Other advantages are that the JSON file doesn't have to be analyzed or linted, and reducing cross-talk between sub-modules is nice (things to do with the phase diagram plotting specifically should be stored with the phase diagram plotting). Some global Plotly templates could be put into utils though, e.g. font, etc. that might be something for us to think about.

mattmcdermott: Got it, that all makes sense to me! Yup I think that would be a good direction to go.

rkingsbury:
Thanks @mattmcdermott ! This is going to be a really nice enhancement. This also includes the uncertainties support (for 2D plots at least), correct?

mattmcdermott:
> Thanks @mattmcdermott ! This is going to be a really nice enhancement. This also includes the uncertainties support (for 2D plots at least), correct?

Yep, actually I am finishing this up at this exact moment. There will be uncertainty error bars for all plots, plus uncertainty shading for binaries.

mattmcdermott:
Ok @mkhorton @rkingsbury -- should be pretty much ready to go! It would be nice if you could test it out on your side to make sure things work (I am sure there are more things layout wise we can play around with as well)

mkhorton:
Thanks @mattmcdermott! This is great. If there are any additional improvements I think they can be put into a separate PR, since this looks feature complete as-is.

PR #1937, opened 2020-08-20
bugfix in surface and composition

gpetretto:
## Summary

Two bug fixes:

* with the current implementation the Surface object can get weird results when instantiating with `reorient_lattice=True `and `coords_are_cartesian=True`. Fixed by switching to frac coords when needed.
* after dec6ecb6a9a80dbd4bfdfb924040cdfc7fe4ebb3 `Composition.to_reduced_dict` is not returning a dict but a Composition instance.


shyuep:
Can you implement some tests to catch the bugs fixed pls? Thanks.

Issue #1935, opened 2020-08-18
MoleculeMatcher is failing too easily due to small perturbations


fekad:
**Describe the bug**
The current version of MoleculeMatcher does not seem to be robust enough for practical applications.

**To Reproduce**
You can find a simplified code snippet below:

```python
import random
from pymatgen import Lattice, Structure, Molecule
from pymatgen.analysis.molecule_matcher import MoleculeMatcher, IsomorphismMolAtomMapper, InchiMolAtomMapper

coords = [[0, 0, 0], [0.75,0.5,0.75]]
lattice = Lattice.from_parameters(a=3.84, b=3.84, c=3.84, alpha=120, beta=90, gamma=60)
struct = Structure(lattice, ["Si", "Si"], coords)
struct.make_supercell([2,2,2])

# Creating molecule for testing
struct = Molecule.from_sites(struct)
```

Using  IsomorphismMolAtomMapper:
```python
N = 10
mm = MoleculeMatcher(tolerance=2., mapper=IsomorphismMolAtomMapper())

# perturbing the atoms' position
for i in range(N):
    struct2 = struct.copy()
    struct_perturbed = struct.copy()
    struct_perturbed.perturb(0.3)
    print(mm.get_rmsd(struct2, struct_perturbed))
    

# perturbing the atoms' order
for i in range(N):
    struct2 = struct.copy()
    struct_perturbed = struct.copy()
    random.shuffle(struct_perturbed)

    print(mm.get_rmsd(struct2, struct_perturbed))
```
The output:
```
inf
inf
inf
inf
0.29595839352301634
inf
inf
inf
0.2830535234793642
inf
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
1.606558315893547e-15
```
Using InchiMolAtomMapper:
```python
N = 10
mm = MoleculeMatcher(tolerance=2., mapper=InchiMolAtomMapper())

# perturbing the atoms' position
for i in range(N):
    struct2 = struct.copy()
    struct_perturbed = struct.copy()
    struct_perturbed.perturb(0.3)
    print(mm.get_rmsd(struct2, struct_perturbed))
    

# perturbing the atoms' order
for i in range(N):
    struct2 = struct.copy()
    struct_perturbed = struct.copy()
[...]

PR #1938, opened 2020-08-21
[WIP] Implemenation  of a robust permutation invariant molecule matcher

fekad:
## Summary

The main goal to implement a robust matching method between molecules. This implementation is based on [Kabsch algorithm]( https://en.m.wikipedia.org/wiki/Kabsch_algorithm) and on an excellent python package called [rmsd](https://github.com/charnley/rmsd). The code has been basically rewritten from scratch and adapted to pymatgen. Also, some small bugs in the referred solution have been fixed.

There are three new methods available:
* `KabschMatcher`: this methods works when the order of the atoms is the same in the two structures. Returns with the optimal rotation and translation which belong to the minimal distance between the molecules.
* `BruteForceOrderMatcher`: This method compares all the possible permutations and it also returns with the optimal order of atoms. Works perfectly with small molecules but not feasible larger systems due to it's factorial complexity.
* `HungarianOrderMatcher`: This method tries to identify the atom pairs by aligning the molecules based on their principle axis and sorted by the (Hungarian algorithm)[https://en.m.wikipedia.org/wiki/Hungarian_algorithm]. Please note that it is not guaranteed that this method always returns with the "best" solution. 
* (NEW) `GeneticOrderMatcher`: This method was inspired by genetic algorithms and tries to match molecules based on their already matched fragments.

Key features:
* A new method to match  `p` onto `q`  molecule
* Returns with the rotation matrix, translation vector,  proper order of atoms (in case of permutation invariant method) and root-mean-square deviation. 
* The implementation is pure python and it is completely independent of `openbabel`.
* closing: #1935 

## TODO/Checklist

in the pull request title.

- [x] Creating a common interface for the `match` and `fit` methods
- [x] Refactoring the code to have a more object-oriented code 
      on your local machine.
      to type check your code.

allowing commits.


fekad:
@mkhorton Do any of you have experience / best practice to use random numbers for testing? I was trying to use seed method and also creating a random generator object with a specific seed value but it doesn't seem to work...

Obviously, I can just use some predetermined random numbers for testing, but in theory, using random generators with seed method should give the same results and you can easily do more test if you want ...

mkhorton:
@fekad I haven't had to do this previously, but using `seed` should work to create a deterministic random number generator, both via the in-built `random` library and `np.random`. I see you're initializing an rng object in some of your kwargs, perhaps that's responsible?

As an FYI, we don't currently use this in pymatgen but [hypothesis](https://hypothesis.readthedocs.io/en/latest/) is a nice library for creating unit tests that incorporates random data (following a specific requirement) as the test input, and asserts something about the output.

fekad:
@mkhorton The initialization of the rng object was not an issue because every time I explicitly constructed one and passed that one to the function.

I also tried to use the - old-fashion - seed method but the result is the same...
I use multiple macs without any issue but It looks like the seed methods are not cross-platform yet.

In conclusion, I will use pre-generated structures for the tests unless somebody has a better solution.

mkhorton:
Hi @fekad, my apologies for the delay in reviewing this, I will try to review this week. From your end, do you consider this PR ready to merge?

Looking at the test failures, I'm seeing a minor linter failure, but also some perhaps more serious test failures:

```
FAILED pymatgen/analysis/tests/test_molecule_matcher.py::KabschMatcherSiTest::test_perturbed_atom_position
129
FAILED pymatgen/analysis/tests/test_molecule_matcher.py::HungarianOrderMatcherSiTest::test_perturbed_atom_position
130
FAILED pymatgen/analysis/tests/test_molecule_matcher.py::GeneticOrderMatcherSiTest::test_all
131
FAILED pymatgen/analysis/tests/test_molecule_matcher.py::GeneticOrderMatcherSiTest::test_perturbed_atom_position
```

These are all of the form:

```
E       AssertionError: 0.2680448948923893 != 0.2628450748567651 within 8 places
77
```

It's possible these tests just need to be updated for the new method.

fekad:
@mkhorton No worries. I didn't fix the tests yet but I will do in the next days.

fekad:
@mkhorton From my side, this PR is ready for review. Please let me know if something is not understandable or if you have better names for the classes :)

mkhorton:
This is a really substantial contribution and nicely tested, thank you! Class names look good to me. Please make sure to include your details in [this form](https://forms.gle/JnisFb38QDR8QTFTA) so that you can be credited appropriately in the [pymatgen documentation](https://pymatgen.org/team.html) too.

kavanase:
Just noting here that I'm now using some of these classes in my research and they are extremely useful. Thanks @fekad !

PR #1939, opened 2020-08-21
[WIP] Jsmol viewer for jupyter notebook

fekad:
## Summary

The tiny interface for visualising structures in jupyter notebook/lab using JSmol. The [`jupyter_jsmol`](https://github.com/fekad/jupyter-jsmol) package provides a widget (fully compatible with any other ipywidgets widgets) which is using the JSmol. 
More information about the features of the package can be found [here](https://github.com/fekad/jupyter-jsmol/tree/master/examples).

Features:
* Although it is not the fastest it is full of features. Any custom [command/script](https://chemapps.stolaf.edu/jmol/docs/) can be sent to the JSmol "applet"
* it is easy to integrate the viewer with other ipywidgets (Layout, slider etc...)
* works offline and  support Jupyter lab as well

## Checklist

      on your local machine.
      to type check your code.

### DIFF
--- a/pymatgen/vis/structure_jsmol.py
+++ b/pymatgen/vis/structure_jsmol.py
@@ -1,11 +1,5 @@
+"""Visualization for structures using jupyter-jsmol.
 """
-Visualization for structures using jupyter-jsmol.
-"""
-
-import numpy as np
-from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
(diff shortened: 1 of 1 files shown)


mkhorton:
> I also think that having multiple options for visualising a structure is always a good thing. 

Agreed, I think a good place to share this would be as a public Gist https://gist.github.com, it's a good place to store re-usable useful code snippets even if they don't end up in the library, and they show up in Google search etc.

PR #1940, opened 2020-08-22
SCAN InputSet changes - ADDGRID, ISMEAR for ScanStaticSet

rkingsbury:
Minor updates to SCAN InputSets per recent communications with Perdew, Kresse, et al.

- Remove ADDGRID
- Enforce `ISMEAR=-5` in all static calculations with `MPScanStaticSet`

Note:
We have learned that when `PREC=Accurate`, setting `ENAUG` has no effect. Although we considered removing ENAUG from the InputSet, we decided to keep it explicitly defined in case a user (or custodian) change the `PREC` setting


mkhorton:
Thanks @rkingsbury

PR #1951, opened 2020-09-10
address DeprecationWarning triggered by get_pourbaix_entries

rkingsbury:
The recent introduction of the `solid_compat` kwarg to `get_pourbaix_entries` began triggering a `DeprecationWarning` when the module was imported, because the kwarg defaulted to an instance of `MaterialsProjectCompatibility()`. This commit changes the interface slightly to avoid instantiating the class on import and therefore avoid triggering the `DeprecationWarning` until `get_pourbaix_entries` is actually called.

* `MaterialsProjectAqueousCompatibility` now accepts `None`, `Compatibility()`, or `Compatibility` as the `solid_compat`
* `MaterialsProjectAqueousCompatibility` now defaults to `solid_compat=MaterialsProjectCompatibility` (was previously `None`)
* allow `get_pourbaix_entries` to accept str as the `chemsys` argument (e.g. "Fe-Cr") for consistency with `get_entries_in_chemsys`
* pylint fixes and updates for `pourbaix_diagram.py`

TODO

* Fix the warnings suppression when `get_pourbaix_entries` calls `MaterialsProjectAqueousCompatibility` (can be addressed in a future PR; not changed by this one).

### DIFF
--- a/pymatgen/analysis/pourbaix_diagram.py
+++ b/pymatgen/analysis/pourbaix_diagram.py
@@ -127,9 +127,6 @@ class PourbaixEntry(MSONable):
         if self.phase_type == "Solid":
             return self.entry.composition.reduced_formula + "(s)"
 
-        if self.phase_type == "Ion":
-            return self.entry.name + "(aq)"
-
         return self.entry.name
 
     @property
--- a/pymatgen/analysis/pourbaix_diagram.py
+++ b/pymatgen/analysis/pourbaix_diagram.py
@@ -15,22 +15,22 @@ Fast computation of many-element Pourbaix diagrams:
 """
 
 import logging
-import numpy as np
 import itertools
 import re
 from copy import deepcopy
-from functools import cmp_to_key, partial, lru_cache
-from monty.json import MSONable, MontyDecoder
-
-from multiprocessing import Pool
 import warnings
+from multiprocessing import Pool
 
+from functools import cmp_to_key, partial, lru_cache
+import numpy as np
+from monty.json import MSONable, MontyDecoder
 from scipy.spatial import ConvexHull, HalfspaceIntersection
-
 try:
     from scipy.special import comb
 except ImportError:
     from scipy.misc import comb
+from tqdm import tqdm
+
 from pymatgen.util.coord import Simplex
 from pymatgen.util.string import latexify
 from pymatgen.util.plotting import pretty_plot
@@ -41,7 +41,6 @@ from pymatgen.entries.computed_entries import ComputedEntry
 from pymatgen.entries.compatibility import MU_H2O
 from pymatgen.analysis.reaction_calculator import Reaction, ReactionError
 from pymatgen.analysis.phase_diagram import PhaseDiagram, PDEntry
-from tqdm import tqdm
 
 __author__ = "Sai Jayaraman"
 __copyright__ = "Copyright 2012, The Materials Project"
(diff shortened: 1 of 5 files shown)


mkhorton:
Thanks @rkingsbury. I noticed you commented out a test? Btw, NERSC is having load issues this morning so there are intermittent rester timeouts.

rkingsbury:
> Thanks @rkingsbury. I noticed you commented out a test? Btw, NERSC is having load issues this morning so there are intermittent rester timeouts.

Yes, that was a linting issue. Previously, the entire test was commented out except for the line that assigned a `Structure` to the variable s. The linter was complaining about the variable s being assigned to but never used. Since it seemed that the remainder of the test had already been disabled, I just commented out the whole thing.

rkingsbury:
There is another warning raised by `get_pourbaix_entries` that needs to be suppressed. I had previously merged a warnings filter to catch it, but it no longer appears to be working (at least in my Jupyter environment). It's not clear to me why not. Still working on that. Here is the relevant code in `matproj.py`

```
        # suppress the warning about supplying the required energies; they will be calculated from the
        # entries we get from MPRester
        with warnings.catch_warnings():
            warnings.filterwarnings(
                "ignore",
                message="You did not provide the required O2 and H2O energies.",
            )
            compat = MaterialsProjectAqueousCompatibility(solid_compat=solid_compat)
        ion_ref_entries = compat.process_entries(ion_ref_entries)
        ion_ref_pd = PhaseDiagram(ion_ref_entries)
```

rkingsbury:
> There is another warning raised by `get_pourbaix_entries` that needs to be suppressed. I had previously merged a warnings filter to catch it, but it no longer appears to be working. It's not clear to me why not. Still working on that. Here is the relevant code in `matproj.py`
> 
> ```
>         # suppress the warning about supplying the required energies; they will be calculated from the
>         # entries we get from MPRester
>         with warnings.catch_warnings():
>             warnings.filterwarnings(
>                 "ignore",
>                 message="You did not provide the required O2 and H2O energies.",
>             )
>             compat = MaterialsProjectAqueousCompatibility(solid_compat=solid_compat)
>         ion_ref_entries = compat.process_entries(ion_ref_entries)
>         ion_ref_pd = PhaseDiagram(ion_ref_entries)
> ```

I can't for the life of me figure out why this warnings filter doesn't work. I've tried several varations (adding `category`, removing the context manager, etc.), but python always emits the warning regardless. I think this should be fixed but perhaps as part of another PR (my suspicion is there's a bigger issue here regarding how `warnings` behaves within pymatgen.

If you agree @mkhorton , this should be ready to go.

mkhorton:
This is only raised if the user calls `get_pourbaix_entries` not on import? So I think it may be less of an issue anyway.

mkhorton:
Did you try `warnings.simplefilter("ignore")`within the `catch_warnings` block btw?

rkingsbury:
> This is only raised if the user calls `get_pourbaix_entries` not on import? So I think it may be less of an issue anyway.

Correct, it's only raised when calling `get_pourbaix_entries`

rkingsbury:
> Did you try `warnings.simplefilter("ignore")`within the `catch_warnings` block btw?

Yes I did, but it has no effect.

mkhorton:
Ok, I'll merge this as-is. Thanks for addressing this issue @rkingsbury

Issue #1941, opened 2020-08-26
Error in fetching band structure for specific materials id


goodwilling:
The following script/URL, either 

mpr.get_bandstructure_by_material_id("mp-867624") 

or

https://www.materialsproject.org/rest/v2/materials/mp-867624/vasp/bandstructure?API_KEY=************

gives the error: 

{
  "valid_response": false,
  "error": "too many indices for array: array is 0-dimensional, but 2 were indexed",
  "version": {
    "db": "2020.08.21",
    "pymatgen": "2020.8.3",
    "rest": "2.0"
  },
  "created_at": "2020-08-26T11:00:09.393429",
  "traceback": "Traceback (most recent call last):\n  File \"/var/www/python/matgen_prod/materials_django/rest/rest.py\", line 94, in wrapped\n    d = func(*args, **kwargs)\n  File \"/var/www/python/matgen_prod/materials_django/materials/rest.py\", line 114, in get_vasp_property\n    loadee=d[\"response\"])\n  File \"/var/www/python/matgen_prod/materials_django/materials/rest.py\", line 56, in mid_ETL\n    value = transform(s)\n  File \"/var/www/python/matgen_prod/materials_django/materials/rest.py\", line 68, in es_transform\n    return extracted.as_dict()\n  File \"/opt/miniconda3/envs/mpprod3/lib/python3.6/site-packages/pymatgen/electronic_structure/bandstructure.py\", line 858, in as_dict\n    d[\"is_metal\"] = self.is_metal()\n  File \"/opt/miniconda3/envs/mpprod3/lib/python3.6/site-packages/pymatgen/electronic_structure/bandstructure.py\", line 307, in is_metal\n    if np.any(values[i, :] - self.efermi < -efermi_tol) and \\\nIndexError: too many indices for array: array is 0-dimensional, but 2 were indexed\n"
}

mkhorton:
Hi @goodwilling, this is a Materials Project issue so would be better discussed at matsci.org/materials-project

In brief, we're going through a series of large database updates, including this week expanding our band structures (from 54k -> 76k) and drastically improving the quality of our densities of states (see the release notes on materialsproject.org). In the process we've introduce a bug, this bug is being fixed currently but it might take a day or two before this will work.

PR #1953, opened 2020-09-15
Bandstructure bug fix and review of from_dict/as_dict

fraricci:
This bug fix should solve the problem in issue #1941,
Example: mp-15077
Problem: for some reason some bandstructures on symmetry line in the MP
database contain the bands as dict(array). The BandStructureSymmLine.from_dict()
cannot handle this format. It creates the object but then the other methods
(e.g. is_metal()) fail because they expect an array.

Solution:
The BandStructure.from_dict() actually can handle this case. 
So I took this opportunity to review the BandStructure.from_dict()/as_dict() 
and make them work for the subclass BandStructureSymmLine as well.
So those methods were removed from the subclass.

Tests are running fine, but I'd like to do some other using data from the database.
I tried to get the json file for the problematic cases from the rester without success. 
It seems there's a check of the data server side that fails (because there the bus is 
not fixed) and it does not allow the download. I get the usual error instead.


mkhorton:
Many thanks @fraricci. For some context on the MPRester issue, some particularly large band structures fail due to a timeout on Cloudflare (which we cannot change). We use Cloudflare since it provides caching to reduce load on our servers and firewalls to prevent abuse. However, we're going to establish a second endpoint that bypasses Cloudflare for the purposes of these large files, but we need to purchase a new certificate first so this may take a week or so.

fraricci:
Hey @mkhorton, is that related with the issue I'm trying to solve with this issue?

mkhorton:
It is not, separate issue

mkhorton:
To add additional information, large band structures that may take more than 30 seconds to pull or process will cause a gateway timeout error. So if you're seeing that, that's why.

fraricci:
Hi @mkhorton, it seems that this issue is not solved yet.
Fetching bands for mp-867624 still returns the same error:
sbs = rester.get_bandstructure_by_material_id('mp-867624')
MPRestError: too many indices for array: array is 0-dimensional, but 2 were indexed.....

If I well remember, to solve this, the server-side version of pymatgen has to contain the patch implemented in this pull request.
Do you have any updates?
Thanks!

mkhorton:
Hi @fraricci, this could simply be because the MP website is still running an older pymatgen (currently, it's `2020.8.13`). We have a release pending so we can update the version of pymatgen then. Note that we're spending most of our development effort on the new API and will be looking for beta testers soon, you might prefer to use that.

PR #1959, opened 2020-09-21
Added code to calculate NGX/NGY/NGZ NGXF/NGYF/NGZF

jmmshn:


jmmshn:
@mkhorton for the tests I just grabbed a bunch of random calculations and check that the grids matched.
I actually check vs all 100K + calculations in my DB but only uploaded 300 to keep the file size small.

mkhorton:
Thanks @jmmshn! Interesting you have to do a prime factorization with this too

jmmshn:
This a was amusing for me as well.  It's because of how VASP interacts with FFTs.

mkhorton:
Thanks @jmmshn !

PR #1960, opened 2020-09-21
populate "Generated by" in EnergyAdjustment generated by MP2020 and fix serialization bug

rkingsbury:
Populate the `cls` kwarg when creating `EnergyAdjustment` within `MaterialsProject2020Compatibility`. This information will be helpful in tracking the provenance of energy adjustments. 

Now that the `cls` kwarg is populated as intended, executing `ComputedEntry.energy_adjustments` on an entry that has been processed with the MP2020 correction scheme should generate output similar to

```
[CompositionEnergyAdjustment:
   Name: MP2020 anion correction (oxide)
   Value: -5.920 eV
   Uncertainty: 0.014 eV
   Description: Composition-based energy adjustment (-0.740 eV/atom x 8.0 atoms)
   Generated by: MaterialsProject2020Compatibility,
 CompositionEnergyAdjustment:
   Name: MP2020 GGA/GGA+U mixing correction (Fe)
   Value: -8.728 eV
   Uncertainty: 0.036 eV
   Description: Composition-based energy adjustment (-2.182 eV/atom x 4.0 atoms)
   Generated by: MaterialsProject2020Compatibility]
```

Whereas before, the "Generated by:" line read `None`

In addition, I discovered that several classes inherited from `EnergyAdjustment` (including `CompositionEnergyAdjustment`, which is used by MP2020) were not serializing properly. This was because the `description` attribute was being modified during init and because some kwargs were not stored as attributes. This PR also fixes those problems.


mkhorton:
Unless you want to communicate that people should be able to change `adj_per_atom` after instatiation, you probably want to set it with `self._adj_per_atom` (similar to eg `self._value` previously), likewise with `uncertainty`. Otherwise this looks good.

PR #1967, opened 2020-09-29
Cube parsing for Bader Analysis

nwinner:
## Summary

Bader caller has been amended to allow for parsing VASP outputs (CHGCAR) as well as the Gaussian Cube format.

* Feature 1: pymatgen.io.cube module created for Cube object, which reads and stores volumetric data from cube file just as Chgcar does for CHGCAR files.

* Feature 2: BaderAnalysis now supports either VASP or Cubes. Unit tests were overridden to explicitly reference the keyword argument desired. Because "from_path" and related methods in the bader_caller.py file are used to assimilate several VASP files, they can remain as they are, and so bader analysis done by drones in atomate will still function the same.

## Note

Passes tests and lints _except_ for one that was recently added, but the master branch also does not pass this so I believe it is out of my hands.


pymatgen/command_line/bader_caller.py
```diff
@@ -251,28 +276,47 @@ def get_charge_transfer(self, atom_index):
...
+        Note, this assumes that the Bader analysis was correctly performed on a file
+        with electron densities
+        """
+        charges = [-self.get_charge_transfer(i) for i in range(len(self.structure))]
+        struc = self.structure.copy()
+        struc.add_oxidation_state_by_site(charges)
+        return struc
+
+    def get_spin_state_decorated_structure(self):
+        """
```
shyamd, line 310: Is there any way to verify this from the data?

shyamd (approved):
Over all looks good.

pymatgen/command_line/bader_caller.py
nwinner: Might work. Is there ever a situation where Bader code outputs a negative charge, or is it always positive and you have to calculate the charge transfer? 

If we're confident that it will always be positive, then we can check for negative charge as a filter for spin density files.

mkhorton: I don't think there's an unambiguous way to figure out what's in a cube file. Perhaps you could sum up and see if it's the number of electrons you expect (therefore charge) but I'm not sure it's worthwhile.

shyamd: Alright lets skip this idea for now then.

pymatgen/command_line/bader_caller.py
```diff
@@ -98,19 +96,18 @@ class BaderAnalysis:
...
+    def __init__(self,
+                 cube_filename=None,
+                 chgcar_filename=None,
+                 potcar_filename=None,
+                 chgref_filename=None,
+                 parse_atomic_densities=False):
         """
         Initializes the Bader caller.
 
         Args:
```
mkhorton, line 108: It looks like this docstring is incomplete

nwinner: Should be fixed now.

mkhorton: Thanks!

mkhorton:
Many thanks @nwinner, this is great! Thanks for breaking it off into a smaller PR too from the CP2k PR too, makes review easier.

Please make sure to include your details in [this form](https://forms.gle/JnisFb38QDR8QTFTA) so that you can be credited appropriately in the [pymatgen documentation](https://pymatgen.org/team.html) too.

PR #1972, opened 2020-10-02
Ignore POTCAR.spec in Vasprun

rkingsbury:
To facilitate testing in the absence of proprietary POTCAR files, `InputSet` have the option to write a `POTCAR.spec` file instead of an actual POTCAR. However, when a `POTCAR.spec` file is present, `Vasprun.get_potcars()` tries to read it, resulting in an error.

This PR causes Vasprun.get_potcars() to ignore POTCAR.spec when present. See also a related atomate PR which requires this change: https://github.com/hackingmaterials/atomate/pull/454


mkhorton:
Thanks @rkingsbury

PR #1974, opened 2020-10-03
[WIP] Add PatchedPhaseDiagram to accelerate handling of large approximately disjoint data sets.

CompRhys:
## Summary

Introduce a new `PhaseDiagram` based class to handle large disjoint data sets. Calculation of the convex hull of a data set scales with the dimensions and amount of data. Breaking up disjoint datasets into smaller patches reduces the complexity in both these regards and would allow for parallel computation of different patches.

Not all the functionality of `PhaseDiagram` will be available in `PatchedPhaseDiagram` as some of these functions can't be quickly calculated in very very large chemical spaces without the full phase diagram. Currently the aim will be to keep all functionality that would be helpful for those investigating Machine Learning workflows who may be interested in the relative stability large spaces I.e. the aim is to allow for `get_quasi_e_to_hull` to be evaluated from scratch for the entirety of the Materials Project on a 5 year old laptop in a timely manner.

#### Functionality
* ~~Use multiprocessing to accelerate the calculation of the convex hulls within the patches~~ **on a 85 dimensional ppd with 4000 entries we take 90 seconds without mp and 60 seconds using mp across 3 cores**
* ~~Add function to stitch patches together when required~~ **deemed impossible (see discussion below)**
* Standardise `__eq__` and `__hash__` for `Entry` and it's child classes.

#### Speed
* ~~Carry out a speed test to show this is actually faster~~ **Much faster provided we have a high rank chemical space e.g.  140msec vs 10.4 sec for a toy set of just 30 materials containing 10 diverse materials and 20 different elements**

## Checklist

- [x] Tests have been added for any new functionality or bug fixes. **Tests expose functionality that was not configured correctly previously - this now needs to be resolved**


mattmcdermott:
Nice -- very timely as I've actually been wanting to solve this issue as well! I frequently need to make phase diagrams of 10+ elements, which is essentially impossible (that's around when it starts taking hours). I came up with a "phase diagram expansion" solution (not actually in pymatgen). It's the `expand_pd()` method here -- is this similar to your approach?

https://github.com/GENESIS-EFRC/reaction-network/blob/96850b013538f36ee310d37c028b9000575d20f7/rxn_network/helpers.py#L611

My thought at the time was to sort the entries in reverse order by size (number of elements), and then check that each next entry is found within one of the existing larger PDs

mattmcdermott:
The problem I'm working on now is figuring out how to do this expansion method, but preserve the information about the simplicial facets (i.e. not just the formation energies), and then combine these facets in each subspace back into one gigantic phase diagram...

CompRhys:
I think what I have is pretty similar to your function. All it does it just break up the data into the different (smaller) chemical spaces, I haven't (yet?) excluded the subsets so I would have Li-O and Li-Fe-O in my dictionary. 

In terms of stitching the patches back together I haven't really thought about it at all. I would need to get more familiar with the qhull output to even give it a go

CompRhys:
As a simple example the problem you're looking at is something like: I have NaCl, MgO, Na2O and MgCl2 - Can I get the Na-Mg-O-Cl quaternary pd from the 4 binary pds? The use would then be that I could get a decomposition for something like Mg2OCl2?

mattmcdermott:
Yep, that is spot on. In my particular case, I'd also like to get predicted reaction products when mixing together two compositions on the phase diagram (which we already have code for via the `InterfacialReactivity` class), but you need to know the facets of the energy hull to be able to get the products. Furthermore, it turns out that it's quite possible (although not super common) to get reactions between two pentanary compounds, meaning that you will occasionally be required to create phase diagrams with 5+5=10 (or more) dimensions.

CompRhys:
Sounds like an interesting problem which if it can be done would improve the functionality of this `PatchedPhaseDiagram` class.

I think the thing to do is introduce a new `PhaseDiagram` child class called `StitchedPhaseDiagram` (working name) that we initialise from a list of pds and stray entries. Then if we can reconstruct all the attributes of `PhaseDiagram` in the `StitchedPhaseDiagram.__init__` we can let it inherit all the functions from `PhaseDiagram` which make the code nice and clean. Although honestly it looks pretty hard to do the simplicies and facets.

I think there's two hard stages to stitching them back together firstly defining a new co-ordinate system in the new space and then secondly you'd need to iteratively check all the possible triangular facets that can be added and take the ones that are furthest away. The one thing I'm not sure about is whether you would need to handle cases where you have facets that bisect as I can visualise it beyond 3D.

CompRhys:
@mattmcdermott Looking at your particular problem I don't think you need to go the whole way with `StitchedPhaseDiagram` straight away (although long term I think it would be good to add if the stitching can be achieved in a timely manner - but I am not sure about this having looked at it a little). The `InterfacialReactivity` code only needs the facets to get the decomposition/hull_energy and it's possible to get these by solving a constrained optimisation problem like I implemented in order to add `PhaseDiagram.get_quasi_e_to_hull`. You could probably even reuse the `_slsqp_decomp_solution` function.

CompRhys:
@mattmcdermott I implemented `get_hull_energy` for `PatchedPhaseDiagram` using the slsqp method if we have examples that don't fit in a given patch. This might help with your problem or at least offer a partial speed up - better (i.e. more narrow) selection of the competing entries would help as the spaces get larger.

Here's an example showing it gets the same results:

```python
from pymatgen import Composition
from pymatgen.entries.entry_tools import EntrySet
from pymatgen.analysis.tests.test_phase_diagram import module_dir
from pymatgen.analysis.phase_diagram import (
    PhaseDiagram,
    PatchedPhaseDiagram,
)

entries = EntrySet.from_csv(str(module_dir / "reaction_entries_test.csv"))

pd = PhaseDiagram(entries=entries)
ppd = PatchedPhaseDiagram(entries=entries)

print(ppd.spaces)

# novel entries not in any of the patches
c1 = Composition("H5C2OP")
c2 = Composition("V2PH4C")

print("Full Pd")
print(pd.get_hull_energy(c1))
print(pd.get_hull_energy(c2))

print("Patched Pd")
print(ppd.get_hull_energy(c1))
print(ppd.get_hull_energy(c2))
```

mattmcdermott:
First of all, thank you so much. Your PR and phase diagram updates are literally a godsend for me.

The code seems to be performing pretty well! Was able to make some gigantic (P)PDs no problem. Unfortunately, I still need to play around with `InterfacialReactivity` to get it working with the new ppds, since it actually relies on using the `get_critical_compositions()` method to get the kinks in tieline across the hull. I'm pretty confident I can get something working though.

Thanks again for all your additions!

CompRhys:
No problem, the fact that the slsqp method can be used for more than the decomposition energy (`quasi_e_to_hull`) justifies merging my last PR despite skepticism about the usefulness of the energy in question!

I have no idea how `get_critical_compositions()` works but it looks like it uses the simiplicies so that may be a problem.

I probably have most of the functionality I want from ppd  but it would be cool to try get as much of the functionality from pd as possible.

Do you envisage there being a need for a `GrandPotentialPatchedPhaseDiagram` in your use case?

mattmcdermott:
I don't frequently do grand potential PDs for high-throughput reaction analysis, but I do use them a lot for looking at single reactions. It would be nice eventually to add, but the reality is I can often get away with the time it takes to create several large (i.e. 9-10) element grand PDs at different chemical potential values.

mattmcdermott:
@CompRhys so I thought a bit more about this idea of "stitching together" a phase diagram, and it may be an impossible effort. Consider this example ternary phase diagram which contains no ternary compounds. The question is -- how would we, given only knowledge of the hull along the binary edges of the triangle, be able to predict which binary compounds are stable against each other? For example, how would we ever determine that there is a tie-line between TiFe-TiCl3, but not one between TiFe-TiCl4?
Screen Shot 2020-10-05 at 1 16 17 PM

My naive idea was to do a triangulation scheme to come up with facets, but it turns out that triangulation (at least Delaunay) is done by computing the convex hull anyway, so you're going to run into the same problem.

The reason why I think it might be truly impossible to "beat" the qhull algorithm is that despite being our best algorithm, it also fails due to large combinatorics; that is partially why qhull begins to fail once it hits around 9-10 dimensions... combinatorial explosion in the number of possible facets. So it appears that if we really want to get EVERY possible simplex we will never get there even if the compositional "middle" of the phase diagram is sparse.

Instead, I think the better idea might be to stick to using your decomposition method (with SLSQP) to get several decomposition product sets along the hull between two compositions (i.e. don't use precomputed facets to find kinks, just grab import ones intelligently -- TBD on how to do this). Effectively what I'd like to do is to say "hey, we don't actually need EVERY facet on the hull", and just find good guesses at what it looks like between certain compounds. Does that make any sense?

mkhorton:
Posting this in case it's useful: approximate convex hull algorithm for high-dimensional data, https://arxiv.org/abs/1603.04422

CompRhys:
I believe (?) that we already have the points that define the convex hull as the union of the stable points from the patch pds so as @mattmcdermott says the issue is that patching it together we don't get the simplicies the code uses for free.

I have done a little refactoring to add this `_get_simplex_intersections` method which if we can implement for the ppd would allow us to inherit `get_critical_compositions` from pd. In terms of implementing it I still don't have a firm grasp of what the `Simplex.line_intersections` returns so haven't made any progress.

shyamd:
You might consider using the methodology in the `CompoundPhaseDiagram` to stitch together lower dimension PDs.

mattmcdermott:
@shyamd The compound methodology is definitely helpful, but in this particular case since there are no entries in the "middle" of the gigantic phase diagram (e.g. you only have binary compounds on a ternary diagram like the pic above), you would never find the facets of the hull. In other words, the compound methodology looks for entries which *directly* exist on the tie-line between two compounds on a PD and in this case there are 0 of those compounds. Compare this to the InterfacialReactivity model, which is a true slice of the bigger convex hull and thus uses knowledge about the facet structure to find product *combinations* of compounds which minimize energy (not just single compounds).

CompRhys:
Doing some reading it's the triangulation that brings in the exponential scaling. For a set of n points in d dimensions the expected number of simplicies is . The only way to avoid this will be to not generate all the simplicies. I would propose that I aim to refactor more of the code to add in private methods that could later be approximated in the ppd if someone (@mattmcdermott) has a good idea about a smart approach

CompRhys:
@mkhorton After deciding that the task of stitching back the patches isn't feasible are you still in principle happy to accept this PR based on the functionality of calculating the `e_above_hull`, `equilibrium_reaction_energy` and `quasi_e_from_hull` more quickly when starting from a high dimensional set of entries?

CompRhys:
I have changed the hash for `PDEntry` to the following:

```python
    def __hash__(self):
        return hash(json.dumps(self.as_dict(), sort_keys=True, cls=MontyEncoder))
```

As a consequence I needed to change two tests that were set based on the hash not removing duplicates. 

As `Entry` is and ABC i think it would be good to move the `__eq__` and `__hash__` from `PDEntry` into `Entry` such that they can be inherited by other classes that use the ABC e.g. `ComputedEntry`. Both these functions are written using generic ideas that should be extensible to newer `Entry` types as everything should be `MSONable`. 

```python
    def __eq__(self, other):
        # NOTE Scaled duplicates are not equal unless normalized separately
        if isinstance(other, self.__class__):
            return self.as_dict() == other.as_dict()
        return False

    def __hash__(self):
        return hash(json.dumps(self.as_dict(), sort_keys=True, cls=MontyEncoder))
```

If all the tests apart from the two mentioned above still pass would this be an acceptable thing to do? I imagine that it may impact some users workflows.

shyamd:
It looks like the test used to assert 492 unique entries and now it only has 490 in the hull data file. Can you double check that there should actually only be 490 unique entries? That is my only major concern.

In general, though, the pattern of putting `__eq__` and `__hash__` in an `ABC` interface class is not a good idea. We can get away with it because of `MSONable`.

CompRhys:
> It looks like the test used to assert 492 unique entries and now it only has 490 in the hull data file. Can you double check that there should actually only be 490 unique entries? That is my only major concern.

@shyamd Yes there are literal duplicates in the `pymatgen/analysis/tests/pdentries_test.csv` file. `Li2O,2,0,1,-14.31361175` appears in rows 8 and 14 and `Fe,0,1,0,-6.59394176` appears in rows 11 and 19.

> In general, though, the pattern of putting `__eq__` and `__hash__` in an `ABC` interface class is not a good idea. We can get away with it because of `MSONable`.

Okay I will leave it in PDEntry!

mattmcdermott:
@CompRhys do you have an idea how the new `__hash__` function impacts the speed of operations with `PDEntry` objects? 

Bringing this up because I had to implement a new `__hash__` function for `GibbsComputedStructureEntry`to deal with a multiprocessing issue. I originally tried that exact code with json and it really slowed down everything drastically (not sure exactly which lines but I think anything involving sets/dictionaries of entries).

CompRhys:
@mattmcdermott I don't know how it might affect anything that isn't implemented for ppd so that should explored before this gets merged. 

I would guess the issue wouldn't be so pronounced here as `GibbsComputedStructureEntry` seems to be a much more complicated object with many more nested layers to encode and dump. What hash operation did you end up using?

The other hash I started with was

```python
    def __hash__(self):
        return hash(str(self))
```
but that hash seemed as though it might to be less discriminative

mattmcdermott:
I ended up doing this:

```python
def __hash__(self):
    data_md5 = hashlib.md5(f"{self.composition}_"
                           f"{self.formation_enthalpy}_{self.entry_id}_"
                           f"{self.temp}".encode('utf-8')).hexdigest()
    return int(data_md5, 16)

```

Not saying this is the best solution by any means but it seemed to work decently for me :) FYI I was having trouble creating entry objects on different CPUs that would hash equivalently so I could use them as keys in a shared dictionary

CompRhys:
@mattmcdermott to try and make the `__eq__` and `__hash__` consistent for `Entry` type classes I have been doing a bit of refactoring. One thing I can't quite get my head around is what is supposed to/should happen to the structure when we call `normalize` on a `ComputedStructureEntry`. After calling `normalize` `self.composition` and `self.structure.composition` are not related anymore. 

One option would be to just allow this disconnect (current solution) or the other I can think of would be to disable the `normalize` method for `ComputedStructureEntry` and then have `self.compositon` always return `self.structure.composition`. Do you have any thoughts on this?

It would also be really great if you could check for me whether adding this `__eq__` to your version of `GibbsComputedStructureEnergy` slows your workflow down:

```python
        # NOTE Scaled duplicates i.e. physically equivalent materials
        # are not equal unless normalized separately
        if id(self) == id(other):
            return True

        if isinstance(other, self.__class__):
            if self.entry_id != other.entry_id:
                return False
            if not Entry.__eq__(self, other):
                return False
            # NOTE this is not performant and should be preceeded
            # if faster rigorous checks are available.
            return self.as_dict() == other.as_dict()
        return False
```

For `GibbsComputedStructureEntry` do you ever have entries with the same `entry_id` but different temperatures? should I add another check in for that?

CompRhys:
Due the issues with `__hash__` and `__eq__` for `Entry` and it's child classes (`PDEntry`, `ComputedEntry`, `GrandPotPDentry`, etc.) this PR has ballooned in scope slightly. I have refactored all the `Entry` type classes to have a nominally private/immutable `_composition` as well as `_energy` such that the `__hash__` and `__eq__` can be defined consistently across all of them. 

I have also moved the reaction used to calculate the dummy species for `TransformedPDEntry` inside the class such that it can be normalized (previously you couldn't normalize based on the formulae unit as there are not electronegativities for dummy species). 

Whilst this PR passes all the tests I am not sure about whether it can be merged yet as I imagine that there could be edge cases not covered by the tests that this might create issues for. The speed of `__eq__` may potentially be problematic for some workflows that contain lots of similar entries such that simple checks don't work and the dictionaries need to be compared.

mkhorton:
I appreciate the question was directed to @mattmcdermott, but regarding:

> One option would be to just allow this disconnect (current solution) or the other I can think of would be to disable the normalize method for ComputedStructureEntry and then have self.compositon always return self.structure.composition. Do you have any thoughts on this?

I imagine that the structure object is retained essentially for provenance (or to calculate, say, what bonding is present so that an appropriate correction can be applied). Normalization to the entry's composition would therefore be fine, even if it results in a disconnect/disagreement. That said, a disagreement between the two certainly might be surprising so it could be clarified in the docstring.

More broadly, my only feeling here is that having Entry's be mutable at all is unfortunate, since it complicates things (rather than normalizing always returning a new Entry). I'm not sure the refactoring that would be necessary to change it at this point however. It's a shame we have to spend time thinking about `__hash__` `__eq__` etc. when something like Entry seems ideal for `@dataclass` where these features would come "for free."

I haven't followed this PR as closely though so will defer to @mattmcdermott on this.

mattmcdermott:
Sorry for such a late response! @CompRhys I checked out your fork of pymatgen and some of the entry operations are ~30% slower, but that is because I put an extremely basic `__eq__()` method on my version of GibbsComputedStructureEntry, which deviated from pymatgen a while back:

```python
def __eq__(self, other):
    if isinstance(other, self.__class__):
        return (self.composition == other.composition) and \
               (self.formation_enthalpy == other.formation_enthalpy) and \
               (self.entry_id == other.entry_id) and (self.temp == other.temp)
    else:
        return False
```
Of course, this doesn't check the structure, nor is it a great idea to write a slightly different version of a pymatgen class in my code, but since checking structure is a huge cost that's why I excluded it. Not sure this should be the way it functions in pymatgen though. Now that I've been thinking some more, I wouldn't worry too much about small performance differences -- just do what you think is logical and if something major comes up in my code that I think will help others, I will raise an issue :)

Regarding the hashing... the current methods you added look good to me. It's really very unfortunate that Entry is mutable as @mkhorton mentioned. IMO I would refactor to make immutable, but I also agree that it's not necessary to do in this PR. Nice work on all this btw!

CompRhys:
@mkhorton 
> Normalization to the entry's composition would therefore be fine, even if it results in a disconnect/disagreement. That said, a disagreement between the two certainly might be surprising so it could be clarified in the docstring.

I have it so that it raises a warning if you normalize a computed structure entry but there are definitely edge cases that exist that I haven't looked at i.e. if I make a `GrandPotPDEntry` from a `ComputedStructureEntry` what happens if I normalize.

> when something like Entry seems ideal for `@dataclass` where these features would come "for free."

two points here, `@dataclass` is py3.7+ which would mean pymatgen would need to drop support for <py3.6 and apparently most people still use 3.6** and in #1985 @rkingsbury suggested that there was some benefit to `ComputedEntry` being mutable in regards to adding corrections. I am not sure if `@dataclass` allows for some things to be frozen i.e. the composition, structure etc and others to be mutable i.e. the corrections applied. This may be something that the dev team wants to think about but as we can mostly solve the `__hash__` and `__eq__` issues with getting the correct sets of stable entries for `PatchedPhaseDiagram` without making any breaking changes I think I should try do the minimum required amount such that this PR works. 

That said I am not in a huge rush to merge this PR as I think more consideration of edge cases (like #1985 for my last PR) are needed as we're changing `ComputedEntry`a little.

@mattmcdermott 
> checking structure is a huge cost that's why I excluded it.

I also don't really think we need to but there's balance to this, if a user is using the code as intended they will be using an unique id to distinguish things in which case it could be faster for everyone to not check structures. I am not sure where on the balance between **robust** and **(fairly) robust** to target. 


**although dropping <py3.6 would make the tests pass as it appears to have a floating point error that the other test actions didn't encounter.

mkhorton:
Brief note on the `@dataclass` point, there's an official back port to 3.6 and it's already in the pymatgen requirements. Nevertheless, it was more a general comment than a strong recommendation.

rkingsbury:
Thanks for all the thoughtful work on this @CompRhys ! I appreciate that all this `Entry` stuff is more than you signed up for when you opened the PR. 

I tested `ThermoBuilder`, energy corrections, and Pourbaix diagram code against your branch merged with the MP2020 corrections "on by default" (see #2006) and had no problems. 

This Fixes #1985 in my use cases at least. I reviewed all the changes you made to `Entry` and `ComputedEntry`, and I don't see anything that should cause a problem. My biggest concern was how it might affect energy corrections and normalization, because we are about to change our correction scheme, but those things seem to still work as expected. I also made sure that we can still load old `ComputedEntry` using `from_dict()`.

Although I suspect you're right that there may be edge cases we'll encounter, I've tested the ones I can think of and would be supportive of merging, pending resolution of the couple of minor test failures you have (they're just decimal issues)

IMO, the bug in #1985 is pretty serious and affects common use cases, so I'd rather fix it with something we think is robust and covers the common cases than leave it open while we hunt for edge cases.

pymatgen/entries/computed_entries.py
```diff
@@ -494,7 +500,7 @@ def __str__(self):
         return self.__repr__()
 
     @classmethod
-    def from_dict(cls, d) -> "ComputedEntry":
+    def from_dict(cls, d: dict) -> "ComputedEntry":
```
rkingsbury, line 503: Found a small bug in `ComputedEntry.from_dict`, not related to your changes. Since we're making a lot of changes to `ComputedEntry` here, would you mind changing line 511 to

```
if d.get("correction", 0) != 0 and not d.get("energy_adjustments"):
```

Otherwise de-serializing `ComputedEntry` that don't have a .correction` attribute fails.

CompRhys:
@rkingsbury @mkhorton okay there are some small tweak/optimisations that might still need to be made to `PatchedPhaseDiagram` but those can be done if a future PR all the functionality I think I can add works and they would just be to try speed it up further. I don't think there are any linting errors in the files I changed even though the test failed. The test coverage is pretty good and now `ComputedEntry` should behave more like `PDEntry` so hopefully there aren't actually any breaking edge cases anywhere but my awareness of the general codebase is limited.

mattmcdermott:
@CompRhys I was just working on filtering large chemical spaces and tested out the PatchedPhaseDiagram class.

I ran it on all entries in the 14-dimensional "Y-Mn-O-Li-Cl-C-Na-Fe-W-Bi-Ti-Be-W-S" chemical system (as acquired from MP), and it performed equivalently fast to the `expand_pd()` method I linked to a while back (about 7 seconds) on a single core, but actually took about 20x longer (~155 seconds) with multiprocessing enabled. I think that the current setup of doing a map to each chemical system and providing all_entries each time is causing a great deal of overhead, as it tries to copy the list of all entries into the memory of each process. 

Wasn't sure if you already had planned on addressing this, but to help speed this up, you might try to do some batching so that there are only a few processes created (maybe # processes = # cores), making sure that each one takes a substantial amount of compute time... this system seemed to work well for me in similar multiprocessing applications :)

rkingsbury:
> @rkingsbury @mkhorton okay there are some small tweak/optimisations that might still need to be made to `PatchedPhaseDiagram` but those can be done if a future PR all the functionality I think I can add works and they would just be to try speed it up further. I don't think there are any linting errors in the files I changed even though the test failed. The test coverage is pretty good and now `ComputedEntry` should behave more like `PDEntry` so hopefully there aren't actually any breaking edge cases anywhere but my awareness of the general codebase is limited.

Hi @CompRhys , can we ask you to please split out the `ComputedEntry` hashing related changes into a separate PR? Those are important bugfixes that we'd like to merge sooner rather than later; whereas the new functionality in this PR may take a little longer to review. I know it's a hassle to do that, but will keep the scope of each PR more focused and facilitate review. Thanks for understanding and please let us know if you have questions.

CompRhys:
I am closing this PR in favour of a new PR that I will fork from my branch for #2014 as otherwise merging this will be riddled with merge-conflicts. I will make that PR once #2014 is merged.

PR #1979, opened 2020-10-07
Remove self.kwargs.get pattern in MP input sets

utf:
A common pattern in many of the MP VASP input sets is to store the input kwargs in the constructor and then access them using `self.kwargs.get` in the `incar` and `kpoint` functions. Unfortunately, this can lead to discrepancies between the actual attributes of the input set and the data stored in the `kwargs` variable.

For example, the following causes an error:
```python
static_set = MPStaticSet(structure, user_incar_settings=None)
static_set.incar()

#  File ".../pymatgen/io/vasp/sets.py", line 974, in incar
#    self.kwargs.get("user_incar_settings", {}).keys()
# AttributeError: 'NoneType' object has no attribute 'keys'
```

This is surprising since the default value of `user_incar_settings` in the `VaspInputSet` constructor is `None`, and the following code works fine.
```python
static_set = MPStaticSet(structure)
static_set.incar()
```

The issue arises because the `incar` and `kpoints` functions call `self.kwargs.get("user_incar_settings", {}).keys()` which evaluates to `None.key()`. This issue can be avoided by just calling `self.user_incar_settings.keys()`. This is because the `VaspInputSet` constructor initialises `user_incar_settings` to as an empty dictionary if `None` is given as the argument.

I've left the `self.kwargs` pattern in the input set constructor as the `kwargs` variable is used by the monty decoder in the `as_dict` function. Note that I only edited MP input sets.

This issue is responsible for a bug in atomate: hackingmaterials/atomate#445.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -883,6 +883,7 @@ class MPRelaxSet(DictSet):
         :param kwargs: Same as those supported by DictSet.
         """
         super().__init__(structure, MPRelaxSet.CONFIG, **kwargs)
+        self.kwargs = kwargs
 
 
 class MPScanRelaxSet(DictSet):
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @utf!

PR #1986, opened 2020-10-20
Add f* diagram generator

jonathanjdenney:
## Summary

Added f* diagram generator.

## TODO (if any)

to be done

## Checklist

in the pull request title.

      on your local machine.
      to type check your code.

allowing commits.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,7 +71,7 @@ repos:
         exclude: tests
 
   - repo: https://github.com/MarcoGorelli/cython-lint
-    rev: v0.11.1
+    rev: v0.12.4
     hooks:
       - id: cython-lint
         args: [--no-pycodestyle]


mkhorton:
Hi @jonathanjdenney, thanks for this.

A few comments before merging, mostly just to follow Python conventions:

* The file `pymatgen/analysis/Fstar/__pycache__/fstar.cpython-36.pyc` should be removed
* The folder `pymatgen/analysis/Fstar/` should be removed `pymatgen/analysis/fstar/`
* The files `neutron_factors.csv` `xray_factors_2.csv` are duplicated in the `tests` folder, this is unnecessary.
* `FStarDiagram_test` -> `test_fstardiagram` (method names should be lower case)
* `x_ray_scatter_df`, `neutron_scatter_df` should be `X_RAY_SCATTER_DF`, `NEUTRON_SCATTER_DF` (global variables upper case)
* instead of `try` `except AttributeError` you can do an `if` statement, e.g. `if isinstance(sp, Species` or `if hasattr(sp, "element")` etc.

I see changes were made via direct file upload, I highly recommend [GitHub Desktop](https://desktop.github.com) for making easier edits.

These are all my suggestions for now, but I'll take another look after the above changes.

mkhorton:
Also, I would add a reference to the papers using f* (e.g. https://doi.org/10.1063/1.5044555 https://doi.org/10.1021/acs.chemmater.9b03646) to the docstring to point newcomers in the right direction.

jonathanjdenney:
Hi @mkhorton ,
I will get to work on your suggestions. Thanks for looking at it.

jonathanjdenney:
Hi @mkhorton ,
I believe I have addressed all of your comments. Please let me know if I missed anything.

mkhorton:
Thanks @jonathanjdenney ! I'll take another look. Were some tests removed in the most recent iteration?

jonathanjdenney:
@mkhorton No tests were actually removed. I had previously accidentally added a file that was a non-functioning version of the test file which I have now removed.

pymatgen/analysis/Fstar/fstar.py
```diff
@@ -0,0 +1,246 @@
...
+    NEUTRON_SCATTER_DF = pd.read_csv(f)
+    # from http://www.ccp14.ac.uk/ccp/web-mirrors/neutrons/n-scatter/n-lengths/LIST~1.HTM
+
+
+class FStarDiagram:
+    """
+    Take a list of structure objects and/or cifs and use them to generate an f* phase diagram.
+    """
+
+    def __init__(self, structure_objects=None, cifs=None, cif_index=0, occupancy_tolerance=5,
```
mkhorton, line 36: I'm reviewing this again before merging, I think this `__init__` should be simplified to be just:

```
def __init__(self, structures, scattering_type='X-ray_simple', custom_scatter=None)
```

If the structures come from a CIF, when a person is using this class they would write something like:

```
structures = [CifParser(file, occupancy_tolerance=5).get_structures(primitive=False)[0] for file in cifs]
fstar = FStarDiagram(structures)
```

I know this is an extra line for the end-user, but I get the feeling that the CIF input (e.g. especially cif-index etc) is very specific to where your CIFs are coming from.

The philosophy is that the `Structure` is the canonical way that pymatgen deals with structural information internally.

pymatgen/analysis/Fstar/fstar.py
```diff
@@ -0,0 +1,246 @@
...
+        for ind1, struct in enumerate(self._equiv_inds):
+            fstar_df = pd.DataFrame(columns=self.site_labels, data=[[0.0 for i in self.site_labels]])
+            for ind2, site in enumerate(struct):
+                occ_f_list = []
+                mult = len(site)
+                site_frac_coord = str(self._structures[ind1][site[0]].frac_coords)
+                column = [label for label in self.site_labels if site_frac_coord in label]
+                elements_and_occupancies = self._symstructs[ind1][site[0]].species_and_occu.items()
+                for sp, occ in elements_and_occupancies:
+                    if self.cif_list:
```
mkhorton, line 171: Likewise, there's a lot of CIF-specific functionality here, which is probably simpler if working from the Structure.

It looks like the purpose of lines 174:196 is to create a list of fractional co-ordinates normalized to the [0, 1] range? This can be done on the Structure object something like:

```
structure = Structure.from_sites(input_structure.sites, to_unit_cell=True)
```

pymatgen/analysis/Fstar/tests/test_fstar.py
```diff
@@ -0,0 +1,38 @@
...
+        self.struct_list = [CifParser(os.path.join(test_dir, file)).get_structures(primitive=False)[0]
+                            for file in self.cif_list]
+        self.fstar = FStarDiagram(structure_objects=self.struct_list)
+
+    def test_edit_fstar_diagram(self):
+
+        self.assertEqual(self.fstar.site_labels, ['[0. 0. 0.]Li', '[0.  0.  0.5]Co', '[0.   0.   0.25]O'])
+        new = FStarDiagram(structure_objects=self.struct_list)
+        self.assertEqual(self.fstar.plot, new.plot)
+        new.edit_fstar_diagram(combine_list=[['[0.  0.  0.5]Co', '[0. 0. 0.]Li']])
```
mkhorton, line 27: Could you explain what the expected input is for `combine_list`? How does the user know to use `'[0.  0.  0.5]Co'` etc.?

jonathanjdenney:
@mkhorton 
Hello Matt,
Sorry for not getting to you comments sooner, I've had other deadlines. The purpose of including a way to input cifs directly into the script was to make it more user friendly to our target audience, experimentalists. Experimentalists will typically be more familiar with cifs over structure objects. In addition, this method allows the user to easily input non-physical values for the occupancies which may have resulted from the refinement. As far as I know, the cif parser will round greater than 1 occupancies to 1 regardless of what tolerance you set. This script can take structure objects as its input without any trouble, so is it really unacceptable that I've given the option to use cifs? 
In regards to your last comment I agree it is a bit confusing. I will add to the code description an explanation that you use the labels generated in the site_labels list, which can be viewed by printing the list. It is input as a list of strings. 
Best,
Jonathan

mkhorton:
> In addition, this method allows the user to easily input non-physical values for the occupancies which may have resulted from the refinement. As far as I know, the cif parser will round greater than 1 occupancies to 1 regardless of what tolerance you set. 

If there's a case where unphysical occupancies are useful, then we should change `CifParser` accordingly.

>  The purpose of including a way to input cifs directly into the script was to make it more user friendly to our target audience, experimentalists.
> This script can take structure objects as its input without any trouble, so is it really unacceptable that I've given the option to use cifs?

Maybe to explain the issue a different way, pymatgen is intended as a library. For someone using it, regardless of their background or expertise (I started off as an experimentalist myself), you have to learn how to use Python, how to install the package, how to import the relevant functionality you want etc.

Therefore, with the recommendation to people using the script will be to say, "first you write one line to import your CIFs" and then "you run the FStarDiagram generator", in my own view the extra directive to write the additional line is not the limiting hurdle for using a library like pymatgen.

The objection to having the CIF input is that it creates confusion and makes it more difficult to reason about what the code is doing internally. The existence of the Structure object is not to create yet another file format (e.g. on Materials Project, we offer CIF downloads, not Structure downloads), it's to create a standardized interface to crystal structures internally within pymatgen.

More importantly however, this discussion has highlighted a limitation (the CIF unphysical occupancies mentioned), and we'd be papering over this limitation rather than addressing it properly. By adding the capability to CifParser, other people can also have the benefit of importing CIF with unphysical occupancies, and the resulting code within FStarDiagram will be a lot simpler too.

PR #1989, opened 2020-10-27
Fix outcar serialization

utf:
## Summary

`Outcar.as_dict` method fails if the OUTCAR doesn't contain the average electrostatic potential at atomic cores. This can occur if `ICORELEVEL = 1` is set. I've fixed this behaviour and added a test. 

      on your local machine.
      to type check your code.


utf:
This should be ready to merge btw.

mkhorton:
Thanks @utf

PR #1991, opened 2020-11-05
Change convention used to generate Abinit typat/znucl variables

gmatteo:
## Summary

* Use  deterministic algorithm (not based on python set) to obtain list of elements before writing Abinit input file.
  This is required to maintain compatibility with the original implementation of structure.types_of_specie and all the DFPT 
  results produced so far.

* Add refine_struct optional argument to CifWriter


shyuep:
Thanks. Can you uncomment abinit in the setup.cfg and make sure the pydocstyle tests pass too? Thanks.

gmatteo:
Hi, I've removed abinit from setup.cfg and fixed the doc strings as requested. All tests seem to be OK.

shyuep:
Thanks.

PR #1996, opened 2020-11-16
Add tests and SCAN variant support to Vasprun.run_type

rkingsbury:
At present, `OUTCAR.run_type` incorrectly sets the `run_type` to 'GGA' when METAGGA is equal to `R2SCAN` or `RSCAN`. This commit adds support for these functionals in `run_type` and also adds a test for regular `SCAN` run_type, which was not previously covered by a test.


shyuep:
I think we can make it stricter. Explicitly set the type to "unknown" with a warning if it is not within the list of defined types.

mkhorton:
Might be more robust to set it to the METAGGA tag if that tag is specified, in terms of if additional meta GGAs are added in future.

mkhorton:
Specifically this line:

```
        elif self.incar.get("METAGGA", "").strip().upper() in METAGGA_TYPES:
```

could instead be:

```
        elif self.incar.get("METAGGA"):
            rt = self.incar.get("METAGGA").strip().upper() in METAGGA_TYPES
```

mkhorton:
I would also break out the +U logic to a separate if statement, so it doesn't have to be duplicated, i.e. at the end of the method:

```
if self.is_hubbard or self.parameters.get("LDAU", True):
    rt += "+U"
```

and maybe also add to the end:

```
IVDW_TYPES = {1: 'DFT-D2', 10: 'DFT-D2', 11: 'DFT-D3', 12: 'DFT-D3-BJ', 2: 'TS', 20: 'TS', 21: 'TS-H', 202: 'MBD', 4: 'dDsC'}
if self.incar.get("IVDW", "").strip().upper() in IVDW_TYPES:
    rt += "vdW-" + IVDW_TYPES[self.incar.get("IVDW", "").strip().upper()]
elif self.incar.get("IVDW"):
   rt += "vdW-unknown"
```

since this is an important case not captured by current logic.

mkhorton:
(The `TODO: Fix for other functional types like PW91, other vdW types, etc.` could be removed then too)

pymatgen/io/vasp/outputs.py
```diff
@@ -658,21 +673,25 @@ def run_type(self):
...
+        elif self.parameters.get("GGA"):
+            incar_tag = self.parameters.get("GGA", "").strip().upper()
+            rt = GGA_TYPES.get(incar_tag, incar_tag)
         elif self.potcar_symbols[0].split()[0] == 'PAW':
             rt = "LDA"
-        elif self.parameters.get("GGA", "").strip().upper() in GGA_TYPES:
-            rt = GGA_TYPES[self.parameters.get("GGA", "").strip().upper()]
-            if self.is_hubbard or self.parameters.get("LDAU", True):
-                rt += "+U"
+
```
shyuep, line 684: We should add a

```python
else:
    rt = "unknown"
    warnings.warn("Unknown Run type!")
```

rkingsbury: @shyuep I added this. Realistically I'm not sure it will ever be triggered, but it's good to catch anyway. Right now if the contents of `METAGGA` or `GGA` are not found in the respective dicts, the code will just pass through whatever that content is to the `run_type`. So if `vasprun.xml` contains `GGA=RANDOMFUNCTIONAL`, the `run_type` that would be returned is `RANDOMFUNCTIONAL`.

So, the only way the `unknown` would ever be triggered were if the `GGA` tag were missing from `vasprun.xml`. I'm not sure it's possible for that to happen and still have a valid .xml file, but again, it's good to cover all bases.

mkhorton: If the GGA tag is missing it defaults to whatever functional is defined in the POTCAR. I'm not sure how that would change the logic here.

Thanks for adding the extra lines @rkingsbury! Appreciate the clean-up.

rkingsbury: I think `vasprun.xml` contains GGA and METAGGA tags even if they are not set in the INCAR. For example, I know that GGA calculations still have `METAGGA=--` in the file. As you say, if GGA is not in the INCAR VASP will default to something, but I think it will always put that something in the vasprun.xml file. Could be wrong though.

shyuep: I think the catch is in case VASP implements new functionals right?

pymatgen/io/vasp/outputs.py
```diff
@@ -658,21 +673,28 @@ def run_type(self):
...
-        elif self.incar.get("METAGGA", "").strip().upper() in METAGGA_TYPES:
-            rt = METAGGA_TYPES[self.incar.get("METAGGA", "").strip().upper()]
-            if self.is_hubbard or self.parameters.get("LDAU", True):
-                rt += "+U"
+        elif self.incar.get("METAGGA") and self.incar.get("METAGGA") != "--":
+            incar_tag = self.incar.get("METAGGA", "").strip().upper()
+            rt = METAGGA_TYPES.get(incar_tag, incar_tag)
+        elif self.parameters.get("GGA"):
+            incar_tag = self.parameters.get("GGA", "").strip().upper()
+            rt = GGA_TYPES.get(incar_tag, incar_tag)
```
rkingsbury, line 681: @shyuep I think this block will already catch any new functionals VASP may implement

pymatgen/io/vasp/outputs.py
mkhorton: I think Ryan's point is the the `GGA_TYPES` variable is just to map the VASP GGA tag names to a human-readable name (e.g. PS -> PBEsol). If it's not in GGA_TYPES because VASP implements a new functional, you would just get the raw VASP tag name instead.

rkingsbury:
@mkhorton @shyuep do you have any other comments / concerns here?

mkhorton:
Thanks @rkingsbury! going to go ahead and merge

PR #2000, opened 2020-11-19
improved Fermi level / bandgap detection in Vasprun

rkingsbury:
- Implement a `smart_efermi` property in `Vasprun` to detect cases in which the Fermi level reported by VASP crosses one of the bands, but the bands otherwise indicate insulating behavior. This issue sometimes causes `Vasprun` to report inaccurate bandgaps (for example, insulators may be reported to have zero bandgap)

- Changes `Vasprun.get_bandstructure` to default to using the `smart_efermi` property to determine Fermi energy.

- Add type hinting and update docstrings in affected methods


rkingsbury:
@utf @mkhorton see `smart_efermi` method based on Alex's code. Fermi level seems to be set in the `init` method of `Outcar`, `BSVasprun`, and `Wavecar`. It is also read from the DOS in `Vasprun`.

Would it make sense to default to this method inside  `Vasprun.get_band_structure()`? Where else in outputs.py would we want to use this by default?

pymatgen/io/vasp/outputs.py
```diff
@@ -874,6 +874,35 @@ def eigenvalue_band_properties(self):
...
+        # inside a band
+        fermi_crosses_band = False
+        for spin_eigenvalues in eigenvalues.values():
+            eigs_below = np.any(spin_eigenvalues < self.efermi, axis=1)
+            eigs_above = np.any(spin_eigenvalues > self.efermi, axis=1)
+            if np.any(eigs_above & eigs_below):
+                fermi_crosses_band = True
+        # if the Fermi level crosses a band, the eigenvalue band properties is a more
+        # reliable way to check whether this is a real effect
+        bandgap, cbm, vbm, _ = self..eigenvalue_band_properties
```
utf, line 893: should only be one `.`?

pymatgen/io/vasp/outputs.py
```diff
@@ -874,6 +874,35 @@ def eigenvalue_band_properties(self):
...
+        bandgap.
+        """
+        # finding the Fermi level is quite painful, as VASP can sometimes put it slightly
+        # inside a band
+        fermi_crosses_band = False
+        for spin_eigenvalues in eigenvalues.values():
+            eigs_below = np.any(spin_eigenvalues < self.efermi, axis=1)
+            eigs_above = np.any(spin_eigenvalues > self.efermi, axis=1)
+            if np.any(eigs_above & eigs_below):
+                fermi_crosses_band = True
```
utf, line 890: Also, `spin_eigenvalues` isn't defined at this point.  You can change this section to
```suggestion
        for spin_eigenvalues in self.eigenvalues.values():
            # drop weights and set shape nbands, nkpoints
            spin_eigenvalues = spin_eigenvalues[:, :, 0].transpose(1, 0)
            eigs_below = np.any(spin_eigenvalues < self.efermi, axis=1)
            eigs_above = np.any(spin_eigenvalues > self.efermi, axis=1)
            if np.any(eigs_above & eigs_below):
                fermi_crosses_band = True
```

pymatgen/io/vasp/outputs.py
```diff
@@ -874,6 +874,35 @@ def eigenvalue_band_properties(self):
...
+        # if the Fermi level crosses a band, the eigenvalue band properties is a more
+        # reliable way to check whether this is a real effect
+        bandgap, cbm, vbm, _ = self..eigenvalue_band_properties
+        if not fermi_crosses_band:
+            # safe to use VASP fermi level
+            efermi = vasprun.efermi
+        elif fermi_crosses_band and bandgap == 0:
+            # it is actually a metal
+            efermi = vasprun.efermi
```
utf, line 899: Also, `vasprun` should be `self` in this section:

```suggestion
        if not fermi_crosses_band:
            # safe to use VASP fermi level
            efermi = self.efermi
        elif fermi_crosses_band and bandgap == 0:
            # it is actually a metal
            efermi = self.efermi
```

pymatgen/io/vasp/outputs.py
```diff
@@ -874,6 +883,38 @@ def eigenvalue_band_properties(self):
                         cbm_kpoint = k
         return max(cbm - vbm, 0), cbm, vbm, vbm_kpoint == cbm_kpoint
 
+    @property
+    def smart_efermi(self):
```
mkhorton, line 887: Rename to `calculate_efermi`, and modify the docstring to add:

> Calculate the Fermi level based on band occupancies, as an alternative to using the Fermi level reported directly by VASP. For a semiconductor, the Fermi level will be put in the center of the gap. This algorithm works by [...]

mkhorton:
Aside from the comment above this seems ready to merge.

pymatgen/io/vasp/outputs.py
mkhorton: It also should not be a property.

pymatgen/io/vasp/outputs.py
```diff
@@ -721,25 +721,31 @@ def get_computed_entry(self, inc_structure=True, parameters=None,
             return ComputedStructureEntry(self.final_structure, self.final_energy, parameters=params, data=data)
         return ComputedEntry(self.final_structure.composition, self.final_energy, parameters=params, data=data)
 
-    def get_band_structure(self, kpoints_filename=None, efermi=None,
-                           line_mode=False, force_hybrid_mode=False):
+    def get_band_structure(self,
+                           kpoints_filename: Optional[str] = None,
+                           efermi: Union[str, float, None] = "smart",
```
utf, line 726: I'm not sure about modifying the Fermi level by default?

Seems like it could lead to a lot of confusion for people who aren't aware of this option.

mkhorton: I agree, thanks @utf.

mkhorton: > The Fermi energy associated with the bandstructure, in eV. By default,
                 uses the 'smart_efermi' attribute of the Vasprun, which in some cases
                 may differ from (but be more accurate than) the value in vasprun.xml.

This docstring should also be reworded to state explicitly what this does (ie remove the reference to "smart"). I don't think we should making claims as to the accuracy though it *can* be more accurate.

The type hint could be `Optional[Union[float, str]]` too.

mkhorton:
Thanks @rkingsbury

PR #2005, opened 2020-12-02
Magmom check

jmmshn:
## Added automatic magmom validation during set creation.
Currently, if there is missing or incompatible magmom data you just get an error when the INCAR is written.
This is not ideal since this often happens on the remote cluster.
So adding an additional validation function should make the behavior smoother.

## Added DBPhaseDiagram
New class for that reading and writing PhaseDiagram to DB

## Added ElementBase
Similar to the case above Element should be a subclass of ElementBase so that other subclasses can be created with a subset of the periodic table (Userful for automatic parsing by the API)


pymatgen/io/vasp/sets.py
```diff
@@ -3179,3 +3190,57 @@ def batch_write_input(
     [[0, 0, 0]],
     site_properties={"magmom": [[0, 0, 1]]},
 )
+
+
+def get_valid_magmom_struct(structure, inplace=True, spin_mode="auto"):
```
jmmshn, line 3195: Function is here

mkhorton: Can you clarify when this happens? Is this for when some site properties are missing or `None`? Not sure if an easier solution might be to cast to the `Magmom` class.

pymatgen/io/vasp/tests/test_sets.py
```diff
@@ -447,6 +450,46 @@ def test_user_potcar_settings(self):
...
+        get_valid_magmom_struct(structure=struct, inplace=True, spin_mode="v")
+        struct.insert(0, "Li", [0, 0, 0], properties={"magmom": 10.0})
+        with self.assertRaises(TypeError) as context:
+            get_valid_magmom_struct(structure=struct, inplace=True, spin_mode="a")
+        self.assertTrue("Magmom type conflict" in str(context.exception))
+
+        # Test the behavior of MPRelaxSet to atomacically fill in the missing magmom
+        struct = self.structure.copy()
+        get_valid_magmom_struct(structure=struct, inplace=True, spin_mode="s")
+        struct.insert(0, "Li", [0, 0, 0])
```
jmmshn, line 479: This is a common way for the magmom error to occur

pymatgen/io/vasp/sets.py
jmmshn: Both cases can happen but the missing properties case is probably more common I flagged a case in the tests that occur pretty often.  It can be avoided but it just bothers me that no error is raised until the job is submitted and running.

Also, my understanding is that Magmom class is focused on SOC stuff.

pymatgen/analysis/phase_diagram.py
```diff
@@ -260,71 +257,52 @@ def from_dict(cls, d):
         return cls(d["composition"], entry)
 
 
-class PhaseDiagram(MSONable):
+class DBPhaseDiagram(MSONable):
```
jmmshn, line 260: Change to PhaseDiagram, Now inherits from DBPhaseDiagram which has the simpler constructor and the stander as_dict and from_dict

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -587,6 +604,25 @@ def test_str(self):
         self.assertIsNotNone(str(self.pd))
 
 
+class DBPhaseDiagramTest(PhaseDiagramTest):
```
jmmshn, line 607: The testing of the new DBPhaseDiagram clas is here

pymatgen/core/periodic_table.py
```diff
@@ -29,130 +27,8 @@
...
-    Ds = "Ds"
-    Rg = "Rg"
-    Cn = "Cn"
-    Nh = "Nh"
-    Fl = "Fl"
-    Mc = "Mc"
-    Lv = "Lv"
-    Ts = "Ts"
-    Og = "Og"
+class ElementBase(Enum):
```
jmmshn, line 30: Small change to how Elment works.  This is so that we can restrict the options on the API

pymatgen/io/vasp/sets.py
jmmshn: BTW the pre-commit stuff is making these reviews a bit nasty... do you wanna run the pre-commit on the entire codebase at some point or just do it piece by piece like this?

Or should we have some kind of simplified workflow like turn off pre-commit until the end right before merge??

Again sorry for the mess.

mkhorton: Yup, I was actually intending to do that soon anyway, stay tuned ...

> Or should we have some kind of simplified workflow like turn off pre-commit until the end right before merge??

Is this possible?

mkhorton: > Also, my understanding is that Magmom class is focused on SOC stuff.

Magmom class is an attempt to reconcile vector vs scalar magmoms, e.g. if you run without SOC it will cast to a float, via float(Magmom) -> scalar.

I'm curious how missing properties actually happen in practice however. Guessing it's a case where you insert an additional site?

jmmshn: Yeah that is the most common case for me, but there are other times when sites get copied and mutate where you can loose data.
Since I don't think we can add validation in all those places this seems like a good place for a sanity check

mkhorton: Ok, have run linters module-wide, it has introduced a few merge conflicts though mostly in module sort order.

jmmshn: Awesome. I saw that.
Does this mean I have to redo this PR?

jmmshn: NVM, I just merge in the new master.

pymatgen/io/vasp/tests/test_sets.py
mkhorton: Do you think there's logic we could add to `struct.insert()` to help there? (e.g. look at what site properties are defined, and warn and/or add a default if there's one missing in the inserted site)

pymatgen/io/vasp/sets.py
mkhorton: Yeah, or you can fix merge conflicts on GitHub.com too, usually only takes a minute or two

setup.cfg
```diff
@@ -17,7 +17,7 @@ statistics = True
 exclude=pymatgen/__init__.py,docs_rst/*.py
 
 [flake8]
-exclude = .git,__pycache__,docs_rst/conf.py,tests,__init__.py
+exclude = .git,__pycache__,docs_rst/conf.py,tests,__init__.py,test*.py
```
jmmshn, line 20: this might be an edit error? not sure where it came from

pymatgen/io/vasp/tests/test_sets.py
jmmshn: You mean instead of doing this? So I think we can leave that up to users, technically there is nothing wrong with some missing properties at a few sites.  Ex: I sometimes get an intercollated structure where all the Li have an 'insertion_energy' property

mkhorton:
Thanks @jmmshn. Renamed `DbPhaseDiagram` to be more clear and to match `BaseElement`, also expanded the docstring.

PR #2007, opened 2020-12-02
fix bug in GGA run_type when METAGGA=None

rkingsbury:
This PR fixes a recently discovered bug in which the `Vasprun.run_type` of a calculation in which `METAGGA=None` would not be set correctly.


mkhorton:
Thanks @rkingsbury

PR #2011, opened 2020-12-09
Add IsayevNN near neighbor class

utf:
## Summary

Added `IsayevNN`, a class for finding near neighbors based on: [10.1038/ncomms15679](http://www.nature.com/articles/ncomms15679).

Two sites are considered coordinated if: (i) they share a Voronoi face and (ii) the near neighbor distance is less than the sum of the covalent radii + 0.25 Å.

Also added tests.


mkhorton:
Thanks @utf!

PR #2014, opened 2020-12-11
feature: more standard approach to __hash__ and __eq__ for Entry and children

CompRhys:
Split out the Entry changes as cleanly as possible from #1974 @rkingsbury


rkingsbury:
Thank you @CompRhys ! Looks like you're missing a newline character in pymatgen.entries.__init__.py. Otherwise looks good to me. I tested with the MP2020 corrections enabled as well and saw no issues.

pymatgen/analysis/phase_diagram.py
```diff
@@ -245,7 +280,7 @@ def as_dict(self):
             "@module": self.__class__.__module__,
             "@class": self.__class__.__name__,
             "entry": self.original_entry.as_dict(),
-            "composition": self.composition,
+            "sp_mapping": self.sp_mapping,
```
CompRhys, line 283: this may be a breaking change for workflows storing transformed entries as dictionaries

mkhorton: @shyamd do you have any comments on this before merged?

pymatgen/entries/computed_entries.py
```diff
@@ -944,11 +1025,22 @@ def from_dict(cls, d) -> "GibbsComputedStructureEntry":
...
+    def __hash__(self) -> int:
+        # NOTE based on hash choice of @mattmcdermott + class name
+        data_md5 = hashlib.md5(
+            f"{self.__class__.__name__}"
+            f"{self.entry_id}"
+            f"{self.composition.reduced_formula}"
+            f"{self._energy}"
+            f"{self.temp}".encode('utf-8')
+        ).hexdigest()
+        return int(data_md5, 16)
```
mkhorton, line 1046: Isn't there an easier way to hash this? e.g. `hash(the_string)`, using hexdigest/hashlib is a bit over complicating I think

CompRhys: I adopted `hashlib` to match what @mattmcdermott said he was already using for his `ComputedGibbsStructureEntry`. I think that the python `hash()` isn't guaranteed to be the same across different processes which was a requirement for matt's work. I don't know whether there are alternatives to making `hashlib` a dependency that would keep this functionality.

mkhorton: Thanks @CompRhys, that was a subtlety I wasn't aware of, I was under the impression `hash` was fully deterministic for strings. I have no objection to using `hashlib`.

pymatgen/analysis/phase_diagram.py
```diff
@@ -68,6 +66,7 @@ def __init__(
         energy: float,
         name: str = None,
         attribute: object = None,
+        per_atom: bool = False,
```
shyamd, line 69:
```suggestion
```

pymatgen/analysis/phase_diagram.py
```diff
@@ -76,8 +75,9 @@ def __init__(
             name (str): Optional parameter to name the entry. Defaults
                 to the reduced chemical formula.
             attribute: Optional attribute of the entry. Must be MSONable.
+            per_atom (bool): Whether the energy given is per atom.
```
shyamd, line 78:
```suggestion
```

pymatgen/analysis/phase_diagram.py
```diff
@@ -76,8 +75,9 @@ def __init__(
...
         """
-        super().__init__(composition, energy)
+        super().__init__(composition, energy, per_atom)
```
shyamd, line 80:
```suggestion
        super().__init__(composition, energy)
```

pymatgen/analysis/phase_diagram.py
shyamd: TransformedPDEntries are usually only used when doing alternative phase diagrams, so this shouldn't affect me.

shyamd (commented):
Some modifications i've suggested. @mattmcdermott  has said he no longer needs the hash consistency and in general, it's not a good idea to override salted hashes for `__hash__`. This was done on purpose early on in python 3 to prevent insecurities in dictionaries and sets. 

It also looks like this PR is doing more than fixing __hash__ and __eq__. What is this new TransformedPDEntry logic for? What is `sp_mapping`? Can you add docstrings and maybe rename `sp_mapping` to something that is a bit more easily readable and understandable? There's a lot of changes to PD. Can you summarize what is going on and why?

pymatgen/entries/__init__.py
```diff
@@ -114,5 +142,25 @@ def as_dict(self) -> dict:
...
+    def __eq__(self, other):
+        # NOTE Scaled duplicates i.e. physically equivalent materials
+        # are not equal unless normalized separately
+        if self is other:
+            return True
+
+        if isinstance(other, self.__class__):
+            # NOTE this is not performant and should be overwritten or
+            # preceeded if faster rigorous checks are available.
+            return self.as_dict() == other.as_dict()
```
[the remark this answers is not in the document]
CompRhys: `as_dict()` also checks name and attribute for `PDEntry` which is why it is used here. In general I left that `as_dict()` was the best fail safe option as it will work for all `Entry` children. The only worry is that `as_dict()` of a `ComputedStructureEntry` is very slow hence the note about preceding with faster checks.

CompRhys:
> What is this new TransformedPDEntry logic for? What is `sp_mapping`? Can you add docstrings and maybe rename `sp_mapping` to something that is a bit more easily readable and understandable? There's a lot of changes to PD. Can you summarize what is going on and why? 

I missed that I hadn't updated the docstring here, I will do it now! I haven't changed the logic behind `TransformedPDEntry` just moved in from inside `CompoundPhaseDiagram`, the reason for this was that previously `TransformedPDEntry` didn't have any information about what the terminal compositions and changing this allowed for normalisation of entries. 

`sp_mapping` is the species mapping between dummy species and the terminal compositions. I inherited the name from `CompoundPhaseDiagram` but can change it to something clearer? 

The other changes to PD are only cosmetic/for clarity 

1. several are due to swapping older `.. attribute:: energy` style docstrings to the google docstring format.
2. some are to use list comprehensions rather than for loops that append to a set/list (not that this speed difference is significant here)
3. `chempot_range_map` had some anti-patterns of using `pd=self` then using `pd.method()` than I took out
4. `finalfacets` became `final_facets` to be more readable also `allfacets` to `all_facets`
5. removing strange auto-formatting edge cases e.g. assigning `test_open` in line 1187 because it was previously being computed multiple times and each time it was being spread over 5 lines by the auto-formatting used.

pymatgen/entries/__init__.py
```diff
@@ -46,16 +52,27 @@ def __init__(self, composition: Composition, energy: float):
...
             energy (float): Energy of the entry.
+            per_atom (bool): Whether the energy given is per atom.
         """
-        self._energy = energy
-        self.composition = Composition(composition)
+        self._composition = Composition(composition)
+        if per_atom:
+            self._energy = energy * self.composition.num_atoms
+        else:
+            self._energy = energy
```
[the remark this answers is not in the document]
CompRhys: I had added this for personal convenience but will remove from the PR if there could be issues it causes in other peoples workflows!

pymatgen/entries/computed_entries.py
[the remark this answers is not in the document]
CompRhys: I will replace this custom `__hash__` with a custom `__eq__` that checks the temperature as an early check then.

pymatgen/entries/computed_entries.py
```diff
@@ -565,6 +570,36 @@ def as_dict(self) -> dict:
...
+        # NOTE there is a trade-off between using `_energy` and `energy`
+        # as `energy` requires the calculation of the corrections. It is
+        # more discriminative but also more expensive.
+        data_md5 = hashlib.md5((
+            f"{self.__class__.__name__}"
+            f"{self.entry_id}"
+            f"{self.composition.reduced_formula}"
+            f"{self._energy}").encode('utf-8')
+        ).hexdigest()
+        return int(data_md5, 16)
```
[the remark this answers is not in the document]
CompRhys: hashes are supposed to be as discriminative as possible whilst still being fast? is there a point to making the hash less discriminative for no real improvement in speed all that does it require more calls to `__eq__` which has to prioritise correctness over speed.

pymatgen/entries/computed_entries.py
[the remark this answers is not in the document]
CompRhys: I believe that the temperature check should be done in both to ensure that the hash is discriminative and prevent unnecessary calls to `__eq__`. If multiple `GibbsComputedStructureEntry` are derived from a single `ComputedStructureEntry` do you assign a new `entry_id` to each?

pymatgen/entries/computed_entries.py
```diff
@@ -565,6 +570,36 @@ def as_dict(self) -> dict:
...
+        if isinstance(other, self.__class__):
+            if self.entry_id != other.entry_id:
+                return False
+            if self._energy != other._energy:
+                return False
+            if self._composition != other._composition:
+                return False
+            # NOTE this is not performant and should be preceeded
+            # if faster rigorous checks are available.
+            return self.as_dict() == other.as_dict()
```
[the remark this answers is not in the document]
CompRhys: If you're happy to set this as a sufficient requirement for equality of `ComputedEntry` I will adopt this as it will be faster. 

I was testing for differences and then full equality using `as_dict()` as I felt uncomfortable returning that two things are equal when they could easily be different.

CompRhys:
> Some modifications i've suggested. @mattmcdermott has said he no longer needs the hash consistency and in general, it's not a good idea to override salted hashes for `__hash__`. This was done on purpose early on in python 3 to prevent insecurities in dictionaries and sets.

shall I remove the `hashlib` dependency and go back to hashing the strings? It depends on whether it's likely that other people encounter the same consistency issues I can't think of any security issues but that doesn't mean they don't exist

pymatgen/entries/__init__.py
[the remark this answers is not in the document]
shyamd: Ok, i was worried that there might be a chance that two `as_dict`s might not be equal and that's very hard to troubleshoot. If you're confident that's not going to happen let's proceed with your version.

pymatgen/entries/computed_entries.py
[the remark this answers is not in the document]
shyamd: Yeah, I think if the user provides an `entry_id` then it's up to them to make sure they don't use it twice.

pymatgen/entries/computed_entries.py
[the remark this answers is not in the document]
shyamd: This is sort of linked to the previous one. Hashes should also not override equals. That is if two entries are equals according to `__eq__` they must have the same hash.

pymatgen/entries/__init__.py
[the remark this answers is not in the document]
shyamd: Mutating data in `__init__` should be last resort. I bet if you loadfn/dumpfn entries with this code in init, the energy will keep growing. That kind of bug can be hard to discover; so I suggest just storing data in `__init__` and mutating elsewhere.

pymatgen/entries/computed_entries.py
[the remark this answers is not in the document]
shyamd: So having a discriminative `__hash__` doesn't eliminate calls to `__eq__` it just reduces it down to 1.

shyamd:
I think unless we have a compelling use case for using unsalted hashes, we should continue to use salted hashes. Basically, the problem has to do with inserting objects into dictionaries, but it also is an issue for sets. This was solved in 2012 by salting the builtin hash function: https://bugs.python.org/issue13703

> > Some modifications i've suggested. @mattmcdermott has said he no longer needs the hash consistency and in general, it's not a good idea to override salted hashes for `__hash__`. This was done on purpose early on in python 3 to prevent insecurities in dictionaries and sets.
> 
> shall I remove the `hashlib` dependency and go back to hashing the strings? It depends on whether it's likely that other people encounter the same consistency issues I can't think of any security issues but that doesn't mean they don't exist

CompRhys:
I ended up not adding temperature to the `GibbsComputedStructureEntry` to try make this PR as simple as possible. If it appears that this affects performance then it could be easily added at a later date.

I removed `hashlib` and now just use the default python hash. 

I removed my `per_atom` kwarg due to the potential issues highlighted by @shyamd.

shyamd (approved):

CompRhys:
I checked all the edge cases I could think of with `get_quasi_e_to_hull` and `ComputedEntry` and tweaked the normalizations to eliminate the few that did cause issues.

I added a note to `__eq__` in the `Entry` class highlighting that the `e.as_dict()` comparison isn't robust to floating point changes in the energy which can arise following repeated normalizations. I tweaked `get_quasi_e_to_hull` to avoid this and doubt that it would come up anywhere else.

shyamd:
The lack of robustness in equality using the dict format is a problem. That's why I suggest explicitly breaking out what is being compared and using the appropriate comparator such as `np.allclose`

shyamd:
I'm going to unapprove this PR @CompRhys because `__eq__` should be 100% reliable. I think the only viable way us to explicitly check each parameter introduced by that class and then doing a super `__eq__` rather than relying on the dict equals.

CompRhys:
A potential problem with that is that lots of the `Entry` children have there own dictionaries optional data arguments i.e. `data` and `parameters` in `ComputedEntry` or `chempots` for `GrandPotPDEntry` and I want to avoid going down that rabbit hole.

One thing to note is that even if the dictionary comparison isn't 100% robust for a lot of use cases it is more robust than 
```return self == other``` 
for which is currently the case for `ComputedEntry` or
```
    def __eq__(self, other):
        if isinstance(other, self.__class__):
            return self.as_dict() == other.as_dict()
        else:
            return False

    def __hash__(self):
        return id(self)
```
which was the original `__hash__` and `__eq__` for `PDEntry`.  

As I was told it was "urgent" to fix this issue and given that this is equivalent or better in terms of robustness than what is currently in the code I think that it would probably be better for someone on the core team to take over this PR as this now seems to be more of a core issue where a wider knowledge of how robust is robust enough in `pymatgen` would be needed -- it's certainly not the small issue that was holding back my `PatchedPhaseDiagram` contribution I initially thought i was trying to fix.

**Comment by janosh** [comment]
We have a project that would be much easier to proceed with if this hashing issue and #1974 were resolved. Would be great if, as @CompRhys suggested, @mkhorton or someone else on the core team could have another look at this PR to help get it merged.

**Comment by shyamd** [comment]
Let's leave it as-is then and merge this in as there isn't a "core" team. Pymatgen is very much a community-driven effort. I will say that while this works out now, it's almost a guarantee that anyone that relies on Entries will hit a bug they can't diagnose because of the floating-point issue.

**Comment by CompRhys** [comment]
Thanks! After finishing #1974 I will come back and have a look at whether we can further improve the robustness here and what equality actually means in terms of `Entries` and how to deal with the edge cases thrown up by the optional `parameters`, `attributes`, `data` attributes.

File: pymatgen/analysis/phase_diagram.py [hunk 10]
```diff
@@ -88,9 +78,6 @@ def energy(self) -> float:
         """
         return self._energy
 
-    def __repr__(self):
-        return "PDEntry : {} with energy = {:.4f}".format(self.composition, self.energy)
```
mkhorton, line 92: Why is this `__repr__` removed?

pymatgen/analysis/phase_diagram.py
```diff
@@ -139,21 +115,47 @@ def __init__(self, entry, chempots, name=None):
             name: Optional parameter to name the entry. Defaults to the reduced
                 chemical formula of the original entry.
         """
-        comp = entry.composition
+        super().__init__(
```
mkhorton, line 118: This `__init__` has changed substantially, what connection does this have to the `__hash__` and `__eq__`?

pymatgen/analysis/phase_diagram.py
```diff
@@ -139,21 +115,47 @@ def __init__(self, entry, chempots, name=None):
...
+        Returns:
+            Composition
+        """
+        return Composition(
+            {el: self._composition[el] for el in self._composition.elements
+             if el not in self.chempots}
+        )
+
+    @property
+    def chemical_energy(self):
```
mkhorton, line 144: New property has been added without tests?

pymatgen/analysis/phase_diagram.py
```diff
@@ -208,23 +199,62 @@ class TransformedPDEntry(PDEntry):
     compositions.
     """
 
-    def __init__(self, comp, original_entry):
+    # Tolerance for determining if amount of a composition is positive.
+    amount_tol = 1e-5
+
+    def __init__(self, entry, sp_mapping, name=None):
```
mkhorton, line 205: `__init__` for `TransformedEntry` has also changed substantially, could you clarify what these changes are for?

pymatgen/analysis/phase_diagram.py
```diff
@@ -319,10 +353,9 @@ def from_entries(cls, entries, elements=None):
     @classmethod
     def _kwargs_from_entries(cls, entries, elements):
         if elements is None:
-            elements = set()
-            for entry in entries:
-                elements.update(entry.composition.elements)
-            elements = sorted(list(elements))
+            elements = sorted(list(set().union(
```
mkhorton, line 356: I believe this can just be `sorted({els for e in entries for els in e.composition.elements})`

pymatgen/analysis/phase_diagram.py
```diff
@@ -450,9 +481,15 @@ def stable_entries(self):
         """
         return self._stable_entries
 
-    def get_stable_entries_normed(self, mode="formula_unit"):
+    @lru_cache(1)  # cache in case of repeated calls
+    def get_stable_entries_normed(self, mode="atom"):
```
mkhorton, line 485: In the interests of backwards compatibility (or, principle of least surprise for other users), why is the default changed here?

mkhorton (changes requested):
Thanks for the comments and work everyone. We're eager to get this merged too since this also solves some issues on our side.

@CompRhys I've started a line-by-line review, however there are a lot more changes in this PR than just `__hash__` and `__eq__`, it's not clear to me how this fits together. Can you clarify?

pymatgen/entries/__init__.py
```diff
@@ -114,5 +141,32 @@ def as_dict(self) -> dict:
...
+        if self is other:
+            return True
+
+        if isinstance(other, self.__class__):
+            if not np.allclose(self.energy, other.energy):
+                return False
+            if self.composition != other.composition:
+                return False
+
+            # NOTE this is not performant and should be preceeded
```
mkhorton, line 159: Responding to @shyamd's concern about equality checking and floating point precision, would something like the following work?

```
self_dict = self.as_dict()
other_dict = other.as_dict()

def is_eq(v1, v2):
    """Will use np.allclose for numerical values,
    otherwise will use default __eq__"""
    if isinstance(v1, Number) and isinstance(v2, Number):
        return np.allclose(v1, v2)
    else:
        return v1 == v2

return all([is_eq(other_dict[k], v) for k, v in self_dict.items()])
```

pymatgen/analysis/phase_diagram.py
CompRhys: Now inherited from the `Entry` baseclass using `self.__class__.__name__`

pymatgen/analysis/phase_diagram.py
CompRhys: This change was to introduce the `_composition` with `composition` property pattern for consistency. I went for this pattern after we concluded that having them be immutable wasn't sensible.

pymatgen/analysis/phase_diagram.py
CompRhys: I think this is tested indirectly by tests on the `energy` property which is now derived from this property and the `_energy`.

Should I add a direct test for coverage?

pymatgen/analysis/phase_diagram.py
CompRhys: Previously the `TransformedPDEntry` didn't give an indication as to what the terminal species were. As a result I brought a section of code from the phase diagram inside the entry so that it could be used in more of a standalone manner if ever needed.

pymatgen/analysis/phase_diagram.py
CompRhys: Default probably doesn't need to change as I have changed the kwarg in `get_quasi_e_to_hull`. Using `atom` rather than `formula_unit` was needed there to resolve an edge case where a copy of the entry that was normalised according  to `atom` first didn't give the same energy to hull as the unnormalised copy.

I will change the default back here

pymatgen/analysis/phase_diagram.py
mkhorton: Thank you, a standalone test would be good -- if we have a new, documented, public property, we should make sure to test it returns what we expect explicitly.

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -222,7 +235,7 @@ def test_get_formation_energy(self):
             self.assertAlmostEqual(energy, stable_formation_energies[formula], 7)
 
     def test_all_entries_hulldata(self):
-        self.assertEqual(len(self.pd.all_entries_hulldata), 492)
+        self.assertEqual(len(self.pd.all_entries_hulldata), 490)
```
mkhorton, line 238: I'm assuming this changed because two entries are now equivalent.

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -90,30 +91,37 @@ def test_str(self):
...
-        self.assertEqual(self.transformed_entry.energy_per_atom, 53.0 / 2)
+        self.assertAlmostEqual(self.transformed_entry.energy_per_atom, 53.0 / 2, 11)
 
     def test_get_name(self):
         self.assertEqual(self.transformed_entry.name, "LiFeO2", "Wrong name!")
 
     def test_get_composition(self):
         comp = self.transformed_entry.composition
-        expected_comp = Composition({DummySpecies("Xa"): 1, DummySpecies("Xb"): 1})
+        expected_comp = Composition({DummySpecies("Xf"): 1, DummySpecies("Xg"): 1})
```
mkhorton, line 124: Is there any reason the DummySpecies changed here? (Not a problem, just wanting to make sure I hadn't missed something important)

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -90,30 +91,37 @@ def test_str(self):
...
-            {DummySpecies("Xa"): 1, DummySpecies("Xb"): 1}, entry
-        )
+
+        terminal_compositions = ["LiO", "FeO"]
+        terminal_compositions = [Composition(c) for c in terminal_compositions]
+
+        sp_mapping = OrderedDict()
+        for i, comp in enumerate(terminal_compositions):
+            sp_mapping[comp] = DummySpecies("X" + chr(102 + i))
+
```
mkhorton, line 109: For this code block, why not just:

```
sp_mapping = {
    Composition("LiO"): DummySpecies("Xa"),
    Composition("FeO"): DummySpecies("Xb")
}
```

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -90,30 +91,37 @@ def test_str(self):
     def test_read_csv(self):
         entries = EntrySet.from_csv(str(module_dir / "pdentries_test.csv"))
         self.assertEqual(entries.chemsys, {"Li", "Fe", "O"}, "Wrong elements!")
-        self.assertEqual(len(entries), 492, "Wrong number of entries!")
+        self.assertEqual(len(entries), 490, "Wrong number of entries!")
 
 
 class TransformedPDEntryTest(unittest.TestCase):
     def setUp(self):
+        # NOTE a 1:1 isn't the best test? maybe a tertiary system should be used.
```
mkhorton, line 99: Agree, thanks for adding note

pymatgen/entries/__init__.py
```diff
@@ -89,21 +110,27 @@ def normalize(
...
-            entry.composition /= factor
-            entry._energy /= factor
-            return entry
+
+        entry = copy.deepcopy(self)
+        entry.normalize(mode, inplace=True)
+        return entry
 
     def _normalization_factor(self, mode: str = "formula_unit") -> float:
+        # NOTE here we use composition rather than _composition in order to ensure
```
mkhorton, line 122: Thanks for note

mkhorton (commented):
Have now reviewed the test files too, added a few more comments.

pymatgen/analysis/tests/test_phase_diagram.py
CompRhys: This is how the sp_mapping is setup automatically so I just copied it over here. It starts from 102 which is `f `rather than 97 which is `a` which is why I start from `f`.

pymatgen/analysis/tests/test_phase_diagram.py
CompRhys: just to mimic the dictionary you would get for this task using the live code

pymatgen/analysis/tests/test_phase_diagram.py
CompRhys: shall I update the test to something more complicated?

pymatgen/analysis/tests/test_phase_diagram.py
CompRhys: Yes in the test file we have two sets of duplicated entries. "Li2O,2,0,1,-14.31361175" on both line 8 and line 14 and "Fe,0,1,0,-6.59394176" is repeated on lines 11 and 19.

pymatgen/analysis/tests/test_phase_diagram.py
CompRhys: I have updated this test now to use three terminal compositions

mkhorton: Thanks!

pymatgen/entries/__init__.py
mkhorton: Right, that's a good point -- my hope here was that the check the line above that both `self` and `other` are the same class would mean that both dicts should be guaranteed to have the same keys (technically, the check above uses `isinstance` so would also pass for subclasses, so perhaps this could be tightened).

pymatgen/analysis/phase_diagram.py
CompRhys: test added

pymatgen/entries/__init__.py
CompRhys: Yep I call `isinstance` first so they would have the same keys but they could "physically" correspond to the same material but one `ComputedEntry` might have the optional `parameters` and the second might have `parameters` be `None`. Currently this means they would not be equal due to a metadata attribute. I think that's probably how it should be as it's closer to how other things are but I just wanted to try highlight it explicitly.

mkhorton:
Thanks @CompRhys! If we ever hand out annual PRs-with-most-comments-responded-to award I'll make sure you're on the list :-)

PR #2013, opened 2020-12-10
Changed electrode document to dataclasses

jmmshn:
## Summary

PR to address #2012 

- Changed the `AbstractVoltagePair` and `AbstractElectrode` into `dataclass`s
- Changed the `InsertionVoltagePair` and `InsertionElectrode`  into `dataclass`s
- For `InsertionVoltagePair` and `InsertionElectrode`, moved original entries based init into `from_entires` function
- For `InsertionElectrode`, moved original `as_dict` `from_dict` to `*_legacy`

PR #2016, opened 2020-12-14
Further cleanup of Electrode objects

jmmshn:
## Summary
- Change `ConversionElectrode` to type `dataclass` similar to the change to `InsertionElectrode` in #2013 
- Added `x_charge` and `x_discharge` properties to report the number of working ion ver formula unit
- Standardized how the summary dictionary is created, now the base Electrode class has a `get_summary_dict` function which can be improved upon by the Insertion and Conversion objects.
- Added some more tests of the `summary_dict`

The old `as_dict_summary` methods are still around but maybe should be deprecated?


mkhorton:
Yes, please add a deprecation decorator to `as_dict_summary()`, and we can remove at a later date. Thanks!

jmmshn:
@mkhorton I made some additional (independent) changes to the plotter.

mkhorton:
Can you make `capacity` synonymous with `capacity_grav`? e.g. just add an `or` in line 73, for backwards compatibility (the docstring can be left as-is), I'm not sure if features in the current materials project website rely on the existing behavior

jmmshn:
OK done.  Also maybe we should consider raising the version number up from 0.1.

mkhorton:
I'm not sure the version numbers are useful, we don't have a policy for what they mean or when they're incremented

mkhorton:
Thanks @jmmshn!

PR #2017, opened 2020-12-15
[WIP] add pretty_parity_plot function

rkingsbury:
## Summary

Adds a new plotting function `pretty_parity_plot` to facilitate creating customizable and nicely-formatted plots that compare two quantities. An example is shown below.

* Most common plot customizations are exposed through explicitly documented kwargs
* Automatically plot or hide linear regression line, regression statistics, and 1:1 line
* 'units' are automatically appended to both axes and error metrics (if visible)
* Robust enough to plot data that includes `np.nan` / `np.inf`
* Optional `ax` kwarg allows plotting directly to a subplot axis of another figure

## Example

```
parityplot(x,
                 y,
                 "$\Delta H_f^{R2SCAN}$",
                 "$\Delta H_f^{SCAN}$", 
                 units="eV/atom", 
                 show_regression_line=False,
                 show_diagonal=True,
                 regression_ls='k-'
                 )
```
image

* `scikit-learn` for calculation of error metrics (see TODO)

## TODO (if any)
- [ ] Find an alternative to `sklearn.metrics` to avoid introducing a new dependency (probably will just hand-code the calculations of RMSE, max error, mean absolute error)
- [ ] Implement the `ax` kwarg
- [ ] Determine what kind of tests are needed

### DIFF
--- a/pymatgen/util/plotting.py
+++ b/pymatgen/util/plotting.py
@@ -79,11 +79,11 @@ def pretty_parity_plot(x: Union[np.ndarray, List],
                        xlim: Optional[Tuple] = None,
                        ylim: Optional[Tuple] = None,
                        width: float = 6,
-                       hegith: float = 6,
+                       height: float = 6,
                        show_diagonal: bool = True,
+                       diagonal_ls: str = "k--",
                        show_regression_stats: bool = True,
                        show_regression_line: bool = False,
-                       diagonal_ls: str = "k--",
                        regression_ls: str = "r-",
                        ax: Optional[Axes] = None,
                        ):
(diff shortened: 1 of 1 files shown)


mkhorton:
This looks useful but I'm trying not to merge new matplotlib plotting into pymatgen in favor of plotly-based plots which allow for more interactive data exploration. If you are using matplotlib, the [seaborn](https://seaborn.pydata.org) wrapper has some nice options for this kind of plot, including marginal histograms and 2D histograms, which can help deal with with data overlap issues.

rkingsbury:
> This looks useful but I'm trying not to merge new matplotlib plotting into pymatgen in favor of plotly-based plots which allow for more interactive data exploration. If you are using matplotlib, the [seaborn](https://seaborn.pydata.org) wrapper has some nice options for this kind of plot, including marginal histograms and 2D histograms, which can help deal with with data overlap issues.

OK, it sounds like the method in this PR needs to find a different home then. I do like plotly but personally find it very difficult to generate polished / publication-ready figures with it. I think matplotlib is still superior in that regard and use the existing `pretty_plot` class regularly for that purpose.

mkhorton:
I personally have found no significant difference between them for publication, plotly's default template seems heavily inspired by ggplot2 so can look a bit different, but it has a white-background template if that's preferred. I find their Jupyter extension useful for quick edits of the chart too and being able to save the chart to JSON and make edits after the fact, independently of the code used to generate the chart, has been useful. But not trying to sell anyone on this; use in pymatgen is purely because it allows easier data exploration, and makes it easier to transfer plots to web apps (so they're 1-to-1 equivalent between publication and website).

PR #2018, opened 2020-12-21
Minor change to definition of InsertionVoltagePair

jmmshn:
## Summary

Moved the entry_charge/discharge into the data model.
This allows the from_dict to recover the entries after serialization.


mkhorton:
Thanks @jmmshn!

PR #2022, opened 2020-12-28
fix MoS test filename in test_graph.py

drew-parsons:
test file is MoS2_single.pdf not MOS2_single.pdf

"MOS2_single.pdf" triggers an error: "No such file or directory: 'MOS2_single.pdf'" from `os.remove(test_file)`


mkhorton:
Thanks for the fix @drew-parsons!

Since this is your first contribution, please do fill out [the form here](https://forms.gle/JnisFb38QDR8QTFTA) so that we can [acknowledge you appropriately](https://pymatgen.org/team.html). I know this is only a quick fix, but we want to make sure we acknowledge all contributions :-)

PR #2028, opened 2021-01-07
Quasi-RRHO Thermochemistry Analysis Module

arepstein:
## Summary

Added quasirrho.py to the analysis subpackage to calculate the Quasi-RRHO free energy from a Gaussian or QChem frequency calculation.

* Calculates Grimme's Quasi-RRHO free energy
* Option for also correcting for solvent concentration 

*None

## TODO (if any)

* The rotational symmetry number is required as an input. To calculate this on the fly, will consider updates to PointGroupAnalyzer 

- [ x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
      on your local machine.
      to type check your code.
- [x ] Tests have been added for any new functionality or bug fixes.
- [ x] All existing tests pass.

allowing commits.


pymatgen/analysis/tests/test_quasirrho.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+import os
+import unittest
+
+from pymatgen.analysis.quasirrho import QuasiRRHO
+from pymatgen.io.gaussian import GaussianOutput
+from pymatgen.io.qchem.outputs import QCOutput
+
+test_dir = os.path.join(os.path.dirname(__file__), "..", "..", "..",
```
shyuep, line 15: Instead of using module level test_dir, pls inherit from PymatgenTest for unit tests and then use self.test_dir instead. This will ensure correct behavior. Thanks.

mkhorton:
@arepstein is this PR ready for review to be merged?

arepstein:
I believe so. The only pending update in my mind is automated detection of the rotational symmetry number, but I think it's okay as an input parameter for now.

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,206 @@
...
+__credits__ = "Steven Wheeler, Trevor Seguin"
+
+from typing import Union
+
+import numpy as np
+from pymatgen.io.gaussian import GaussianOutput
+from pymatgen.io.qchem.outputs import QCOutput
+
+# Define useful constants
+kb = 1.380662E-23  # J/K
```
mkhorton, line 27: Are these already defined in pymatgen.util.units?

arepstein: Looks like yes! I was not aware of pymatgen units -- will fix and commit the changes

arepstein: A lot of the units I'm using are slightly different and don't lend themselves well to the subclasses in pymatgen.core.units. For example, converting amu*angs^2 to kg*m^2 for moments of inertia. Would you recommend still using Units subclasses when I can?

mkhorton: I would recommend supplementing the existing units if you don't have the ones you want.

To be specific, I'm talking about the units *constants* in pymatgen.core.units specifically, rather than suggesting you use the classes in that file.

The reason for this recommendation is simply that if we expand our unit functionality later (we've explored a few options here), it's a lot easier for us to spot where units are being used in pymatgen if unit constants are imported from the same place.

htz1992213:
Hey @arepstein , I get to know this PR in today's subgroup meeting. I am just wondering how the functionalities here comparing to those in the Goodvibes package https://github.com/bobbypaton/GoodVibes?

arepstein:
> Hey @arepstein , I get to know this PR in today's subgroup meeting. I am just wondering how the functionalities here comparing to those in the Goodvibes package https://github.com/bobbypaton/GoodVibes?

Thanks for pointing this out, I didn't know about Goodvibes when I put in this PR. This PR should be identical to the Grimme Quasi-RRHO approximation for the entropy that's implemented in GoodVibes, but does not include any other methods  that GoodVibes implements. One big difference I see in terms of implementation into pymatgen infrastructure is that this PR can calculate Quasi-RRHO entropeis for Q-Chem output files, Gaussian output files, or manual input parameters, which is useful for Atomate integration. It would be a good idea to check that this agrees with GoodVibes.

arepstein:
Hi @mkhorton , I don't understand why linting is failing, could you please help? I do use the method _get_avg_mom_inertia later in the module. Thanks!

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,219 @@
...
+
+# Define useful constants
+kb = kb_ev * const.eV  # pymatgen kb from ev/K to J/K
+c = const.speed_of_light * 100  # cm/s
+h = const.h  # Planck's constant J.s
+R = const.R / const.calorie  # Ideal gas constant cal/mol
+
+# Define useful conversion factors
+kcal2hartree = 0.0015936  # kcal/mol to hartree/mol
+
```
rkingsbury, line 38: Is there a way to define this conversion factor using `scipy.constants`? If not, just make sure you carry enough decimal places to not lose precision due to the very different magnitudes of Ha and kcal.

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,219 @@
...
+    iav = np.average(inertia_eigenvals)
+
+    return iav, inertia_eigenvals
+
+
+class QuasiRRHO:
+    """
+    Class to calculate thermochemistry using Grimme's Quasi-RRHO approximation.
+    All outputs are in atomic units.
```
rkingsbury, line 67: Can you be explicit about what atomic units are? For thickheaded engineers like myself, I have to go and google this :)

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,219 @@
...
+    inertia_eigenvals = np.multiply(np.linalg.eig(inertia_tensor)[0], amu_to_kg * 1e-20).tolist()  # amuangs^2 to kg m^2
+
+    iav = np.average(inertia_eigenvals)
+
+    return iav, inertia_eigenvals
+
+
+class QuasiRRHO:
+    """
+    Class to calculate thermochemistry using Grimme's Quasi-RRHO approximation.
```
rkingsbury, line 66: May add the citation to the paper in this docstring so that it's visible in, e.g., Jupyter Notebook and documentation pages?

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,219 @@
...
+        :param output: Requires input of a Gaussian output file,
+                        QChem output file, or dictionary of necessary inputs:
+                        {"mol": Molecule, "mult": spin multiplicity (int),
+                        "frequencies": list of vibrational frequencies [a.u.],
+                        elec_energy": electronic energy [a.u.]}
+        :param sigma_r (int): Rotational symmetry number
+        :param temp (float): Temperature [K]
+        :param press (float): Pressure [Pa]
+        :param conc (float): Solvent concentration [M]
+        :param v0 (float): Cutoff frequency for Quasi-RRHO method [cm^1]
```
rkingsbury, line 84: If possible, please convert to a [google style docstring](https://gist.github.com/redlotus/3bc387c2591e3e908c9b63b97b11d24e) 

Since all of the outputs are stored as class attributes, it would also be especially nice to have a `Attributes` section that describes each output, with units

rkingsbury: Alternatively, you could define python `@property` methods for each of the relevant outputs (e.g., `g_corrected`, h_corrected`, which would let you give each one an explicit docstring. Each property would not do anything other than contain the docstring, e.g.

```
@property
def g_corrected(self):
'''
Corrected free energy in Ha
'''
return self.g_corrected
```

so the only benefit to this structure would be visibility of the outputs to users

rkingsbury (commented):
Thanks for putting this together @arepstein ! I used it recently and I think it's a very useful addition to the molecular DFT infrastructure in pymatgen. I made a few comments; if you can pull in the latest pymatgen and address these hopefully we can get @janosh or another maintainer to review soon; I know this one has been open for a while.

rkingsbury:
See also some small edits I made when I used the code, in a PR against your fork: https://github.com/arepstein/pymatgen/pull/1

rkingsbury:
I will also note for other reviewers that although the GoodVibes package contains similar functionality, it _only_ accepts Gaussian output files, and hence is not useful with our high-throughput Q-Chem infrastructure. So this PR is valuable because it gives us those capabilities.

janosh:
@arepstein Is this still WIP?

arepstein:
@janosh Yes, still WIP, I haven't incorporated Ryan's comments yet. Thanks @rkingsbury for the comments!

rkingsbury:
> @janosh Yes, still WIP, I haven't incorporated Ryan's comments yet. Thanks @rkingsbury for the comments!

No problem! Recall that I also opened a [small PR](https://github.com/arepstein/pymatgen/pull/1/files) against your branch with some other edits, notably exposing `h_corrected` as an attribute.

I had another thought about the call signature which I think would be better than what I initially proposed in that PR. Currently you have

```
def __init__(self, output: GaussianOutput | QCOutput | dict, sigma_r=1, temp=298.15, press=101317, conc=1, v0=100):
```

I think it would be cleaner to make the regular `__init__` method generic by using explicit kwargs, e.g.
```
def __init__(self, mol: Molecule, frequencies: List, energy: float, sigma_r=1, temp=298.15, press=101317, conc=1, v0=100):
```
Where `mol`, `frequencies`, and `energy` are the items currently extracted from the `GaussianOutput` / `QCOutput` / `dict`. Then, you would add two `@classmethod`: `.from_gaussian(self, output: GaussianOutput)` and `.from_qchem(self, output: QCOutput)` to allow instantiating from the respective classes.

This structure allows for 1) use of the module with output from arbitrary codes (i.e., any molecule and frequency combo) and 2) leaves the door open for additional `.from_xxxx` methods to cleanly expand support to new codes in the future.

samblau:
Funny timing - I now would very much like to use this capability! One thing I want to note is that it would be very nice if, in addition to an actual Q-Chem output file, we could pass it a summary document that comes out of the new Molecules Emmet infrastructure. Happy to provide an example JSON.

arepstein:
> 
Hey @samblau , I'm getting to this now. Could you please provide an example JSON? Right now it can take a manual dictionary input, so it should be easy to incorporate.

arepstein:
Hey @rkingsbury and @janosh, I've incorporated Ryan's edits and recommendations. While making these edits it became clear that a Class might not be the best choice for QuasiRRHO. As it stands now, it might be better as a simple function. Ideally, I think it could be nice to have a `Thermochemistry` class that is a `Molecule` plus some extra information like frequencies. We could then have different treatments of thermochemistry, like GoodVibes implements, but in a way works well with our ecosystem. 

That being said, hopefully QuasiRRHO is still useful for now and can be a starting point for later changes if desired.

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,262 @@
...
+
+from pymatgen.core.units import kb as kb_ev
+
+if TYPE_CHECKING:
+    from pymatgen.core import Molecule
+    from pymatgen.io.gaussian import GaussianOutput
+    from pymatgen.io.qchem.outputs import QCOutput
+
+# Define useful constants
+kb = kb_ev * const.eV  # Pymatgen kb [J/k]
```
rkingsbury, line 35: [J/k] -> [J/K]

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,262 @@
...
+# constant [Ha/K]
+R_volume = const.R / const.atm * 1000  # Ideal gas constant [L.atm.K^-1.mol^-1]
+
+# Define useful conversion factors
+amu_to_kg = const.value("atomic mass unit-kilogram relationship")  # AMU to kg
+# kcal2hartree = 0.0015936  # kcal/mol to hartree/mol
+kcal2hartree = 1000 * const.calorie / const.value("Hartree energy") / const.Avogadro
+
+
+def get_avg_mom_inertia(mol):
```
rkingsbury, line 49: Not required for this PR, but I wonder if this should be made a @property of the `Molecule` class? Any strong opinions one way or another?

pymatgen/analysis/quasirrho.py
rkingsbury: Thanks for adding to the docstring! Since you opened this PR, we have also added a nice system for citations in `pymatgen` based on [duecredit](https://github.com/duecredit/duecredit). Would you mind adding a @due.dcite decorator to this class? ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

pymatgen/analysis/quasirrho.py
```diff
@@ -0,0 +1,262 @@
...
+            srotor = 1 / 2 + np.log(np.sqrt(8 * np.pi**3 * mu_prime * kb * self.temp / h**2))
+            weight = 1 / (1 + (self.v0 / vt) ** 4)
+            sv_quasiRRHO += weight * sv_temp + (1 - weight) * srotor
+
+        sv_quasiRRHO *= R
+        sv *= R
+        ev *= R
+        etot = (et + er + ev) * kcal2hartree / 1000
+        self.h_corrected = etot + R * self.temp * kcal2hartree / 1000
+        molarity_corr = R_ha * self.temp * np.log(R_volume * self.temp * self.conc)
```
rkingsbury, line 255: It looks like the molarity correction is `RT log (RTC)`. Shouldn't it just be `RT log(C`? I may be ignorant, but please double check / confirm.

rkingsbury: Also, this formulation implicitly assumes a standard state of 1 M. It might be good to state that in the docstring in case someone is using a calc with a different standard state

rkingsbury (approved):
Thanks @arepstein and nice work on this. Should be a good addition to pymatgen.

pymatgen/analysis/quasirrho.py
arepstein: I was following Wheeler's script for the concentration correction. I believe it comes from a thermodynamic cycle for cluster continuum models that leads to an extra `RT log(R_volume T)`. It's equations 7 and 8 in the [Ho and Coote pKa paper](https://link.springer.com/article/10.1007/s00214-009-0667-0). 
Thanks for the point about the 1M assumption. I will add that.

rkingsbury: Ah, thanks for clarifying. So looking at that section of the paper:
image

I'm pretty sure the extra `-RT log [R~T]` term is there to adjust the reference states (the paragraph above and the different units of R and R~ are the clues). The solvation energy assumes 1 mol/L in both gas and solution, so you need a correction for that difference. The numerical value of R~T is 24.43 atm / (mol/L), which is the same value one might use to make a standard state correction.

So I think it should be `RT ln C`, and I'm glad we have the standard state in the docstring to hopefully call attention to the need to be careful interpreting the number. Does that make sense?

pymatgen/analysis/quasirrho.py
```diff
@@ -83,7 +84,7 @@ class QuasiRRHO:
     Attributes:
         temp (float): Temperature [K]
         press (float): Pressure [Pa]
-        conc (float): Solvent concentration [M]
+        conc (float): Solvent concentration. Assumes 1M unless specified [M]
```
rkingsbury, line 87: I would add "Concentration correction is reference to a 1 M standard state."

pymatgen/analysis/quasirrho.py
arepstein: Your explanation of the source of the extra term makes sense to me, but I don't understand why we shouldn't include the standard state correction when providing a concentration-corrected free energy in solvent. If RTln(concentration) uses mol/L, shouldn't we also correct for the 1 atm  standard state used in the DFT code?

rkingsbury: That's a fair point. To me what makes it tricky is that you only need to do the standard state correction when you are writing a _reaction energy_ between species with different reference states, but this module just computes a single correction for a single species. (in other words, the reason for the correction is to make sure all species use the same reference state). We don't necessarily know that users of this class are going to take this free energy and combine it with a gas phase energy.

The other small issue is that I think not every DFT code necessarily uses the same reference state. But I think both Q-Chem and Gaussian use 1 atm for gas phase calcs, and as long as we document what we're doing I'm not too worried about that.

Personally I think it will be more transparent and less confusing to users to just do RT ln C and let them adjust as needed if they have to change reference states. But again, as long as we document what's going on it could be OK to build it in.

If you want to build it in, I suggest writing it as `( RT log ( C) - RT log (1/24.55) )` and adding a comment line to make explicitly clear that you're dividing C by the molarity of an ideal gas at 1 atm.

Do others like @samblau or @espottesmith or @materialsproject/second-foundation have thoughts?

arepstein: I definitely agree that we should be more specific in the documentation, and I'll put that in. Maybe with a reference to describe it in detail if I can find a really good one. I also think of the free energy in implicit solvent as already containing an assumed change in free energy from vacuum to solution, and would thus include the standard state correction when using implicit solvent. Is that correct?

I would also appreciate input from everyone you tagged! 

I think another option would be to just entirely remove the concentration correction for now. It is somewhat just a legacy from converting a script that was useful to me into Pymatgen.

rkingsbury: > I definitely agree that we should be more specific in the documentation, and I'll put that in. Maybe with a reference to describe it in detail if I can find a really good one. I also think of the free energy in implicit solvent as already containing an assumed change in free energy from vacuum to solution, and would thus include the standard state correction when using implicit solvent. Is that correct?

I just think of the free energy of an isolated molecule as "DFT energy (~ 0K enthalpy) plus T-dependent enthalpy and entropy".

My understanding is that implicit solvent models adjust the free energy for _interactions between the solute and the medium_ but do not account for changes in standard state (vac -> solution). Unless we're talking about _formation free energy_ or a _reaction energy_, the free energy of an isolated molecule begs the question "relative to what"? And you don't have to introduce a standard/reference state until you answer that question. I know some solvent models output a solvation (reaction) energy, but I believe that implicitly assumes the same reference state for both vacuum and solvent. There's nothing inherently wrong with that but it's often not what the end user needs for comparison to experiment.

> 
> I think another option would be to just entirely remove the concentration correction for now. It is somewhat just a legacy from converting a script that was useful to me into Pymatgen.

Yeah, the more I think about it, that might be the cleanest thing to do here. This module is for calculating QRRHO thermochemistry, which (unless I'm mistaken) does not _specifically_ have anything to do with an implicit solvent model and does not make any assumptions about the concentration; it's just a way of dampening low-frequency contributions to rot and vib entropy. So in the same way that standard thermochemistry output in Gaussian/Q-Chem doesn't make any adjustments for concentration, maybe this shouldn't either.

arepstein: I'm still not entirely convinced, mainly because implicit solvent models are made to capture solvation free energies, but I definitely understand what you're saying. I'd love to have this conversation offline more too, given how fundamentally important it is to get these corrections correct, but I agree that that can be beyond the scope of this PR. I'll remove it for now and maybe if Pymatgen wants to start recommending free energy treatments we can implement more general thermochemistry!

rkingsbury: Happy to discuss offline!

rkingsbury:
I think this is ready to merge, right @arepstein ? @janosh can you remove the 'stale' label?

arepstein:
Yes, I believe ready to merge!

janosh (changes requested):
There's been a recent change of the `pymatgen` test folder structure. The file `test_files/molecules/co2.log` now needs to in `tests/files/...`. Would also be good to `gzip` this file.

janosh (approved):
Thanks for this PR @arepstein! 👍

PR #2031, opened 2021-01-13
Optimizing pymatgen start-up time

mkhorton:
## Summary

This PR will be to explore strategies for optimizing pymatgen's start-up and import times.

Current root-level import is ~ 1 second (specifically, 1,100,534 µs as profiled by `python -X importtime`). Reducing this will also have consequences for speed of test suite.

Initial investigations suggest this could be sped up by at least 2x.


mkhorton:
This should speed up import by 2x.

Removing all top-level imports from pymatgen would be the more forward-looking approach, and would maybe add an extra 10x. This would break backwards compatibility for anyone doing a root import (e.g. `from pymatgen import Structure` would have to become `from pymatgen.core.structure import Structure`). To leave the door open for this, I've made sure all tests etc. are using the canonical import.

I tried lazy importing these top level imports, but the user experience wasn't there.

PR #2033, opened 2021-01-15
Parse OUTCAR core potentials in fixed width format

utf:
## Summary

The electrostatic core potentials in the OUTCAR file are written based on a fix width format. The potentials are numbered from 1 to natoms, and sometimes the fixed width means there is no space between the index and the potential. For example:

```
 average (electrostatic) potential at core
  the test charge radii are     1.0894  0.9698
  (the norm of the test charge is              1.0000)
       1-101.5055       2-101.5055       3-101.5055       4-101.5055       5 -85.7708
       6 -85.7708       7 -85.7708       8 -85.7708
```

This PR changes the parsing of the potentials to be based on a fix width format (rather than `str.split()`). I've added a new test.


shyuep:
Thanks.

PR #2034, opened 2021-01-15
Wavefunctions from Lobster

JaGeo:
## Summary

This is a implementation of the new class Wavefunction for the Lobster module that can read in Lobster wavefunctions and transfer them to a VolumetricData object (This allows plotting with VESTA and compability with other VASP files). It can also calculate densities based on the Lobster output.

Thanks to @rnels12 for helpful comments on the Lobster implementation.


JaGeo:
From my side, this code would now be ready to merge. Let me know, if there is anything else to do.

mkhorton:
Thanks @JaGeo, this looks great!

I did have a question on the `set_volumetric_data`, wasn't sure what the `runner` variable and the following block was doing.

JaGeo:
I need to filter out some values from the Lobster output (at x=1.0, y=1.0 or z=1.0). These values are written out twice if I use a box output of the wavefunction between the fractional coordinates 0.00 0.00 0.00 and 1.00 1.00 1.00. As far as I understood the volumeticdata object, they should not be included.

mkhorton:
Yes, that makes sense, thanks. And yes, VolumetricData wouldn't want values for both.

PR #2039, opened 2021-01-21
[WIP] add kwarg to pass in a plt ax to plot_brillouin

janosh:
## Summary

Adds the ability to plot Brillouin zones into a `plt.subplot`

## TODO (if any)

- Does this need a test?

## Checklist

      on your local machine.
      to type check your code.

allowing commits.

### DIFF
--- a/pymatgen/electronic_structure/plotter.py
+++ b/pymatgen/electronic_structure/plotter.py
@@ -929,7 +929,7 @@ class BSPlotter:
             plt.legend(handles=handles)
         return plt
 
-    def plot_brillouin(self):
+    def plot_brillouin(self, ax=None):
         """
         plot the Brillouin zone
         """
@@ -949,7 +949,7 @@ class BSPlotter:
                 ]
             )
 
-        plot_brillouin_zone(self._bs[0].lattice_rec, lines=lines, labels=labels)
+        plot_brillouin_zone(self._bs[0].lattice_rec, lines=lines, labels=labels, ax=ax)
 
 
 class BSPlotterProjected(BSPlotter):
--- a/pymatgen/phonon/plotter.py
+++ b/pymatgen/phonon/plotter.py
@@ -529,7 +529,7 @@ class PhononBSPlotter:
 
         return plt
 
-    def plot_brillouin(self):
+    def plot_brillouin(self, ax=None):
         """
         plot the Brillouin zone
         """
@@ -549,7 +549,7 @@ class PhononBSPlotter:
                 ]
             )
 
-        plot_brillouin_zone(self._bs.lattice_rec, lines=lines, labels=labels)
+        plot_brillouin_zone(self._bs.lattice_rec, lines=lines, labels=labels, ax=ax)
 
 
 class ThermoPlotter:


janosh:
I noticed that `get_ax3d_fig_plt` still prevents Brillouin zones from being plotted into a subplot because matplotlib appears unable to find the current figure and instead creates a new one:

```py
def get_ax3d_fig_plt(ax=None, **kwargs):
    """
    Helper function used in plot functions supporting an optional Axes3D
    argument. If ax is None, we build the `matplotlib` figure and create the
    Axes3D else we return the current active figure.

    Args:
        kwargs: keyword arguments are passed to plt.figure if ax is not None.

    Returns:
        ax: :class:`Axes` object
        figure: matplotlib figure
        plt: matplotlib pyplot module.
    """
    from mpl_toolkits.mplot3d import axes3d
    if ax is None:
        fig = plt.figure(**kwargs)
        ax = axes3d.Axes3D(fig)
    else:
        fig = plt.gcf()
        # preventing the call to plt.gcf() avoids creating new figures
        # fig = None
```

```py
def gcf():
    """
    Get the current figure.

    If no current figure exists, a new one is created using
    `~.pyplot.figure()`.
    """
    figManager = _pylab_helpers.Gcf.get_active()
    if figManager is not None:
        return figManager.canvas.figure
    else:
        return figure()
```

Currently: producing lots of empty figures

empty-figs


Goal: subplots

subplots

If the user passes in an `plt.axis` instance, presumably, they don't need `BSPlotter.plot_brillouin()` to return the current `fig`. Would it be alright if we get rid of the call to `get_ax3d_fig_plt` altogether if the user provided their own axis?

@mkhorton Pinging you as this is related to https://github.com/materialsproject/api/issues/181#issuecomment-763514303.

janosh:
@shyuep What's your take on this?

utf:
It isn't clear to me why `gcf` is creating a new figure. Is `plt.show()` being called before the subplots are added?

janosh:
> It isn't clear to me why gcf is creating a new figure.

@utf That confuses me as well.

> Is plt.show() being called before the subplots are added?

I looked for that as well. Not in my code and not (as far as I can tell) by `pymatgen` either. `BSPlotter` _does_ have a `show` method (with an unused argument `smooth_tol` btw):

```py
    def show(self, zero_to_efermi=True, ylim=None, smooth=False, smooth_tol=None):
        """
        Show the plot using matplotlib.

        Args:
            zero_to_efermi: Automatically subtract off the Fermi energy from
                the eigenvalues and plot (E-Ef).
            ylim: Specify the y-axis (energy) limits; by default None let
                the code choose. It is vbm-4 and cbm+4 if insulator
                efermi-10 and efermi+10 if metal
            smooth: interpolates the bands by a spline cubic
            smooth_tol (float) : tolerance for fitting spline to band data.
                Default is None such that no tolerance will be used.
        """
        plt = self.get_plot(zero_to_efermi, ylim, smooth)
        plt.show()
```

but that's not being called by any of the other methods.

fraricci:
Hello @janosh,
just to understand the problem you're point out.
I managed to create subplots of brillouin zones with not change in the code already there.
Here my simple approach:
kpath = HighSymmKpath(structure)
plt.subplot(121,projection='3d')
plot_brillouin_zone_from_kpath(kpath,plt.gca())
plt.subplot(121,projection='3d')
plot_brillouin_zone_from_kpath(kpath,plt.gca())

here the result:
image

janosh:
@fraricci You're not using the code I'm suggesting to modify. I'm aware that `plot_brillouin_zone_from_kpath` already has an `ax` kwarg. But the most user-facing method `BSPlotter.plot_brillouin()` doesn't, hence my PR.

Issue #1985, opened 2020-10-20
PhaseDiagram.get_equilibrium_reaction_energy no longer works for ComputedEntry


rkingsbury:
**Describe the bug**
`PhaseDiagram.get_equlibrium_reaction_energy()` fails when called on a stable entry from the phase diagram, and raises

```ValueError: None is unstable, the equilibrium reaction energy is available only for stable entries```

**To Reproduce**
```
with MPRester() as m:
    entries = m.get_entries_in_chemsys("Zn-S")

pd = PhaseDiagram(entries)
pd.get_equilibrium_reaction_energy([e for e in pd.stable_entries][0])
```

**Desktop (please complete the following information):**
 - OS: Mac
 - Version 2020.10.9

**Additional context**
The problem is caused by `entry.normalize(inplace=False)`, here:
https://github.com/materialsproject/pymatgen/blob/055805bc5e58a9951add78c441f5cb275da60679/pymatgen/analysis/phase_diagram.py#L665

`self.get_stable_entries_normed()` is populated by calling `entry.normalize(inplace=False)`, which creates copies of each entry. `entry.normalize(inplace=False)` creates _another_ copy of the entry that is not in the list.

I am working on a fix.

**Comment by rkingsbury** [comment]
@shyamd  feel free to add anything I missed

**Comment by rkingsbury** [comment]
I think this might actually be a bug in `.normalize`. The unittest we have passes, and it builds a `PhaseDiagram` from very simple `PDEntry` - they just have an energy and a composition, and
```
[e.normalize(inplace=False) for e in pd.stable_entries][0] in [e.normalize(inplace=False) for e in pd.stable_entries]
```
returns `True`. So somehow, using `ComputedEntry` instead of `PDEntry` causes this to fail.

**Comment by mkhorton** [comment]
The behavior of `ComputedEntry` did change recently, see https://github.com/materialsproject/pymatgen/commit/513eddb3fe32a43270df1f51cd58b93f9cc1aa9b

**Comment by rkingsbury** [comment]
I think this might be a subtle memory referencing thing. Using `ComputedEntry` (not `PDEntry`), I can do

```
>>> with MPRester() as m:
         entries = m.get_entries_in_chemsys("Zn-S")

>>> pd = PhaseDiagram(entries)
>>> [e.normalize(inplace=False) for e in pd.stable_entries][0] in pd.get_stable_entries_normed()
False
[...]

PR #2040, opened 2021-01-24
fix: #1985 and add test to stop computed entry being used for get_quasi_e_to_hull

CompRhys:
@rkingsbury 

As the proper fix in #2014 seems like it might still take a while to get merged this is the temp fix I should have done for #1985 when I first saw it. It avoids the bug that I introduced by not (at that point) being aware that `ComputedEntry` was supposed to be interoperable with `PDEntry` and adds a catch to stop another bug in `get_quasi_e_to_hull` that would be caused by the same issues with `__eq__` and `__hash__` for `ComputedEntry` that caused this bug.

### DIFF
--- a/pymatgen/analysis/phase_diagram.py
+++ b/pymatgen/analysis/phase_diagram.py
@@ -25,6 +25,7 @@ from pymatgen.analysis.reaction_calculator import Reaction, ReactionError
 from pymatgen.core.composition import Composition
 from pymatgen.core.periodic_table import DummySpecies, Element, get_el_sp
 from pymatgen.entries import Entry
+from pymatgen.entries.computed_entries import ComputedEntry
 from pymatgen.util.coord import Simplex, in_coord_list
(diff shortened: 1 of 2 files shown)


rkingsbury:
Thanks @CompRhys ! I appreciate you doing this and I know that #2014 has become more involved than originally expected. This fixes #1985 at least insofar as it affects my use cases and looks like a very reasonable stopgap solution. @mkhorton , do you have any concerns about merging this as a temporary fix?

mkhorton:
@rkingsbury I plan to merge both asap, stay tuned... I think #2014 is in good shape, though it did indeed end up longer than originally intended :-)

mkhorton:
Given that #2014 is merged, shall we close this?

rkingsbury:
OK with me @mkhorton !

CompRhys:
not needed due to #2014!

PR #2042, opened 2021-01-27
[WIP] Add PatchedPhaseDiagram to accelerate handling of large approximately disjoint data sets

CompRhys:
## Summary

This PR takes over from #1974. 

Introduce a new `PatchedPhaseDiagram` based class to handle large disjoint data sets. Calculation of the convex hull of a data set scales with the dimensions and amount of data. Breaking up disjoint datasets into smaller patches reduces the complexity in both these regards and would allow for parallel computation of different patches.

The code has been adapted such that as much as possible can be inherited from `PhaseDiagram`. However, not all the functionality of `PhaseDiagram` will be available in `PatchedPhaseDiagram` as some of these functions can't be calculated without access to the facets. My primary aim is to be able to extract `get_decomp_and_quasi_e_to_hull` for large high dimensional phase diagrams in a timely manner to allow for more robust testing workflows of ML models extending on ideas in doi.org/10.1038/s41524-020-00362-y . 

Where functionality is not able to be replicated I have attempted to modularise the code such that if approximate methods can be devised then only smaller modular methods need to be re-written for functionality to be inherited correctly. e.g. `_get_simplex_intersections` and `_get_all_facets_and_simplexes`

More of the code has been converted into google docstring format.

#### Speed

The code has been tested for speed using the following `ppd_test.py` script: 

```python

import pymatgen
from pymatgen.analysis.phase_diagram import (
    PhaseDiagram,
    PatchedPhaseDiagram,
)

with pymatgen.MPRester("API_KEY") as m:
    entries = m.get_entries_in_chemsys("Y-Mn-O-Li-Cl-C-Na-Fe-W-Bi-Ti-Be-S")

def foo():
    pd = PatchedPhaseDiagram(entries, workers=0)

def bar():
    pd = PatchedPhaseDiagram(entries, workers=1)

def flob():
    pd = PatchedPhaseDiagram(entries, workers=3)

def fum():
    pd = PatchedPhaseDiagram(entries, workers=-1)  # i.e. 4 workers

def fifo():
    pd = PhaseDiagram(entries)
```

The options were tested on a laptop with 4 threads

```
% python -mtimeit -s'import ppd_test as test' 'test.foo()'
1 loop, best of 5: 14.6 sec per loop
```
```
% python -mtimeit -s'import ppd_test as test' 'test.bar()'
1 loop, best of 5: 15.3 sec per loop
```
```
% python -mtimeit -s'import ppd_test as test' 'test.flob()'
1 loop, best of 5: 10 sec per loop
```
```
% python -mtimeit -s'import ppd_test as test' 'test.fum()'
1 loop, best of 5: 10.2 sec per loop
```

I gave up testing `fifo` as it took too long.

On a dataset of 56788 entries drawn from MP it took ~1hr40m to construct a ppd on my laptop. CPU: Intel i5-5200U (4) @ 2.700GHz

With 20 cores on the cds3 cluster it takes ~5m. CPU: Intel(R) Xeon(R) Gold 6142 CPU @ 2.60GHz 

## Checklist


CompRhys:
> @CompRhys I was just working on filtering large chemical spaces and tested out the PatchedPhaseDiagram class.
> 
> I ran it on all entries in the "Y-Mn-O-Li-Cl-C-Na-Fe-W-Bi-Ti-Be-S" chemical system (as acquired from MP), which has 13 elements and about 3,595 entries. The PatchedPhaseDiagram code performed equivalently fast to the `expand_pd()` method I linked to a while back (about 7 seconds) on a single core, but actually took about 20x longer (~155 seconds) with multiprocessing enabled. I think that the current setup of doing a map to each chemical system and supplying `all_entries` as an argument each time is causing a great deal of overhead, as it tries to copy the list of all entries into the memory of each process.
> 
> Wasn't sure if you already had planned on addressing this, but to help speed this up, you might try to do some batching so that there are only a few processes created (maybe total # processes = # cores), making sure only to enable multiprocessing when each process takes a substantial amount of compute time (maybe 10+ seconds?)... this system seemed to work well for me in similar multiprocessing applications :)

@mattmcdermott I have changed the multiprocessing a bit now do you still encounter this slow-down? I made the `__get_pd_for_space` function into a method which means it might not need to t it shouldn't need to copy `all_entries` each time but I am not sure? I didn't manage to replicate the slow down with the older setup

pymatgen/analysis/phase_diagram.py
```diff
@@ -731,42 +779,34 @@ def get_decomp_and_quasi_e_to_hull(
...
-                    )
-                )
-                + list(self.el_refs.values())
-            )  # terminal points
+            reduced_space = list(set.intersection(
+                set(competing_entries),         # same chemical space
+                set(self.qhull_entries),        # negative E_f
+                set(self.unstable_entries),     # not already on hull
+            )) + list(self.el_refs.values())    # terminal points
+            inner_hull = PhaseDiagram(reduced_space)
```
CompRhys, line 787: maybe this should be a ppd? this could be problematic

CompRhys:
@mkhorton I've been using this code for a while now and I think this is probably now stable enough to think about reviewing/merging. I refactored get `phase_separation_energy` a bit on this branch and added a fix to an edge case I found today, I fixed it here as this is the code I have been using but can make a separate PR for the fix in prefered

CompRhys:
a small note is that I did some testing when swapping `tqdm` for the `pymatgen` `PBar` and am pretty sure that `PBarSafe` isn't safe or at least not how I was using `tqdm` so could be worth making `tqdm` a requirement?

CompRhys:
Actually tried applying to whole of MP today and hit some issues speed issues that I am not sure can be fixed to make this generally useful, will close for the time being and maybe open again if I can speed it up

mkhorton:
Could you clarify what the speed issues were, and what the unsafe PBar behavior was? I'm not against merging this if it's useful for intermediate cases, even if it can't handle very large data sets.

CompRhys:
When I purposefully removed tqdm from my environment PBar should default for PBarSafe but PBarsafe threw an error. I didn't really explore further than that as PBar worked fine as a direct substitution for tqdm.

I can get the `ppd` for the whole of MP in <40minutes using 20 workers. I don't know how it compares to the equivalent stages in your database builds for getting the pds for the get e_above_hull calcs. The main bottleneck is selecting the entries to build the relevant patches which scales badly to larger entry sets due to it being done in serial (calls to elements and frozenset is superset in profile are due to this step).

The other speed issues of concern are a mix-match quite a lot due to me not knowing what was slow in pymatgen and trying to write pretty code that kept walking into speed-traps. These effect calls to `get_e_above_hull` and `get_phase_separation_energy`. Mostly the problem was that I was using a lot of calls to fractional_composition which ends up being very slow and running checks to avoid computation that are slower than the computation being avoided. Tweaking this I think i've got a 4x speed-up on `get_phase_separation_energy` but still haven't worked out how to make use of multi-processing with the ppd to evaluate lots of these in parallel. The issue is that the object for the whole of MP ppd can be v large (esp if using ComputedStructureEntries) so don't necessarily want to make too many copies.

[ppd-profile.txt](https://github.com/materialsproject/pymatgen/files/6400552/ppd-profile.txt)

PR #2043, opened 2021-01-27
Read Eigenval for Vasp non spin polarized Calculations

mdforti:
## Summary

* add a ine to include the case where the eigenval corresponds to non spin polarized

### DIFF
--- a/pymatgen/__init__.py
+++ b/pymatgen/__init__.py
@@ -7,11 +7,6 @@ Pymatgen (Python Materials Genomics) is a robust, open-source Python library
 for materials analysis. This is the root package.
 """
 
-#MDF-COMMENT my version for test
-
-import sys
-sys.path.insert(0, '/home/mariano/git/pymatgen/')
-
 import os
 import warnings
 from fnmatch import fnmatch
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @mdforti! Do you have an example non-spin-polarized EIGENVAL file to hand too? We can include it as a test case

arosen93:
Could you clarify what issue the EIGENVAL reader has with spin unpolarized calculations? To the best of my knowledge, it has always worked for such systems.

mdforti:
Thank you for taking this. It has been a long time since I was working on it. The version of vasp I was working was older and the EIGENVAL  looks exactly by the EIGENVAL_old  provided by @arosen93 , where the occupations are missing. In that case, also for spin polarized there are no occupations listed, so the reading would not be correct either.

janosh:
From @arosen93's comment, sounds like even with this merged, spin-polarized calcs (3 cols) run with pre 5.4.x Vasp would still be parsed incorrectly as spin-unpolarized by `pymatgen`? I assume the unspoken conclusion here was `pymatgen` EIGENVAL reader doesn't support parsing pre 5.4.x Vasp?

janosh:
Closing this as abandoned.

PR #2044, opened 2021-01-28
method to estimate NBANDS + tests

rwoodsrobinson:
## Summary

* created method to estimate NBANDS for VASP input set
* simplified `nelect` property
* added tests

## Checklist

      on your local machine.
      to type check your code.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -632,20 +632,12 @@ class DictSet(VaspInputSet):
         """
         Gets the default number of electrons for a given structure.
         """
-        # if structure is not sorted this can cause problems, so must take
-        # care to remove redundant symbols when counting electrons
-        site_symbols = list(set(self.poscar.site_symbols))
-        structure = self.structure
-        nelect = 0.0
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @rwoodsrobinson !

utf:
If possible, I'd also take into account whether spin-orbit coupling is included in the calculation (`LSORBIT = True`), as that will ~double NBANDS.

rwoodsrobinson:
Thanks for mentioning @utf, I'll add it. Know of any other cases where VASP changes the # of NBANDS, or how they are estimated as a function of # of cores?

mkhorton:
@utf it can definitely change drastically in some cases, e.g. imagine a small cell with few bands on something like KNL (64 cores) with default parallelization (I think; I would have to check specific circumstances, but I've definitely seen it change significantly, perhaps this was with OMP threads too)

utf:
Apologies, I meant with a specific INCAR setting rather than the number of cores.

mkhorton:
Good point about SOC btw, VASP manual doesn't mention that but makes sense

PR #2055, opened 2021-02-04
Additions to Q-Chem IO

espottesmith:
Add ability to handle more diverse job types (including potential energy surface scans and transition-state optimizations) to the Q-Chem IO module.

One other small change was moving the metal_edge_extender function from pymatgen.analysis.fragmenter to pymatgen.analysis.local_env. As this function determines bonding and doesn't directly have to do with creating molecular fragments, it belongs with the NearNeighbor classes.

## TODO (if any)

None

## Checklist

in the pull request title.

- [X ] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [ X] Tests have been added for any new functionality or bug fixes.


shyuep:
Pls just run black through your code first. Also, can you try to add some type annotations? That would be helpful.

espottesmith:
I'll work on the linting problems.

Are there any specific areas that you think need type hints? Mostly I've just modified existing functions or made close copies (such as the new Sets classes in Q-Chem IO). I didn't add a lot of new functions in places that already had type hints.

shyuep:
Well, type hints are not compulsory, but if you can add them wherever in the qchem package, that would be helpful.

shyuep:
Hmm... I am not sure what you need to do with black. Simply running `black <filename>` should fix all the black related lining errors. Then you can focus on the rest of the stuff. That was what I did for the entire pymatgen code base in ~10 seconds.

espottesmith:
Yeah, I have this problem pretty consistently where running black doesn't completely resolve all issues, so I have to go back and fix by hand.

pymatgen/io/qchem/inputs.py
```diff
@@ -458,7 +458,7 @@ def find_sections(string: str) -> List:
         return sections
 
     @staticmethod
-    def read_molecule(string: str) -> Molecule:
+    def read_molecule(string: str) -> Union[Molecule, str]:
```
shyuep, line 461: This cannot be the correct type. You actually return a Molecule, not a string.

espottesmith: It depends. It's possible for the molecule section to just be "read", meaning you're reading in the molecule from a previous calculation.

shyuep: Then it should be Optional[Molecule]. Not str. Reading string to return a string is bad.

espottesmith: I suppose we could do Optional, but "read" is really the correct value. It doesn't just indicate the lack of a molecule, but that the molecule is being read in from somewhere else.

mkhorton: Following that through, the type hint for `QCMolecule` `__init__` would be incorrect too since that says it only accepts `Molecule` not `"read"`.

Btw, if there's only one possible string, the correct hint is `Literal["read"]`.

mkhorton: I do also concur that `None` is a more natural value than `"read"`, but I'm not familiar with QChem, so perhaps `"read"` makes more sense to QChem users?

mkhorton: Yup, that makes sense, given its usage. Just make sure the `QCMolecule.__init__` has a consistent type hint, since currently it states it can only accept `Molecule`, when it seems like it could be passed `"read"` instead.

espottesmith: Ooh, unfortunate. Looks like Literal was only added in Python 3.8 ([see here](https://www.python.org/dev/peps/pep-0586/)).

Since that's not an option, are we okay with "str" (there is a check on that in QCInput.__init__ to make sure that, if it's a string, it's "read")? Or should I just go with Optional at that point?

Also, not sure what you're referring to by QCMolecule, @mkhorton. I don't think that's a class in pymatgen?

mkhorton: Meant `QCInput`, apologies.

There's a backport with typing-extensions, we can add it -- I think it's important enough to do so.

mkhorton: Actually, @shyuep already added typing-extensions two days ago, so we're good. I just added to the setup.py too and made it only install for older versions of Python.

espottesmith: I'll add that to requirements.txt and make the change.

mkhorton:
Thanks for all the work on this PR @espottesmith !

Issue #2053, opened 2021-02-03
Bug: GibbsComputedStructureEntry returning incorrect energies, running slowly


mattmcdermott:
**Describe the bug**
`GibbsComputedStructureEntry` is now returning incorrect formation energies and running very slowly, I think due to refactoring in #2014 (tagging @CompRhys for reference, but I should be able to address).

**To Reproduce**
```python
from pymatgen.analysis.phase_diagram import PhaseDiagram
from pymatgen import MPRester
from pymatgen.entries.computed_entries import GibbsComputedStructureEntry

with MPRester() as mpr:
    all_entries = mpr.get_entries_in_chemsys("C-O", inc_structure="final")

pd_mp = PhaseDiagram(all_entries)

temp = 900
gibbs_entries = GibbsComputedStructureEntry.from_pd(pd_mp,temp)
pd_gibbs = PhaseDiagram(gibbs_entries)

for e in sorted(list(pd_gibbs.stable_entries),key=lambda x: x.composition.reduced_formula):
    print(e.composition.reduced_formula, pd_gibbs.get_form_energy_per_atom(e))
```

**Expected behavior**
The formation energy of CO2 should be around -1.367 eV/atom at T=900 K, according to NIST-JANAF data. Instead, the energy output currently shows -3.147 eV/atom.

**Desktop (please complete the following information):**
 - OS: Mac

**Additional context**
I will be making a PR to address ASAP -- will also try to add more tests to catch bugs like this in the future.

CompRhys:
The phase diagram parts running more slowly will be due to the equality now checking the structure in getting stable_entries. It might be worth removing the structure check from the `__eq__` if the slow down in prohibitive?  

I can't think of any edge cases with the gibbs energy correction I went through all the edge cases I could think of.

mkhorton:
Thanks for making a PR @mattmcdermott -- could you add an explicit test for the CO2 data too? Just so we can catch a bug like this in future.

Re. equality checking, open to suggestions on the cleanest way to do that.

mattmcdermott:
@CompRhys the issue with the refactoring is double-counting the energy for experimental entries. i.e. if the composition appears in one of the expt. tables, that expt Gibbs Free Energy is added as a `gibbs_correction` to the formation enthalpy, when it should really just be the final energy!

CompRhys:
@mkhorton cleanest way would probably be to replace
[...]

PR #2058, opened 2021-02-05
Fix bug in GibbsComputedStructureEntry, add test

mattmcdermott:
## Summary

Addresses bug in #2053, where experimentally-derived Gibbs free energies (for gases like CO2) were being added incorrectly to DFT energies. Adds a test to check that energy of CO2 is correct, even after normalization of entry.

Note: a future PR should address the slow equality checking, as well as the additional computational delay from having to do the Gibbs computation each time.

## Checklist


mkhorton:
Thanks @mattmcdermott

PR #2061, opened 2021-02-08
Add tqdm as requirement

jan-janssen:
Because it is used in pymatgen.analysis.pourbaix_diagram

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -129,6 +129,7 @@ setup(
         "pandas",
         "plotly>=4.5.0",
         "uncertainties>=3.1.4",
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks for bringing this to our attention @jan-janssen, I'll close this PR since I fixed it via https://github.com/materialsproject/pymatgen/commit/5126b5bc73173b911acb7d4c20a65655d2a9fe1c instead.

`tqdm` should be an optional dependency, we have a wrapper called `PBar` which will fall back gracefully if tqdm is not available.

PR #2062, opened 2021-02-10
Fix KeyError when initializing XRDCalculator with integer wavelength

flaviu-gostin:
## Summary

* Fix 1: attempting to create a XRDCalculator instance with an integer wavelength (e.g. 1, when wavelength is 1 Angstrom) throws a KeyError.  Made changes to the __init__ method to accept integers for wavelength.  Also raise TypeError if wavelength is neither of float, int or str.  Added test to test if TypeError is raised.

- [x ] Tests have been added for any new functionality or bug fixes.
- [x ] All linting and tests pass. --- Some linting does not pass.  However, it was inherited.  It is not due to this contribution.

### DIFF
--- a/pymatgen/analysis/diffraction/xrd.py
+++ b/pymatgen/analysis/diffraction/xrd.py
@@ -130,9 +130,13 @@ class XRDCalculator(AbstractDiffractionPatternCalculator):
         """
         if isinstance(wavelength, float):
             self.wavelength = wavelength
-        else:
+        elif isinstance(wavelength, int):
+            self.wavelength = float(wavelength)
+        elif isinstance(wavelength, str):
             self.radiation = wavelength
             self.wavelength = WAVELENGTHS[wavelength]
+        else:
+            raise TypeError("'wavelength' must be either of: float, int or str")
         self.symprec = symprec
         self.debye_waller_factors = debye_waller_factors or {}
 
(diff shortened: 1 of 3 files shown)


pymatgen/analysis/diffraction/xrd.py
```diff
@@ -130,9 +130,13 @@ def __init__(self, wavelength="CuKa", symprec=0, debye_waller_factors=None):
         """
         if isinstance(wavelength, float):
             self.wavelength = wavelength
-        else:
+        elif isinstance(wavelength, int):
+            self.wavelength = float(wavelength)
```
shyuep, line 134: I think int and float can be handled in one if statement. The subsequent calculations don't really care whether it si a float or an int.

flaviu-gostin: Thanks, that sounds good.  I pushed another commit which brings the two together.

shyuep:
Thanks. Can you merge main master into your branch pls? I believe that's why your linting failed.

flaviu-gostin:
> Thanks. Can you merge main master into your branch pls? I believe that's why your linting failed.

I just checked and it seems my master is based on the latest commit on the main master, so that wasn't the issue.  I checked the Linting workflow on GitHub and it seems there were not enough blank lines in composition.py.  Run black, pycodestyle and flake8 on that file.  It should be ok now.  I wonder how this file got through in the previous commit when the linting error was introduced.

shyuep:
Thanks!

PR #2065, opened 2021-02-16
Parse projected magnetisation from non-collinear calculations

utf:
## Summary

Added the ability to parse the projected magnetisation from VASP calculations that include spin-orbit coupling or non-collinear magnetism.

For calculations that include orbital projections (`LORBIT = 11`), the projections are written to the vasprun.xml file in the same format as in the PROCAR file (https://www.vasp.at/wiki/index.php/PROCAR). The format of the projections depends on the other calculation parameters:
- For spin-paired calculations (`ISPIN = 1`): There is only one spin channel present.
- For spin-polarised calculations (`ISPIN = 2`): There are two spin channels present.
- For non-collinear or SOC calculations: There are 4 spin channels present.

In the last case, the first spin channel is the usual projection of the wavefunctions onto the atomic orbitals. The last 3 spin channels are actually the projected magnetisation of those orbitals in the 3 cartesian directions.

Previously, the projected eigenvalues from SOC calculations incorrectly had two spin channels, while the remaining spin channels were ignored. This PR fixes the parsing so that the projected magnetisation is extracted correctly for SOC calculations.

I've added a unit test.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -431,7 +431,10 @@ class Vasprun(MSONable):
                 elif parse_eigen and tag == "eigenvalues":
                     self.eigenvalues = self._parse_eigen(elem)
                 elif parse_projected_eigen and tag == "projected":
-                    self.projected_eigenvalues, self.projected_magnetisation = self._parse_projected_eigen(elem)
+                    (
+                        self.projected_eigenvalues,
+                        self.projected_magnetisation,
+                    ) = self._parse_projected_eigen(elem)
                 elif tag == "dielectricfunction":
                     if (
                         "comment" not in elem.attrib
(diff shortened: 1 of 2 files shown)


utf:
I've run black on the files, so I don't know why linting is failing saying "Black would make changes". It doesn't...

mkhorton:
@utf different set of black rules maybe? https://github.com/materialsproject/pymatgen/blob/1e36e835967fb8665a6ede344f4e608af50e93c2/pyproject.toml#L10

utf:
I can run black on outputs.py using a line length of 120 but that will cause lots of unrelated changes. Are you sure you'd like me to do that?

shyuep:
Just run black from the root level of pymatgen.

utf:
> Just run black from the root level of pymatgen.

I did. There were no changes.

shyuep:
I have merged this for now

utf:
Thanks!

shyuep:
I ran black on my system on your pulled code and it definitely reformatted the file. Pls check your black version. It might be out of date.

utf:
I am using the latest version of black, so it can't be that.

```
› black --version                                                                                                                                                                       
black, version 20.8b1
```

mkhorton:
Also seeing this issue myself actually.

PR #2072, opened 2021-02-21
Gaussian: allow None for functional, bset, charge and multiplicity

eimrek:
Hi pymatgen community!

I would like to use the Gaussian input generator for post-processing runs, such as this one 

~~~
%chk=g.chk
%oldchk=../g.chk
#P chkbasis geom=allcheck guess=(only,read) pop=naturalorbital

test
~~~

This would read the geometry, functional, basis set, charge and multiplicity from `../g.chk` and perform natural orbital analysis.

This PR enables this by making 
* functional and basis set are not outputted when set to `None`
* if geometry is `None`, then charge or multiplicity set to `None` will not be displayed

I also added a test for this and I had to modify an existing test.

### DIFF
--- a/pymatgen/io/gaussian.py
+++ b/pymatgen/io/gaussian.py
@@ -461,13 +461,13 @@ class GaussianInput:
         bset_str = '' if self.basis_set is None else self.basis_set.strip()
 
         if func_str != "" and bset_str != "":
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks!

PR #2074, opened 2021-02-24
More Informative Message with PhaseDiagramError

CompRhys:
## Summary

update `PhaseDiagramError` message to tell users which elements are missing.


shyuep:
Please fix the linting error. Also, I suggest you change the error to ValueError instead of PhaseDiagramError. I am planning to remove custom errors in favor of the common ValueError, RuntimeError and KeyError unless there is a good reason to create a new error class.

CompRhys:
I have removed PhaseDiagramError and update the tests.

I always get hit by the linter even when I auto-format the code... have you considered something like https://pre-commit.ci/  to autoformat as an action? It might stop most of the lining issues that are seen?

shyuep:
There is already a pre-commit hook provided. But I prefer not to make it autoformat since the coder might have other ideas.

shyuep:
Thanks.

PR #2075, opened 2021-02-24
[WIP] reconfigure `get_quasi_e_to_hull` when polymorphs are present 

CompRhys:
## Summary

Currently for the stable entry in the toy hull shown below in the presence of a polymorph (blue filled circle) the current code returns the red energy from `get_quasi_e_to_hull` whilst the intended behaviour should be that it gives the orange energy. I have checked this with @CJBartel who agrees that the orange energy is the correct energy. This PR should now return the orange energy.

image

## TODO (if any)

1. Double check edge cases
2. Maybe rename the function to have a more informative name `get_phase_separation_energy` may be appropriate
1. Some materials are not directly tested as the only way to get a test would rely on calling the method being tested for a different material

## Checklist


pymatgen/analysis/phase_diagram.py
```diff
@@ -449,18 +449,6 @@ def stable_entries(self):
...
-        """
-        Args:
-            mode (str): type of normalization to perform.
-                Allowed ["atom", "formula_unit"]
-
-        Returns:
-            list of normalized stable entries in the phase diagram.
-        """
-        return [e.normalize(mode) for e in self._stable_entries]
-
```
CompRhys, line 463: I introduced this in order to get this to work before the changes to `__hash__` and how this method deals with polymorphs therefore in order to keep the code clean as I do not see other workflows method currently I have taken it out.

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -304,16 +308,40 @@ def test_get_quasi_e_to_hull(self):
...
+            toy_pd.get_phase_separation_energy(PDEntry("Li2O", -5)),
+            -1.0,
+            7,
+        )
+        # polymorph
+        self.assertAlmostEqual(
+            toy_pd.get_phase_separation_energy(PDEntry("Li2O", -4)),
+            -2.0 / 3.0,
+            7,
+        )
```
CompRhys, line 332: This is a minimal test for a binary hull that checks the polymorph behaviour

pymatgen/analysis/tests/test_phase_diagram.py
```diff
@@ -323,14 +351,14 @@ def test_get_quasi_e_to_hull(self):
         stable_entry = [e for e in self.pd.stable_entries if e.name == "Li2O"][0]
 
         self.assertEqual(
-            self.pd.get_quasi_e_to_hull(duplicate_entry),
-            self.pd.get_quasi_e_to_hull(stable_entry),
+            self.pd.get_phase_separation_energy(duplicate_entry),
+            self.pd.get_phase_separation_energy(stable_entry),
             "Novel duplicates of stable entries should have same decomposition energy!",
         )
```
CompRhys, line 357: This test was to check that the hashing work around worked. I have left the test in but it doesn't test anything that shouldn't already be covered by another test.

pymatgen/core/tests/test_composition.py
```diff
@@ -397,12 +397,13 @@ def test_equals(self):
...
+        c1 = Composition(f"O{0.2}Fe{0.8}Na{Composition.amount_tolerance*0.99}")
         c2 = Composition(f"O{0.2}Fe{0.8}Na{Composition.amount_tolerance*1.01}")
-        c3 = Composition(f"O{0.2}Fe{0.8}")
+        c3 = Composition(f"O{0.2}Fe{0.8+Composition.amount_tolerance*0.99}")
 
-        self.assertEqual(c1 == c3, hash(c1) == hash(c3), "Hash not robust")
-        self.assertEqual(c2 == c3, hash(c2) == hash(c3), "Hash not robust")
+        self.assertTrue(c1 == c3, "__eq__ not robust")
+        self.assertEqual(c1 == c3, hash(c1) == hash(c3), "Hash doesn't match eq when true")
+        self.assertFalse(hash(c1) == hash(c2), "Hash equal for different chemical systems")
```
CompRhys, line 406: improvements to the tests I added in #2077

CompRhys:
@mkhorton I think this can be reviewed/merged. I added some notes to the code to explain the small changes that may seem unrelated to the PR title.

mkhorton:
Thanks @CompRhys! I'm happy to merge this.

One note regarding the method re-name: given that it's a recently added method and the re-name makes sense, I'm ok to do it here, but as a rule I'd prefer not to merge in changes that rename methods so that we don't cause frustration for users when methods go missing. For future PRs, if renaming does make sense, one option is just to add a stub for the old method name to point to the new method name and give a DeprecationWarning.

Issue #2083, opened 2021-03-05
please consider switching back to semantic versioning


ltalirz:
**Is your feature request related to a problem? Please describe.**

![image](https://user-images.githubusercontent.com/1053245/110118408-9dd05f80-7dba-11eb-8272-66e1b5fd6951.png)

As the github landing page nicely illustrates, `pymatgen` is used as a dependency in many other projects, some of which are again used as components of other projects and so on.

pymatgen's [calendar versioning](https://calver.org/) approach essentially leaves developers of such projects with two options:

 1. lock the version of the pymatgen dependency completely, e.g. `pymatgen==2021.3.4`.
    This has the disadvantage that your tool becomes incompatible with any other tool following the same strategy because their pymatgen dependency versions would clash.
 2. provide a minimal required pymatgen version, e.g. `pymatgen>=2020.3.4`.
    This means your project can (and likely will) break when pymatgen makes breaking changes (such as the recent https://github.com/materialsproject/pymatgen/commit/8f097cf7ea20b8fb42635452b02f520152218394 )

These problems would be solved by switching back to [semantic versioning](https://semver.org/), which would allow developers of upstream tools to depend on pymatgen without risking to break their projects  (e.g. `pymatgen~=5.0`) while still making it possible to work together with other tools that share the pymatgen dependency.

I know that pymatgen originally used semantic versioning and then switched to calendar versioning in https://github.com/materialsproject/pymatgen/commit/51411b19f27906d162de2328214ca3ffed16fdda#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7 in 2017.

I would kindly ask the developers to elaborate on the reasons for this decision and, potentially, to reconsider it.

sphuber:
> These problems would be solved by switching back to [semantic versioning](https://semver.org/)

Should note that this solves problems _provided_ that minor and patch versions do not break backwards compatibility. Might be obvious given the definition of semantic versioning but is in practice not always easy nor guaranteed.

shyuep:
The reason why calendar versioning is used instead of semantic versioning is because the maintenance costs of semantic versioning far outweigh the benefits for us. It is rare that pymatgen introduces breaking changes (other than the recent v2021.3.4, I can't remember really when was the last time this happened). There are also breaking changes that can be introduced by codes that pymatgen depend on, e.g., the recent numpy v1.20 that dropped support for Python < 3.7 and introduced support for the ArrayLike typing object.

[...]

PR #2086, opened 2021-03-08
docs: update compatibility guidance

ltalirz:
## Summary

Drafting an update of the Compatibility guide as requested by @mkhorton in
https://github.com/materialsproject/pymatgen/issues/2083#issuecomment-791597868

## Checklist

in the pull request title.

### DIFF
--- a/docs_rst/compatibility.rst
+++ b/docs_rst/compatibility.rst
@@ -8,11 +8,33 @@ and capabilities are constantly being added.
 However, pymatgen is also used as a library by other tools, and as such breaking
 changes such as the removal or renaming of existing functionality, or substantive
 changes in the output of existing code, are tried to be kept to a minimum. This is
-especially true of all classes contained in the pymatgen.core module.
+especially true of all classes contained in the `pymatgen.core` module.
 
-Despite this, it is sometimes necessary to make breaking changes to enable future
-development, or because other libraries we depend upon might change
-their own requirements. When this happens this page will be updated with guidance.
+Despite this, it is sometimes necessary to make breaking changes to enable
+future development, or because other libraries we depend upon might change
+their own requirements. If a breaking change is causing significant issues,
+please post on the GitHub Issues page to see if it can be resolved.
+
+Depending on Pymatgen
+---------------------
+
+Pymatgen versions (e.g. `2021.3.4`) follow the `semantic versioning <https://semver.org/>`_ 
+scheme and, as such, will increase the **major** version number (here: `2021`)
+whenever breaking changes are made to the API. 
+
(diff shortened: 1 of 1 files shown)


shyuep:
I don't think the writeup is necessarily true. I don't want to give up calendar version just because we are temporarily doing semantic versioning due to the switch over to a namespace package. In future, it will still be calendar versioned.

I think a more important statement is that if you happen to just use pymatgen for its core classes, it really doesn't matter in most instances whether you are using v2018, 2019 or 2021 or 2022. None of the core classes have undergone any backwards incompatible changes in the past few years.

Those who work on the bleeding age of pymatgen will of course have to live with the nature of a research code, i.e., it is always changing!

ltalirz:
> I don't think the writeup is necessarily true. I don't want to give up calendar version just because we are temporarily doing semantic versioning due to the switch over to a namespace package. In future, it will still be calendar versioned.

Oh, I see.

Sorry, I had understood you were going to follow the principle of incrementing the major version number upon breaking changes not just until 2022 but also going forward.

Feel free to disregard this PR then.

> I think a more important statement is that if you happen to just use pymatgen for its core classes, it really doesn't matter in most instances whether you are using v2018, 2019 or 2021 or 2022. None of the core classes have undergone any backwards incompatible changes in the past few years.

Since you mention this, I went back through the aiida-core issue tracker looking for pymatgen-related issues from the last two years

https://github.com/aiidateam/aiida-core/issues/3244
https://github.com/aiidateam/aiida-core/issues/3297
https://github.com/aiidateam/aiida-core/issues/3263
https://github.com/aiidateam/aiida-core/issues/4339
https://github.com/aiidateam/aiida-core/issues/4614
https://github.com/aiidateam/aiida-core/issues/4687
https://github.com/aiidateam/aiida-core/issues/4793

On closer inspection, you are right that most of these are not related to breaking changes. Most were caused by issues with the `numpy` dependency (which should hopefully be fixed now with the `pyproject.toml`) and other dependencies.
Only two were breaking changes - the `species_and_occu` and the recent removal of imports from the core.

All in all, I still think that for a tool like aiida it would still be useful if pymatgen was using semantic versioning... 
but of course announcing them via warnings well in advance already helps as well.

mkhorton:
Adding the `species_and_occu` example to the compatibility page may be a good idea though, if at least to give people a better idea of how often these changes have happened in the past.

shyuep:
@ltalirz Thanks. I will merge this and update the doc to reflect what Matt suggested.

shyuep:
@ltalirz I have clarified the compatibility doc based on your suggested edits. Thanks.

PR #2088, opened 2021-03-09
correct get_neighbor_list speed claim

janosh:
`Structure.get_neighbor_list` claims to be orders of magnitude faster than `get_all_neighbors` which probably refers to `get_all_neighbors_old`. It's actually only about 2-3x faster than `get_all_neighbors`.


shyuep:
Thanks.

PR #2101, opened 2021-03-24
small fix for insertion electrode

jmmshn:
When the decomposition energy are missing the `get_max_instability` and `get_min_instability` functions all error out.
Now they will just return None


mkhorton:
Thanks @jmmshn

PR #2006, opened 2020-12-02
[WIP] enable MaterialsProject2020Compatibility by default

rkingsbury:
## Summary
Make `MaterialsProject2020Compatibility` the default energy correction scheme in pymatgen.

* `MPRester().get_entries()` and `MPRester().get_pourbaix_entries()` now use MP2020 to correct energies
* `MaterialsProjectAqueousCompatibility` now uses MP2020 instead of legacy MP as the `solid_compat`
* Updated Pourbaix tests to reflect energy shifts due to above change
* removed deprecation warning for legacy `MaterialsProjectCompatibility`

## TODO (if any)

- [x] Refit energy corrections after upcoming data release
- [ ] Update publication detail after resubmission
- [ ] validate output of emmet `ThermoBuilder` when using MP2020 corrections to build all phase diagrams

PR #2104, opened 2021-03-25
Update mp2020 to v2021.03.22 DB release, without activating by default

rkingsbury:
## Summary

* Refit MP2020 corrections to the 2021.03.22 DB release using updated experimental data (see #2006 )
* Add the ability to point `MaterialsProject2020Compatibility` to an alternate configuration file
* Various docstring changes and updates ported from #2006

The purpose of this PR is to enable test builds of the MP database using the new corrections without requiring a new pymatgen release. With this PR, we will be able to iterate on correction values in an alternate configuration file, separate from the one distributed with pymatgen.

## TODO (if any)

* There is one test in `test_matproj.py` that is failing, and I think this is a result of the new DB release rather than any changes in this PR.


mkhorton:
Thanks @rkingsbury 

> There is one test in test_matproj.py that is failing, and I think this is a result of the new DB release rather than any changes in this PR.

I'll handle this ...

PR #2108, opened 2021-04-16
Radial site perturbation and local_env patch

nwinner:
## Summary

* Feature 1: I've added a new site transformation called the RadialSiteDistortionTransformation which takes a selected site, and perturbs the atoms around it. The way I've written is that the first nearest neighbors are perturbed by the specified distance, and then subsequent sites follow a 1/r decay, which seems like the most intuitive way to me.
* Feature 2: A rudimentary unit test has been added
* Fix 1: While writing this transformation, I discovered a bug in MinimumDistanceNN from the local_env module. While this is tagged as being applicable to both structures and molecules, it failed for Molecules. This is because it assumed it could find a periodic image of sites. I added a check to ensure that that when its not a structure, this is not needed.

* None

## TODO (if any)

Nominally not a WIP, unless some recommendations exist for the new transformation.

## Checklist

- [ ]  All linting and tests pass.


mkhorton:
Thanks @nwinner, this looks useful! Apologies this was left to sit for a while

PR #2111, opened 2021-04-20
rm spurious warnings filter from GibbsComputedStructureEntryTest

rkingsbury:
Remove a no-longer-needed warnings filter from a test in `GibbsComputedStructureEntryTest`. This had been removed previously but was somehow reinstated amidst the shuffle of MP2020-related merges, and was causing tests to fail.


mkhorton:
Thanks Ryan

Issue #2112, opened 2021-04-22
Numpy numerical types in electronic_structure.bandstructure.Kpoint.as_dict result


zooks97:
In the [`electronic_structure.bandstructure.Kpoint.as_dict` method](https://github.com/materialsproject/pymatgen/blob/6737f23caaf646cc58425561cf08339acf77b396/pymatgen/electronic_structure/bandstructure.py#L127), the numpy arrays `frac_coords` and `cart_coords` are converted to lists using `list()`, which leaves the coordinates as numpy data types.

This causes issues when using the `as_dict` method to convert the object into a dictionary for storing in mongodb using [mincepy_sci](https://github.com/muhrin/mincepy_sci).

This issue is inherited by objects which contain `Kpoint`s and use their `as_dict` method in their own `as_dict` methods, such as `BandStructure`.

A possible solution is to use the numpy array method `tolist()` which also converts the list contents to python types.

PR #2113, opened 2021-04-22
Use numpy tolist method in electronic_structure.bandstructure.Kpoint.as_dict

zooks97:
Use `np.ndarray.tolist()` method instead of python `list()` to convert k-point coordinates in `pymatgen.electronic_structure.bandstructure.Kpoint.as_dict()`.

This fixes issue #2112, where numpy numerical types (e.g. numpy.float64) exist in the result of `Kpoint.as_dict()`.

### DIFF
--- a/pymatgen/electronic_structure/bandstructure.py
+++ b/pymatgen/electronic_structure/bandstructure.py
@@ -130,8 +130,8 @@ class Kpoint(MSONable):
         """
         return {
             "lattice": self.lattice.as_dict(),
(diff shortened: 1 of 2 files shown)


shyuep:
Pls add a unittest to check for this pls. Thanks.

zooks97:
> Pls add a unittest to check for this pls. Thanks.

Done!

shyuep:
Thanks!

PR #2115, opened 2021-04-23
Xdatcar fix

nkeilbart:
## Summary

* Updated Xdatcar class from pymatgen.io.vasp.outputs to update the lattice constants with each new step. Previously, only a single preamble was read in and used for the rest of the structures.
* Created a new XDATCAR test file named XDATCAR_6 to stay consistent with other filenames
* Added a test for Xdatcar that will assure that the lattice constants are different for the first and last image


shyuep:
Thanks.

PR #2117, opened 2021-04-27
Update bader_caller.py

nwinner:
## Summary

My last PR for the bader caller had some issues, which this PR resolves. (edited after discussion below)

* Fix 1: Previously, I had "self.nelects" for cube analysis created by looking for the site property with that name. It didn't occur to me that the self.structure object is not user-defined, but created from the cube file, and this property will never exist. I removed this, and changed the functions to have an optional nelect(s) argument for when parsing cube files.

* Feature 1:  I have added get_decorated_structure method. Which is like the previous "get_spin_decorated_structure", but more general. Assuming you ran bader with an appropriate reference using `chgref_filename`, this lets you assign a generic property to a structure using the reference file for the partitioning and the main file for the integrating. This can be, for example, be used to parse a spin density cube file to get spin decorated structure or a hartree cube file to get a electrostatic potential decorated structure.

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -326,12 +326,12 @@ class BaderAnalysis:
 
     def get_decorated_structure(self, property_name, average=False):
         """
-        Get the property decorated structure from the Bader analysis.
+        Get a property-decorated structure from the Bader analysis.
 
         This is distinct from getting charge decorated structure, which assumes
         the "standard" Bader analysis of electron densities followed by converting
         electron count to charge. The expected way to use this is to call Bader on
-        a non-charge density file such as a spin denisty file, electrostatic potential
+        a non-charge density file such as a spin density file, electrostatic potential
         file, etc., while using the charge density file as the reference (chgref_filename)
         so that the partitioning is determined via the charge, but averaging or integrating
         is done for another property.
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @nwinner. For point 2, doesn’t the Bader binary already support this? (Ie getting volumes from the charge density, then integrating the spin density using those volumes) I seem to recall this feature was available in the Bader caller to do this automatically.

nwinner:
I hadn't seen that before, but after googling it looks like you can use 

bader CHGCAR_spin -ref CHGCAR_sum.

If you generated ACF using that method, then you would get a spin decorated structure by calling get_charge_decorated_structure(), but you would have to flip the sign... 

I'm not sure how to best make that a part of this module though. Thoughts?

mkhorton:
For the Chgcar interface, we automatically do that analysis when a spin
channel is present (I believe it gets added to the summary dictionary). I’m
not sure how that could be done for cube. One option would be to something
like, get_decorated_structure with two args: the field to integrate and,
optionally, a reference field?

nwinner:
Okay so I don't think we need a reference field in decorated structure. It does look like just specifying cghcar_ref when initializing BaderAnalysis, will let you parse a spin density file using the electron density as a reference. So it works as is. 

The name is not great, as chgcar_ref does seem vasp-specific, but its really a "general" reference as far as bader is concerned. 

As for how to change get_decorated_structure, should I put "spin decorated structure" back in, with with the knowledge that you have to set the reference chgcar appropriately, or maybe, should it get_charge_decorated-structure be changed to allow for this? I'm hesitant to rename things because of compatibility.

nwinner:
Okay the newest commit should has a function called get_decorated_structure(). What do you think @mkhorton ?

mkhorton:
Thanks @nwinner, I clarified the docstring a bit otherwise this looks good to me -- ok to merge?

nwinner:
Looks okay to me. Hopefully this doesn't need to be updated again anytime soon.

PR #2119, opened 2021-04-28
Use numpy tolist() in DOS-family object as_dict() methods

zooks97:
Use `np.ndarray.tolist()` method instead of python `list()` to convert density of state (`Dos`, `FermiDos`, `CompleteDos`) energies and densities in their `as_dict()` methods.

This fixes an issue where numpy numerical types (e.g. numpy.float64) exist in the results of {`Dos`, `FermiDos`, `CompleteDos`}`.as_dict()`.

This allows these objects to be saved and stored more easily by [mincePy Sci](https://github.com/muhrin/mincepy_sci).


shyuep:
Thanks.

PR #2121, opened 2021-04-28
Update cube.py

nwinner:
## Summary
 
* Feature 1: Spherical masks of the cube data is now supported
* Feature 2: Using spherical masks, atomic site averages and totals are now supported natively. These can be used, respectively, for getting the average electrostatic potential around a site using a V_HARTREE file from CP2K or the total spin on a site using a SPIN_DENSITY file from CP2K. These features are useful as an alternative to Bader partitioning for these properties when the bader executable is not available.
* Fix 1: Previously, cube files were loaded using for loops, which was prohibitively slow for files >1Gb. Now it is done with numpy and is much faster.

## TODO (if any)

* While not necessary, it would be nice to support multiprocessing for the atomic site functions. These can be pretty expensive when the cube files are >1Gb. I previously used the multiprocessing module for this, but ran into some errors while trying it out. It was exceeding the memory available for the Drone, and since the Drone is supposed to be automatic/not require user modifications, I removed multiprocessing for robustness, but it would be nice to have it back at some point. 
* New tests for the class methods have not been added yet.

## Checklist

in the pull request title.

- [x ] All linting and tests pass.

### DIFF
--- a/pymatgen/io/cube.py
+++ b/pymatgen/io/cube.py
@@ -1,21 +1,24 @@
 """
-Module for reading Gaussian cube files, which have become one of the standard file formats for volumetric data
-in quantum chemistry and solid state physics software packages (VASP being an exception).
+Module for reading Gaussian cube files, which have become one of the standard file formats
+for volumetric data in quantum chemistry and solid state physics software packages
+(VASP being an exception).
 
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks!

PR #2122, opened 2021-04-28
Update vdW radii to CRC handbook values

rkingsbury:
## Summary

This PR updates the `van_der_waals_radius` data used by the `Element` class from Wikipedia data to the latest values in the in the CRC Handbook of Chemistry and Physics. Important changes from the previous values include:

- Revised the value for `H` down by 0.1 angstrom as recommended in a critical review by Rowland and Taylor
- Added values for many lanthanides, actinides, and transition metals that were previously absent
- Updated values or many transition metals

All data, and the justification for which values were chosen, can be found in

"Atomic Radii of the Elements" in CRC Handbook of Chemistry and Physics, 91st Ed.; Haynes, W.M., Ed.; CRC Press: Boca Raton, FL, 2010.


rkingsbury:
Git does not seem well-equipped to diff `.json` files, so you can use this online tool to better see the changes:

https://jsoncompare.com/#!/diff/id=6924b719277a2a2b7a630352828d155c&fullscreen/

click the 'lint' button on both panes and you will see each change nicely highlighted

mkhorton:
Thank you @rkingsbury , definitely happy to have data sourced from somewhere that isn't wikipedia here :-)

PR #2123, opened 2021-04-28
bug fix in SpacegroupAnalyzer

gpetretto:
There is a problem in the `get_conventional_standard_structure` method of the `SpacegroupAnalyzer`.

The code below demonstrates the issue:

```python
from pymatgen.analysis.structure_matcher import StructureMatcher

s = MPRester().get_structure_by_material_id("mp-761986")

sm = StructureMatcher(ltol=0.0001, stol=0.0001, angle_tol=0.1, primitive_cell=False, scale=False)

spga = SpacegroupAnalyzer(s)
sc = spga.get_conventional_standard_structure()
s_ref = spga.get_refined_structure()

print(sm.fit(s, s_ref))
print(sm.fit(s, sc))
```

Output:
```
True
False
```

This comes from the fact that in the triclinic case the reference structure can be modified by the call to `struct.get_reduced_structure("LLL")`, but the lattice object is not updated accordingly.


mkhorton:
Thanks for fixing @gpetretto -- would it be possible to add a quick test case to catch this too? Thank you

PR #2128, opened 2021-04-30
Add auto oxidation state detection to MP2020 ; incremental update to correction values

rkingsbury:
- Automatically try to guess the oxidation state of compounds in `process_entries` if it is not already provided
- Small update to the MP2020 correction values as we finalize for the upcoming database release.


mkhorton:
Thanks @rkingsbury

PR #2129, opened 2021-05-01
Symmetrically equivalent surfaces in slabs

richardtran415:
This pull request contains an improvement for analyzing surface symmetry of slabs. A slab now has symmetrically equivalent surfaces if (hkl) is a mirror plane, if [hkl] is a screw axis or if the slab has inversion symmetry.

* Modifications to is_symmetric() Slab class method. If slab contains inversion symmetry, the (hkl) plane is a mirror plane, or there is a screw axis about [hkl], the surfaces of the slab are symmetrically equivalent.
* Test for generating slabs that don't need inversion
* Remove really inefficient Slab class method for checking equivalent surfaces have_equivalent_surfaces()
* Modify SlabGenerator class method nonstoichiometric_symmetrized_slab() to use is_symmetric() when generating slabs

### DIFF
--- a/pymatgen/core/surface.py
+++ b/pymatgen/core/surface.py
@@ -313,8 +313,8 @@ class Slab(Structure):
         sg = SpacegroupAnalyzer(self, symprec=symprec)
         symmops = sg.get_point_group_operations()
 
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks richard.

PR #2131, opened 2021-05-03
Fix QChem gzipped parsing and add gzipped test

samblau:
My last PR fixed one QChem parsing problem, but introduced a new one - the zopen call could no longer accommodate gzipped output files. I have fixed this problem and also gzipped one of my already present test output files to ensure that the ability to parse gzipped outputs is now tested.


samblau:
@mkhorton ready for review. Thanks!

mkhorton:
Thanks @samblau

PR #2133, opened 2021-05-04
Add $van_der_waals input block to QCInput

rkingsbury:
## Summary

Adds support for custom vdW radii to `QCInput` and `QChemDictSet`. These radii are used in the construction of PCM cavities and when calculating charges.

## TODO

- [x] Feedback requested from @samblau regarding the `vdw_mode` kwarg that I had to add. This construction is a little bit awkward. An alternative would be to include a special 'mode' key in the `van_der_waals` dict. Please let me know if you have a preference.


samblau (approved):

samblau:
I'm fine with it as-is. It's a little awkward, but it would also be a little awkward to shoehorn mode into the van_der_waals dict.

PR #2135, opened 2021-05-07
Add van_der_waals support to QCInput.from_file

rkingsbury:
Follow up to #2133 , adds support for reading the `van_der_waals` block from a QChem output file.

This PR also refactors `test_write_file`. Previously, this test used `QCinput.from_file` to construct a dict, and then compared the contents of the dicts. The problem with this approach is that if `QCInput.from_file` has a bug in it that prevents parsing a section (as was the case when I was developing this code), the test will still pass.

The refactored test performs a direct comparison of the text file written by `write_input` and the reference text file.


mkhorton:
Many thanks @rkingsbury!

PR #2141, opened 2021-05-13
MP2020 corrections post-release housekeeping

rkingsbury:
## Summary

Some housekeeping changes following up the MP2020 database release v2021.05.13

* Silence the oxidation states warning when using `MPRester`
* Update the Pourbaix Diagram tests to be less dependent on specific energy values from the database
* Add ChemRxiv link for MP2020 manuscript to the `MaterialsProject2020Compatibility` docstring


rkingsbury:
@shyuep as part of this PR I'd like to address your request to minimize the dependency of Pourbaix Diagram tests on specific energy values from the database. However, I noticed that the test in question ([link](https://github.com/materialsproject/pymatgen/blob/9953bbe3431865bc637d7aff8a5199b21a3e7912/pymatgen/analysis/tests/test_pourbaix_diagram.py#L275)) appears to be designed specifically to test the pipeline from `MPRester` to `PourbaixDiagram`.

So, I left the the test intact, but I changed [this line](https://github.com/materialsproject/pymatgen/blob/9953bbe3431865bc637d7aff8a5199b21a3e7912/pymatgen/analysis/tests/test_pourbaix_diagram.py#L287) to test the identity of the stable entry rather than its energy. I also moved the custom ion test in [this line](https://github.com/materialsproject/pymatgen/blob/9953bbe3431865bc637d7aff8a5199b21a3e7912/pymatgen/analysis/tests/test_pourbaix_diagram.py#L298) to a different part of the code, and integrated that data into the local test file because this test doesn't have anything to do with MPRester at all.

Let me know if you think this will accomplish what you were looking for.

shyuep:
Pls remove the MPRester dependency. The pipeline to MP should be tested in the ext.matproj so that we can isolate it. I don't think we should have external dependencies in test in the rest of pymatgen.

rkingsbury:
> 
> 
> Pls remove the MPRester dependency. The pipeline to MP should be tested in the ext.matproj so that we can isolate it. I don't think we should have external dependencies in test in the rest of pymatgen.

Done. All `MPRester` related tests have been moved to `test_matproj.py`

shyuep:
Thanks.

PR #2143, opened 2021-05-14
Add new compact style for hkl annotations to peaks in diffraction patterns

flaviu-gostin:
Introducing a new optional style for the hkl labels in diffraction patterns (see example figure below).

The default style was not changed.  However, I would recommend making this new more-compact style the default because it prevents overlap of labels as in https://matgenb.materialsvirtuallab.org/2013/01/01/Calculating-XRD-patterns.html (a similar image can be seen here at the bottom)

```python
from pymatgen.analysis.diffraction.xrd import XRDCalculator

a = 4.209
latt = Lattice.cubic(a)
struct = Structure(latt,  ["Cs", "Cl"], [[0, 0, 0], [0.5, 0.5, 0.5]])
c = XRDCalculator()
```

1. The new more-compact style:
```python
c.show_plot(struct, annotate_peaks='compact')
```
Figure_1

2. The current default style (previously: True, now: 'full'):
```python
c.show_plot(struct, annotate_peaks='full')
```
Figure_2

### DIFF
--- a/pymatgen/analysis/diffraction/core.py
+++ b/pymatgen/analysis/diffraction/core.py
@@ -79,7 +79,7 @@ class AbstractDiffractionPatternCalculator(abc.ABC):
         self,
         structure,
         two_theta_range=(0, 90),
-        annotate_peaks="full",
+        annotate_peaks="compact",
         ax=None,
         with_labels=True,
         fontsize=16,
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks. Feel free to make it the default!

shyuep:
Thanks!

flaviu-gostin:
Thank you for accepting this change!

shyuep:
We should be thanking you! The XRDs now look much better!

PR #2149, opened 2021-05-20
Overhaul interfaces

shyamd:
This PR overhauls all the interface code including the ZSL Generator and the substrate matcher:

- Creates a new `Interface` data class in core
- Moves interface algorithm to `analysis.interfaces`; imports and warnings in the old modules ensure they should still work
- Cleans up `ZSLGenerator` and `SubstrateAnalyzer` by implementing functional dataclasses for the matches making it easier to understand how to extra important quantities like the strain and transformation matrix
- Implements a `CoherentInterfaceBuilder` which builds `Interface` objects using the lattice matching of the Zur and McGill algorithm


pymatgen/analysis/interfaces/coherent_interfaces.py
```diff
@@ -0,0 +1,293 @@
+# coding: utf-8
+# Copyright (c) Pymatgen Development Team.
+# Distributed under the terms of the MIT License.
+"""
+This module provides classes to store, generate, and manipulate material interfaces.
+"""
+
+from __future__ import annotations
```
shyuep, line 8: Pls remove this unless it is really needed for Py3.7+

mkhorton:
Nice PR!

PR #2160, opened 2021-06-02
fix another CombinedData bug

htz1992213:
## Summary

* Fix 1: The CombinedData constructor previously creates new ff dict only according to the keys in the first mol in the mols llist, which may cause the ff information in the remaining mols lost. The new implementation create the ff dict using a set of keys containing all ff keys in mols
* Fix 2: The CombinedData constructor previously traverses the topo and ff info of mol in mols list without checking None type, which may cause error. The new implementation checks for None before iteration, and set ff/topo of the CombinedData to be None if all mol in mols list has None ff/topo.

Tests have been created accordingly.


mkhorton:
Thanks @htz1992213. Rather than edit the pylintrc with this specific case, I think you can add an in-line comment (something like `# ignore`).

Can you comment on what the issue was with `ComputedStructureEntry._composition`? This is a protected member anyway, so shouldn't be access outside the class.

htz1992213:
Thanks for the advice @mkhorton! The access comes from the following line, I am not familiar with these classes so I will change the code with the # ignore workaround.

https://github.com/materialsproject/pymatgen/blob/463e4f1b577287c593e493b2b1138320418f0bc7/pymatgen/entries/computed_entries.py#L700

rkingsbury:
> 
> 
> Thanks @htz1992213. Rather than edit the pylintrc with this specific case, I think you can add an in-line comment (something like `# ignore`).
> 
> Can you comment on what the issue was with `ComputedStructureEntry._composition`? This is a protected member anyway, so shouldn't be access outside the class.

@mkhorton I will note that `ComputedStructureEntry._composition` seems to be causing a `1pylint` failure in another PR as well, in which the changes are completely unrelated. Does this test pass in master?
https://github.com/materialsproject/pymatgen/pull/2135/checks?check_run_id=2754058779

htz1992213:
Hey @mkhorton, could you please let me know if there are other things that I should address for the PR to be merged? Thanks!

mkhorton:
This is all good, thank you Tingzheng!

PR #2163, opened 2021-06-04
Fix KSPACING formula in MPScanRelaxSet

ab5424:
## Summary

Change rmin formula to match Eq. 25 from the paper.


mkhorton:
FYI @rkingsbury

mkhorton:
Thanks for the catch @ab5424

rkingsbury:
Thanks for catching @ab5424 . @mkhorton I can confirm; I looked at the paper and the workflow again. The difference in KSPACING appears to be small, and the previous (incorrect) equation would have resulted in a tighter / more stringent KSPACING value than the one that will be computed now.

mkhorton:
Great, will merge. Thanks both!

PR #2164, opened 2021-06-06
reorder SETTINGS load priority

ardunn:
## Summary

When both a .pmgrc.yaml and environment variables are specified, any pmg environment variable will not be read into `pymatgen.core.__init__`'s `SETTINGS`.

E.g., I have a `PMG_MAPI_KEY` as an environment variable and my .pmgrc.yaml does not contain that key - it looks like

```
#.pmgrc.yaml
MAPI_DB_VERSION:
  LAST_ACCESSED: '2021_05_13'
  LOG: {'2021_05_13': 6}
```

`PMG_MAPI_KEY` will not be found in `SETTINGS`. 

This PR assumes this is not the desired behavior. `SETTINGS` here are now loaded as the union of env vars and pmgrc.yaml values, with the priority being pmgrc.yaml in case of a setting being defined in both sources.

If the desired behavior is in fact to void env variables for `SETTINGS` if .pmgrc.yaml exists, please close this PR. 

## Checklist

- [X] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). The easiest way to handle this
- [X] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [X] N/A Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to type check your code.
- [X] N/A Tests have been added for any new functionality or bug fixes.
- [X] All linting and tests pass.

### DIFF
--- a/pymatgen/core/__init__.py
+++ b/pymatgen/core/__init__.py
@@ -37,18 +37,24 @@ SETTINGS_FILE = os.path.join(os.path.expanduser("~"), ".pmgrc.yaml")
 
 
 def _load_pmg_settings():
+    # Load environment variables by default as backup
+    d = {}
+    for k, v in os.environ.items():
+        if k.startswith("PMG_"):
+            d[k] = v
+        elif k in ["VASP_PSP_DIR", "MAPI_KEY", "DEFAULT_FUNCTIONAL"]:
+            d["PMG_" + k] = v
+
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @ardunn

PR #2165, opened 2021-06-06
QChem: detect NLebdevPts error in QCOutput

rkingsbury:
## Summary

Detect `need to increase the array of NLebdevPts` error in `QCOutput` that can occur when computing RESP charges. Detecting this error is necessary to enable a handler in custodian. See [custodian #170](https://github.com/materialsproject/custodian/pull/170)

### DIFF
--- a/pymatgen/io/qchem/outputs.py
+++ b/pymatgen/io/qchem/outputs.py
@@ -1264,6 +1264,15 @@ class QCOutput(MSONable):
             == [[]]
         ):
             self.data["errors"] += ["premature_end_FileMan_error"]
+        elif (
+            read_pattern(
+                self.text,
+                {"key": r"need to increase the array of NLebdevPts"},
+                terminate_on_match=True,
+            ).get("key")
+            == [[]]
+        ):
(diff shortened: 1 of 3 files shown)


rkingsbury:
Note that there seems to be a new pylint error causing linting to fail on all PRs
```
************* Module pymatgen.entries.computed_entries
pymatgen/entries/computed_entries.py:700:8: E1101: Instance of 'ComputedStructureEntry' has no '_composition' member (no-member)
```

shyuep:
You can try to fix this error? Technically, there should be a _compositon variable. If not, if you are sure this is purely a linter issue, you can set the ignore.

mkhorton:
@shyuep this is not related to this PR, this is a pre-existing issue. I've taken a look at it already + will fix soon. I think it's just started appearing in logs because mypy is getting smarter.

mkhorton:
Thanks @rkingsbury! Have merged since linter issue was unrelated to this PR and otherwise looks good. Will try to get linters passing in the main branch again soon.

PR #2171, opened 2021-06-10
Avoid error exit for ALGO=CHI

KazMorita:
## Summary

* Added ALGO=CHI in place of ALGO=BSE to avoid the code from error exiting.
* This change is minor, so no change to the document is required.
 
* non

## TODO (if any)

* non


shyuep:
Can you add a simple unittest for this? Thanks.

KazMorita:
> Can you add a simple unittest for this? Thanks.

I added a file, "test_files/vasprun.xml.chi.gz", and implemented a test to read it.
Please check.

PR #2178, opened 2021-06-20
QE IO: allow to distinguish same element with different oxidation state

FCMeng:
Allow to distinguish same element with different oxidation state, which might have different pseudopotentials.

## Summary

The only change I made is at line 153 of pwscf.py file:
     out.append("  %s %.6f %.6f %.6f" % (site.specie.symbol, site.a, site.b, site.c))
-> out.append("  %s %.6f %.6f %.6f" % (site.specie, site.a, site.b, site.c))

For example, if in the structure, there is a "Ti+" state, it will be exported as "Ti" if we use site.specie.symbol to export the pwscf input file. In real QE calculations, it is possible that "Ti" and "Ti+" elements have different pseudopotentials. 
By making the change from site.specie.symbol to site.specie, the oxidization state information is preserved.


shyuep:
Can you add a simple unittest pls?

FCMeng:
> Can you add a simple unittest pls?
@shyuep  I uploaded a simple unittest here.  Thanks!

[test.zip](https://github.com/materialsproject/pymatgen/files/6682884/test.zip)

shyuep:
Hi there, you do not need to send it as a zip file. All you need to do is to edit pymatgen/io/tests/test_pwscf.py to add this test to it. Then the checks will automatically run to make sure that the code is running as it should.

FCMeng:
> Hi there, you do not need to send it as a zip file. All you need to do is to edit pymatgen/io/tests/test_pwscf.py to add this test to it. Then the checks will automatically run to make sure that the code is running as it should.

@shyuep Thank you, Dr. Ong. I have committed a modified test_pwscf.py file, where there are three tests for the oxidation states: one with fully oxidation(like Li+, Li+, O2- in the test), second without any oxidation (like Li, Li, O in the test), and the third with a mixed oxidation (like Li, Li+, O in the test).

mkhorton:
Thanks @FCMeng, merging because the listing issue is not related to this PR and otherwise this looks good!

Issue #2185, opened 2021-06-29
eigenvalue_band_properties based on separate spin channels


arosen93:
**Is your feature request related to a problem? Please describe.**
Currently, the band gaps computed using the `eigenvalue_band_properties` function in `pymatgen.io.vasp.outputs` (e.g. for the `Vasprun` or `Eigenval` classes) are based on the VBM and CBM independent of spin. For some spin-polarized systems, this might result in a band gap that is formally a spin-forbidden transition, which may not be the gap of interest depending on the use-case. Additionally, it is not currently possible to quickly determine if a material might be something like a [half-metal](https://en.wikipedia.org/wiki/Half-metal), where one spin channel has metallic behavior and the other does not. [Disclaimer: I'm by no means an expert on this topic, so apologies if I've butchered it.]

As an example, see the DOS below. Band gap 1 (and its corresponding CBM and VBM) is the current value computed in Pymatgen. Band gap 2, however, may be a desirable property to compute since there is no spin transition. Band gap 3, while larger than band gap 2, may also be of interest. Currently, there is no way to compute band gap 2 or band gap 3.

![image](https://user-images.githubusercontent.com/8674072/123741669-11b5dc80-d870-11eb-8629-4b68455c1282.png)

**Describe the solution you'd like**
I propose to add a flag `separate_spins` that, when set to False (default), does not change the behavior of the eigenvalue band properties. However, when `separate_spins` is set to True, the returned tuple will contain properties for the spin-up and spin-down channels separately. The user would most likely be interested in the minimum value of band gap 2 and 3, but they can take the minimum on their own if desired.

**Describe alternatives you've considered**
Nothing major to note.

**Additional context**
Here is an example of how the band gap can currently be computed.
```python
from pymatgen.io.vasp.outputs import Eigenval
eig = Eigenval('EIGENVAL') # I am proposing to add a new keyword argument here
bg = eig.eigenvalue_band_properties[0]
print(bg)
```

**Note**
I am currently working on the PR. Hopefully I can submit it soon if you think this is desirable.

arosen93:
Addressed in #2187.

PR #2187, opened 2021-06-29
Add support for spin-dependent eig band props

arosen93:
## Summary

PR to address #2185 (please refer to this Issue for additional context). This PR allows for the band gap, CBM, VBM, and is_direct_band_gap to be computed for each individual spin channel in a spin-polarized calculation when using the VASP Eigenval or Vasprun parsers via a new `separate_spins` keyword argument. Test cases are also included.

As an example, one can run:
```python
from pymatgen.io.vasp.outputs import Eigenval
eig = Eigenval('EIGENVAL', separate_spins=True)
bg = eig.eigenvalue_band_properties[0]
print(bg)
```
which would print out a list of length two where index 0 and index 1 are the band gaps for the up- and down-spin channels, respectively. Analogous behavior would occur if the Vasprun parser were used.

## TODO (if any)

~Setting `separate_spins=True` should raise an error if the calculation is not spin-polarized (i.e. if ISPIN != 2). I have done this already for the Eigenval class (as shown [here](https://github.com/materialsproject/pymatgen/blob/42156a757ad9a52cd28d42005f4c02c6289e6286/pymatgen/io/vasp/outputs.py#L5276)) but wasn't sure the best way to introduce this for the Vasprun and BSVasprun classes.~

**Any suggestions?**


mkhorton:
Thanks for this PR @arosen93, it looks like a good addition to me.

Note there's possibly a more serious logic bug in the current `eigenvalue_band_properties` in that I believe it's broken on metals, in that it will give a very small but non-zero gap.

arosen93:
@mkhorton -- thanks for the head's up. I think I may have observed similar behavior before.

I think the PR is pretty much done. I am going to use it for a day or so more to make sure nothing unexpected pops up. I'll reach out here once I'm done with that.

Regarding the TODO I mentioned, do you know the best way for me to check if ISPIN = 2 in the Vasprun and BSVasprun classes so I can raise an error if it's not equal to 2 but `separate_spins` is called?

arosen93:
Okay, @mkhorton -- I'm feeling good about this one whenever you wish to proceed with the PR.

mkhorton:
Thanks for adding this! Have merged.

While reviewing, I noticed we have two copies of this implementation (one for `Vasprun`, one for `Eigenval`), I'm not sure if we should simplify this at some point to have a common implementation.

PR #2190, opened 2021-07-02
Gruneisen classes

JaGeo:
## Summary

@ab5424 , @gpetretto and I have worked on classes to handle Grüneisen parametrs in pymatgen. We have added a module phonopy.phonon.gruneisen.py and also plotter classes. The parameters can also be imported from phonopy. 

Please let me know if we can do anything further to get this pull request merged soon.


JaGeo:
Will fix the rest of the problems on Monday... (sorry)

mkhorton:
Looking at the logs, it's unclear why so many linting issues are coming up for classes that are not touched by this PR -- only linting issues associated with the files modified by this PR need to be addressed.

One request is that when a new plotter is added, that it supports a `plotly` backend (see `PDPlotter` for an example). This is not required for the PR to be merged. The reason for this is that plotly plots can be more easily integrated with web apps, and also provide useful interactive features in Jupyter notebooks.

Otherwise, this looks like a great addition! Thank you @ab5424 @gpetretto @JaGeo.

shyuep:
@mkhorton The linting issues are caused by the pylint upgrade, which now detects more instances of .item used,.

mkhorton:
Ah, thanks @shyuep

ab5424:
@mkhorton Thanks for your comment.

Regarding the `plotly` backend: the new band structure plotter inherits from [`PhononBSPlotter`](https://github.com/materialsproject/pymatgen/blob/b789d74639aa851d7e5ee427a765d9fd5a8d1079/pymatgen/phonon/plotter.py#L242), which has no `plotly` backend either. Would it be ok to merge this first and add the new backend later? I think all classes in phonon/plotter only have a `matplotlib` backend, so it might be easier to rework everything in a separate PR.

mkhorton:
Yes, absolutely. As I said above that’s not required for this to be merged, it was more a comment to share our roadmap.

We actually do have plotly code for the BSPlotter that could be merged in, I’m not sure without checking to what extent PhononBSPlotter could make use of this.

ab5424:
Ok, thanks. `PhononBSPlotter` definitely uses some code from `BSPlotter` so this seems like a good start.

JaGeo:
I have fixed all problems in the linting part and also a problem with the test of the Gruneisen classes. The other errors seem to come from a different part of pymatgen that we haven't touched. Is it okay like this for now or do I need to do something else?

Thanks in advance. 

JG

JaGeo:
There are still some problems from our side. They result from the fact that phonopy is needed for some of the tests. I hope that we can get back to them in the next days.

JaGeo:
Has one of you an idea why only one test crashes? https://github.com/materialsproject/pymatgen/pull/2190/checks?check_run_id=3001365472
Is there something different with regard to the requirements? Locally, the test runs correctly with and without phonopy installed.

JaGeo:
Okay, found the problem... Let me know if we should do something more before this can be merged. 

JG

pymatgen/io/phonopy.py
```diff
@@ -196,6 +189,33 @@ def get_ph_bs_symm_line_from_dict(bands_dict, has_nac=False, labels_dict=None):
     return ph_bs
 
 
+def get_eig_b(b, q, structure):
+    """
+
```
mkhorton, line 194: This docstring needs to be completed before merging.

pymatgen/io/phonopy.py
```diff
@@ -430,3 +450,246 @@ def get_phonon_band_structure_symm_line_from_fc(
...
+    The order is structure > structure path > structure from gruneisen dict.
+    Newer versions of phonopy include the structure in the yaml file,
+    the structure/structure_path is kept for compatibility.
+
+    Args:
+        gruneisen_path: Path to gruneisen.yaml file (frequencies have to be in THz!)
+        structure: pymatgen Structure object
+        structure_path: path to structure in a file (e.g., POSCAR)
+
+    Returns:
```
mkhorton, line 467: This docstring also needs the "Returns:" section specified.

pymatgen/io/phonopy.py
JaGeo: Decided to remove this function again and include this in the main code. At a certain stage of the code development, I needed this function more than once but this is now obsolete.

pymatgen/io/phonopy.py
JaGeo: Added returns GruneisenParameter object

JaGeo:
@mkhorton, I have now included all suggestions. I hope it is fine now.

mkhorton:
Thanks all!

JaGeo:
Thank you as well!

Issue #2188, opened 2021-06-29
CombinedData serialization with monty is broken


rkingsbury:
**Describe the bug**
I am unable to serialize and deserialize a `CombinedData` object using `loadfn` / `dumpfn`. I believe this occurs because `CombinedData` inherits `from_dict` from `LammpsData`, but `CombinedData.init()` takes different kwargs.

**To Reproduce**

Create a `CombinedData` object called `data`
```
from monty.serialization import loadfn, dumpfn

dumpfnn(data, 'some_path.json')
data = loadfn('some_path.json')
```

will fail with

```
------------------------------------------------------------------------
TypeError                              Traceback (most recent call last)
<ipython-input-26-b8f560ec17df> in <module>
      1 # data = LammpsData.from_file(pipeline_dir / '2021-05_prelim_lammps_TMA-Cl' / 'md.data')
----> 2 data = loadfn(pipeline_dir / '2021-05_prelim_lammps_TMA-Cl' / 'combined_data.json')

~/miniconda3/envs/md/lib/python3.9/site-packages/monty/serialization.py in loadfn(fn, *args, **kwargs)
     83                 if "cls" not in kwargs:
     84                     kwargs["cls"] = MontyDecoder
---> 85                 return json.load(fp, *args, **kwargs)
     86 
     87             raise TypeError("Invalid format: {}".format(fmt))

~/miniconda3/envs/md/lib/python3.9/json/__init__.py in load(fp, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    291     kwarg; otherwise ``JSONDecoder`` is used.
    292     """
--> 293     return loads(fp.read(),
    294         cls=cls, object_hook=object_hook,
    295         parse_float=parse_float, parse_int=parse_int,

~/miniconda3/envs/md/lib/python3.9/json/__init__.py in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    357     if parse_constant is not None:
    358         kw['parse_constant'] = parse_constant
--> 359     return cls(**kw).decode(s)

~/miniconda3/envs/md/lib/python3.9/site-packages/monty/json.py in decode(self, s)
    381         """
    382         d = json.JSONDecoder.decode(self, s)
--> 383         return self.process_decoded(d)
    384 
    385 

~/miniconda3/envs/md/lib/python3.9/site-packages/monty/json.py in process_decoded(self, d)
[...]

PR #2191, opened 2021-07-03
CombinedData.as_dict/from_dict and other methods overwrite; CombinedData.as_lammpsdata deepcopied

htz1992213:
## Summary

* Add `as_dict`, `from_dict` for CombinedData to overwrite superclass methods
* Overwrite `structure`, `disassemble`, `from_ff_and_topologie`, `from_structur` methods for CombinedData
* Rewrite `as_lammpsdata` method for CombinedData so that each attribute is deep copied
* Update docs
* Join string frags
* Added tests


htz1992213:
Fixes #2188

htz1992213:
Hey @mkhorton, I have added tests for the new methods.

htz1992213:
The serialization of the `CombinedData` class should work as expected with the update @rkingsbury

rkingsbury:
> The serialization of the `CombinedData` class should work as expected with the update @rkingsbury

Thanks @htz1992213 ! This looks like a great update that will really improve robustness and clarity of `CombinedData`.

pymatgen/io/lammps/data.py
```diff
@@ -351,7 +351,7 @@ def get_string(self, distance=6, velocity=8, charge=4, hybrid=True):
             velocity (int): No. of significant figures to output for
                 velocities. Default to 8.
             charge (int): No. of significant figures to output for
-                charges. Default to 3.
+                charges. Default to 4.
```
rkingsbury, line 354: If there is no significant penalty in terms of size or performance, I suggest we default to 5 decimal places here. The only reason I say that is that for the TIP4P-FB water model I'm using, 5 decimals are required in order to keep the water molecule neutral. Obviously it's not a big deal to set this via kwarg if there are strong reasons to keep only 4 decimals as the default.

rkingsbury:
Longer-term, I still think we should work on getting `DataFrame` supported in `monty` so that we can do away with custom `as_dict` and `from_dict` methods (and this comment applies throughout pymatgen). But for now this is a good solution. `CombinedData` has been inheriting too many methods from `LammpsData` that don't make sense, so I'm glad you added these as well as the `AttributeError` where appropriate.

mkhorton:
I think we should prefer the monty solution over the custom `as_dict`/`from_dict` here actually; let's try and get the monty solution first, and fall back to this if that's impractical? I think it would be an equivalent number of lines of code.

I'm trying to actively discourage custom `as_dict`, `from_dict` methods wherever possible, they're creating a lot of technical debt.

rkingsbury:
> I think we should prefer the monty solution over the custom `as_dict`/`from_dict` here actually; let's try and get the monty solution first, and fall back to this if that's impractical? I think it would be an equivalent number of lines of code.
> 
> I'm trying to actively discourage custom `as_dict`, `from_dict` methods wherever possible, they're creating a lot of technical debt.

@mkhorton I agree with you, however this PR fixes several pretty serious flaws in `CombinedData` that are affecting our workflows, and brings it into line with `LammpsData` which also has custom `as_dict` / `from_dict`. I'd encourage you to consider merging this as a stopgap solution until we get monty patched, which may take more time. Once that happens, we can remove `as_dict` / `from_dict` from all the Lammps I/O classes at once.

mkhorton:
Unfortunately "stopgap" solutions are how we accrue technical debt, because we forget to to back and remove them. I am reluctant to merge this when there has been a clear alternative suggested, with specific code proposed, and that could be implemented on a comparable timeframe.

mkhorton:
Note that this comment is for the `as_dict`, `from_dict` specifically, not for the remainder of the PR which is fine to merge.

pymatgen/io/lammps/data.py
htz1992213: The previous default was 3, changed to 4 as a safe buffer. Would you prefer to change it to even more? I am actually working on a charge writer class in `mdgo.forcefield`, it can auto-detect significant digits so that users don't have to specify that explicitly to make it right.

htz1992213:
> Unfortunately "stopgap" solutions are how we accrue technical debt, because we forget to go back and remove them. I am reluctant to merge this when there has been a clear alternative suggested, with specific code proposed, and that could be implemented on a comparable timeframe.

@mkhorton I think modifying `monty` is beyond my expertise because it involves making test cases that I may not familiar. I don't think this is a stopgap because it is for sure built on the latest dependencies of` pymatgen`. But I would be happy to modify both `LammpsData`/`CombinedData` immediately once the patch to `monty` is in place. I am quite actively maintaining this submodule so no worries about forgetting to go back here. And I think the urgent is to make sure the serialization of `CombinedData` works.

mkhorton:
> @mkhorton I think modifying monty is beyond my expertise because it involves making test cases that I may not familiar. 

Ok, that's not a problem -- I'm happy to propose this change as a PR to monty myself but it'll take me a few days until I can get to it.

htz1992213:
Hey @mkhorton @rkingsbury, Shyue merged the Monty update, so I was able to deprecate the as_dict and from_dict methods for LammpsData and CombinedData. The MontyEncoder/Decoder works fine here. One thing I want to add is that, the pandas.DataFrame.to_dict method will convert the int type "index" of the df to str type. This won't cause a fundamental error here because LammpsData will just output that column as str anyway. But I think in general it is something that could be fixed in Monty.

rkingsbury:
Thanks @htz1992213 ! Good catch re: the index type. I don't have a well-informed opinion of whether that type conversion is a problem or not.

mkhorton:
Thanks @htz1992213! I'll go ahead and merge.

The index conversion is interesting to know -- this is done by pandas JSON export itself rather than monty, I believe?

PR #2193, opened 2021-07-07
Bugfix for defect BandFillingCorrection

kavanase:
## Summary
This is a bugfix for two significant errors in the BandFillingCorrection class for defects.
I have attached a PDF, notebook and corresponding data to demonstrate the errors, and how these are now fixed with this PR. I have changed the value in the BandFillingCorrection test to match the correct value as well, and have attached a PDF & notebook for this, calculating this correction manually to verify that it is the correct result.

* **Fix 1**: Fix the potential alignment shifting of the band extrema.
* **Fix 2**: Fix the eigenstate occupancy spinfactor calculation, so that spin-orbit coupling (SOC) calculations can be parsed correctly (previously giving a spurious doubling of the occupancies in these cases).

I have added a new test for the BandFillingCorrection with a SOC defect calculation, and have attached files here demonstrating that this gives the correct value (by also calculating it manually).

[BandFillingCorrection_Tests.zip](https://github.com/materialsproject/pymatgen/files/6779207/BandFillingCorrection_Tests.zip)
[Te_i_Pymatgen_Bandfilling_Error_Demo.zip](https://github.com/materialsproject/pymatgen/files/6779213/Te_i_Pymatgen_Bandfilling_Error_Demo.zip)

### DIFF
--- a/pymatgen/analysis/defects/corrections.py
+++ b/pymatgen/analysis/defects/corrections.py
@@ -828,9 +828,9 @@ class BandFillingCorrection(DefectCorrection):
                         VBM of bulk calculation (or band structure calculation of bulk);
                         calculated on same level of theory as the defect
                         (ex. GGA defects -> requires GGA vbm)
-                        
+
                     run_metadata["defect_incar"] (dict)
-                        Dictionary of INCAR settings for the defect calculation, 
+                        Dictionary of INCAR settings for the defect calculation,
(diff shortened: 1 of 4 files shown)


mkhorton:
Thanks @kavanase! Both for catching and fixing. Tagging @nwinner in case this is relevant to your work.

PR #2198, opened 2021-07-08
Fix substrate analyzer not actually returning objects

### LINKED ISSUE #2195: Inconsistent use of output file path in Structure.to() method
**Describe the bug**
Structure.to() method ignores the parent path of filename for certain file formats, say, .xsf format.
For example, in the following scenario,
```python
struct = Structure(...)
struct.to(filename="/path/to/struct.ext") 
```
the output file is correctly written to /path/to/struct.ext when .ext is .cif, .mcif, .json, .cssr, .poscar. But, for some formats including .xsf, it is written at ./struct.ext (current directory).

This is because, in the source code of to(), fname (which only contains the basename of the given filename) is used instead of filename variable for some extensions.
For example,  in core/structure.py:2303-2310, 
```python
        elif fmt == "xsf" or fnmatch(fname.lower(), "*.xsf*"):
            from pymatgen.io.xcrysden import XSF

            s = XSF(self).to_string()
            if filename:
                with zopen(fname, "wt", encoding="utf8") as f:
                    f.write(s)
            return 
```
zopen uses fname instead of filename. It is also the case for some other extensions.

**Expected behavior**
I think to() method should always use filename as the output path.

**Desktop (please complete the following information):**
 - OS: MacOS Big Sur

shyamd:
This should fix #2198 , which was a bug introduced in the interfaces reorganization where the substrate analyzer was constructing matches to return, but not actually returning them... DOH!

I've added a test to check for this in the future. 

Edit: Should be #2198. Thanks @bernstei

### DIFF
--- a/pymatgen/analysis/interfaces/substrate_analyzer.py
+++ b/pymatgen/analysis/interfaces/substrate_analyzer.py
@@ -64,16 +64,24 @@ class SubstrateMatch(ZSLMatch):
         else:
             elastic_energy = 0
 
-        match_dict = match.as_dict()
-
-        cls(
+        return cls(
             film_miller=film_miller,
             substrate_miller=substrate_miller,
             strain=strain,
             von_mises_strain=von_mises_strain,
             elastic_energy=elastic_energy,
             ground_state_energy=ground_state_energy,
-            **{k: match_dict[k] for k in match_dict.keys() if not k.startswith("@")},
+            **{
+                k: getattr(match, k)
+                for k in [
+                    "film_sl_vectors",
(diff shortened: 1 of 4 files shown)


bernstei:
> This should fix #2195, which was a bug introduced in the interfaces reorganization where the substrate analyzer was constructing matches to return, but not actually returning them... DOH!
> 
> I've added a test to check for this in the future.

Are you sure it's #2195 ?  That seems to be about some file I/O bug.

mkhorton:
Thanks @shyamd !

Issue #2216, opened 2021-08-15
Wrong POSCAR and .cif file generated when using `Structure.to` method


755452800:
**Describe the bug**
Hi, I was using Pymatgen `Structure` object to write a POSCAR and cif file using `to` method, but the files generated were wrong when using cartesian coordinates.

**To Reproduce**
1. generate a pymatgen `Structure` object using cartesian coordinates.
```
from pymatgen.core import Structure

species = [13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 45.0, 7.0, 1.0]

[...]

PR #2217, opened 2021-08-15
fix issue #2216

755452800:
## Summary

Add a few lines code to exam the coords fed to `Structure` object are cartesian or not.
The setting of `coords_are_cartesian` was defaulted to `False` and problem may occur when feeding cartesian coords to `Structure`.

* Fix  issue #2216 

None

## TODO (if any)

I noticed there are several codes set `coords_are_cartesian` to `False` by default, but I'm not sure if they are properly set when cartesian coords are fed and may be examed first like this case.

## Checklist

in the pull request title.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -3151,6 +3151,13 @@ class Structure(IStructure, collections.abc.MutableSequence):
                 have to be the same length as the atomic species and
                 fractional_coords. Defaults to None for no properties.
         """
+        # To check whether the coords are cartesian or not
+        for coord in coords:
+            for i in coord:
+                if i > 1 or i < 0:
+                    coords_are_cartesian = True
+                    break   
+                    
         super().__init__(
             lattice,
             species,


shyuep:
This is not a valid check. Fractional coordinations can also be less than 0 and greater than 1, even if they map to the unit cell. This is frequently used when you are working with trajectories. The reason why the coordinates are defaulted to fractional for crystals is because this is by far the most common way coordinates are specified.

755452800:
Oh, really sorry for the mistake and the wrong pull request.

So is there any valid way to check the coords to be cartesian or not? Or it's just not worth to do it. But in some cases it may have issue like #2216

shyuep:
There is no way to check this. A set of coordinates can be cartesian or fractional. Of course, the likelihood is that small numbers mean fractional and large numbers are more likely to be cartesian. But this is not always the case. It is the responsibility of the user to be awre what coordinates they are entering.

755452800:
Thanks for answering!

PR #2230, opened 2021-08-31
Update bader_caller.py

nwinner:
Patch. The chgref option to the bader caller should be applicable to parsing cube files as well.


mkhorton:
Thanks @nwinner

PR #2231, opened 2021-08-31
indent list for `Composition.contains_element_type` documentation

CompRhys:
should fix a bug in the automatic documentation markup for `Composition.contains_element_type`


mkhorton:
Thanks @CompRhys !

PR #2233, opened 2021-09-02
Refactor `interface_reactions` module, add support for Plotly

mattmcdermott:
## Summary

* Add: support for Plotly: calling `plot()` now provides a `backend` option for choosing matplotlib or Plotly.
* Add:`GrandPotentialInterfacialReactivity` class to separate code based on the use of grand potential phase diagrams (this greatly decreases the complexity of the `__init__` args) and is a more inherited approach in general
* Add: `get_dataframe()` method for plotting possible reactions in a Pandas dataframe. Users would typically have to write/copy boilerplate code to do this each time
* Fix: Better matplotlib plotting, included annotations 
* Fix: Cleaner, more readable code

## Checklist

### DIFF
--- a/pymatgen/analysis/chempot_diagram.py
+++ b/pymatgen/analysis/chempot_diagram.py
@@ -32,10 +32,11 @@ from monty.json import MSONable
 from plotly.graph_objects import Scatter, Scatter3d, Mesh3d, Figure
 from scipy.spatial import ConvexHull, HalfspaceIntersection
 
-from pymatgen.analysis.phase_diagram import PDPlotter, PhaseDiagram, PDEntry
+from pymatgen.analysis.phase_diagram import PhaseDiagram, PDEntry
 from pymatgen.core.composition import Composition, Element
 from pymatgen.entries.computed_entries import ComputedEntry
 from pymatgen.util.coord import Simplex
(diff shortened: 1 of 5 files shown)


mattmcdermott:
@mkhorton My plan is to have this PR done by Wed, September 8, based on the previous timeline goal we set

mkhorton:
Fantastic, thank you for the update @mattmcdermott !

mattmcdermott:
Should be pretty much done; tests work on my laptop. Not sure if things are broken on Github...

mkhorton:
Thanks @mattmcdermott! I think that method is also available on `Composition` etc. directly (via the `Stringify` mixin), just an fyi.

mattmcdermott:
> Thanks @mattmcdermott! I think that method is also available on `Composition` etc. directly (via the `Stringify` mixin), just an fyi.

Ah true. The only downside is for compositions like `Composition("FeO2")`, it automatically puts "1" as a subscript:
e.g.,`Fe_1O_2`

mkhorton:
Ah, I guess it acts on the Composition directly rather than formula, fair enough.

mkhorton:
Merged. Thanks again @mattmcdermott, this is great!

PR #2241, opened 2021-09-07
improve Structure.from_str fmt doc string and type hint

janosh:


janosh:
Sorry about this hasty PR. I'd never looked into [contravariance](https://www.python.org/dev/peps/pep-0483/#covariance-and-contravariance) before. I still think there would be some value in having a more restrictive type hint (same for docs) but seems that would require use of `# type: ignore` on [line 2338](https://github.com/materialsproject/pymatgen/pull/2241/files#diff-5ea334b067925ce7079b569683cf7cc7ee088d0b19d62e4257f79cbb6aa73914R2338). If you disagree, feel free to close.

shyuep:
I think type hints are useful but we should not introduce them and then ignore them?

janosh:
The type hint would be for IDEs and users reading the code. It would only be ignored by linters, i.e. `mypy`. And I think the doc change in this PR is an improvement in any case.

mkhorton:
I like this change, I think we should re-open the PR.

However, *I agree type hints should not be ignored* -- can we fix this? We shouldn't merge it with `ignore` present. I'm not familiar with the contravariance issue, can you explain what the problem is here?

janosh:
Paraphrasing from [Wikipedia](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)), a typing rule is covariant if it preserves the type ordering from more specific to more generic and contravariant if it reverses this ordering.

That wasn't very intelligible to me. But it becomes easier to grasp, if you break it down for our specific case of inheritance and overriding a class method. A method is covariant in its _output_ (i.e. a more restrictive output type means the subclass method is a subtype of the parent method) but contravariant in its _input_, i.e. a more restrictive type rule for inputs means the subclass method becomes more generic and is no longer a subtype of the parent method.

In this specific case, `fmt: Literal["cif", "poscar", ...]` is a more specific type than `fmt: str` in the parent method so the subclass method is not a proper subtype of the base method and that's why `mypy` is complaining. Tbh, this seems like a face palm for OOP to me.

shyuep:
Personally, I think the type problem would be solved if we add an "Any" type to some of the type hints?

janosh:
> Personally, I think the type problem would be solved if we add an "Any" type to some of the type hints?

You mean set the type of `fmt` to `Any` in the parent method, i.e. `SiteCollection.from_str(fmt: Any)`? I think that would work.

shyuep:
I mean you can add proper type hints as what you have done, but allow an "Any" as a final option. That would still give people lots of information of the normal expected types (e.g., the literals) but at the same time provide enough flexibility if someone introduces a strange new type.

janosh:
Personally, I'd go with adding the `Any` to the parent method. Having the `Any` in the `Structure

shyuep:
This is fine.

janosh:
@shyuep Made the change. You prob need to reopen for it to sho up here.

janosh:
Unrelated question: Why import all of

```py
from pymatgen.io.atat import Mcsqs
from pymatgen.io.cif import CifParser
from pymatgen.io.cssr import Cssr
from pymatgen.io.xcrysden import XSF
```

in `from_str()` if you're only going to use 1 (or none) of them? Why not import inside the matching `if` statement?

janosh:
The remaining CI failures seem unrealted to this PR.

mkhorton:
The linting failure should have been fixed in the main branch, so please disregard.

Regarding the imports, I'm not sure it matters too much -- generally it's bad form to import within a method anyway instead of the top of the file (but I see why it's done here), I think they could get moved inside the respective `if` statement, so by all means make the change.

Perhaps the bigger issue with `Structure.from_str` is how it differs from `Structure.from_file` in non-obvious ways. I think this might be improved by abstracting a `_detect_fmt_from_filename()` function and then having `from_file` call `from_str` directly.

shyuep:
I would suggest deferring the changes other than the type hints. My preference is to make these `from_str` or `from_filename` into proper plugins. That way, if a person adds a new io class, `Structure.from_str` would automatically support the new fmt. This requires some inspection methods and probably an abstract class.

The reason why the imports are in the methods in this case is due to circular dependencies. E.g., Poscar imports Structure, so Structure cannot import Poscar. You can move them to the respective if statements and I have no objections to that.

mkhorton:
Ok, sounds sensible, let's defer any additional changes (aside from the `if`) for a later date.

janosh:
While were on the topic of types, how about adding type checking as a pre-commit hook. I noticed you're already using `pre-commit`. I can recommend the [`mypy mirror`](https://github.com/pre-commit/mirrors-mypy), use it on all my projects.

janosh:
Anything blocking this PR?

janosh:
@mkhorton The linting and test errors both appear unrelated to this PR.

PR #2247, opened 2021-09-11
Type hints for literal string kwargs

janosh:
Follow up to #2241. Here are a few more examples where `Literal` type hints make sense. I think there are ~80 more places throughout pymatgen. Happy to address more of them in this PR if you guys will accept.


mkhorton:
I think that would be great, no objections from me as long as any mypy failures are also dealt with, thank you!

PR #2257, opened 2021-10-08
InsertionElectrode bug fix & doc clarification

acrutt:
The InsertionElectrode.from_entries() method can use either ComputedEntries or ComputedStructureEntries - this has been added to the documentation. Setting strip_structures=True only works with ComputedStructureEntries so this has also been clarified.

In constructing the PhaseDiagram to compare the energies of the provided battery entries, the element energies were not always set high enough if there was a large energy difference between the empty host material and intercalated material. As a result, no intercalated entry was found to be stable and voltage pairs could not be formed. The factor for increasing the element_energy was set higher to 1e9 to avoid this problem.


acrutt:
Note for future reference on the appropriate value for the terminal element entries when constructing the phase diagram. This value cannot be set infinitely high otherwise there will be precision errors in constructing the ConvexHull. 

[http://www.qhull.org/html/qh-impre.html](http://www.qhull.org/html/qh-impre.html)

> Your input coordinates start with the same five or more digits (i.e., it is shifted relative to the origin). This reduces the available precision.

Therefore the terminal elements need to have an energy comparable to the provided battery material energies. 10 can be an appropriate value while 1e9 is too big. If the value is set too big (values greater than 1e6), pymatgen's tests will fail due to incorrect convex hull construction that results in misidentification of the stable battery entries.

mkhorton:
Thanks @acrutt and for providing additional context here

acrutt:
@mkhorton I believe this PR is ready to merge. Ultimately the changes are now limited to some documentation clarifications. Let me know if any other work is needed!

mkhorton:
Thanks!

yantar92:
FYI, I just ran into this. + 10 appears to be not good enough. I casually set it to +100 to fix voltage calculation on my system, but probably there should be smarter way than random constant.

PR #2261, opened 2021-10-13
Literal-types

janosh:
Follow up to #2247.

@CompRhys informed me that due to #2247, Pymatgen is no longer installable in Google Colab which is locked to Python 3.7 while `typing.Literal` was only added in Python 3.8.

There's a relatively recent but quiet issue tracking the ability to select different Python versions in Colab: https://github.com/googlecolab/colabtools/issues/2165

In the meantime though, I assume Pymatgen wants to keep supporting Google Colab. I think the best solution for that would be to protect `Literal` imports with

```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Literal
```

and then use the string form `"Literal"` in type hints. Happy to add those changes to this PR if that's acceptable.

### DIFF
--- a/pymatgen/analysis/interface_reactions.py
+++ b/pymatgen/analysis/interface_reactions.py
@@ -23,8 +23,9 @@ References:
 
 import json
 import os
+import sys
 import warnings
-from typing import List, Literal, Tuple, Union
+from typing import List, Tuple, Union
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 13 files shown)


mkhorton:
You can install `typing_extensions`, which is actually already a requirement of pymatgen if you're running in Python 3.7.

The correct way of resolving this is following this example:

https://github.com/materialsproject/pymatgen/blob/f97868791cb8f01aadf1eb3620d1df63fb0ff49a/pymatgen/io/qchem/inputs.py#L19-L22

Apologies, I should have mentioned this in your earlier PR but forgot.

PR #2263, opened 2021-10-16
Use all short-circuit rather than loop to make code more pythonic

CompRhys:
in-built all short circuits when checking a boolean list which means a for loop isn't needed here.

I am making this PR as I am looking at trying to get my `PatchedPhaseDiagram` ready to be merged in near future, I have several small QOL changes like this on that branch and think it would be easier if they are merged separately first to making reviewing the `PatchedPhaseDiagram` PR easier (or I can revert on ppd branch if there are issues).


mkhorton:
I'm fine with either, with a mild preference to breaking them out.

PR #2264, opened 2021-10-16
Remove method and function that are no longer used for entry equality

CompRhys:
Previously when dealing with entry equality we wrote these function to do a robust eq check. After these changes didn't fix the results @mkhorton and @mattmcdermott came up with a better solution for the `__hash__` and `__eq__` but these functions were not removed. They are not referenced in `pymatgen` beyond their definitions and so I removed them in my `ppd` branch for code hygiene and think that it might also be sensible to remove on `main`.


mkhorton:
Thanks!

PR #2278, opened 2021-10-23
update docstring of CutOffDictNN

ltalirz:
follow-up on @mkhorton 's comments


mkhorton:
Thanks! Sorry for delay here too

PR #2283, opened 2021-10-28
Add PatchedPhaseDiagram to allow for handling of Large Disjoint Chemical systems

CompRhys:
## Summary

Handler Class to allow for thermodynamic properties to be calculated across large chemically diverse phase diagrams whilst only generating relevant phase diagrams once. 

Designed to inherit as much as possible and be modular to allow additional functionality to be approximated if solutions can be devised.

```python
m = MPRester(MP_API_KEY)
# Materials project provides a python API via pymatgen
# API key obtained from https://materialsproject.org/open
# 

mp_data = m.get_entries(
    chemsys_formula_id_criteria={"nelements":{"$lt":3}},
    inc_structure=False,
)
```
```bash
100%|██████████| 21160/21160 [00:14<00:00, 1504.11it/s]
```

```python
%timeit ppd = PatchedPhaseDiagram(mp_data, verbose=True)
```
```bash
100%|██████████| 2127/2127 [00:12<00:00, 172.51it/s]
100%|██████████| 2127/2127 [00:12<00:00, 174.56it/s]
100%|██████████| 2127/2127 [00:12<00:00, 174.54it/s]
100%|██████████| 2127/2127 [00:12<00:00, 172.13it/s]
100%|██████████| 2127/2127 [00:12<00:00, 169.93it/s]
100%|██████████| 2127/2127 [00:12<00:00, 172.21it/s]
1 loop, best of 5: 26.2 s per loop
```

c.f. the equivalent `PhaseDiagram` is intractable.

When calling thermodynamic properties, e.g. `e_above_hull`, class first looks to see if sufficient pd exists in patch library, if not we use the `_get_slsqp_decomp` method to get the decomp on the minimal set of competing entries.

If I compare the `e_above_hull` I calculate there are 68 materials where they differ from the MP values. I think these are most likely due to manual fixes/rounding in MP (see below)

### TODO
- [ ] ~~Refactor to allow serialisation in a manner that mimics new `PhaseDiagram`~~


CompRhys:
Just to update this I am pretty sure there is no nice solution to the problem I outline re serialising to json when we have an object containing multiple references to other objects inside itself. Pickle has the correct behaviour in that it serialises a single copy and then keeps track of references for the copies.

As such I think this PR is okay to be merged provided you're happy with the serialisation and de-serialisation to/from json having to repeat the calculations of the hulls/patches.

JosephMontoya-TRI:
Very excited about this contribution - I've been using OQMD's implementation of GCLP-based phase diagrams to handle chemically diverse datasets, but would love to simplify my stack by using this instead.

mkhorton:
@montoyjh do you want to give a run-down of the pros/cons of the GCLP approach for the uninitiated?

@CompRhys only just saw this PR was marked as good-to-go. Will review shortly.

CompRhys:
> @montoyjh do you want to give a run-down of the pros/cons of the GCLP approach for the uninitiated?
> 
> @CompRhys only just saw this PR was marked as good-to-go. Will review sho

JosephMontoya-TRI:
Re: GCLP - it's a linear programming approach to energy minimization that does so for each queried composition, see [this paper from OQMD](https://onlinelibrary.wiley.com/doi/abs/10.1002/adma.200700843) and a [recent paper](https://chemrxiv.org/engage/api-gateway/chemrxiv/assets/orp/resource/item/629135424d8b597d2ea14f76/original/environmental-stability-of-crystals-why-be-greedy-when-you-can-be-exact.pdf) that describes a method that's basically the same a little more explicitly. I think if you want to come up with the stability of *any* composition in a large chemical space (say a 12-element composition in a 96 dimensional space) - GCLP scales better, but on a practical basis we only ever deal with <8 elements and I think it's easier to understand a strategy like the one that's in this PR - plus having the convex hulls of the subspaces is useful.

mkhorton:
Thanks @montoyjh :)

CompRhys:
We have an upfront cost to generate the ppd and then getting the decomposition is cheap for entries within a patch and for entries outside a patch I think the SLSQP I mentioned is just an algorithm to solve the GCLP as opposed to using gurobi to solve the LP as in the second paper.

mkhorton:
Thanks for this, it's excellent! Apologies for the delay, longer PRs always take a moment to review.

PR #2285, opened 2021-10-29
Fix oddly split strings and a few typos

janosh:
`"Lots of " "oddly split" " string in pymatgen"`, presumably as a result of `black` formatting.


mkhorton:
Thanks @janosh !

PR #2286, opened 2021-10-30
ElementBase.X drop NaN warning

janosh:
The warning about missing electronegativities is by far the most prevalent one in my use of pymatgen and it has never once been useful. Would be great to get rid of it.

It comes up all the time and usually repeated a dozen times or so when iterating over a dataframe column involving creation of new `Elements()`.

### DIFF
--- a/pymatgen/core/periodic_table.py
+++ b/pymatgen/core/periodic_table.py
@@ -211,9 +211,7 @@ class ElementBase(Enum):
             float: Electronegativity of element. If an element does not have an electronegativity,
                 NaN is returned.
         """
-        if "X" in self._data:
-            return self._data["X"]
-        return float("NaN")
+        return self._data.get("X", float("NaN"))
 
     @property
     def atomic_radius(self) -> Optional[FloatWithUnit]:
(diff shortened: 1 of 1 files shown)


shyuep:
No the warning needs to be there. BEcause a Nan is still a float, it can lead to strange errors if you use it in the wrong way. If you wish, you can set a proper warning type and on your end, you can capture all the warnings. I personally found it useful to know if I have elements that do not have electronegativities in my dataset.

janosh:
> BEcause a Nan is still a float, it can lead to strange errors if you use it in the wrong way.

Then it would be better to return `None` to avoid tripping up inexperienced users.

shyuep:
It used to return None. That caused more problems because Noke is a type that needs to be caught. Which means everyone who uses electro negativity needs to use try catch.

janosh:
Could also use an electronegativity scale that assigns values to all elements:

https://wikipedia.org/wiki/Electronegativity#Allen_electronegativity

shyuep:
You can of course add Allen electro negativity as a separate property. But most people's understanding of electronegativity refers to Pauling.

PR #2291, opened 2021-11-05
Replace-species

janosh:
`Structure().replace_species` doc string should make it clear that original structure is modified in place.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -459,9 +459,9 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
         for site in self.sites:
             del site.properties[property_name]
 
-    def replace_species(self, species_mapping: Dict[SpeciesLike, SpeciesLike]):
+    def replace_species(self, species_mapping: Dict[SpeciesLike, SpeciesLike]) -> None:
         """
-        Swap species.
+        Swap species. Note that this method modifies the structure in place.
 
         Args:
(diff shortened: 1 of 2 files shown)


pymatgen/core/structure.py
```diff
@@ -1045,28 +1044,30 @@ def matches(self, other, anonymous=False, **kwargs):
...
-            return True
-        if other is None:
-            return False
-        if len(self) != len(other):
-            return False
-        if self.lattice != other.lattice:
-            return False
-        for site in self:
-            if site not in other:
+    def __eq__(self, other: object) -> bool:
```
shyuep, line 1047: Please do not make this change. I am unclear why your PR is mentioning a replace method but you are modifying the way structure checks for equality.

janosh: Sorry, this wasn't meant to be bunched in with the other small changes in this PR. I'll revert it.

In general, updating all Pymatgen rich comparison functions to send `NotImplemented` follows recommended practice. See https://docs.python.org/3/library/constants.html#NotImplemented. So maybe worth tackling in a separate PR.

janosh: The PR mentioned the `Structure.replace_species()` method cause my original intent was just to update its doc string to make it clearer it makes in-place modifications.

mkhorton:
Thanks @janosh, I see you reverted the other commit so going to merge this

PR #2297, opened 2021-11-14
Add warning if LASPH != True for meta-GGA/hybrid/vdW/+U

arosen93:
This PR implements a simple `BadInputSetWarning` if `LASPH = False` (or is not set) yet the user is running a meta-GGA/hybrid/DFT+U/vdW-DF calculation.


mkhorton:
Thanks @arosen93!

PR #2301, opened 2021-11-20
Rosen spin

arosen93:
Currently, the handling of the MAGMOM INCAR flag in `pymatgen.io.vasp.sets` could be made clearer. This became apparent to me when trying to modify MAGMOM in an atomate workflow, which did not at all do what I expected it to do (and led to some confusing errors).

1.  The data type for the "MAGMOM" setting is not immediately obvious. In VASP, MAGMOM is given in the format "1\*5.0 1\*5.0" or "2\*5.0". Naively, one might think then to supply {"MAGMOM": [5.0, 5.0]} or maybe even {"MAGMOM": "2*5.0"}, neither of which work appropriately with pymatgen.io.vasp.sets.incar(), which expects something like {"MAGMOM": {"Fe": 5"}}. An error is now raised if MAGMOM is not a dict, and the user is appropriately notified that this is an element-specific flag, not a site-specific flag like in the typical VASP INCAR.

2. The `constrain_total_magmom` flag, when set to `True`, sums up the magnetic moments on each site and sets that as NUPDOWN. When pulling from the MP sets, this is not a concern because the default magnetic moments are integers (except for 0.6 values that are floored to 0 for low-spin). However, more generally, this is almost never desirable. [NUPDOWN](https://www.vasp.at/wiki/index.php/NUPDOWN) is meant to specify the spin multiplet, i.e. 2*S where S is the total spin angular momentum, which is an integer value. While NUPDOWN technically can take non-integer values, that is [not what NUPDOWN is meant to do](https://www.vasp.at/forum/viewtopic.php?f=4&t=2910).  If the user specified a custom MAGMOM, then there is no guarantee the summed values are integers. I have therefore added a warning if `round(sum(magmoms)) != sum(magmoms)`.

3. The warning about Co being initialized as low-spin could be made clearer. Currently, Pymatgen prints out this warning whenever MAGMOM has an entry for Co and there is no site-specific `magmom` or `spin`. If one were to change the config to have high-spin Co or modify MAGMOM directly (e.g. `{"MAGMOM": {"Co":5}}`), the warning would still appear despite the fact that it is no longer relevant. I have clarified the wording in the warning.

4. The documentation says that if the symbol is not in the config file, VASP's default of 0.6 is used. The [VASP default is 1.0](https://www.vasp.at/wiki/index.php/MAGMOM); the Pymatgen default is 0.6. I've clarified the phrasing.

This is WIP because I need to add a test or two.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -282,8 +282,8 @@ class DictSet(VaspInputSet):
     structure and the configuration settings. The order in which the magmom is
     determined is as follows:
 
-    1. If the site itself has a magmom setting, that is used.
-    2. If the species on the site has a spin setting, that is used.
+    1. If the site itself has a magmom setting (`site.magmom`), that is used.
+    2. If the species on the site has a spin setting (`site.spin`), that is used.
     3. If the species itself has a particular setting in the config file, that
        is used, e.g., Mn3+ may have a different magmom than Mn4+.
     4. Lastly, the element symbol itself is checked in the config file. If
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @arosen93!

For the PR itself, if WIP, please either put [WIP] in the title or you can explicitly "convert to draft" in the top-right.

For the issue, this looks like an important thing to clarify. I did also want to make sure you knew about the [`Magmom` class](https://pymatgen.org/pymatgen.electronic_structure.core.html?#pymatgen.electronic_structure.core.Magmom) which was my own old attempt at starting to unify magmom representations, specifically between non-collinear and collinear cases.

In terms of handling of magmom, we've ended up in a slightly awkward situation:

1. It has been more typical to use the `magmom` site property, however this does not allow the representation of disordered magnetic materials.

shyuep:
Thanks.

PR #2303, opened 2021-11-22
More robust smart fermi method

utf:
## Summary

VASP 6.1+ sometimes puts the Fermi level just inside a valence or conduction band. This is related to a change in VASP tetrahedron which is described here: https://www.vasp.at/forum/viewtopic.php?f=4&t=17981

There is already logic inside pymatgen to try and catch this bug, unfortunately it is not reliable and after some discussion with @mkhorton, we are not sure it can ever work.

I've made a new implementation that checks whether the Fermi level is just inside of band (controlled by a tolerance, by default set to 1 meV) and next to a band gap. If so, this is likely an error on VASPs part. In my testing I found this method much more reliable. 

I'll note this is not enabled by default and is only included when calling `Vasprun.get_band_structure(efermi="smart")`.

@rkingsbury this may be relevant to you.


mkhorton:
Thanks @utf, this is very helpful. I followed up with the VASP folks on this, see [here](https://www.vasp.at/forum/viewtopic.php?p=20686). It sounds hopeful that this might be addressed in a future version.

Issue #2307, opened 2021-11-25
VASP 6.2.1 OUTCAR not properly parsed for run stats


arosen93:
**Describe the bug**
When parsing the attached OUTCAR (from VASP 6.2.1) with the current version of Pymatgen, the number of cores is not output correctly because instead of the OUTCAR reading `running on 64` it reads `running 64`. (The average memory used here is `N/A` by VASP, so `None` is appropriate.)
```
{'Average memory used (kb)': None,
 'Maximum memory used (kb)': 212968.0,
 'Elapsed time (sec)': 148.577,
 'System time (sec)': 6.188,
 'User time (sec)': 142.096,
 'Total CPU time used (sec)': 148.284,
 'cores': 'mpi-ranks,'}
```

**To Reproduce**
```python
from pymatgen.io.vasp import Outcar
out = Outcar('OUTCAR.txt')
print(out.run_stats)
```
[OUTCAR.txt](https://github.com/materialsproject/pymatgen/files/7600610/OUTCAR.txt)

PR #2308, opened 2021-11-25
Fix VASP 6.2.0+ # of cores in run_stats

arosen93:
Addresses #2307 wherein the number of cores was not being parsed properly for a VASP 6.2.1 OUTCAR file. I have added a test for this new VASP file.

## Summary
- Correctly parses number of cores from VASP 6.2.0+ files
- Stores the number of cores as an `int` instead of `str`


arosen93:
@mkhorton, something appears to be up with the tests both here and in #2312. Here, for example, the tests passed successfully earlier. All I did was squash my commits, and now one of the tests simply stalls. Same thing but for a different test in #2312.

mkhorton:
Thanks @arosen93 !

PR #2310, opened 2021-11-25
fix write Trajectory XDATACAR with variable lattice

gpetretto:
## Summary

In the current implementation of the write_Xdatcar in Trajectory the lattice is either written only once if `constan_lattice==True` or never if `constan_lattice==False`. Fix to write it multiple times if the lattice is not constant.
Added a specific check in the tests.


mkhorton:
Many thanks @gpetretto

mkhorton:
Many thanks @gpetretto

PR #2312, opened 2021-11-28
Clearer handling of magmom flag

arosen93:
### Clearer Handling of Magnetic Moments
I have made a few additional clarifications to `pymatgen.io.vasp.sets`, primarily with regards to the handling of magnetic moments as a follow-up to #2301:
1. Clearer docstrings. This is rather self-explanatory.
2. Only print the warning about low magnetic moments on Co if the value is <= 1. If someone were to modify the default config to have high-spin Co, there is no longer a need for this warning. I had done this, and I found it somewhat bothersome.
3. If the user provides an integer for the magnetic moment, e.g. `site["magmom"] = 1`, then convert this to a float when `get_valid_magmom_struct()` is run. Previously, Pymatgen would crash when creating an `Incar()` object if the structure had an integer magnetic moment due to the lines below. https://github.com/materialsproject/pymatgen/blob/dd3737011e76520da1347d5db75db3a3f87e520f/pymatgen/io/vasp/sets.py#L3102-L3106

### Slight Adjustment to Handling of LASPH Warning
4. Extremely minor, but as a follow-up to #2297, I now check the `Incar()` object (rather than the `settings` variable) to see if `LASPH` is present. This is more robust.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -589,7 +589,7 @@ class DictSet(VaspInputSet):
 
         if self.constrain_total_magmom:
             nupdown = sum(mag if abs(mag) > 0.6 else 0 for mag in incar["MAGMOM"])
-            if nupdown != round(nupdown):
+            if abs(nupdown - round(nupdown)) > 1e-5:
                 warnings.warn(
                     "constrain_total_magmom was set to True, but the sum of MAGMOM "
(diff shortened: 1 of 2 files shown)


arosen93:
@mkhorton, thoughts on the handling of blank (omitted) magmom values in a structure (see above)? Keep as is? Change? Indifferent?

arosen93:

mkhorton:
Thanks @arosen93 !

PR #2316, opened 2021-12-03
Add missing elements to MPScanRelaxSet PBE .54 potentials

arosen93:
@rkingsbury and @mkhorton: Some elements that are currently in the VASP PAW .54 PBE library have defined pseudopotentials but aren't in the `MPScanRelaxSet.yaml` file because they never existed in the original PAW PBE set used in the GGA calculations on MP. This change won't influence anything in your R2SCAN workflow because: a) The jobs would have never run with these elements since no pseudopotentials were defined; b) none of these elements exist on MP right now because of the aforementioned statement.

Most of the elements I added have only one PAW PBE potential to use. In the case of multiple, I used the VASP-recommended set.

I know this breaks the hash for `MPScanRelaxSet.yaml`. It would need your blessing.

### DIFF
--- a/pymatgen/io/vasp/MPSCANRelaxSet.yaml
+++ b/pymatgen/io/vasp/MPSCANRelaxSet.yaml
@@ -29,8 +29,10 @@ POTCAR:
   Ac: Ac
   Ag: Ag
   Al: Al
+  Am: Am
   Ar: Ar
   As: As
+  At: At
   Au: Au
   B: B
   Ba: Ba_sv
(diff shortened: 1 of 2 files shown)


mkhorton:
Since these are all additions not changes, they have my blessing!

Open to changes too but would be more discussion required.

rkingsbury:
As long as it doesn't change the workflow I have no objections either. Thanks for the docstring cleanup, too @arosen93 !

arosen93:
Great, thanks! I updated the hash accordingly.

mkhorton:
Thanks both!

I see an unrelated test failure on this PR that will have to be addressed separately so will go ahead and merge.

PR #2320, opened 2021-12-08
Absorption

yang-ruoxi:
## Summary
* Added optical absorption input set for independent-particle approximation (IPA) and random-phase approximation (RPA), expected to be used with AbsorptionFW in atomate 

* Added the AbsorptionSet.yaml for the default input set (IPA)

* Two modes are possible to choose from: IPA or RPA
   * If IPA is chosen,  it is assumed a previous 'ALGO = Normal' calculation needs to be done with the WAVECAR saved. This IPA calculation will use the previous WAVECAR, and set 'ALGO = Exact' and 'LOPTICS = TRUE' with a large number of NBANDS. One obtains the macroscopic frequency dependent dielectrics. Normally only `nbands_factor` is expected to be modified rather than directly modifying the `nbands`. 
   * If RPA is chosen, it is assumed an IPA calculation has to be done before hand with WAVECAR and WAVEDER saved.


mkhorton:
Thanks for opening this PR @yang-ruoxi!

Since you're a new contributor, I have to manually approve the automatic tests. These should now be running.

mkhorton:
Overall this looks good!

The main thing this is lacking right now is tests. You may be able to copy a test from the existing input sets (see `test_sets.py`) as a starting point. It would just need to test the main features of the input set, e.g. the `IPA` and `RPA` modes.

The other comment would be for `MPAbsorptionSet.yaml`, if it's only a few edits to the `MPRelaxSet.yaml` it would be better to make those changes inside `MPAbsorptionSet.__init__` so that you can inherit all the other settings from `MPRelaxSet.yaml`. The benefit of this approach is that if `MPRelaxSet.yaml` _does_ change in future, `MPAbsorptionSet` will automatically remain consistent.

yang-ruoxi:
* added AbsorptionSetTest and the corresponding test files

mkhorton:
Thanks @yang-ruoxi !

PR #2323, opened 2021-12-17
Add new automatic kpoint scheme that normalizes by lattice lengths

arosen93:
## Summary
This PR creates a new automatic k-point generation scheme, `automatic_density_by_lengths`, which allows the user to specify a density of k-points in each dimension (rather than just for the entire volume). This is particularly important for slab calculations, because you almost always want 1 k-point in the vacuum dimension but don't want the vacuum space volume to influence your k-point generation otherwise.

For instance, doing `automatic_density_by_lengths(struct, [50, 50, 1])` would specify a k-point density of 50/a x 50/b x 1/c, with each rounded up to the next integer.

This is a more general implementation of the k-point scheme used in `MVLSlabSet` (reproduced below) and can be found in several papers from the Persson group, such as [this one](https://www.nature.com/articles/s41467-019-08356-1#Sec7).
https://github.com/materialsproject/pymatgen/blob/fcb84a1e129da3a3010b9f6b9cf18beddf39f245/pymatgen/io/vasp/sets.py#L2213-L2244

Tagging @oxana-a, @mhsiron in case they're curious.

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -473,7 +473,7 @@ def bader_analysis_from_path(path, suffix=""):
 
     This method will:
 
-    1. Look for files CHGCAR, AECAR0, AECAR2, POTCAR or their gzipped
+    1. Look for files CHGCAR, AECCAR0, AECCAR2, POTCAR or their gzipped
     counterparts.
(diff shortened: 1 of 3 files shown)


arosen93:

mkhorton:
Many thanks for this @arosen93! Will be investigating failing tests shortly

PR #2324, opened 2021-12-18
Pass magmoms and selective dynamics between ASE Atoms and Pymatgen Structure in ASEAtomsAdaptor

arosen93:
Currently, the `pymatgen.io.ase.ASEAtomsAdaptor` does not pass the magnetic moments or selective dynamics tags between the ASE `Atoms` and Pymatgen `Structure` objects (in either direction). As an example of where this came up for me recently, I was using the `mp_api` to fetch `Structure` objects from MP. When I convert them to ASE `Atoms` objects, they lose their magnetic moment site properties. It's not too hard to fix, so here it is.

This PR enables the passing back and forth of these site properties. Tests have been added. I also fixed the test `POSCAR` file, which had the wrong formula string at the top. This caused problems when comparing the ASE `Atoms` object and Pymatgen `Structure` object.


arosen93:

mkhorton:
Thanks!

Issue #2326, opened 2021-12-26
get_conventional_standard_structure strips the structure of all site properties


arosen93:
When using `pymatgen.symmetry.analyzer.SpacegroupAnalyzer`'s `get_conventional_standard_structure()`, all site properties in the pymatgen structure object (e.g. magnetic moments) are stripped.

Importantly, this also means that when using the MP API, doing `mpr.get_structure_by_material_id(mp_id, conventional_unit_cell=True)` does not return a structure with site properties.

Note: I think the issue is that when `get_refined_structure()` is called in `get_conventional_standard_structure()`, the site properties aren't re-introduced after the call to spglib [here](https://github.com/materialsproject/pymatgen/blob/2aeb9845102cae9bbc466a52d71f9df3ca28cad4/pymatgen/symmetry/analyzer.py#L305).
```python
from pymatgen.core import Structure
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer

# Build structure
struct = Structure(
    lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],
    species=["Mg", "O"],
    coords=[[0, 0, 0], [0.5, 0.5, 0.5]],
)
struct.add_site_property("magmom",[3.14]*len(struct))

# Test conventional cell transform
conventional_struct = SpacegroupAnalyzer(struct).get_conventional_standard_structure()
print(conventional_struct.site_properties.get("magmom",None))
# ^ The above returns None
```

arosen93:
Closing because of the discussion in #2330.

PR #2330, opened 2022-01-04
Keep site properties after calling spglib in get_refined_structure() and find_primitive()

arosen93:
In `pymatgen.symmetry.analyzer.SpacegroupAnalyzer`, the `get_refined_structure()` and `find_primitive()` functions both strip Pymatgen structure objects of all their site properties, including magnetic moments. This PR carries over the site properties after calling `spglib`. Test have been added.

Addresses #2326.

### DIFF
--- a/pymatgen/symmetry/analyzer.py
+++ b/pymatgen/symmetry/analyzer.py
@@ -55,6 +55,7 @@ class SpacegroupAnalyzer:
         self._symprec = symprec
         self._angle_tol = angle_tolerance
         self._structure = structure
+        self._siteprops = structure.site_properties
         latt = structure.lattice.matrix
         positions = structure.frac_coords
         unique_species = []
@@ -69,16 +70,13 @@ class SpacegroupAnalyzer:
                 unique_species.append(species)
                 zs.extend([len(unique_species)] * len(tuple(g)))
 
-        if "magmom" in structure.site_properties:
-            for site in structure:
-                if hasattr(site, "magmom"):
-                    magmoms.append(site.magmom)
-                elif site.is_ordered and hasattr(site.specie, "spin"):
-                    magmoms.append(site.specie.spin)
-            if len(magmoms) != len(structure):
-                raise ValueError("Number of site with magmoms does not match number of sites")
-        else:
-            magmoms = None
+        for site in structure:
+            if hasattr(site, "magmom"):
+                magmoms.append(site.magmom)
+            elif site.is_ordered and hasattr(site.specie, "spin"):
+                magmoms.append(site.specie.spin)
+            else:
+                magmoms.append(0.0)  # needed for spglib
 
         self._unique_species = unique_species
         self._numbers = zs
@@ -114,9 +112,7 @@ class SpacegroupAnalyzer:
             SpacgroupOperations object.
         """
         return SpacegroupOperations(
-            self.get_space_group_symbol(),
-            self.get_space_group_number(),
-            self.get_symmetry_operations(),
+            self.get_space_group_symbol(), self.get_space_group_number(), self.get_symmetry_operations(),
         )
 
     def get_hall(self):
@@ -285,9 +281,7 @@ class SpacegroupAnalyzer:
         """
         ds = self.get_symmetry_dataset()
         sg = SpacegroupOperations(
-            self.get_space_group_symbol(),
-            self.get_space_group_number(),
-            self.get_symmetry_operations(),
+            self.get_space_group_symbol(), self.get_space_group_number(), self.get_symmetry_operations(),
         )
         return SymmetrizedStructure(self._structure, sg, ds["equivalent_atoms"], ds["wyckoffs"])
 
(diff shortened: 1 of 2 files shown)


mkhorton:
How does this work with magnetically ordered structures @arosen93? e.g. the primitive structure may not be large enough to retain the magnetic ordering from the input structure

arosen93:
@mkhorton:

In this PR, the site properties are carried over to any symmetrically equivalent site in the refined or primitive structure. However, the question you raised about longer range magnetic ordering is interesting to consider. I believe this is accounted for in `spglib`. 

As you can see below, the `cell` variable is populated with the magnetic moment information:
https://github.com/materialsproject/pymatgen/blob/4ae7d1b8013274b6848f09474839704890f4014a/pymatgen/symmetry/analyzer.py#L83

When spglib is called, it should take into account the magnetic information when generating the primitive or refined unit cell.
https://github.com/spglib/spglib/blob/0d2a1a50cdbd3413931b1376607b0fd1d83505ae/python/spglib/spglib.py#L93-L97

arosen93:
@mkhorton, ah, do you mean the magnetic ordering site property specifically? I'm not too familiar with that one. Perhaps that shouldn't necessarily be carried over after the call to `spglib`. Instead of transferring all site properties, we could do just the `magmom` and the `selective_dynamics` or we could just strip out any of the magnetic ordering site properties as a note of precaution. Do you have a preferred approach?

mkhorton:
No, I do mean the `magmom` site property.

e.g., consider a cell with two magnetic sites, one with a magmom +5, another with a magmom -5. Calling `get_primitive()` may return a cell with only a single magnetic site, but with what magmom value?

arosen93:
I was under the impression that `spglib` would not return a primitive cell in this case because of the provided magnetic character.  This is perhaps worth a direct investigation to confirm. I'll put it on the to-do list.

mkhorton:
My understanding is that spglib's treatment of magnetism is fairly basic. I believe there's some discussion on the spglib mailing list about this.

For our case, if we respect the magnetic moments (magmom) but pass them along as-is, but they're noisy (as they would be from VASP), then it would mean the returned primitive structure is likely incorrect. It's a tricky situation.

arosen93:
I see. Yes, that does add some complications to think about. 

If we focus on `get_refined_structure()`, which itself is used in `get_conventional_standard_structure()`, does the same concern apply? For instance, if I have a one atom primitive cell with a magnetic moment of 5 and then make a conventional cell that has two atoms, they both would have magnetic moments of 5. Of course, in reality, the real magmoms might be [-5, 5], but that's a separate issue. I don't think there's any information loss this way, right? There's only information less when making the primitive? If so, then maybe it makes sense to only pass properties for the `get_refined_structure()`, which in fact is the only one of those two that I was using anyway.

[Correct me if I'm wrong, of course. Trying to think about this visually].

mkhorton:
Yes, it's not immediately obvious to me what the correct answer is here. For the refined structure, unfortunately the same logic applies I think (eg if getting the conventional crystallographic structure of a magnetically-ordered supercell, AFM NiO is the good test case here).

arosen93:
Ah, you're totally right. Good example. I don't think there is a satisfying solution in the end. Feel free to close this if you wish! I think it's somewhat of an ill-defined problem, unfortunately.

mkhorton:
Ok, I'll close for now, but I am in favor of addressing this if we can find a good solution (and include appropriate tests for some of these tricky cases). I don't think you're the only person who's been tripped up by this.

mkhorton:
Also my apologies for the work spent on this PR since we're not merging, but hopefully it can help someone pick up from here in the future.

arosen93:
Not at a problem at all! You brought up some important points, and it didn't take too long to do anyway.

arosen93:
@mkhorton: This has still been in the back of my mind, and I wanted to ask -- would a possible middle-ground solution be to provide an optional keyword argument that keeps the site magmoms where they are (just based on whatever spglib decides to do with the transformation)? You're right that there are edge-cases where we don't want to do this in general, but there are also a lot of systems where it would be perfectly fine.

The reason why I think something like this would be helpful is that if one is screening a large number of materials (e.g. from MP), a common approach is to take the converged magnetic moments and use those as the MAGMOM guess for downstream calculations. That's not possible if you need to convert the MP unit cell to a conventional cell in your workflow. The best option is to recalculate magnetic moments (naively with some trivial spin initialization or more exhaustively using the magnetic moment workflow), but this is time-consuming. A middle-ground solution, which is imperfect, is to just keep the magnetic moments on the atoms as they are from spglib. Sure, for antiferromagnetic NiO, it poses a problem. But for ferromagnetic bulk Fe, for example, it's perfectly fine.

mkhorton:
Sure, I think that would be reasonable. Let's just make sure the docstring is explicit to warn users about caveats.

arosen93:
Sounds good! Thanks for the very helpful discussion.

PR #2336, opened 2022-01-10
Mixing Scheme

rkingsbury:
## Summary

Implementation of the SCAN / GGA(+U) mixing scheme. Mixing scheme code lives in the new file `mixing_scheme.py` and is implemented as a `Compatibility` class.

In addition to the core mixing scheme, this PR contains the following minor supporting changes:

* Changes to `MateraialsProjectAqueousCompatibility` that facilitate building Pourbaix diagrams with the mixing scheme. Previously, all H2 polymorphs were assigned the same energy by this class. Changes in this PR preserve energy differences among different polymorphs
* Additional kwargs to adjust the style of plots returned by `PourbaixPlotter`
* `EntrySet`: new methods `group_entries_by_composition`, `remove_non_ground_states`, and `is_ground_state`
* `MaterialsProjectCompatibility`: ensure that legacy corrections are only applied to GGA or GGA+U entries


rkingsbury:
@mkhorton could you please re-trigger the tests? It appears the only failure is coming from XAS and if I understood you correctly, that should be fixed now?

```
___________________________ XASTest.test_stitch_l23 ____________________________

self = <xas.tests.test_spectrum.XASTest testMethod=test_stitch_l23>

    def test_stitch_l23(self):
        self.l2_xanes.y[0] = 0.1
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            XAS.stitch(self.l2_xanes, self.l3_xanes, 100, mode="L23")
>           self.assertEqual(len(w), 6)
E           AssertionError: 5 != 6

pymatgen/analysis/xas/tests/test_spectrum.py:105: AssertionError
```

mkhorton:
I've re-triggered the tests, but may be worth merging from main branch again just in case. I see you did this yesterday.

rkingsbury:
Thanks! I did a git pull right before posting the comment today, but there
appeared to be no changes.

mkhorton:
I have not been able to reproduce this failing test, and other PRs are passing successfully. Can you try merging from the main branch again?

rkingsbury:
Done @mkhorton ; thanks for investigating. I can reproduce the failure locally; it only occurs on this branch and not on my local main branch. The question now is why? I can investigate but am not very familiar with the XAS code. Do you have any thoughts on why/how `XAS.stitch` would be interacting with `ComputedEntry`?

rkingsbury:
> Done @mkhorton ; thanks for investigating. I can reproduce the failure locally; it only occurs on this branch and not on my local main branch. The question now is why? I can investigate but am not familiar with the XAS code. Do you have any thoughts on why/how `XAS.stitch` would be interacting with `ComputedEntry`?

Nevermind, found the problem; it was the changes in `StructureMatcher` (replacing np.bool and np.int with bool and int). Not sure why, but I reverted those and the test passes locally.

pymatgen/entries/compatibility.py
```diff
@@ -888,7 +910,7 @@ def __init__(
             self.u_corrections = {}
             self.u_errors = {}
 
-    def get_adjustments(self, entry: Union[ComputedEntry, ComputedStructureEntry]):
```
mkhorton, line 891: Noting the type hint is removed here, was there a reason?

pymatgen/analysis/pourbaix_diagram.py
```diff
@@ -1030,7 +1030,16 @@ def show(self, *args, **kwargs):
...
         plt.show()
 
-    def get_pourbaix_plot(self, limits=None, title="", label_domains=True, plt=None):
+    def get_pourbaix_plot(
+        self,
+        limits=None,
+        title="",
+        label_domains=True,
+        label_fontsize=20,
+        show_water_lines=True,
```
mkhorton, line 1039: Thanks for adding this!

mkhorton (approved):

pymatgen/entries/compatibility.py
rkingsbury: Nope, I think that was probably an accident 🙃. Restored.

arosen93:
> Nevermind, found the problem; it was the changes in `StructureMatcher` (replacing np.bool and np.int with bool and int). Not sure why, but I reverted those and the test passes locally.

There are a bunch of pymatgen warnings about `np.bool` and `np.int` being deprecated that are pretty annoying. I tried to change them all over in pymatgen the other day but gave up...

mkhorton:
Yes--how a type hint change could cause a test failure in XAS I have no idea too. Anyway, separate issue to this PR which looks good, thank you!

arosen93:
CONGRATS, @rkingsbury!!!!!!

rkingsbury:
> Yes--how a type hint change could cause a test failure in XAS I have no idea too. Anyway, separate issue to this PR which looks good, thank you!

Thanks @mkhorton ! To clarify, it wasn't the type hint but rather the change from `np.bool` and `np.int` to `bool` and `int`, respectively, in `StructureMatcher` that caused the XAS failure. Maybe a floating point precision issue or something? Not sure; I just know there are a ton of warnings about using the numpy types in pymatgen as @arosen93 said.

mkhorton:
Thanks for the excellent test coverage for new features here too, definitely an example :)

rkingsbury:
Ahhh, nothing like merging a big PR on a Friday afternoon to close out the workweek 😄 ☺️

Issue #2327, opened 2021-12-28
Group numbers for the lanthanides and actinides are strange


arosen93:
When accessing the `.group` property of a `specie` that is a lanthanide or actinide, the returned group number is quite unconventional. For instance, the group number for Yb is 16. For Lu it's 17. While there is certainly debate to be had around the group numbers for the lanthanides and actinides, I don't believe this choice would ever be considered one of the [conventional options](https://www.chemeurope.com/en/encyclopedia/Group_number_of_lanthanides_and_actinides.html). 

```python
from pymatgen.core import Structure
struct = Structure(lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],species=["Yb", "O"],coords=[[0, 0, 0], [0.5, 0.5, 0.5]])
print(struct[0].specie.group)
# ^outputs 16
```

shyuep:
Thanks. Feel free to propose a conventional and implement a change. For me, group numbers don't really make any kind of sense of the TM and rare earth series. If you are planning to use it for ML purposes, I would say group number is probably one of the worst choices you can use. I for one don't believe the large body of work out there that utilizes this kind of features for ML.

arosen93:
I definitely agree with you re: group number being used as a descriptor for ML. Besides, it's often highly correlated with many other properties anyway.

In this case, I wasn't even thinking of ML. If one wanted to query a database of structures to find all the entries containing halogens, one might write a query that looks for group 17 elements. But Lu and Lr would unexpectedly come up too. There are of course better ways to do the query, but we should make it easy on the user.

To back up what I'm going to suggest, here are a few points:
- IUPAC understandably makes no explicit statement on the group number.
- Currently, in pymatgen La and Ac are group 3 whereas Lu and Lr are group 17. However, Lu and Lr are more group-3-like than La and Ac, as argued [here](https://pubs.acs.org/doi/10.1021/ed059p634), so this is a point that should be resolved in a proposed change.
- The lanthanoids and actinoids share common chemistry and should, in my view, not have different group numbers. 

With the above points, the only solutions I can think of are:
1. All lanthanoids and actinoids are in group 3.
2. All lanthanoids and actinoids are in a ficticious group 19, which would be mentioned in the documentation.
3. All lanthanoids and actinoids are not in group 3 but have `None` as the group number. This will likely pose a problem for down-stream ML tasks, and I do not advise it.
[...]

PR #2343, opened 2022-01-12
Update row and group number for lanthanoids/actinoids

arosen93:
## Purpose
This PR addresses #2327 wherein the group number for lanthanoids and actinoids was set to unconventional values of 3-17 depending on the element. The row was also set to the fictional values of 8 and 9, which is equally unconventional (although convenient for plotting).

## Changes
- The group number for all lanthanoids and actinoids are now 3 as a matter of convention. Usually, a value of 3 or none is accepted. I chose 3 primarily to avoid `None`.
- The row number for all lanthanoids and actinoids is now 6 and 7, respectively. This is generally accepted.
- The documentation for `pymatgen.core.periodic_table.ElementBase.from_row_and_group` has to be modified accordingly, as there is no longer a 1:1 map between row+group and element. Here, the old convention for row and group (i.e. row 8 and 9 for lanthanoids/actinoids and groups 3-17) should be used. This also means the function does not change in its inputs/outputs compared to before. 
- I have added tests for row and group attributes as well as `from_row_and_group()`, none of which had tests to the best of my knowledge.

## Additional Checks
Since this change influences the way the periodic table is represented, I confirmed that the `print_periodic_table()` function still functions. Indeed, it does:
```python 
from pymaten.core.periodic_tale import ElementBase
ElementBase.print_periodic_table()
H                                                                   He
Li  Be                                          B   C   N   O   F   Ne
Na  Mg                                          Al  Si  P   S   Cl  Ar
K   Ca  Sc  Ti  V   Cr  Mn  Fe  Co  Ni  Cu  Zn  Ga  Ge  As  Se  Br  Kr
Rb  Sr  Y   Zr  Nb  Mo  Tc  Ru  Rh  Pd  Ag  Cd  In  Sn  Sb  Te  I   Xe
Cs  Ba      Hf  Ta  W   Re  Os  Ir  Pt  Au  Hg  Tl  Pb  Bi  Po  At  Rn
Fr  Ra      Rf  Db  Sg  Bh  Hs  Mt  Ds  Rg  Cn  Nh  Fl  Mc  Lv  Ts  Og
        La  Ce  Pr  Nd  Pm  Sm  Eu  Gd  Tb  Dy  Ho  Er  Tm  Yb  Lu
        Ac  Th  Pa  U   Np  Pu  Am  Cm  Bk  Cf  Es  Fm  Md  No  Lr
```

I also ran `from pymatgen.util.plotting.periodic_table_heatmap` on a dataset of my own and checked that it stayed the same before and after the proposed PR.


arosen93:
Before accepting this PR, I ask that one of the maintainers please give it a look to see if there are any obvious spots where this could have unintended side-effects (like the periodic table plotting, which is specifically addressed).

pymatgen/core/periodic_table.py
```diff
@@ -601,20 +601,38 @@ def from_Z(z: int) -> "Element":
         raise ValueError("No element with this atomic number %s" % z)
 
     @staticmethod
-    def from_row_and_group(row: int, group: int) -> "Element":
+    def from_row_and_group(pseudo_row: int, pseudo_group: int) -> "Element":
```
mkhorton, line 604: This is a breaking change. Can we keep the old argument names? Clarifying in the docstring as you have should be sufficient here. I appreciate the numbers returned are also a breaking change, but no need to introduce an API change on top of it.

Otherwise this PR looks great, thanks for addressing this!

arosen93: Definitely can revert that! Since it they were positional args, I wasn't sure that it mattered much, but you're right that it should be kept the same and just have the docstrings be clear. Will get on that.

mkhorton: Positional args can also optionally be used as keyword arguments -- actually useful, if you want to be very explicit and notified of API changes (which is an argument _against_ keeping them the same actually). I sometimes do it in my own code for readability.

arosen93: Also, if you're like me, you often just forget they're optional keywords and specify them anyway sometimes...

arosen93:
@mkhorton, should be good now. Thanks for the review!

mkhorton:
Thanks!

PR #2344, opened 2022-01-12
Add caching to all GH workflows

### LINKED ISSUE #2340: Pip cache not used in GitHub CI
The GitHub test workflows for Linux, Mac and Windows all cache installed pip dependencies but then never check if there's a cache hit from previous runs. Would be great for saving time and energy to fix this soon.

https://github.com/materialsproject/pymatgen/blob/81e96a4a5739e383f921b7272db7ebf8b1710cb9/.github/workflows/test-linux.yml#L38-L44

Reusing cache requires setting an `id` on the cache step and then checking if there was a cache hit before rerunning the install:

```yml
      - name: Cache pip
        uses: actions/cache@v2
        id: pip-cache
        with:
          path: ~/.cache/pip
          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
          restore-keys: |
            ${{ runner.os }}-pip-

      - name: Install dependencies
        if: steps.pip-cache.outputs.cache-hit != 'true'
        run: |
          pip install -r requirements.txt
```

Also, in the Mac and Windows workflows, there's no need for `if: startsWith(runner.os, 'macOS')` etc.

https://github.com/materialsproject/pymatgen/blob/81e96a4a5739e383f921b7272db7ebf8b1710cb9/.github/workflows/test-mac.yml#L34-L47

Since v20.1, `pip` provides a platform agnostic way of getting the cache location: [`pip cache dir`](https://pip.pypa.io/en/stable/cli/pip_cache).

Happy to submit a PR for this.

janosh:
Closes #2340.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
     - uses: actions/cache@v2
       id: pip-cache
       with:
-        path: ${{ pip cache dir }}
+        path: ~/.cache/pip
         key: ${{ runner.os }}-pip-${{ hashFiles('requirements-dev.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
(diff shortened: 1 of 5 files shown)


shyuep:
Thanks.

PR #2349, opened 2022-01-14
Provide an option to keep site properties when calling spglib

arosen93:
As discussed in greater detail in #2330, this PR provides a new keyword option (`keep_site_properties`) in various `structure.symmetry.analyzer` functions to keep the site properties on the sites after a transformation. The docstring makes it clear that this may lead to undesirable side-effects as a result of the transformation, such as a loss of long-range magnetic ordering or antiferromagnetism.


arosen93:
@mkhorton, hopefully this should be better.

shyuep:
Great. Thanks.

PR #2353, opened 2022-01-17
Improve passing of site properties between ASE Atoms object and Pymatgen Structure/Molecule objects (and vice versa)

arosen93:
- When creating a Pymatgen `molecule` object from an ASE `Atoms` object using `pymatgen.io.ase.AseAtomsAdaptor.get_molecule()`, the `molecule.charge` and `molecule.spin_multiplicity` are now appropriately set. Before, they would always be set to their default values of 0 and 1, respectively.
- Now, _all_ Pymatgen site properties are passed to the ASE Atoms object (and vice versa).


shyuep:
Thanks @arosen93 !

PR #2356, opened 2022-01-19
Minor cleanup of bader_caller.py + new get_partial_charge() convenience function

arosen93:
I'll be adding a `chargemol_caller.py` soon, so this is a PR that does some minor cleanup of things I've found in `bader_caller.py` in the process.

- Add a new convenience function `get_partial_charge()`, which is just the negative of the `get_charge_transfer()`. This is to parallel the forthcoming `chargemol_caller.py`.
- Removes unused function kwarg. Adds missing docstrings for function args
- Adds code comments to describe the sign convention of "charge" and "charge transferred", which is a bit unclear on its own.
- Moved the `get_filepath()` function out of `bader_caller.py` (where it's not being used) to `critic2_caller.py` (where it is being used).


shyuep:
Thanks.

PR #2357, opened 2022-01-19
Create a ChargemolAnalysis command line caller (for DDEC6, CM5 analysis)

arosen93:
# Purpose
This PR introduces `pymatgen.command_line.chargemol_caller` with a `ChargemolAnalysis` class that functions in an analogous way as `BaderAnalysis` but for DDEC6/CM5 population analyses. It analyzes the output files from [Chargemol](https://sourceforge.net/projects/ddec/files/) and can also execute Chargemol on its own given a VASP directory of CHGCAR, AECCAR0, AECCAR2, and POTCAR files. Tests have been added for all user-facing functions with NaCl (spin-unpolarized) and O3 (spin-polarized) test systems. The only place a test wasn't added was for the execution of Chargemol, as it requires Pymatgen to have the same license if we were to integrate the binary into the Pymatgen repo.

Currently, the `ChargemolAnalysis` class can return: partial atomic charges (both DDEC and CM5), spin moments (if a spin-polarized calculation was run), dipoles, the sum of bond orders, a bond order mapping between every pair of atoms (including the spin-polarization of those bonds), and R^2/R^3/R^4 moments. Currently, the `ChargemolAnalysis` is written to be compatible with VASP, but it should only require minor updates if someone wanted to make it compatible with other codes (e.g. Gaussian, CP2K).

# Minimal Example
```python
from pymatgen.command_line.chargemol_caller import ChargemolAnalysis

path = "/path/to/my/vasp/run" 
ca = ChargemolAnalysis(path)
print(ca.summary)
```
which would return
```python
{
    "ddec": {
                "partial_charges": List[float],
                "spin_moments": List[float],
                "dipoles": List[float],
                "rsquared_moments": List[float],
                "rcubed_moments": List[float],
                "rfourth_moments": List[float],
                "bond_order_sums": List[float],
                "bond_order_dict": Dict
            },
    "cm5": {
                "partial_charges": List[float],
            }
}
```
-----
Note: This is built off of work originally by @mhsiron (with permission) -- see [here](https://github.com/mhsiron/pymatgen/blob/ads_wf/pymatgen/command_line/ddec6_caller.py) for the original. The PR here adds a number of features not available in the original version (e.g. CM5 charges), refactors the code a fair bit, and fixes several bugs (e.g. related to bond orders, data types).


arosen93:
@oxana-a, @mhsiron -- this is essentially completed. If you have any feature requests or changes you suggest, please let me know! Thanks to @mhsiron for kicking this one off.

shyuep:
Great thanks!

PR #2358, opened 2022-01-19
Don't perform dict type-check on the MAGMOM flag in io.vasp.sets (fixes downstream Atomate issue)

arosen93:
- Addresses https://github.com/materialsproject/pymatgen/commit/9cfef441c16fd4c68290b80bb0383d52a0a8c0fe#r64164646 by reverting the type-check on the MAGMOM key-value pair.


mkhorton:
Thanks @arosen93. Noting that this reverts the warning recently introduced in pymatgen v2022.0.17 via #2301

PR #2360, opened 2022-01-19
flake8 add per-file-ignores = __init__.py: F401

### LINKED ISSUE #2359: Fix yaml.safe_load() deprecation warnings
My `pytest` and `fireworks`  logs are full of `yaml.safe_load()` deprecation warnings originating in `pymatgen` and `monty`.

```py
site-packages/monty/serialization.py:85: PendingDeprecationWarning: 
  load will be removed, use
  
    yaml=YAML(typ='unsafe', pure=True)
    yaml.load(...)
  
  instead
    return yaml.load(fp, *args, **kwargs)

site-packages/pymatgen/analysis/local_env.py:51: PendingDeprecationWarning: 
  safe_load will be removed, use
  
    yaml=YAML(typ='safe', pure=True)
    yaml.load(...)
  
  instead
    default_op_params = yaml.safe_load(f)

site-packages/pymatgen/analysis/local_env.py:54: PendingDeprecationWarning: 
  safe_load will be removed, use
  
    yaml=YAML(typ='safe', pure=True)
    yaml.load(...)
  
  instead
    cn_opt_params = yaml.safe_load(f)
```

Would be nice to fix those to have readable logs again. Should be doable by turning this

https://github.com/materialsproject/pymatgen/blob/905fa8592423f51784b9e83e02c53bb0c6ea0aa3/pymatgen/core/__init__.py#L11-L17

into this

```py
try:
    from ruamel.yaml import YAML

    yaml = YAML(typ="safe", pure=True)
except ImportError:
    try:
        import ruamel_yaml as yaml  # type: ignore  # noqa
    except ImportError:
        import yaml  # type: ignore # noqa
```

and then replace all `yaml.safe_load()` with `yaml.load()`. What do you say?

janosh:
~Closes #2359.~ Closes nothing.

#2359 was closed in d7cfd8a4a6a00fce1860f1a4d2ed0a3830520dbc.

### DIFF
--- a/pymatgen/analysis/interfaces/__init__.py
+++ b/pymatgen/analysis/interfaces/__init__.py
@@ -1,4 +1,7 @@
 """ Module that implements various algorithms related to interface construction and analysis """
-from pymatgen.analysis.interfaces.zsl import ZSLGenerator, ZSLMatch  # noqa
-from pymatgen.analysis.interfaces.coherent_interfaces import CoherentInterfaceBuilder  # noqa
-from pymatgen.analysis.interfaces.substrate_analyzer import SubstrateAnalyzer, SubstrateMatch  # noqa
+from pymatgen.analysis.interfaces.coherent_interfaces import CoherentInterfaceBuilder
+from pymatgen.analysis.interfaces.substrate_analyzer import (
+    SubstrateAnalyzer,
+    SubstrateMatch,
+)
+from pymatgen.analysis.interfaces.zsl import ZSLGenerator, ZSLMatch
--- a/pymatgen/core/__init__.py
+++ b/pymatgen/core/__init__.py
@@ -9,14 +9,14 @@ operations on them.
 import os
 
 from ruamel.yaml import YAML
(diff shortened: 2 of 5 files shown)


janosh:
This has become quite the opinionated PR with more monkey patching than expected.

Since the `safe_load` method will be removed soon, something like the mess I'm creating here will become necessary unless `pymatgen` decides to be more opinionated about the YAML parser it requires users to have. I.e. by saying everyone must install `ruamel.yaml` you'd have a more dependable interface.

shyuep:
Let me do this. ruamel is already a requirement in pymatgen anyway.

shyuep:
I just pushed a fix. This PR should not be necessary unless you see I missed something.

janosh:
@shyuep Cool thanks! Fix looks good. I left the

```
per-file-ignores =
  # F401: imported but unused
  __init__.py: F401
```

part of this PR cause I think that's an improvement. What do you think?

shyuep:
Ok!

Issue #2367, opened 2022-01-22
Vasprun().final_energy does not report energy(sigma->0) for VASP 6 & Vasprun and Outcar final energies are inconsistent


arosen93:
@shyuep, @mkhorton: There is a bug with the `final_energy` property in `Vasprun` for VASP 6+, and the `final_energy` property in `Outcar` is somewhat unintuitive.

## Background
For the attached VASP 6 files below, there are the three types of energies in the output for the final step.

Vasprun.xml:
```
   <i name="e_fr_energy">    -33.16377435 </i>
   <i name="e_wo_entrp">    -33.15237262 </i>
   <i name="e_0_energy">    -33.15807349 </i>
```

OUTCAR:
```
  free  energy   TOTEN  =       -33.16377435 eV

  energy  without entropy=      -33.15237262  energy(sigma->0) =      -33.15807349
```

Here is a quote from the VASP manual in the [ISMEAR section](https://www.vasp.at/wiki/index.php/ISMEAR) to motivate what I'm going to discuss below, although getting the sigma->0 extrapolated energy is important regardless of ISMEAR.

> The Gaussian smearing method leads to very reasonable results in most cases. Within this method **it is necessary to extrapolate from finite SIGMA results to SIGMA=0 results**. You can find an extra line in the OUTCAR file: energy( SIGMA→0 ), giving the extrapolated results.

## Vasprun
For the attached files, the `Vasprun` prints out the `e_wo_entrp` as the `final_energy`. It's because of the codeblock below, where it detects that `e_wo_entrp` and `e_0_energy` are different and takes the former.

It's not particularly clear to me, however, why it's taking the former when you'd almost universally want the energy in the sigma --> 0 limit (there are no vdW corrections here despite the warning message). **NOTE**: Turns out the code is not what it looks like at first glance. It's to fix a confusing `vasprun.xml` bug in VASP 5 (see comment below), but it ends up breaking in VASP 6.

https://github.com/materialsproject/pymatgen/blob/ae3f1b77d75b397735ee7cf16fcd77d8f487d85e/pymatgen/io/vasp/outputs.py#L621-L629

## Outcar
To add to the confusion a bit, the `Outcar` prints out the `e_fr_energy` instead as the `final_energy`. Again, I would have naively assumed that it would output the `e_0_energy`. There are few situations where I'd actually want `e_fr_energy`. Either way, it is not great that it is a different `final_energy` than that generated with `Vasprun`.

[...]

PR #2369, opened 2022-01-23
Return the correct Vasprun().final_energy for VASP 6+ and change the default Outcar().final_energy to be energy(sigma->0)

arosen93:
## The Problem
This is to address #2367. In short, there are two issues:
1. `Vasprun().final_energy` does not report `e_0_energy` (the desired energy quantity) for VASP 6+. In an attempt to address the VASP 5.x.x bug documented [here](https://www.vasp.at/forum/viewtopic.php?f=3&t=16942) where `e_wo_entrp` and `e_0_energy` were swapped, Pymatgen seems to have an unintended side-effect of returning energies that are not extrapolated to sigma-->0 for VASP 6+ when the bugfix is no longer needed.

2. `Outcar().final_energy` always reports `e_fr_energy`, which is rarely the desirable quantity. Fortunately, there is no bug with `e_0_energy` in the OUTCAR. This was just a design decision but one that is not intuitive and also inconsistent with the expected behavior of `Vasprun().final_energy`. 

## Changes in this PR
1. Bugfix: `Vasprun().final_energy` now returns `e_0_energy` as expected. For VASP < 6, the energies should be the same as before to within 7 decimal places. For VASP 6+, the energies will differ by whatever the difference between `e_wo_entrp` and `e_0_energy` was, as the former was incorrectly returned before. 

2. Improvement: `Outcar().final_energy` now prints out `e_0_energy` (also called "energy(sigma->0)" in the OUTCAR) rather than `energy_fr_energy` (also called "free  energy   TOTEN" in the OUTCAR). This is to be consistent with `Vasprun().final_energy` and because it is generally the desired quantity.

3. Feature addition: `Outcar()` now has two new attributes: `.final_energy_wo_entrp` and `final_fr_energy`, which correspond to `e_wo_entrp` and `e_fr_energy`, respectively. The user can fetch whichever one of the three they desire now, although `.final_energy` gives you the value you'd want in virtually all cases.

## Tests
It is crucial that we get this right, so I am including here some tests (attached) that use different versions of VASP (as well as different IVDW and different ISMEAR for good measure). Four of these are added to the test suite. The desired solution is the `energy(sigma->0)` column, which is taken from the last `energy(sigma->0)` line of the OUTCAR, which has no known issues.

### Tests with Main Branch Pymatgen
I draw your attention to Run 3 and Run 6. The `Vasprun` is not producing the correct energy for VASP 6+.

Also, Run 2 and Run 3 are the same material and settings. If you relied solely on the `Vasprun` final energy, you might incorrectly think that VASP 6.2.1 changed the energy by 0.05 eV when in reality it was a parsing error. 

The differences in `Outcar` are just because it was outputting `e_fr_energy` (design decision).
```
Run, energy(sigma->0), Vasprun, Outcar, VASP version
1, -15.89355325, -15.89355325 eV, -15.92106087, 5.4.1
2, -11.18981538, -11.18981538 eV, -11.217323, 5.4.1
---> 3, -11.18986774, -11.1348525 eV, -11.21737536, 6.2.1 <---
4, -15.02557351, -15.02557351 eV, -15.05308113, 5.4.1
5, -15.92852842, -15.92852842 eV, -16.01840921, 5.4.1
---> 6, -33.15807349, -33.15237262 eV, -33.16377435, 6.2.1 <---
```
### Tests with this PR
Everything is as expected now, regardless of version number.

There is slight fluctuation in the eighth decimal place for the `Vasprun` final energy with VASP 5.x.x because I updated how the `vasprun.xml` bugfix is handled to be more robust, at the expense of a rounded 8th decimal place. See the [VASP forum post](https://www.vasp.at/forum/viewtopic.php?f=3&t=16942) for details.
```
Run, energy(sigma->0), Vasprun, Outcar, VASP version
1, -15.89355325, -15.89355324 eV, -15.89355325, 5.4.1
2, -11.18981538, -11.18981538 eV, -11.18981538, 5.4.1
3, -11.18986774, -11.18986774 eV, -11.18986774, 6.2.1
4, -15.02557351, -15.02557351 eV, -15.02557351, 5.4.1
5, -15.92852842, -15.92852843 eV, -15.92852842, 5.4.1
6, -33.15807349, -33.15807349 eV, -33.15807349, 6.2.1
```
[tests.zip](https://github.com/materialsproject/pymatgen/files/7919790/tests.zip)


arosen93:
@mkhorton -- this is ready for your review at your next convenience.

mkhorton:
Hi @arosen93, this looks great. My only questions:

1. Can you add some comments (or change file names) to make clear what version of VASP was used to generated e.g. `vasprun.xml.etest1.gz`
2. The VASP forum post was perhaps a little vague on this, do you know if this behavior definitely changed with 6 (6.1.0?) vs 6.1.2, 6.2.1 etc?

For point 2., in particular I'm aware of a large quantity of calculations that were generated with a VASP 6 beta which may incorrectly report its version number. I assume there's no alternative we can take here, i.e. that we just have to trust the version number reported and use that for the `if` statement, rather than another method of being able to figure out if the bug is present?

arosen93:
@mkhorton:
1. Certainly, I can make that clearer. 
2. I have checked and confirmed that VASP 6.1.0 (and the VASP 6 pre-release used by @rkingsbury) both have the `vasprun.xml` change.

Re your point 2 follow-up: There is indeed a workaround. The workaround would be to just delete the `if` statement entirely. The codeblock below works for any version of VASP. It's just kind of messy and unnecessary for VASP 6.1.0+, but it will still work for VASP 6.1.0+. The only side-effect is there will be rounding in the 8th digit of the energy (negligible). We can just have this block regardless of version number if you'd prefer. This is what ASE does, for what it's worth.

https://github.com/materialsproject/pymatgen/pull/2369/files#diff-0d68150cedc59844e2cd1ea9f15543fe522b072531e5050fde8a4fe4bee3baf3R625-R628

mkhorton:
I won't refer to the ASE code on this for licensing reasons (though I'm sure it's wonderful!)

> Actually, I could just check to see if both solutions yield the (essentially) the same answer. If they do, it's probably VASP 6+ so just use the normal e_0_energy. If they differ, it's probably a messed up vasprun.xml and take the latter.

This sounds like a great solution.

mkhorton:
I'm perhaps over-paranoid about edge cases (eg old results that are reporting an incorrect VASP version) but I know in a large enough database these edge cases always happen sooner or later.

arosen93:
Yes, good points all around. I will implement a check that is not dependent on the version number just in case. Will get back to you on this!

arosen93:
This should do the trick! I now compare the energy computed from two methods: 1) By taking the final `e_0_energy` from the last ionic step, and 2) from the "bugfix" approach discussed on the VASP forum. If Option 1 and Option 2 are the same, we'll just take Option 1 because that's what it should be if the `vasprun.xml` is correct (and should occur for VASP 6+). If they differ by 1e-6 or more, we'll take Option 2 because it's a sign that there's something wrong with the `vasprun.xml` (expected for VASP < 6).

I also added comments to the tests to indicate the version numbers.

mkhorton:
Perfect, thanks @arosen93 !

PR #2371, opened 2022-01-24
Pass oxidation states between Pymatgen structure/molecule and ASE Atoms

arosen93:
This hopefully should be my last update in my quest to enhance the `AseAtomsAdaptor`. 

In this PR:
1. Oxidation states are now passed from Pymatgen structure/molecule objects to ASE atoms objects. Since these aren't `site_properties`, they never got transferred. This is the main change.
2. Charges passed from Pymatgen to ASE now get assigned "intelligently" in the ASE Atoms object such that the ASE-recommended `atoms.get_charges()` function retrieves them.
3. Several new tests have been added to the test suite, most importantly a check that doing Atoms --> Structure --> Atoms as well as Structure --> Atoms --> Structure result in identical objects before/after the transformation cycle. This is an important check that was missing, as it's very easy for a property to be lost in the translation if not coded appropriately. This will also help future-proof the module if/when others make changes.

### DIFF
--- a/pymatgen/io/ase.py
+++ b/pymatgen/io/ase.py
@@ -69,13 +69,32 @@ class AseAtomsAdaptor:
         # Note: ASE distinguishes between initial and converged
         # magnetic moment site properties, whereas pymatgen does not. Therefore, we
         # have to distinguish between "magmom" and an "initial_magmom" site property.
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks.

PR #2374, opened 2022-01-24
Added property symbol_list to class SiteCollection.

exenGT:
## Summary

* Added a new property, symbol_list, within the class SiteCollection in core/structure.py. This property returns the list of chemical symbols of all atoms in the collection. This property may be useful for postprocessing purposes, as the "species" property does not return a list of strings.

No additional dependency is needed.

## TODO (if any)

N/A

## Checklist

in the pull request title.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -270,6 +270,11 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
         """
         return tuple(sorted(specie.symbol for specie in self.types_of_species))  # type: ignore
 
+    @property
+    def symbol_list(self) -> List[str]:
+        """List of element symbols."""
+        return [specie.symbol for specie in self.species]
+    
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks for the suggested contribution. But since this is a method that is a one-liner and not that frequently used, I am not in favor of adding it to the already long list of Structure properties/methods.

exenGT:
Thanks for your comment @shyuep. I agree that there are already plenty of methods / properties in `Structure`. In this particular case though, I think it would be really nice to have a method / property that outputs a list of *strings* of atom identifiers (e.g. chemical symbols). The rationale is that, as many other postprocessing codes need this information as an input, the unique format of `species` in pymatgen makes it difficult to directly use `species` as input to other codes. I would imagine such routine to be rather common, as from my own experience, I have to write this one line of code many times. and would really appreciate if you could provide some comment / suggestion. Thanks again!

PR #2375, opened 2022-01-24
Extend get_symmetric_site to non laue slabs

bcamino:
## Summary

* The current version of get_symmetric_site relies on slabs to have Laue symmetry. Treating a slab as a 3D object introduces symmetry operators that duplicate a site to a neighbouring cell in the z direction. This introduces two possible issues:
    * some of the operators will duplicate the point to be on the same side of the slab in a neighbouring cell rather than on the other side;
    * if the slab is to be used in a code such as CRYSTAL17 that treats slabs as objects with 2D symmetry, all the atoms need to be in the same unit cell (at least along the z direction)
* To solve the issue, I propose this approach: 
    * center the slab and the point at z=0
    * get the symmetry operators to this structure
    * apply the symmops to the point coordinates. If the resulting point is at -z coordinate (where z was the coordinate of the original point), check the slab is symmetric (rather than laue)
    * if the slab is symmetric return the coordinates of the point

* No additional dependencies were added

## TODO (if any)

* None for this functionality

## Checklist

in the pull request title.

[New_get_symmetric_site.ipynb.zip]
Jupyter Notebook showing the tests I performed for the new functionality.

(https://github.com/materialsproject/pymatgen/files/7935219/New_get_symmetric_site.ipynb.zip)


shyuep:
Thanks. Can you add a unittest for the non-Laue part of the code pls? The tests should be in pymatgen/core/tests/test_surface.py

bcamino:
Hello, I modified the function test_get_symmetric_sites in pymatgen/core/tests/test_surface.py to test the new version of the get_symmetric_site function.

> Thanks. Can you add a unittest for the non-Laue part of the code pls? The tests should be in pymatgen/core/tests/test_surface.py

bcamino:
Hi @shyuep, is there anything else I should do about this PR? Thanks for your help (first time submitting a PR here).
Cheers,
Bruno

shyuep:
I apologize for this, but there is a conflict due to some recent changes. Can you resolve the conflict? @CFritz145 Do you mind taking a look and see if there are any issues with this change to support non-laue slabs? Thanks.

CFritz145:
Sure thing, I'll take a look at the PR later during the day. Thanks for letting me know

bcamino:
> I apologize for this, but there is a conflict due to some recent changes. Can you resolve the conflict? @CFritz145 Do you mind taking a look and see if there are any issues with this change to support non-laue slabs? Thanks.

Hello, the only conflict I am seeing is the piece of code I replaced in get_symmetric_site. Is there more to that? Cheers, Bruno

CFritz145:
I seen no issue with compatibility, however I thinks there's a simpler solution to all of this. 

The point of this PR is to identify symmetric surface sites by generalizing what we consider a "symmetric slab". A slab doesn't need Laue symmetry (centrosyymetric) to be symmetric as implied by the PR, e.g. a mirror plane perpendicular to the c-axis of the slab or a screw axis along c would make it possible to symmetric sites between two surfaces. The get_symmetric_site() method of the Slab class is a bit outdated and uses the get_laue() class to determine slab symmetry, so yes that needs to be changed. However the operation you proposed to identify slab symmetry is believe is already taken care of by the .is_symmetric() function which **previously operated under the assumption that a slab is symmetric IFF it has Laue symmetry**. This was modified a few months ago to now take into account any symmop along the c-axis that gets you the same sites on both sides of the slab (essentially treating symmetry the slab like its 2D symmetry as you proposed). @bcamino  correct me if I'm wrong but would replacing the operations proposed in the PR with the .is_symmetric() function accomplish the same thing?

bcamino:
Hi @CFritz145, thank you for your response. The goal of this PR is twofold:

the first, as you identified, is to use is_symmetric() rather than is_laue in defining the symmetry of a slab. 

the second is being able to get the symmetric site that sits on the other side of the same slab. In the current version of get)symmetric_site(), there are instances in which the point found by this function is symmetric indeed, but in the neighbouring cell. If you want to replicate this behaviour you can use:

bulk = Structure.from_spacegroup("Fm-3m", Lattice.cubic(6.63), ["Cu"], [[0, 0, 0]])
adsorbate = Molecule('O',[[0.0, 0.0, 0.0]])
substrate = SlabGenerator(initial_structure=bulk,miller_index=[1,0,0],min_slab_size=15.,min_vacuum_size=10.,center_slab=True).get_slab()
ads_sites = AdsorbateSiteFinder(substrate)
sub_ads = ads_sites.adsorb_both_surfaces(adsorbate,repeat=[1,1,1])[0] 

In the first adsorption structure, the first oxygen is placed at [0.5000, 0.5000, -0.8103] and duplicated at [0.0000, -0.0000, 0.8103] (all fractional coordinates). This latter point was obtained by applying the 
Rot:
[[ 1.  0.  0.]
 [ 0. -1.  0.]
 [ 0.  0. -1.]]
tau
[0.5 0.5 0. ]
symmetry operator. This duplicated oxygen atom is equivalent to the first one, but it is in the cell (0,0,-1). This doesn’t cause problems in a plane wave code where slabs are simulated as 3D structures with a vacuum region between slabs. However, I’m interested in generating structure with pymatgen for the CRYSTAL code. CRYSTAL uses localised atomic orbital generated Bloch functions as basis set and treats the slabs as 2D systems. Therefore, I do need to have the symmetric adsorbate on the side of the same slab. In addition, having both adsorbates on the same slab makes it easier to visualise the structures. 

The proposed methodology (centering the slab at z=0 and getting the symmetry operators) ensures this is true.

Please let me know if you require any more information about this PR.

Cheers,

Bruno

CFritz145:
Hi @bcamino 

Thank you for the explanation. 

So the heart of the issue here seems to be that the symmetry operators return a fractional coordinate outside the periodic boundary condition. In the block of code you sent, for example, the return slab had an O atom appended to a position at c < 0 which poses problems in CRYSTAL. This PR solves this issue by ensuring the fractional coordinates remain inside the PBC ie between 0 and 1. Let me know if this may work as an alternative:

After identifying a symmetric surface frac_coord using .get_symmetric_site(), the function will create a PeriodicSite with the dummy site. This is just a crystal site with the lattice information in it so that it can determine the frac_coords. With our dummy PeriodicSites, we can use the class method .to_unit_cell() which ensures the site will remain inside the bounds of the lattice. Then we return the frac_coord or cartesian_coord. This will avoid any symmetric positions being added outside the PBC. 

Just trying to seeing if there might be a more parsimonious solution, but if there's an underlying issue that .to_unit_cell() can't solve here, I see no issue in the PR from my side.

bcamino:
Hi @CFritz145,

thank you! That does sound like a better plan. I’m going to modify this PR to use this approach. 

One more thing: in the current implementation, if no slab satisfied the is_laue() condition, the site2 that was generated by the last symmop is returned. However, this would be a non symmetrical slab. What do you think about adding a check at the beginning of get_symmetric_site that checks if the initial slab is symmetric? I believe having an initial symmetric slab is a condition to obtain a symmetric one with adsorbates on both surfaces (please correct me if I’m wrong). 

If you think this is a good idea, in case of an initial non symmetric slab, should the function raise a warning message and return None?

Thank you so much for your help.

Cheers,

Bruno

CFritz145:
Hi @bcamino 

Glad I could help. 

In regards to returning an assymetric site, you are right, it will return something regardless of whether or not it found a symmetric site. I originally planned submit a PR on this but lost track of it somewhere down the line. 

If you want, you can copy this block of code i modified along time ago into the PR, it should be able to address everything talked about here. Feel free to modify as you see fit.

`    def get_symmetric_sites(self, points, cartesian=False):
        """
        This method uses symmetry operations to find equivalent sites on
            both sides of the slab. Works mainly for slabs with Laue
            symmetry. This is useful for retaining the non-polar and
            symmetric properties of a slab when creating adsorbed
            structures or symmetric reconstructions.

        Arg:
            point: Fractional coordinate.

        Returns:
            point: Fractional coordinate. A point equivalent to the
                parameter point, but on the other side of the slab
        """

        ops = SpacegroupAnalyzer(self).get_symmetry_operations(cartesian=cartesian)

        # Each operation on a point will return an equivalent point.
        # We want to find the point on the other side of the slab.
        found_site = False
        for op in ops:
            # skip any symmops that don't bring the site to the other side of the slab 
            if all((op.rotation_matrix[2] == [0, 0, 1])):
                continue

            slab = self.copy()
            sites2 = op.operate_multi(points)

            # Add dummy site to check the overall structure is symmetric
            for site2 in sites2:
                slab.append("O", site2, coords_are_cartesian=cartesian)
            for point in points:
                slab.append("O", point, coords_are_cartesian=cartesian)

            try:
                if slab.is_symmetric():
                    found_site = True
                    break
            except TypeError:
                # error related to spacegroup analysis
                pass
                
        if found_site:
            psite = PeriodicSite('O', site2, slab.lattice, to_unit_cell=True,
                                 coords_are_cartesian=cartesian)
            site2 = psite.coords if cartesian else psite.frac_coords
            return sites2
        else:
            warnings.warn('Cannot add symmetric site!')
            return None
`

bcamino:
Hello @shyuep , after some more tests I don't think this PR will be necessary, so I will close it now. Thank you, Bruno

PR #2381, opened 2022-01-26
Weighted average position with PBC

jmmshn:
## Summary

* Added function to find the weighted average of positions in a periodic lattice
* Added test

This is in preparation for a provenance agnostic defect position identification


shyuep:
Thanks. But I am not sure this shuld be in Lattice. I am not even sure how the weighted average is defined - fractional or cartesian coords? If this method is mainly for defect finding, it can be implemented in the defects package.

jmmshn:
So this is coded for `frac_coords` but I think that since the transformation is linear the weighted average position can just be transformed with `get_cartesean_coord`.

I need this for some defect code development but I imagine that people might have a need to average some scalar property on a few sites and get an effective coarse-grained value.
This function will allow us to easily to that.

shyuep:
Well, the main purpose of such a function that I can think of is to do something like calculating moment of inertias (atomic mass weighted position). In any case, I don't think it should be done in Lattice. It is at best a SiteCollection method. And everything should be done with cartesian coords (fractional coord averaging doesn't make much sense since the PBC can cause the correct fractional coordinates for neighbors to be quite different).

jmmshn:
> It is at best a SiteCollection method
I'll look at this and put it there.

> fractional coord averaging doesn't make much sense since the PBC can cause the correct fractional coordinates for neighbors to be quite different
The code in this PR specifically deals with PBC so that even if the provided positions in the list "wraps" the average point does not. 
An example will be averaging the following points.
```
                [0.6, 0.7, 0.5],
                [0.9, 0.8, 0.5],
                [0.1, 0.2, 0.5],
                [0.1, 0.95, 0.5],
```
Gives the same result as averaging:
```
[[0.6, 0.7, 0.5],
 [0.9, 0.8, 0.5], 
[1.1, 1.2, 0.5], 
[1.1, 0.95, 0.5]]
```
Because the code understands that the last 2 points should be wrapped.  This allows users to not be super worried about providing the exactly correct period image index.

mkhorton:
What is the practical usage for allowing for weighted positions? To calculate centre of mass?

jmmshn:
The origin of this was that I needed to take the weighted average of like scalar disturbance field on a bunch of atoms.
Then I realized that crossing the boundary condition what going to be a problem so this function is just the math of how to deal with average positions across PCB.

jmmshn:
Although I do acknowledge that if people use a set of points that are spread across the entire unit cell, the answer of where the COM is non-unique.  It's only unique if the points are close together to begin with.

jmmshn:
Will include this code with the defect position finder

PR #2387, opened 2022-01-26
Fix readme badges

janosh:
- uniform shield.io readme badges
- brought onto 1 line
- fix GH CI badge
- add PyPI download badge

## Before

Screen Shot 2022-01-26 at 14 42 56

## After

Screen Shot 2022-01-26 at 14 50 33


janosh:
While we're at it, what do you think of converting the readme to Markdown? RST is a PITA IMO.

shyuep:
Sphinx uses RST. I would prefer not to have to handle two different documentation formats. The README is basically a cut and paste of the introduction RST.

PR #2389, opened 2022-01-26
PDEntry.__repr__ include name attr if not fallback value

janosh:
Override `Entry.__repr__` in `PDEntry` to include the `name` if its not the fallback value of reduced composition.

**Before**

```py
PDEntry("LiFeO2", 53, name="mp-757614")
>>> PDEntry : Li1 Fe1 O2 with energy = 53.0000
```

**After**

```py
PDEntry("LiFeO2", 53, name="mp-757614")
>>> PDEntry : Li1 Fe1 O2 (mp-757614) with energy = 53.0000

PDEntry("LiFeO2", 53)
>>> PDEntry : Li1 Fe1 O2 with energy = 53.0000
```

Added a test for both cases.

We could also do

```py
PDEntry mp-757614 : Li1 Fe1 O2 with energy = 53.0000
```

if preferred.

### DIFF
--- a/pymatgen/analysis/phase_diagram.py
+++ b/pymatgen/analysis/phase_diagram.py
@@ -69,6 +69,12 @@ class PDEntry(Entry):
         self.name = name if name else self.composition.reduced_formula
         self.attribute = attribute
 
+    def __repr__(self):
+        name = ""
+        if self.name != self.composition.reduced_formula:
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks. I don't have a problem with this. The repr is purely for end user readability.

PR #2397, opened 2022-01-29
Delete overdue deprecations

janosh:
Noticed that a bunch of stuff has deprecation messages saying it would be removed in pmg `2020.*`. This PR removes those functions and methods since we're already at v2022.1.24.


arosen93:
Beat me to it. I was just about to point out the AiiDA deprecation.

janosh:
There were some more things where I couldn't easily figure out from `git blame` when it was last changed and if it might be time to get rid of that as well. Maybe @shyuep can comment on those:

https://github.com/materialsproject/pymatgen/blob/07794afb624f7613b3dbcebdf0034cab4c9ee7ea/pymatgen/analysis/interface_reactions.py#L629-L634

https://github.com/materialsproject/pymatgen/blob/ce3615803666b28726e56f1a1fd11c3676b06c60/pymatgen/analysis/defects/dilute_solution_model.py#L42-L56

https://github.com/materialsproject/pymatgen/blob/0af8a0c66bd54a3dc83a3f33792d48ca0bf22632/pymatgen/analysis/elasticity/elastic.py#L408-L414

arosen93:
The first one at least seems like a very recent module from @mattmcdermott.

shyuep:
Thanks.

shyuep:
It doesn't matter. If it has been indicated as depreacted, we can remove it now. If someone complains, we then add it back as necessary. Most of these are very old indeed.

PR #2400, opened 2022-01-30
Update zeo++ installation steps

arosen93:
The Zeo++ installation steps in `pymatgen.io.zeopp` are extremely out of date. I've modernized them and added the reference that should be cited when using Zeo++.


shyuep:
Can you put these in the docs_rst instructions too? A lot of people just search on pymatgen.org/installation for these instructions.

arosen93:
Good idea! Should be done now.

mkhorton:
Thanks @arosen93 !

PR #2401, opened 2022-01-30
Use setup-python automatic pip caching in CI

janosh:
Since Nov 2021, `setup-python` GitHub action supports automatic `pip` dependency caching:

https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,6 +20,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         cache: pip
+        cache-dependency-path: requirements-dev.txt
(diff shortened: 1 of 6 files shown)


janosh:
Note by default this caching mechanism uses the combined hash of all files matching  `**/requirements.txt` as the restore key. In case of the `linting` workflow for example, we actually want this to be `requirements-dev.txt`:

```yml
- uses: actions/setup-python@v2
  with:
    cache: pip
    cache-dependency-path: requirements-dev.txt
```

https://github.com/actions/setup-python#caching-packages-dependencies

PR #2404, opened 2022-01-31
Finding Symmetrically Equivalent Bonds (Neighbors) in Structure

nheinsdorf:
## Summary

Added a method that gets all the neighbors up a maximum distance for a Structure, and groups these 'bonds' according to their symmetry. This is useful when building a physical model to see which bonds should have the same attributes by symmetry (like a tight-binding hopping amplitude). Furthermore, for each bond, a symmetry operation is given that connects it to the first bond of this symmetry group. This is useful if a directional attribute (like a Dzyaloshinskii-Moriya (DM) interaction in a spin model) is associated with a given bond. The operations can then be used to rotate a given DM vector obeying the symmetry of the crystal. For a set of bonds with the same symmetry, these operations basically constitute what is known as the orbit for a Wyckoff position. 
This was written in response to the discussion on the pymatgen forum:
https://matsci.org/t/finding-symmetry-equivalent-bonds-neighbors/38171/3

* Added the 'are_symmetrically_related_bond'-method to SymmOp-class to check if the operation relates two bonds.
* Added the 'get_symmetric_neighbor_list'-method to Structure that gives a similar output as 'get_neighbor_list' but groups the bonds according to symmetry.

## TODO 

Similar to the method 'get_all_neighbors', it would be nice to have similar method that does not return the output as a tuple of lists, but returns a list of 'PeriodicBond' instances (that could inherit from PeriodicNeighbor). In addition to the properties of PeriodicNeighbor, this class would also store its own 'index' (as opposed to the index of the neighbor), as well as the space group of the structure. Then, the '==' operator of that class can be overloaded to see whether they are symmetrically related.


mkhorton:
Thanks for this PR @nheinsdorf!

Another place to put this may even be `pymatgen.analysis.graphs.StructureGraph`. That way the method can be more agnostic to _how_ the bond is determined, e.g. a neighbor list is not necessarily equal to a list of bonds.

nheinsdorf:
Hey @mkhorton, 
yeah I know that you gave the incentive back then on the FAQ to use pymatgen's graph functionality. I agree that due to the different environments it is more flexible, and also that you are free to define some properties of a node. Here, I wanted to start with the simplest example of just considering connections between points with no intrinsic properties that might also transform under the symmetry. E.g. it would be nice if one could put orbitals on a site and then the algorithm could also check whether some symmetry is able to map the orbitals on top of each other (same with spin or some magnetic configurations).

nheinsdorf:
Just made sure that always the first instance of a bond of a certain symmetry index in the output is the one that is assigned the identity matrix as a symmetry operation. That's more convenient when constructing models with the neighbor list that are generated by the symmetries that connect the bonds.

nheinsdorf:
Hey @mkhorton, so I added some extra functionality to the two functions I implemented. I did this because I noticed I run into problems with the previous implementations. Let me illustrate: 
I used the new 'get_symmetric_neighbor_list' to build a physical model. Each neighbor (which I will call bond from now on) was given some exchange vector D that also has to obey the symmetries of the model. In the previous version, get_symmetric_neighbor_list essentially gave every bond twice. From i to j along R and from j to i along -R. Usually, if a physical model is built, only one of these bond is considered, since the inverted direction just corresponds to the hermitian conjugate. This is why I implemented the option 'unique'. If True, the code will output each bond once. 

Then I ran into the problem that 'are_symmetrically_related_bond' does not check for the direction of the bond, since from i to j along R is equivalent to j to i along -R. I added a bool to the method's output that allows the user to trace whether a bond has been inverted or not. 'get_symmetric_neighbor_list' makes use of that. If unique is set to True, it always gives the bond in the output that has NOT been inverted. Now it is ensured that applying the symmetry operation associated with given bond to the first bond with the same symmetry_index, yields the actual bond and not its inverted counterpart. 

If I now associate the vector D to a given bond, applying the symmetry operations from the output of 'get_symmetric_neighbor_list' to D generates the right vector (pointing in the right direction) for each bond. 

I can gladly provide a code example where this is used to conveniently build a magnetic low-energy model with DM exchange.

mkhorton:
Thanks for the added context @nheinsdorf. Indeed, some of the issues you mention, namely bond "directionality" and equivalency, are handled by the StructureGraph class which encodes a bond as a directed edge and ensures that duplicate bonds cannot be added. However, that code is not symmetry-aware, so definitely inferior in that sense to the code suggested in this PR. The advantage of the StructureGraph is simply that it acts a container for this kind of bond information, providing a unified interface and allowing it to be easily serialized and deserialized alongside the corresponding Structure, though at the cost of efficiency in terms of instantiating the class.

This PR looks to be high quality, so I would be glad to merge it in some form.

For code review, I have two questions:

(1) what is the purpose of the commented-out test block, should this be retained?

(2) The docstring for `get_symmetric_neighbor_list` is excellent, but the code itself is a little difficult to follow, some comments throughout would be welcome.

If you were happy to share the code example to test, that would definitely be useful. I believe you can attach files to a Github comment.

nheinsdorf:
Hey @mkhorton , 
Thank you very much for your feedback. Concerning your points
1) You mean the block above `test_get_symmetric_neighbor_list`? I did not add it. It was contained in the repo I cloned. 
2) I can add comments! Browsing through the source code it felt like that most of the commenting is done in the docstring, but not in the actual code. But I can add some comments in between steps, no problem!

When I have some time, I'll put to together an example script with maybe some images to show the intended use of the code!

nheinsdorf:
As promised:

An example script that uses the new functionality to generate generate spin exchanges for a toy model. The following code generates the figure that is given below: 

```

import numpy as np
from mpl_toolkits.mplot3d import Axes3D
from matplotlib.lines import Line2D

class SpinModelExample(object):
    """
    Example for using the 'get_symmetric_neighbor_list'-method.
    We use it to first find the all the next-nearest neighbors in 
    a localized magnetic moment model with space group 198. We use
    the method's functionality to first distinguish two symmetrically
    inequivalent hoppings to assign different exchanges. Next, we assign
    an antisymmetric exchange (Dzyalioshinskii-Moriya) to the bonds and 
    use the method to rotate the exchange vector in accordance with the
    space groups symmetry. 
    
    """

    def __init__(self):

        # Lattice parameters
        a = 8.908
        self.SG = 198
        lattice = [[a, 0, 0], [0, a, 0], [0, 0, a]]

        # Initialize structure
        self.struc = Structure.from_spacegroup(self.SG, lattice, ['Cu'], [[0, 0, 0]])

        self.bonds = None
        self.Js = None
        self.Ds = None

    def examplify_symmetry_indices(self, J1=-1, J2=-1.5):
        """
        Use 'get_symmetric_neighbor_list' to find all next-nearest neigbhors.
        Note that we provide the space group to the method. If that is not 
        done, the method will try to determine the space group based on the 
        Structure, which in our case has higher symmetry than space group 198, because
        we only provided the atoms that have a finite, localized magnetic moment.
        Further, we set unique to True, because we only want to consider each 
        bond once (not back and forth).
        """

        self.bonds = self.struc.get_symmetric_neighbor_list(7, sg=self.SG, unique=True)

        # All nearest neighbors have the same distance.
        print(f'Unique distances: {np.unique(self.bonds[3])}\n')

        # Naively, we would consider all of these equivalent. However, checking
        # the symmetry indices assigned by the methods, we see that there are 
        # actually two groups of symmetrically inequvialent nearest neighbors.
        print(f'Symmetrically inequivalent neighbors {int(np.max(self.bonds[4]) + 1)}\n')

        # Let us use this information to assign two hopping parameters to the 
        # symmetrically distinct nearest neighbors
        sym_idcs = self.bonds[4]
        self.Js = np.zeros(sym_idcs.shape)
        self.Js[np.where(sym_idcs == 0)] = J1
        self.Js[np.where(sym_idcs == 1)] = J2

        print('#\tSymmetry\tExchange\n')
        for _, (sym_idx, J) in enumerate(zip(sym_idcs, self.Js)):
            print(f'{_}\t{int(sym_idx)}\t{J}')
        print('\n')

    def examplify_symmetry_ops(self, D1=[0, 0, 0.1], D2=[0, 0, -0.2]):
        """
        Use 'get_symmetric_neighbor_list' to assign vector quantities to 
        symmetrically equivalent bonds and make sure they transform according
        to the symmetry of the provided space group.
        """

        # Again, we start by constructing all the nearest neighbors
        self.bonds = self.struc.get_symmetric_neighbor_list(7, sg=self.SG, unique=True)
        sym_idcs = self.bonds[4]
        sym_ops = self.bonds[5]
        self.Ds = np.zeros((sym_idcs.shape[0], 3))
        for _, (sym_idx, sym_op) in enumerate(zip(sym_idcs, sym_ops)):
            if sym_idx == 0:
                self.Ds[_] = sym_op.apply_rotation_only(D1)
            if sym_idx == 1:
                self.Ds[_] = sym_op.apply_rotation_only(D2)

        print('#\tSymmetry\tExchange\n')
        for _, (sym_idx, D) in enumerate(zip(sym_idcs, self.Ds)):
            print(f'{_}\t{int(sym_idx)}\t{D}')
        print('\n')

    def plot(self, Js, Ds):
        """
        We plot the structure to illustrate the neighbors that
        we found in the two examples above
        """

        fig = plt.figure()
        ax = fig.add_subplot(111, projection='3d')
        # plot the conventional (cubic) unit cell
        lines = [[(0, 1), (0, 0), (0, 0)],
                 [(0, 0), (0, 1), (0, 0)],
                 [(0, 0), (0, 0), (0, 1)],
                 [(1, 0), (1, 1), (1, 1)],
                 [(1, 1), (1, 0), (1, 1)],
                 [(1, 1), (1, 1), (1, 0)],
                 [(1, 1), (0, 0), (0, 1)],
                 [(1, 1), (0, 1), (0, 0)],
                 [(0, 0), (0, 1), (1, 1)],
                 [(0, 0), (1, 1), (0, 1)],
                 [(0, 1), (1, 1), (0, 0)],
                 [(0, 1), (0, 0), (1, 1)]]
        for line in lines:
            ax.plot(line[0], line[1], line[2], c='black')

        # plot the 4 (magnetic) sites
        # and its periodic images in the 1x1x1 cube
        images = [(1, 0, 0), (0, 1, 0), (0, 0, 1),
                  (1, 1, 0), (1, 0, 1), (0, 1, 1),
                  (1, 1, 1), (0.5, 0.5, 1), (0.5, 1, 0.5),
                  (1, 0.5, 0.5)]
        size = 100
        for site in self.struc:
            ax.scatter(*site.frac_coords, s=size, c='blue')
        for image in images:
            ax.scatter(*image, s=size, c='blue')

        # plot the bonds and the DM vectors
        tol = 0.01
        for (i, j, R, d, sym_idx, sym_op, D) in zip(*self.bonds, self.Ds):
            ri = self.struc[i].frac_coords
            rj = self.struc[j].frac_coords + R
            m = rj < (0 - tol)
            if sym_idx == 0:
                c='red'
            else:
                c='green'
            ax.plot(*zip(ri + m, rj + m), c=c)
            ar = (rj + m - (ri + m)) / 2 + (ri + m)
            ax.quiver(*ar, *D, length=0.1, normalize=True)

        ax.set_xticks([])
        ax.set_yticks([])
        ax.set_zticks([])
        ax.set_axis_off()
        legend = [Line2D([0], [0], color='red', lw=2),
                  Line2D([0], [0], color='green', lw=2),
                  Line2D([0], [0], color='lightblue', lw=2)]
        ax.legend(legend, ['J1', 'J2', 'DM'])
        fig.tight_layout()

        plt.show()

if __name__ == '__main__':

    example = SpinModelExample()
    example.examplify_symmetry_indices()
    example.examplify_symmetry_ops()
    example.plot(example.Js, example.Ds)
```



image

mkhorton:
Thanks for the comprehensive example @nheinsdorf, I'll try and review this shortly, and thanks for your patience as well.

nheinsdorf:
Have you tried running the tests on your system? I still couldn't figure out what the problem is. :thinking:

mkhorton:
Apologies @nheinsdorf for the extreme delay here -- we don't usually let PRs sit for so long, especially ones that are so detailed + pass tests (and from a first time contributor like yourself), so my apologies, I can only say I've been busy recently.

I've gone ahead and run your example and explored it a bit, and it runs beautifully.

I have the following comments:

* Would prefer lowercase `R_a`, `R_b`, etc. for PEP8 compliance, I appreciate this is mathematically less ideal
* I still think, to avoid confusion with existing functionality that is specifically *for* bonds, the `are_symmetrically_related_bond` function should be renamed `are_symmetrically_related_vectors` -- this makes it sufficiently clear that it's a general function.
* I noticed a commented-out test, could this be restored?

Otherwise, I am very happy to merge! The functionality itself is great.

> Next, I might tackle implementing it in a way that returns neighbors in an own class as suggested above.

This would be good, but please familiarize yourself with `NearNeighbor` and `StructureGraph` classes first, so that we may integrate any additions cleanly.

nheinsdorf:
Hey @mkhorton, 
don't worry! 
I don't mind changing R_a, and R_b to lowercase and renaming the function. Will you take care of this when merging, or you want me to edit and update the PR? 

Concerning your third remark, this commented-out block was not written by me. It was contained in the repo when I cloned it. Do with it as you please. 

Happy I could help!

mkhorton:
> Will you take care of this when merging, or you want me to edit and update the PR?

If it's straight forward for you to do, it'd be appreciated if you could this (along with the function rename to `vector`).

> Concerning your third remark, this commented-out block was not written by me. It was contained in the repo when I cloned it. Do with it as you please.

Understood! It showed up in the diff, I guess mistakenly.

nheinsdorf:
Done! But changed it 'vectors' as suggested in your first comment and not to 'vector'. Hope that both is fine. I guess the plural is implied anyway in 'are_symmetrically_related' since it compares two sites?! 
Well anyway, looking forward to see it merged!

mkhorton:
Merged!

PR #2405, opened 2022-01-31
Remove more deprecated stuff

janosh:
Follow up to #2397 and #2401.

Deletes some more code [@shyuep said is very old](https://github.com/materialsproject/pymatgen/pull/2397#issuecomment-1025976675).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,6 +65,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
+          cache-dependency-path: |
+            requirements.txt
+            setup.py
       - name: Install dependencies
         run: |
           pip install --upgrade pip wheel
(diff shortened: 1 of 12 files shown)


shyuep:
The get_neighbors old should not be removed. It is a debugging code that is very useful. Basically it is a brute force version of the get_neighbors that is very slow but guaranteed to be correct because of the very simple logic.

janosh:
@shyuep Yeah wasn't sure about whether that one should go. Put it back in.

mkhorton:
Thanks for this @janosh. If removing old code (even if marked as deprecated), please also add a note to https://github.com/materialsproject/pymatgen/blob/master/docs_rst/compatibility.rst in the PR.

janosh:
@mkhorton Ah cool, wasn't even aware of that file (even though I'd made auto-changes to it 😅). Good thing it exists! Though it appears to be a bit out of date. The file says pmg still supports py3.7. Will bring it up to date.

janosh:
I placed the removed functions under a new heading "Unreleased". Whoever publishes the next version should fill in the right version number.

janosh:
Screen Shot 2022-01-31 at 21 34 52

mkhorton:
Thanks @janosh!

Issue #2408, opened 2022-02-02
Vasprun().complete_dos.spin_polarization yields a KeyError for ISPIN=1


arosen93:
For the a ISPIN=1 calculation with VASP, the `complete_dos` from the `Vasprun` parser yields a `KeyError`. The `spin_polarization` property is not applicable for ISPIN=1, but we should make it clearer for the user rather than throwing a `KeyError`.

There are also no tests for this property.

```python
from pymatgen.io.vasp.outputs import Vasprun
v = Vasprun('vasprun.xml.gz')
v.complete_dos.spin_polarization
```
```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-23-84089bcd748d> in <module>
----> 1 v.complete_dos.spin_polarization

~/software/miniconda/envs/rosen/lib/python3.8/site-packages/pymatgen/electronic_structure/dos.py in spin_polarization(self)
    800
    801         n_F_up = n_F[Spin.up]
--> 802         n_F_down = n_F[Spin.down]
    803
    804         if (n_F_up + n_F_down) == 0:

KeyError: <Spin.down: -1>
```

PR #2409, opened 2022-02-02
Return None for complete_dos.spin_polarization if ISPIN=1

arosen93:
Addresses #2408.

If a VASP calculation is carried out with ISPIN=1, the `Vasprun('vasprun.xml').complete_dos.spin_polarization` property throws a `KeyError`. This PR makes it so that `None` is returned instead.

The `.spin_polarization` property also had no tests. I added them.


shyuep:
Thanks.

Issue #2361, opened 2022-01-20
Incorrect Gvec generation in Wavecar class of pymatgen.io.vasp.outputs module


RyanNKHall:
**Describe the bug**
Within the `pymatgen.io.vasp.outputs` module's `Wavecar` class is a function called `_generate_G_points` (line 5012) that generates the G vectors for the plane wave expansion of the wavefunction using the cutoff energy from the Wavecar file and the kpoints. This function fails to generate the correct number of Gpoints when tested against a range of VASP version 5 Wavecar files. I found the issue arises due to the _encut attribute, which represents the cutoff energy for the plane wave expansion being cast to an integer type in the Wavecar class init function (line 4833). I was able to fix the bug by leaving encut as a float.

**To Reproduce**
1. Use the example WAVECAR file provided. It was generated for Silicon (mp149 in materials project) with hybrid functionals and a Monkhorst-pack grid of 9 x 9 x 9. The resultant Wavecar file contains 35 kpoints, 96 bands and has an encut of 245.345 when not cast to an integer -> this is cast to 245 in the original code
2. I have provided a test Python script called Wavefunction.py that creates an instance of the Wavecar class with the WAVECAR file to reproduce the error
3. A value error "Incorrect value of vasp_type given..." is produced on the third kpoint (ink = 2) (the casting/rounding does not affect the first two iterations likely because the kpoint magnitude is small)
4. Adding a print statement to outputs.py if statement that raises the value error provides some more verbose information on the issue. Specifically, adding: `print("\n# Gpoints:", len(self.Gpoints[ink]), "# plane waves:", nplane, "# extra Gpts:", len(extra_gpoints), "\n")` before the value error is raises will display to the terminal the number of Gpoints generated versus the required number nplane. In this example, 244 Gpoints are generated instead of the expected 253.
5. If you modify line 4833 of outputs.py from:
```
# extract kpoint, bands, energy, and lattice information
self.nk, self.nb, self.encut = np.fromfile(f, dtype=np.float64, count=3).astype(np.int_)
```
to:
```
self.nk, self.nb = np.fromfile(f, dtype=np.float64, count=2).astype(np.int_)
self.encut = np.fromfile(f, dtype=np.float64, count=1)
```
The error/issue is resolved. I have not tested this with earlier VASP WAVECAR outputs

Provide any example files that are needed to reproduce the error, 
especially if the bug pertains to parsing of a file.

**Expected behavior**
[...]

PR #2410, opened 2022-02-02
Allow for float ENCUTs in Wavecar class

arosen93:
Addresses #2361. 

As noted by @RyanNKHall, the `Wavecar` class requires that ENCUT be an `int` when strictly speaking, this is not necessary within VASP. This PR implements the change suggested by @RyanNKHall to read ENCUT as a `float`. 

If a test is requested, I can run a super cheap H2 calc, although I'm not sure it's necessary for this one. The change I made to the test is only a cosmetic change for the reader.


mkhorton:
Many thanks for addressing this issue @arosen93, this looks good.

On a technical note, I don't think this can be classed as a cosmetic change -- the return type is changing, this will have downstream consequences for anyone using the pymatgen parser (eg to store results in a database). However, I can't anticipate a case where this would cause issues so happy to merge.

arosen93:
@mkhorton, apologies! I wasn't clear in my original response. I meant the change to the test file I made is a cosmetic change. You're right about the change in return behavior.

mkhorton:
Actually, the one test case that may be useful is a test case that reproduces @RyanNKHall's original bug report, so we can confirm it's fixed

arosen93:
Sure, I can run a quick calculation on H2 with an encut of like 25.5. I want to keep the filesize small.

mkhorton:
Thanks, I appreciate the concern about filesize. We do have a few wavecar in the test_files dir, so not unprecedented, but definitely good to keep to minimum.

RyanNKHall:
Thanks @arosen93 for the mention here. I'm surprised and pleased to see someone else encountering the same issue within the same period of time. Let me know if you need/require any of my original files to replicate and/or test the bug and associated fix. I believe, however, that everything I provided in the original bug report should suffice.

arosen93:
@RyanNKHall, I opened this PR because of your issue! :) I wanted to make sure it got fixed in future version of pymatgen. Of course, if you want to do the PR, you can by all means do so! Otherwise, I'll finish this up in the next day or two. Thanks for reporting!

RyanNKHall:
@arosen93 I'm more than happy for you to make the PR :) Thanks for your help.

arosen93:
I managed to make a 5 kb WAVECAR file. A teeny-tiny one! Not sure what `pip install` is complaining about though.

arosen93:
Sorry, I was battling the pre-commit CI there for a little bit. Hopefully should be sorted out now.

arosen93:
@mkhorton, just a reminder that this one is done.

mkhorton:
Thanks @arosen93 :)

PR #2411, opened 2022-02-03
More string format tidying

utf:
Continuing on from #2379


arosen93:
Thanks for fixing my `EnvironmentError`'s too. I just looked it up in response and it does not mean what I thought it meant.

utf:
There is a version difference for black between `.pre-commit-config.yaml` (21.12b0) and in requirements-dev.txt (22.1.0). Its also quite confusing that there is both a `.pre-commit-config.yaml` and a `pre-commit` script that do different things.

It could be better to only use `.pre-commit-config.yaml` and then call pre-commit in the linting GH action. E.g., [I do that in atomate2](https://github.com/materialsproject/atomate2/blob/8d75d850fb3c95c6662fd1833183fe8e6951d5c7/.github/workflows/testing.yml#L29) with a line like:

```
      - name: Lint
        run: pre-commit run --all-files --show-diff-on-failure
```

And you could always run `pylint` after pre-commit if it is too long to put in the pre-commit.

shyuep:
I fixed the pre-commit black version. This is the first time we are using it. I will make it consistent in future.

utf:
There still seems to be an issue with the different pre-commit configurations?

Also, windows tests are failing for an unrelated reason.

shyuep:
As I said, I pinned the correct black version in the precommit. Pls pull the latest master into your fork.

shyuep:
Thanks.

PR #2413, opened 2022-02-04
Update minimal basis

JaGeo:
This is a very small change of the configuration files for the Lobster basis. We removed the 2p orbitals for Be as they led to problems in our computations and probably should be optional during the projection.

This should hopefully not affect any of the tests.

Best, 
JG


JaGeo:
Not sure what is going on with the tests as I did not change anything on the code :).

JaGeo:
One other thing: the coverage always seems to be reduced by -0.6% independet of what we are doing. Here, I haven't changed anything on the code.

shyuep:
Don't worry about the coverage. I made it an optional check. This is more for our information rather than a strict check.

JaGeo:
Thank you very much.

PR #2414, opened 2022-02-04
pre-commit.ci readme badge

janosh:
Nice to see you adopted pre-commit.ci!

Normally it auto-updates the hooks in `.pre-commit-config.yaml` weekly. Set it to monthly to keep down the noise.


janosh:
@shyuep Better to leave the `flake8`, `autoflake` and `mypy` hooks in place so that users running `pre-commit` locally can still check new code they write against these hooks and only disable them in CI. As errors are fixed slowly over time, they can be re-enabled.

shyuep:
How do you disable them in CI? As far as I can see, mypy and flake8 does not do autofix but only checks. This is unlike black where changes are made automatically,.

janosh:
>  mypy and flake8 does not do autofix but only checks. This is unlike black where changes are made automatically

That's correct.

I already made the change that disables these hooks in CI only: d70d096

Here are the config docs: https://pre-commit.ci/#configuration-skip

shyuep:
Ok thanks.

janosh:
There's 1 more file that offends the `check-yaml` hook:

```
check yaml...............................................................Failed
- hook id: check-yaml
- exit code: 1

could not determine a constructor for the tag 'tag:yaml.org,2002:python/tuple'
  in "pymatgen/analysis/vesta_cutoffs.yaml", line 1, column 3
```

shyuep:
Yeah, I am not sure that one is solvable. There is no "tuple" in yaml per se. For this particular use case, we need it to be a hashable tuple as it is a dict key. We can just disable the yaml check. YAML is only used as a data file in most cases and if the yaml is not right, the parser won't even load it. So this check is kind of redundant.

shyuep:
That is fine. Yes a yaml failure occurs at run time. But when someone includes a yaml config file in pymatgen, it is typically actually USED somewhere. The tests will catch these yaml failures. I am not sure we want to wring ourselves into knots creating file exceptions. Ultimately, this is a python code base and the yamls are not the key component.

PR #2415, opened 2022-02-04
isort

janosh:
Add `isort` pre-commit hook and apply auto-fixes across code base.


utf:
I think you should run isort with the black preset:

```
  - id: isort
    args:
    - --profile=black
```

PR #2417, opened 2022-02-04
Remove last remaining star imports

janosh:
`flake8` complains about star imports (for good reason imo, they make code harder to understand and debug).

One step closer to enabling `flake8` in CI.

I noticed one odd thing, 

```
pymatgen/io/abinit/netcdf.py
pymatgen/io/abinit/pseudos.py
```

both define `class AbinitHeader(dict)`. The 2nd starred import was overwriting the class from the first file in `pymatgen/io/abinit/__init__.py` so I imported that one for backwards compat.

### DIFF
--- a/pymatgen/analysis/chemenv/utils/scripts_utils.py
+++ b/pymatgen/analysis/chemenv/utils/scripts_utils.py
@@ -10,16 +10,6 @@ import re
 
 import numpy as np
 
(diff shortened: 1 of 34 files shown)


shyuep:
Interesting. I didn't know that they introduced which in Python 3.3. I coded that which command back in Python 2.7+.

PR #2435, opened 2022-02-16
Remove unused mypy ignores

janosh:
Might want to add

```cfg
[mypy]
warn_redundant_casts = true
warn_unused_ignores = true
check_untyped_defs = true
```

 to `setup.cfg` in future though currently emits too many type errors to deal with.

### DIFF
--- a/docs_rst/usage.rst
+++ b/docs_rst/usage.rst
@@ -97,8 +97,8 @@ add these keys if they are not present, but for better long term stability
 through the encoder), the easiest way is to add the following to any to_dict
 property::
 
-    d["@module"] = self.__class__.__module__
-    d["@class"] = self.__class__.__name__
+    d["@module"] = type(self).__module__
+    d["@class"] = type(self).__name__
 
 To use the MontyDecoder, simply specify it as the *cls* kwarg when using json
 load, e.g.::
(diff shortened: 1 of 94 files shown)


janosh:
@shyuep Given how slow `mypy` is, is there a good reason for deleting the cache on every run?

https://github.com/materialsproject/pymatgen/blob/9bdf9fb8eebf4bef66f8a3541cfd27ba8ee1bccf/.github/workflows/lint.yml#L28-L32

shyuep:
When it is cached, the results may be inaccurate. In any case, the speed is not a big limiting factor for CI.

janosh:
CI error seems unrelated?

```py
def predict_mp(self, model_name, mp_id):
        """
        Predict using the http://megnet.crystals.ai/ API.
        """
        response = None
        url = self.url + f"/predict_mp/{model_name}/{mp_id}"
        try:
            response = self.session.get(url)
            if response.status_code in [200, 400]:
                return response.json()
            raise ValueError(f"REST query returned with error status code {response.status_code}")
        except Exception as ex:
            msg = f"{ex}. Content: {response.content}" if hasattr(response, "content") else str(ex)
>           raise ValueError(msg)
E           ValueError: REST query returned with error status code 503.  Content: ...
```

shyuep:
The CI errors are related to the MP website being down.

PR #2439, opened 2022-02-21
added wswq parser

jmmshn:
# WSWQ Parsing

VASP has a "secret" file WSWQ representing the matrix elements of <W | S | W(Q)>
Where W is the plane-wave wavefunction of the present calculation and W(Q) is the wavefunction of some perturbed calculation.

These matrix elements can be used to calculate the finite-difference of the wavefunction overlaps as you distort as structure.
This has applications for evaluating the electron-phonon interactions as we have done in the NONRAD code:
https://github.com/mturiansky/nonrad

Having this parser here I think makes a bit more sense especially if someone wants to make a workflow involving this stuff.


mkhorton:
Thanks @jmmshn :)

PR #2440, opened 2022-02-21
Add pydocstyle pre-commit hook

janosh:
Brings pre-commit.ci more in line with GHA `lint.yml` so people can be more confident CI won't complain once checks pass locally.

Adds a few type hints along the way.


shyuep:
Thanks.

PR #2441, opened 2022-02-21
Add CITATION.cff

janosh:
GitHub supports adding citation metadata to repos. This PR moves instructions on how to cite `pymatgen` from readme into the `CITATION.cff` file which GH uses to then display citations in BibTeX and APA formats in the side bar.  See [docs for details](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files).

Looks like this:

Screen Shot 2022-02-21 at 19 24 57

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,42 @@
+
+cff-version: 1.2.0
+message: |
+  If you use pymatgen in your research, please consider citing the following:
+
+  In addition, some of pymatgen's functionality is based on scientific advances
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks. But let's still keep the readme citation as well since some people may not be aware of the CFF file.

janosh:
> Thanks. But let's still keep the readme citation as well since some people may not be aware of the CFF file.

Sure thing. Though its not just the file, it's also prominently displayed on the repo's landing page side bar.

PR #2448, opened 2022-02-24
Allow to switch off print of ChemEnv citation

JaGeo:
## Summary

I am adding a small parameter to switch off the ChemEnv citation print. This will make it easier to use ChemEnv in other program packages. I am also updating the citation of ChemEnv.

I am tagging @davidwaroquiers  here, so that he knows.

Let me know if I should do anything additionally.


shyuep:
I would say - why print citations at all? I don't think any other part of the code prints citations. I would suggest turning it off by default.

JaGeo:
I would be fine with this. I can remove it and add the reference for the technical paper in the documentation.

PR #2450, opened 2022-03-03
Add monty `cached_class` to other compatibility schemes

munrojm:
## Summary

This PR adds the `@cached_class` decorator from monty to the `MaterialsProject2020Compatibility` and `MaterialsProjectAqueousCompatibility` classes. This fixes slow decoding of large lists of entry object which greatly impacts the new `MPRester` client.

None

## Checklist

in the pull request title.


mkhorton:
Thanks @munrojm!

For the record, in private communication @munrojm noted that the decoding changes from ~7s to ~0.3s in our benchmark case of a set of entries in the `Fe-O` system, so a huge improvement.

janosh:
@mkhorton @munrojm fwiw, we just did performance testing using the new `cached_class` proposed in https://github.com/materialsvirtuallab/monty/pull/691 comparing to the old or no `cached_class` and the performance stays basically unchanged in all 3 cases.


```py
from mp_api.client import MPRester

%%timeit
MPRester().get_entries("Fe-O")
>>> 1.17 s ± 181 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

@mkhorton do you remember what kind of testing you did? perhaps changes in the `mp_api` mean the `cached_class` no longer offers any speed benefits? in which case removing it altogether from `MaterialsProject2020Compatibility` would also solve @CompRhys's pickling issue reported in https://github.com/materialsproject/pymatgen/issues/3898

PR #2451, opened 2022-03-05
Type hints

janosh:
Add some more type hints in `electronic_structure/plotter.py`, `bandstructure.py`, `io/vasp/outputs.py`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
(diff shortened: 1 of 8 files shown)


arosen93:
@janosh -- nice update to the typehints. I have a general question based on this. For something like `kwarg = None`, which is better programming practice: `kwarg: int = None` or `kwarg: int | None = None`?

janosh:
@arosen93 Thanks! `kwarg: int | None = None` is strictly speaking more correct but unless you set `mypy`'s [`--no-implicit-optional`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional) flag (which I usually don't) then the two are equivalent.

arosen93:
Good to know, thanks! :)

mkhorton:
Thanks!

PR #2452, opened 2022-03-05
Future annotations

janosh:
Convert annotations to future syntax available through import in Python 3.7 upwards. [See PEP 563](https://www.python.org/dev/peps/pep-0563/#id6).


janosh:
Note to future `pymatgen` maintainers: `pymatgen` makes heavy use of variadic args (`**kwargs` in particular, 438 hits occurrences across the code base, and another 57 `*args`) which unfortunately can't yet be annotated. There's a `mypy` issue https://github.com/python/mypy/issues/4441 from 2018 about this limitation still unclosed. Though it looks like the recently accepted PEP 646 slated for py3.11 with its `Unpack` type might make it easier to allow using `TypedDict`s to annotate `**kwargs` (at least `pyright` seems keen to try it: https://github.com/microsoft/pyright/issues/3002). IMO definitely sth to explore once that becomes possible.

mkhorton:
Hi @janosh, thanks for the PR -- can you point me to where in the PEP the `x | None` syntax is preferred over `Optional[None]`?

I was reading elsewhere that the new preferred option for kwargs may be to omit the `Optional` entirely, e.g. just `your_kwarg: x = None`, and that `mypy` will respect this.

janosh:
> I was reading elsewhere that the new preferred option for kwargs may be to omit the `Optional` entirely, e.g. just `your_kwarg: x = None`, and that `mypy` will respect this.

@mkhorton That's right! As I was discussing with @arosen93 over in https://github.com/materialsproject/pymatgen/pull/2451#issuecomment-1060296902, `kwarg: int | None = None` and `kwarg: int = None` are equivalent unless you set `mypy`'s [`--no-implicit-optional`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional) flag.

janosh:
@mkhorton Regarding your 1st question, I believe [604 is the PEP](https://peps.python.org/pep-0604/#proposal) you're looking for.

> We will also be able to write `t | None` or `None | t` instead of `Optional[t]`:
> 
> ```py
> isinstance(None, int | None)
> isinstance(42, None | int)
> ```

mkhorton:
Thanks, so just to summarize:

* `Optional[x]` and `x | None` are equally accepted, with neither specifically preferred (presumably the latter is easier since it skips the import)
* The implicit optionals, though accepted by `mypy` by default, are not currently "officially" preferred (i.e., no guidance from Python/PEPs on this)

Happy to merge this regardless since it looks cleaner, I just want to make sure we have guidelines for future PRs.

janosh:
Yes, whether you allow implicit optionals is up to personal preference afaik. Seems more pythonic to me so I would leave it enabled.

> I just want to make sure we have guidelines for future PRs.

We already do actually. This will all be handled automatically by pre-commit.ci in the future since we're using the `pyupgrade` hook. It will auto rewrite all type hints to use the new leaner syntax in files that have the `__future__`'s import.

The only thing left to decide really is whether that `from __future__ import annotations` should be added to all files automatically.

PR #2463, opened 2022-03-18
Fix ValueError when cations, anions are not detected (LobsterEnv)

JaGeo:
## Summary

This fixes a problem in the tool to determine coordination environments based on a Lobster analysis. In case, there are no cations, anions in the compound, a more useful error message is raised. In such a case, cation-anion bonds cannot be used to determine coordination environments. I have also added tests for this case.


shyuep:
Thanks.

PR #2467, opened 2022-03-22
Allow for nearest neighbor algorithms other than VoronoiNN in get_surface_sites

arosen93:
The `get_surface_sites()` function when dealing with surfaces is hard-coded to use `VoronoiNN`. However, this is no particular reason to do so, and in fact it would be nice to let the user choose their own preferred method (e.g. `CrystalNN`). This PR gives the user that option with a new kwarg called `local_env`.


arosen93:
Ah, I see this is using the weight VN polyhedra. This is a bit too much work for me to want to mess around with it. Abandoning for now.

PR #2475, opened 2022-03-28
Cp2k 1.0

nwinner:
## Summary

First major update to the CP2K IO module. Unless some changes are desired by devs. I think I'll title this v1.0 for the module.

inputs.py:
* Feature 1: New SectionList allows for repeated section names. Analogous to KeywordList
* Feature 2: New strict set action for less error prone custodian modifications.
* Feature3: Kpoints can know be generated automatically using SpacegroupAnalyzer. While CP2K generally doesn't use kpoints, this is convenient for the few times where you do.

* Fix 1: Resolved input file parsing issues for list of sections and for potential info in restart files

sets.py
* Feature 1: Updated defaults to align with expected usage of CP2K better. 

outputs.py:
 * Features 1: Parse calculation type

 * Fix 1: Resolved errors with stress and ionic step parsing
 * Fix 2: Fixed issues in DOS parsing. 
 * Fix 3: Previously would fail for metallic systems, since they are not that common in cp2k calculations
 
utils.py:
* Feature 1: Updated to align with new set. 
* Feature 2: new truncation radius function to correct for user error in setting radius for hybrid functionals.  

* No new dependencies, but this version changes several defaults as compared to CP2K internal defaults. This version assumes a modern version of CP2K is being used, and so better flags can be set than the legacy defaults.

- [ x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). The easiest way to handle this
- [ x] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [ x] Tests have been added for any new functionality or bug fixes.
- [ x] All linting and tests pass.


nwinner:
Linting says that X | Y syntax requires python 3.10. This syntax was added in by Shyue, so I think that it should be alright. Please let me know if I should revert.

shyuep:
I think this problem can be solved by using `from __future__ import annotations`?

nwinner:
Oh you're right I didn't realize I'd removed that. Added it back in and also replaced ruamel with the newer import.

PR #2476, opened 2022-03-29
Make `Composition` and `ChemicalPotential` special binary methods return `NotImplemented`

janosh:
Follow up to https://github.com/materialsproject/pymatgen/pull/2291#discussion_r746839280.

Quoting the docs:

> [NotImplemented](https://docs.python.org/3.6/library/constants.html#NotImplemented) should be returned by the binary special methods (e.g. __eq__(), __lt__(), __add__(), __rsub__(), etc.) to indicate that the operation is not implemented with respect to the other type

I also added tests for this.


pymatgen/core/tests/test_composition.py
```diff
@@ -395,6 +401,10 @@ def test_equals(self):
         )
```
janosh, line 401: @shyuep While updating the tests, I found this odd `random.randint` code in `CompositionTest.test_equals()`. I don't understand what it's doing.

Couldn't that whole block be replaced by this?

```py
        c1, c2 = self.comp[:2]
        self.assertTrue(c1 == c1)
        self.assertFalse(c1 == c2)
```

janosh: I see. I'll add a comment for future readers.

PR #2478, opened 2022-03-30
[WIP] Relaxation Transformation

nwinner:
Sometimes, we have calculations with continue from a previous calculation. For example, you run a coarse relaxation followed by a fine relaxation in separate fireworks. In that case, it would be convenient to have a Transformation which represents that relaxation that can be part of the second calculation's transformation history, rather than starting from the coarse-relaxed structure.

Perhaps this should be named something else, though? Relaxation is probably the most common potential use case, but it could be more general perhaps. Open to suggestions.

### DIFF
--- a/pymatgen/transformations/standard_transformations.py
+++ b/pymatgen/transformations/standard_transformations.py
@@ -1131,7 +1131,7 @@ class RelaxationTransformation(AbstractTransformation):
         """
         self.final_structure = final_structure
         self.metadata = metadata if metadata else {}
-        
+
     def __str__(self):
         return "RelaxationTransformation"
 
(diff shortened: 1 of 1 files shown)


shyuep:
Hmmm... this is an odd transformation since it does not have any rules associated with it. All it does is to log that you have performed a process and this is the final structure. While conceptually I see the reason for it, I definitely wouldn't call it RelaxationTransformation and I would definitely suggest including metadata related to the transformation - at minimum, a name plus a dict of metadata. Maybe ExternalTransformation or MetaTransformation? The name can be something like "VaspRelaxation", and the dict can be the parameters. I can definitely see other use cases like "ExtSlabGenerator" or "MDJiggle", etc. I guess this is the mother of transformations that does some arbitrary thing to the structure that is not easily encoded in a series of Python steps.

janosh:
Closing this as outdated. @nwinner Feel free to ping for reopen if you want to pick this up again.

PR #2482, opened 2022-03-31
Fix `{StructureGraph,MoleculeGraph}.sort()`

janosh:
Plus add tests for `{StructureGraph,MoleculeGraph}.sort()`.


mkhorton:
Thanks @janosh !

PR #2483, opened 2022-04-02
Fix errors on entries_charge/entries_discharge attributes

peikai:
The results that entries_discharge and entries_discharge attributes returned in _pymatgen/apps/battery/conversion_battery.py_ were confused. I fixed this problem that entries in charge/discharged state should be the entries from step1/step2, respectively.

We can find evidence in similar attribute, vol_discharge, that properties in discharge states should be retrieved from the step2.
```python
        vol_discharge = sum(
            abs(curr_rxn.get_coeff(e.composition)) * e.structure.volume
            for e in step2["entries"]
            if e.composition.reduced_formula != working_ion
        )
```

### DIFF
--- a/pymatgen/apps/battery/conversion_battery.py
+++ b/pymatgen/apps/battery/conversion_battery.py
@@ -374,8 +374,8 @@ class ConversionVoltagePair(AbstractVoltagePair):
         frac_discharge = totalcomp.get_atomic_fraction(Element(working_ion))
 
         rxn = rxn
-        entries_charge = step2["entries"]
-        entries_discharge = step1["entries"]
+        entries_charge = step1["entries"]
+        entries_discharge = step2["entries"]
 
         return ConversionVoltagePair(  # pylint: disable=E1123
             rxn=rxn,


shyuep:
Pls add a test to check for this bug.

peikai:
> Pls add a test to check for this bug.

We can test discharge profiles of the entry mp-246 (TiF3) with these codes:

```python
composite_A_list = [entry.composition.reduced_formula for entry in voltage_pair.entries_discharge]
composite_B_list = [entry.composition.reduced_formula for entry in voltage_pair.entries_charge]
```
which is only two steps discharge based on MP Database V2021.05.13,

| reaction                                            | composite_A  | composite_B  |
|----------------------------------------|----------------|----------------|
| 1.5 Li + TiF3 -> 0.5 Li3TiF6 + 0.5 Ti  | ['TiF3']            | ['Li3TiF6', 'Ti'] |
| 1.5 Li + 0.5 Li3TiF6 -> 3 LiF + 0.5 Ti | ['Li3TiF6', 'Ti'] | ['Ti', 'LiF']        |

Obviously, the composite_A should be the composite in charged state, and composite_B is the composite in discharged state. However, the bug is that the composite_A actually comes from voltage_pair.entries_discharge in the codes above. 

Hence, I found the aforementioned bug and fixed it in this PR.

shyuep:
Thanks. What I meant is that since you detected this bug, a unittest should be added to the test_conversion_battery.py to check that this fix is done properly. It also ensures that future people will not make changes that result in this bug again. Thanks.

peikai:
I have found out the meaningfulness of a unittest. Now, a test function has been added to check phase components at each discharge step. This unittest will be able to notice the above bug. Thanks!

peikai:
I have found out the meaningfulness of a unittest. A test function have been added to check phase components at each discharge step. This unittest will be able to notice the above bug.

shyuep:
Thanks.

Issue #2485, opened 2022-04-05
Vasp 6.3 and new units for dipole moments


fraricci:
I noticed that (at least) version 6.3 of VASP has changed the units for the dipole (LCALC=True) from, i.e.:
Ionic dipole moment: p[ion]=(     0.00000     0.00000    19.70306 ) **electrons** Angst
to
Ionic dipole moment: p[ion]=(     0.00000     0.00000    -19.70306 ) **|e|** Angst

This means that these vectors needs to be multiplied by -1 to recover previous values.
In particular, to get the correct value of the polarization (as computed in the analysis.ferroelectricity.polarization),
the electronic dipole need to be multiplied by -1, since the ionic part is computed separately (p_ion is not from outcar).

I propose a fix in this pull request  #2486

But, any comments and suggestion are welcome.

mkhorton:
Thanks for catching this! Concerning to have such a big change that wasn't in the release notes / change log.

PR #2486, opened 2022-04-05
fix dipole moments units

fraricci:
Fix for the new units in recent vasp versions (at least 6.3, maybe even before).

Discussed in #2485


mkhorton:
Could you add a quick test to this @fraricci? Then will merge ASAP

mkhorton:
Thanks @fraricci! Looks like test failures are unrelated (CrystalAIResterTest.test_query) so will go ahead and merge

PR #2492, opened 2022-04-13
Add pylint pre-commit hook

lan496:
## Summary

[The linter workflow](https://github.com/materialsproject/pymatgen/blob/master/.github/workflows/lint.yml) in GitHub actions uses `pylint`, but `.pre-commit-config.yaml` did not call it. This PR adds `pylint` in `.pre-commit-config.yaml`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: ^(docs|.*test_files|cmd_line|dev_scripts)
 
 default_language_version:
-  python: python3.8
+  python: python3
 
 ci:
   autoupdate_schedule: monthly
@@ -54,3 +54,18 @@ repos:
     rev: v0.942
     hooks:
       - id: mypy
+
+  # You need to install pylint locally.
+  # See https://pylint.pycqa.org/en/latest/user_guide/pre-commit-integration.html
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args:
+          [
+            "-sn",
+            "--rcfile=pylintrc",
+          ]


shyuep:
Thanks.

janosh:
`pylint` doesn't play nicely with `pre-commit` for the reasons discussed in https://github.com/PyCQA/pylint/issues/5293. I don't think adding it to the `.pre-commit-config.yaml` is a good idea as it means everyone who has installed or will install commit hooks on the pymatgen repo will now encounter this error when trying to commit:

```
pylint...................................................................Failed
- hook id: pylint
- exit code: 1

Executable `pylint` not found
```

The situation before was that people would only get an error in CI if the code they added offended `pylint`.

lan496:
@janosh Hi. Thanks for pointing out the discussion on `pylint`.
I understand your concern is the situation that people who are unaware or forget `pip install -r requirements-dev.txt` encounter the error when trying to commit. I agree that this PR additionally requires people to install `requirements-dev.txt`, but I think it is a good thing to prepare linters in their local. So, I prefer adding instructions in [Coding Guidelines](https://pymatgen.org/contributing.html) or introducing extra dependencies (like `pip install -e ".[dev]"`) to removing `pylint` from `.pre-commit-config.yaml`.

janosh:
@lan496 Adding setup instructions for contributors is definitely an improvement. But in practice those are often not consulted and even if, they would still increase the burden on contributors. Given how many community contributions `pymatgen` receives, I think it's very valuable to keep the barrier to entry as low as possible.

I'm not a `pylint` user myself so don't know if there are any downsides to this approach, but I think it should work to just change these two lines:

```diff
  - repo: local
    hooks:
      - id: pylint
        name: pylint
        entry: pylint
-       language: system
+       language: python
        types: [python]
        args: [-sn, --rcfile=pylintrc]
+       additional_dependencies: [pylint]
```

That will make pre-commit install `pylint` automatically.

Also, since `pylint` has been running on this project continuously, there are no legacy violations. So with this change, I think there wouldn't be a need to skip the `pylint` hook in pre-commit.ci.

PR #2501, opened 2022-04-20
No complex

jmmshn:
# Removed complex numbers from the definition of WSWQ
- split the complex numbers into a real and imaginary parts
- added test for proper orthonormality of WSWQ


mkhorton:
Thanks @jmmshn! More descriptive PR titles would be appreciated though :) 

Was this because of issues with serialization when leaving them as complex?

jmmshn:
> Was this because of issues with serialization when leaving them as complex?

Sorry for the short title.  Yes, I'm trying to make this more JSON friendly.

PR #2505, opened 2022-04-25
FEFF: support Molecule as input

rkingsbury:
## Summary

This PR includes minimal changes to FEFF I/O to support the use of non-periodic input structures (e.g., `Molecule` rather than `Structure` objects). Tested with `MPXANESSet`. This also includes a fix for an edge case in which there is only a single absorbing atom present. When that occurs, some changes were necessary to ensure that the single absorbing atom does not get listed twice in the POTENTIALS section of the input file.


rkingsbury:
This is ready for review. Note that there is one none-deterministic test failure that sometimes occurs (see below)

```
E       AssertionError: 'PATH[72 chars]6884 1 Fe\n-5.133990 -0.916498 -5.379918 1 Fe\[188 chars]0 Fe' != 'PATH[72 chars]6884 2 Fe\n-5.133990 -0.916498 -5.379918 2 Fe\[188 chars]0 Fe'
E       Diff is 831 characters long. Set self.maxDiff to None to see it.

pymatgen/io/feff/tests/test_inputs.py:411: AssertionError
```

This is because the potential indices of Fe and O sometimes get reversed (where Fe is 2 and O is 1, but sometimes they don't. It's not completely clear to me whether the changes in this PR have triggered the non-determinism (I don't think so), but perhaps we can restructure the test to make it more robust? I'm open to suggestions there.

rkingsbury:
This failure also appears to be non-deterministic:

```
______________________ FeffInputSetTest.test_get_feffPot _______________________

self = <feff.tests.test_sets.FeffInputSetTest testMethod=test_get_feffPot>

    def test_get_feffPot(self):
        POT = str(self.mp_xanes.potential)
        d, dr = Potential.pot_dict_from_string(POT)
>       self.assertEqual(d["Co"], 1, "Wrong symbols read in for Potential")
E       AssertionError: 2 != 1 : Wrong symbols read in for Potential

pymatgen/io/feff/tests/test_sets.py:61: AssertionError

```

rkingsbury:
I think I've found a fix for the non-deterministic test failures. Please hold off on merging though, as I want to also double check handling of charged clusters (and whether the `ION` card should be set when there is a charge). I have an email in to one of the devs about this and will follow up soon.

rkingsbury:
OK, ready for review! I have gotten clarification about how calculations involving charged structures should be set up. Full support for that may be a bit involved, so in this PR I have opted to 1) detect when charged structures are used to instantiate an InputSet and 2) raise an exception or a warning as appropriate to alert the user that some manual editing is needed in that case. I've also added in the code comments what is supposed to be done.

I will experiment with adding support for automatic `ION` tags but I think that's best for a separate PR.

rkingsbury:
Friendly reminder about this PR, which is ready for review. If any changes are needed, I'd like to do them soon b/c I have shifted to other kinds of calculations and my memory about the details of FEFF IO will become fuzzy soon :)

Regarding support for charged `Molecule`:

> I will experiment with adding support for automatic `ION` tags but I think that's best for a separate PR.

Instead of waiting for a separate PR, I have added here minimal support for ION tags via `user_tag_settings`, with an associated warning message that is triggered if 1) the input `Molecule` has a net charge and 2) `the user has not supplied the IONS key in `user_tag_settings`. The changes correctly produce multiple `ION` tags in the output file when `IONS` is included in the `user_tag_settings`.

mkhorton:
Thanks @rkingsbury, apologies for the delay with review -- features are very welcome.

Issue #2506, opened 2022-04-26
Corruption of `.pmgrc.yaml` still happening


janosh:
Pinging @shyuep as I think this was supposed to be fixed in https://github.com/materialsproject/pymatgen/issues/2421.

Using `MPRester` still corrupts `.pmgrc.yaml`:

```yaml
dictitems:
  MAPI_DB_VERSION:
    LAST_ACCESSED: '2020_09_08'
    LOG: {'2020_09_08': 1}
  PMG_DEFAULT_FUNCTIONAL: PBE_54
  PMG_MAPI_KEY: **********************
  PMG_VASP_PSP_DIR: /Users/janosh/dev/vasp/pseudo-potentials
state:
  _yaml_format: !!python/object/new:ruamel.yaml.comments.Format
    state: !!python/tuple
    - 
    - {_flow_style: false}
  _yaml_line_col: !!python/object:ruamel.yaml.comments.LineCol
    col: 0
    data:
      PMG_DEFAULT_FUNCTIONAL: [0, 0, 0, 24]
      PMG_MAPI_KEY: [1, 0, 1, 14]
      PMG_VASP_PSP_DIR: [2, 0, 2, 18]
    line: 0
MAPI_DB_VERSION:
  LOG:
    '2020_09_08': 2
  LAST_ACCESSED: '2020_09_08'
```

leading on subsequent `MPREster` calls to

```py
pymatgen.ext.matproj.MPRestError: REST query returned with error status code 403
```

Is this due to an outdated `ruamel.yaml` on my machine?

```
pip list | grep ruamel  
ruamel.yaml                     0.17.20
ruamel.yaml.clib                0.2.6
```

If so, should update the min version in `setup.py`:

```py
"ruamel.yaml>=0.15.6"
```

shyuep:
Well, I searched the entire code for any dump commands using Pycharm. The only dump commands operating on the pmgrc are in pmg_config and MPRester. These are the only places where corruption can occur.

mkhorton:
This is/was definitely due to a change in behavior in `ruamel.yaml`.

Specifically, data from .pmgrc.yaml was now being loaded as non-native Python data types, and then when written back out, was written out with a lot of extra metadata to preserve these non-native data types. In my view this is a bug, because the .pmgrc.yaml should have been loaded as a `dict` and this was the original behavior.

[...]

PR #2507, opened 2022-04-26
Report exceptions in `MPRester.__init__()` when trying to read/write `SETTINGS_FILE`

janosh:
Relates to #2506.

Changes `MPRester.__init__()` to print exceptions when trying to read/write `SETTINGS_FILE` instead of swallowing them silently. This might help identify the underlying issue leading to `.pmgrc.yaml` corruption.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,9 +62,5 @@ repos:
         entry: pylint
         language: python
         types: [python]
-        args:
-          [
-            "-sn",
-            "--rcfile=pylintrc",
-          ]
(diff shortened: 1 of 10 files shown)


shyuep:
This has been superseded by recent changes.

janosh:
@shyuep The changes were largely orthogonal. You don't think it would be a good idea to catch `TypeError` correctly when trying to read an empty `MP_LOG_FILE` (formerly `SETTINGS_FILE`)?

janosh:
That was my motivation as well except I thought it'd be more likely to get fixed properly when there's a warning than when it fails silently.

PR #2509, opened 2022-04-27
fix bug with nmr set writing units

### LINKED ISSUE #2508: QUAD_EFG includes units for efg mode of MPNMRSet
**Describe the bug**
The MPNMRSet (with mode="efg") writes units to file for the QUAD_EFG setting. This occurs because it is using FloatWithUnit class under the hood.

**To Reproduce**
``` python
from pymatgen.io.vasp.sets import MPNMRSet

# Random structure to test with
s = Structure(
    lattice=[
        [3.485437, 0.0, 2.012318],
        [1.161812, 3.286101, 2.012318],
        [0.0, 0.0, 4.024635],
    ],
    species=["Na", "Cl"],
    coords=[[0.0, 0.0, 0.0], [0.5, 0.5, 0.5]],
)

MPNMRSet(s, mode="efg").incar.write_file("INCAR")
```

This produces the following INCAR:
```
ALGO = FAST
EDIFF = -1e-10
ENCUT = 520
IBRION = -1
ISIF = 3
ISMEAR = -5
ISPIN = 2
ISYM = 0
LAECHG = True
LASPH = True
LCHARG = False
LEFG = True
LORBIT = 11
LREAL = Auto
LVHAR = True
LWAVE = False
MAGMOM = 2*0.6
NELM = 100
NELMIN = 10
NSW = 0
PREC = ACCURATE
QUAD_EFG = 104.1 mbarn -81.65 mbarn   # <------------ Note the bug here
SIGMA = 0.01
```

**Expected behavior**

The INCAR should instead have the following result for QUAD_EFG:
```
QUAD_EFG = 104.1 -81.65
```

**Desktop:**
 - OS: Linux
 - Version 2022.4.19

**Additional context**
This is a quick, one-line fix. We need to just convert from FloatWithUnit to Float here:
https://github.com/materialsproject/pymatgen/blob/1b6d1d2212dcf3a559cb2c489dd25e9754f9f788/pymatgen/io/vasp/sets.py#L1942

jacksund:
resolves #2508

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -1939,7 +1939,9 @@ class MPNMRSet(MPStaticSet):
 
             isotopes = {ist.split("-")[0]: ist for ist in self.isotopes}
 
-            quad_efg = [Species(p).get_nmr_quadrupole_moment(isotopes.get(p, None)) for p in self.poscar.site_symbols]
+            quad_efg = [
+                float(Species(p).get_nmr_quadrupole_moment(isotopes.get(p, None))) for p in self.poscar.site_symbols
+            ]
 
             incar.update(
                 {


shyuep:
Please add a Unittest to catch this bug and prevent it from recurring.

jacksund:
Yep, I'm working on that now :smile:

jacksund:
I was trying to add a lower-level check to all float-type INCAR settings, but am unsure the best place to do this. Maybe this would be in the `get_string` method of the `Incar` class? Might be something to consider in order to avoid issues with other vasp sets.

https://github.com/materialsproject/pymatgen/blob/6c84eb53f00de42dc2110ce39f3ce4c083c5e0a9/pymatgen/io/vasp/inputs.py#L693

And the addition would be something like...
``` python
elif isinstance(self[k], float):
    # note the added float() which ensures FloatWithUnit conversion
    lines.append([k, " ".join([str(float(i)) for i in self[k]])])
# plus another check for a list of floats
```

shyuep:
Thanks.

PR #2514, opened 2022-05-04
Replace IALGO with ALGO in MVLNPTMDSet 

qianchenqc:
## Summary

* Replaced the IALGO tag with ALGO as recommended in the vasp documentation https://www.vasp.at/wiki/index.php/IALGO. 
* Removed IALGO tag to enable Davidson first and then switch to the RMM-DIIS with ALGO=Fast.


shyuep:
Thanks.

PR #2517, opened 2022-05-10
Fixed MSONable + InputSet's str(object)

davidwaroquiers:
## Summary

* Added functionality to write the __str__ contents of an object in the contents of one of the InputSet's files, if this object is not an instance (subclass) of InputFile.
* Added unit tests for InputFile and InputSet.
* Fixed MSONable for InputSet when kwargs are given.

None


shyuep:
Thanks.

PR #2518, opened 2022-05-11
Fix reading of orbitalwise ICOHP

JaGeo:
## Summary
By accident, the wrong line in ICOHPLIST.lobster was read to assess whether orbitalwise interactions are included in these files. This lead to erros in case the first atom only has one relevant valence orbital. I fixed the problem. And, I included a test to fix this for the future. 

JG


JaGeo:
From my side, this would be ready. Let me know if I can do anything else.

shyuep:
Thanks.

PR #2522, opened 2022-05-13
fixed deserialization of phase diagrams

jmmshn:
JSON reconstruction bug:
The problem is: Because MSONables from_dict was replaced, there is no recursive reconstruction of the objects when you hit `el_ref` .  When you do `as_dict` -> `from_dict`  in python everything looks fine.
but the object was never completely serialized.
If you completely serialized it like putting it into as json file or database then you cannot reconstruct the phase diagram.


mkhorton:
Thanks @jmmshn!

PR #2539, opened 2022-06-01
small bug in gruneisen.py

antonf-ekb:
At converting adt from K to Hz we get numbers of the order 10^12 (TeraHz) but since w are not yet multiplied by 10^12, the value adt * const.value ("Boltzmann constant in Hz/K") should be divided by the const.tera. The error lead to that all values of w satisfy the condition while this should not be the case.

## Summary

* Feature 1
* Feature 2
* Fix 1
* Fix 2

## TODO (if any)

to be done

* Feature 1 supports A, but not B.

## Checklist

in the pull request title.

### DIFF
--- a/pymatgen/phonon/gruneisen.py
+++ b/pymatgen/phonon/gruneisen.py
@@ -107,7 +107,7 @@ class GruneisenParameter(MSONable):
 
         if limit_frequencies == "debye":
             adt = self.acoustic_debye_temp
-            ind = np.where((w >= 0) & (w <= adt * const.value("Boltzmann constant in Hz/K")/const.tera))
+            ind = np.where((w >= 0) & (w <= adt * const.value("Boltzmann constant in Hz/K") / const.tera))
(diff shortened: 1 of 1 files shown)


JaGeo:
Thank you for finding the bug. I think you are correct. We had developed/adapted this module within a Master thesis and hadn't used this option for further studies.

Would you be able to add a short test to avoid this bug in the future? Otherwise, @ab5424 or I would try to get back to it in one of the next weeks.

antonf-ekb:
> Thank you for finding the bug. I think you are correct. We had developed/adapted this module within a Master thesis and hadn't used this option for further studies.
> 
> Would you be able to add a short test to avoid this bug in the future? Otherwise, @ab5424 or I would try to get back to it in one of the next weeks.

Hi, I'm sorry, but I don't quite understand what test do you mean.

JaGeo:
Thank you very much again @antonf-ekb . As we have fixed the issue in another pull-request, this could be closed now, I think.

PR #2551, opened 2022-06-14
Add `flake8` plugin `bugbear` as pre-commit hook

janosh:
Just discovered [`flake8-bugbear`](https://github.com/PyCQA/flake8-bugbear) and tried it out on pymatgen. I think it makes sense to add it to `pre-commit`. it points out a lot of mutable function default values, unused loop control variables (stuff like `for i, val in enumerate(values): print(value)` where `i` is never used) and other code quality issues ([full list of warnings](https://github.com/PyCQA/flake8-bugbear#list-of-warnings)). Started fixing some of them in this PR but having it as a commit hook means we avoid adding new ones in the future.


janosh:
Something keeps shaving off the last line from `requirements.txt` and then pre-commit.ci adds it back. @shyuep What's that 1st thing doing the shaving?

shyuep:
No idea....

janosh:
Didn't realize this would turn into such a hornets' nest of `mypy` errors... 😅

janosh:
@shyuep pre-commit.ci complains about `CHANGES.rst` which I didn't touch so I think this is ready for review.

mkhorton:
Thanks! Looks like this caught some important things

PR #2552, opened 2022-06-17
Remove vertical lines from VoltageProfilePlotter.get_plotly_figure()

acrutt:
Change VoltageProfilePlotter.get_plotly_figure() so connecting vertical lines are not drawn in voltage profile. These vertical lines are misleading for interpreting voltage profile plots. 

@mkhorton this change was discussed with Kristin at the Jun 13 meeting on the current state of the battery code infrastructure. Please let me know if any revisions are needed but otherwise I think this PR is ready to merge.


shyuep:
Thanks.

PR #2563, opened 2022-06-30
Molecule: warn when writing charged mol to xyz

rkingsbury:
## Summary

`Molecule.to()` supports many formats including .xyz; however .xyz format does not store charge information; only atomic positions. For users not intimately familiar with details of the .xyz format, this can lead to unexpected behavior when working with charged `Molecule`, e.g.

```
>>> m = Molecule([Element("K")], [[0,0,0]], charge =1)
>>> m.charge
1
>>> m.to(filename='test.xyz')
>>> m2=Molecule.from_file('test.xyz')
>>> m2.charge
0
```

This PR simply adds a warning to alert the user that they will lose information when writing to .xyz. The warning is only raised when 1) the `Molecule` has a charge and 2) the user tries to write to .xyz format.

```
>>> m.to(fmt='test.xyz')
UserWarning: Your molecule has a charge of 1, but the XYZ format only stores atomic positions, not charge. 
Using Molecule.from_file() with this file will result in an uncharged Molecule.
'1\nK1\nK 0.000000 0.000000 0.000000'
```

Other minor changes to `test_structure.py` were made to satisfy the `pylint` pre-commit hooks

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -3219,6 +3219,12 @@ class IMolecule(SiteCollection, MSONable):
         fmt = "" if fmt is None else fmt.lower()
         fname = os.path.basename(filename or "")
         if fmt == "xyz" or fnmatch(fname.lower(), "*.xyz*"):
+            if self.charge != 0:
+                warnings.warn(
+                    f"Your molecule has a charge of {self.charge}, but the XYZ "
+                    "format only stores atomic positions, not charge. Using Molecule.from_file() "
+                    "with this file will result in an uncharged Molecule."
+                )
             writer = XYZ(self)
         elif any(fmt == r or fnmatch(fname.lower(), f"*.{r}*") for r in ["gjf", "g03", "g09", "com", "inp"]):
             writer = GaussianInput(self)
--- a/pymatgen/core/tests/test_structure.py
+++ b/pymatgen/core/tests/test_structure.py
@@ -1566,6 +1566,11 @@ Site: H (-0.5134, 0.8892, -0.3630)"""
             self.assertEqual(m, self.mol)
             self.assertIsInstance(m, IMolecule)
 
+        # test warning when writing charged molecule to .xyz
+        m_charged = Molecule([Element("K")], [[0, 0, 0]], charge=1)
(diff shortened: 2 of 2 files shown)


shyuep:
Again, I have to question whether this is necessary at all. Surely a user would know the XYZ format doesn't store charge and use something else if he wants the charge?

rkingsbury:
Because I think in general, the expectation throughout pymatgen is that you can use `to_file` / `from_file` methods to reversibly write out and reconstruct pymatgen objects. This is a particular case where that will not work, which is why I think it warrants an extra warning.

shyuep:
I don't actually think that statement is true. For example, you can specify all kinds of site properties for Structures. Except for the json format, most of the other formats do not preserve all that information. 

I get that you can add warnings all over the place. But it does contribute to noise when using the code. I think many of us have been irritated by this in the past both in pymatgen and other codes. E.g., when Excel tells you that saving in csv loses information as opposed to saving in xlsx. I have probably seen that warning more than a hundred times in my lifetime and it still bugs me today when I see it. 

In the end, the Python representations in pymatgen are by definition a superset of the features of specific output formats (ideally speaking). Other than the pickled/json serialized forms where we are explicit in supporting full reconstruction, all bets are off when you output to another format since support depends on the specifications of that format. Moreover, if you didn't realize XYZ doesn't support charge, you probably didn't need the charge until the day you actually found that out.

mkhorton:
The `XYZ` class does have a [docstring note](https://github.com/materialsproject/pymatgen/blob/c30082466191abcb9ca714a2d871a0dd8e92b13b/pymatgen/io/xyz.py#L23-L26) about losing periodicity, so perhaps reasonable to add an additional note here (and in the `.to` docstring) about charge, as an alternative?

shyuep:
I have no problems with a docstring comment. It is more warnings that I think we should make judicious use of. If anything, I am all for using Python logging to log some of these messages and people can set a logging.basicConfig(BEGINNERMODE) if they want all the noise. BTW, non-beginner people have been complaining about such noise from pymatgen and other codes.

shyuep:
Closing this. If docstring is desired, pls go ahead and implement and submit a new PR. Thanks.

PR #2565, opened 2022-06-30
Molecule: warn user about non-triplet O2

rkingsbury:
## Summary

When a user creates a `Molecule` to represent O2 with default keyword arguments, the spin_multiplicity will be set to 1. However, the ground state of the O2 molecule is triplet (unlike most molecules with a single covalent bond).

```
>>> mol = Molecule(["O"] * 2, [[0, 0, 0], [0, 0, 1.2]])
>>> m.spin_multiplicity
1
````

This PR adds a warning to remind users of this fact. The warning is only invoked when 1) the `Molecule` is O2 (not O4, or O8, etc.), 2) the charge is zero, and 3) the spin_multiplicity is not equal to 3. It does not change the internal spin multiplicity setting either; it merely alerts the user that they may have made a mistake.

```
>>> mol = Molecule(["O"] * 2, [[0, 0, 0], [0, 0, 1.2]])
UserWarning: You have specified a spin multiplicity of 1 for an O2 molecule; however the ground state of 
molecular O2 has a spin multiplicity of 3. If this was intentional, you can safely ignore this message.
>>> m.spin_multiplicity
1
````

This may be controversial; I recognize that we have to expect users to know what they are doing and can't anticipate every mistake or non-obvious behavior. That said, given how common O2 is to many different types of simulation work and that I'm sure the vast majority of users want the ground state when they make an O2 Molecule, I think it's valuable to include this warning to 1) help users catch their own mistakes (preventing wasted time and having to re-do calculations) and 2) educate less savvy users that this is something they need to think about.

Flagging @samblau and @mkhorton and @arosen93  for any additional thoughts

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -2782,7 +2782,7 @@ class IMolecule(SiteCollection, MSONable):
         if self.species == [Element("O"), Element("O")] and self.charge == 0 and self._spin_multiplicity != 3:
             warnings.warn(
                 f"You have specified a spin multiplicity of {self._spin_multiplicity} for "
-                "an O2 molecule; however the ground state of molecular O2 has a spin multiplicity of 3."
+                "an O2 molecule; however the ground state of molecular O2 has a spin multiplicity of 3. "
                 "If this was intentional, you can safely ignore this message."
             )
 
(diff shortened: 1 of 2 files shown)


samblau:
No harm in a warning. Seems good to me.

PR #2575, opened 2022-07-06
InputSet: implement equality method and fix __iter__

rkingsbury:
## Summary

This pull request fixes two items in `InputSet`:

1. `__iter__` was previously iterating an iterator over `self.inputs.items()` rather than `self.inputs`. This was not a correct implementation (see example for implementing a `dict`-like class from `MutableMapping` [here](https://stackoverflow.com/questions/21361106/how-would-i-implement-a-dict-with-abstract-base-classes-in-python)) and prevented `__eq__` from working properly
2. After fixing item 1), I realized that the built-in `__eq__` method was only comparing the `inputs` and not the kwargs. Therefore, I added a custom `__eq__` method to ensure that `InputSet` are considered equal only if both the `inputs` and all kwargs are equal.

Note that a small change in some of the pre-existing tests was required due to the changed behavior of `__iter__`.

Thanks @orionarcher  for identifying this bug. Also flagging @utf in case the changed behavior of `__iter__` will affect your new input sets in atomate2

### DIFF
--- a/pymatgen/io/core.py
+++ b/pymatgen/io/core.py
@@ -144,7 +144,7 @@ class InputSet(MSONable, MutableMapping):
         return len(self.inputs.keys())
 
     def __iter__(self):
-        return iter(self.inputs.items())
+        return iter(self.inputs)
 
     def __getitem__(self, key):
         return self.inputs[key]
@@ -154,6 +154,9 @@ class InputSet(MSONable, MutableMapping):
 
     def __delitem__(self, key):
         del self.inputs[key]
+
+    def __eq__(self, other):
+        return (self.inputs == other.inputs) and (self.__dict__ == other.__dict__)
 
     def write_input(
         self,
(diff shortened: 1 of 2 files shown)


rkingsbury:
> Thanks @rkingsbury, and thanks @orionarcher for reporting.
> 
> I'm going to merge this because both changes seem sensible, but if there's an issue please alert me and we can resolve.

Thanks @mkhorton . The only potential issue is that the change in behavior of `__iter__` might break existing code. However I don't think this new style `InputSet` has seen wide use yet (certainly not within pymatgen - it's only the packmol and `TemplateInputGen' afaik), and those tests all pass without modification. If others have developed `InputSet` outside of pymatgen and they happen to be iterating over the object, they would be affected. But other than @utf in atomate2 and @orionarcher in his OpenMM addon, I don't believe anyone has.

PR #2581, opened 2022-07-09
Add "External tools" section to `docs_rst/addons.rst`

janosh:
Add https://github.com/janosh/pymatviz to https://pymatgen.org/addons.

### DIFF
--- a/docs_rst/addons.rst
+++ b/docs_rst/addons.rst
@@ -1,19 +1,48 @@
-Pymatgen Add-ons
-================
+Pymatgen Add-ons and External Tools
+===================================
+
+Add-ons
+-------
 
 With effect from v2022.0.3, pymatgen, pymatgen.analysis, pymatgen.ext and and pymatgen.io are now
-`namespace packages <http://packaging.python.org/guides/packaging-namespace-packages/>`_. You may refer to the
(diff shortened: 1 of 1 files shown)


mkhorton:
The add-ons page is specifically for packages within the pymatgen namespace, so this is not appropriate here.

We can definitely add a page on complementary codes, or codes that use pymatgen, somewhere else in the docs though, perhaps under a heading "Third-party codes that use pymatgen" and list `pymatviz` there.

janosh:
A new list for 3rd party tools built with/for `pymatgen` sounds like a cool idea. @JaGeo Might be interested in listing https://github.com/JaGeo/LobsterPy and I'm sure there are many more.

@mkhorton Just to confirm, you're suggesting this as a new section on the same docs page?

mkhorton:
Ambivalent about location, same page sounds fine.

Maybe could be renamed to “Add-ons and External Tools” and have a new
subheading and a little text to clarify the distinction. Also add some text
to say that if anyone would like their own tool listed to submit a PR, and
also include a link to
https://github.com/materialsproject/pymatgen/network/dependents?package_id=UGFja2FnZS01MjQ5NTY2Nw%3D%3Dtoo
.

PR #2585, opened 2022-07-16
Added 'get_ir_reciprocal_mesh_map'-method

nheinsdorf:
## Summary

In physical simulations it is often desirable to make use of the lattice symmetries to save computational time. The 'get_ir_reciprocal_mesh' function (which is basically an interface to spglib) returns an irreducible mesh and the weight, which just tells how many other kpoints a given irreducible kpoint is mapped to. 

This function does the same, but instead of returning the irreducible kpoints and their weights it returns the full grid and the mapping that is used to generate the irreducible grid. This is useful if physical objects are calculated that do not include a summation over all kpoints, because in that case the weight of an irreducible kpoint is not enough and the mapping is needed to relate any kpoint to the irreducible mesh. 

## TODO (if any)

It would be nice if that function could also return a list of symmetry operations that show which symmetry maps a given k-point to its irreducible counterpart. This would be very useful if the properties calculated do not transform as scalars. 

## Checklist
Implemented a test that compares to the already tested 'get_ir_reciprocal_mesh'-method.

### DIFF
--- a/pymatgen/symmetry/analyzer.py
+++ b/pymatgen/symmetry/analyzer.py
@@ -386,6 +386,30 @@ class SpacegroupAnalyzer:
         for i, count in zip(*np.unique(mapping, return_counts=True)):
             results.append(((grid[i] + shift * (0.5, 0.5, 0.5)) / mesh, count))
         return results
+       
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @nheinsdorf, this looks useful!

PR #2584, opened 2022-07-15
Structure __eq__ bugfix

KirbyBroderick:
Ran into an obscure bug where `other` was a length-18 string and `self` was a structure with 18 atoms

## Summary

Include a summary of major changes in bullet points:

* Check that the structure being compared against has a lattice parameter before trying to access the lattice parameter.

## Additional dependencies introduced (if any)

* None

## TODO (if any)

None

## Checklist

Work-in-progress pull requests are encouraged, but please put [WIP]
in the pull request title.

Before a pull request can be merged, the following items must be checked:

- [ X] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). The easiest way to handle this
      is to run the following in the **correct sequence** on your local machine. Start with running
      [black](https://black.readthedocs.io/en/stable/index.html) on your new code. This will automatically reformat
      your code to PEP8 conventions and removes most issues. Then run
      [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/), followed by
      [flake8](http://flake8.pycqa.org/en/latest/).
- [ ] Docstrings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
      Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [ ] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to type check your code.
- [ ] Tests have been added for any new functionality or bug fixes.
- [ ] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more efficient if you already fix most
errors prior to submitting the PR. It is highly recommended that you use the pre-commit hook provided in the pymatgen
repository. Simply `cp pre-commit .git/hooks` and a check will be run prior to allowing commits.

PR #2588, opened 2022-07-21
Fix edge cases in Structure equality

### LINKED ISSUE #2587: `IStructure.__eq__` fails for integers
There are more edge cases like #2584 in `IStructure.__eq__`. E.g. comparing against `int`s also throws.

```py


coords = [[0.5, 0.5, 0.5], [0, 0, 0]]
species = ["Na", "Cl"]
lattice = Lattice.from_parameters(a=4, b=4, c=4, alpha=90, beta=90, gamma=90)
structure = Structure(lattice, species, coords)

structure == 42
```

```py
    def __eq__(self, other) -> bool:
        if other is self:
            return True
        if other is None:
            return False
>       if len(self) != len(other):
E       TypeError: object of type 'int' has no len()
```

I think a fix to handle all such edge cases would be to start with an `isinstance` check that returns `NotImplemented` if `other` is not a `(I)Structure`.

```py
    if not isinstance(other, IStructure):
        return NotImplemented
```

janosh:
Closes #2587.

Related #2584.

Also adds tests for `IStructure.__eq__()`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,24 +4,24 @@ Change log
 v2022.7.19
 ----------
 
-This will be the final release with the pymatgen.analysis.defects 
-module included in the standard pymatgen package. This release will 
-include the older defects code by default, but can also be replaced with 
+This will be the final release with the pymatgen.analysis.defects
+module included in the standard pymatgen package. This release will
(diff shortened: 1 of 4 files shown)


mkhorton:
Thanks @janosh!

PR #2590, opened 2022-07-22
Update absorption coefficient

yang-ruoxi:
### Summary
The previous absorption coefficient lack a 2*pi constant. It is corrected in this PR and the results compare well with experiments.


mkhorton:
Thanks for catching this @yang-ruoxi !

PR #2595, opened 2022-07-30
Correct doc description of electron affinity

755452800:
The electron affinity description is  duplicate with ionization energy. Updated the definition according to wiki (https://en.wikipedia.org/wiki/Electron_affinity).


shyuep:
thanks.

PR #2601, opened 2022-08-03
Remove `__ne__` methods

janosh:
Since they're auto-defined by python as `not __eq__()`.


janosh:
To quote [this SO answer](https://stackoverflow.com/a/30676267): 

> should I implement `__ne__()` operator based on `__eq__`?
> 
> Short Answer: Don't implement it, but if you must, use `==`, not `__eq__`

janosh:
Wait! Still need to fix a test. :)

shyuep:
Don't worry about it.

janosh:
I'm in a call atm, will do it after.

shyuep:
I already fixed it.

PR #2603, opened 2022-08-04
Add option for 90 degree "cubic" supercells

JaGeo:
## Summary

I have added an additional option to the CubicSupercellTransformation to only return cells with three 90 degree angles. I think this might be useful for tasks where supercells are required.

I have also added two test cases and documentation.


shyuep:
Thanks.

PR #2604, opened 2022-08-04
Change default number of significant digits to write VASP POSCAR file

henriquemiranda:
## Summary

* Changed the default number of significant digits to be used when writing the POSCAR file.

Writing the POSCAR with the current default can cause all kinds of different nasty issues in VASP for example:
- VASP might not be able to find the correct space group
- VASP might lose the symmetry during relaxation
- Might lead to a slightly different number of G vectors for symmetry equivalent k-points
- ...

https://www.vasp.at/wiki/index.php/Precision,_symmetry_and_POSCAR
These issues are particularly painful for beginners.
Since there is not much to be gained by using a low number of digits I propose to use 16 significant digits.

Then, when a user starts off with a POSCAR with low precision he can symmetrize the structure using spglib and write a new POSCAR with higher precision with a simple python script using pymatgen before starting the VASP calculation:
https://gist.github.com/henriquemiranda/e4a1b616693aac339ef011af6484f890


shyuep:
Thanks @henriquemiranda. This change breaks quite a number of tests. They need to be fixed before I can merge it.

henriquemiranda:
Ok, I see... I don't see where to start to fix those issues.
@shyuep Do you think any of the devs could help me?

shyuep:
Click on the failed tests and look at the logs. There will be error messages indicating what test failed. From that, you just need to correct the tests. Most of it is likely due to the fact that you changed the number of significant figures and that changed all the output file comparisons.

henriquemiranda:
Actually... It seems the error is triggered by a test where the lattice parameters must not match because of the precision of the POSCAR files...
https://github.com/materialsproject/pymatgen/blob/master/pymatgen/analysis/defects/tests/test_core.py#L33
That's certainly a useful test to have

henriquemiranda:
@shyuep I think I fixed all the tests that were failing because of the change in the precision of the POSCAR.
The checks here on GitHub actions are failing but I am not sure if its related to the changes in this PR.

shyuep:
Great. Thanks. I will merge now.

PR #2610, opened 2022-08-06
Correct link to Fleur website in addon entry for `pymatgen-io-fleur`

janssenhenning:
The link in the docs was mistakenly interpreted as relative inside the pymatgen documentation. Also added `rel/`, which will always point to the documentation for the latest version of the code


shyuep:
Thanks,

PR #2614, opened 2022-08-09
Class to handle Thermal displacements matrices

JaGeo:
## Summary

I have added a class and a function to handle and convert thermal displacement matrices from phonopy and potentially other codes. 
This is a partial re-implementation of this matlab code from my PhD https://github.com/JaGeo/MolecularToolbox/
Tests and documentation have been added.


JaGeo:
I don't think the failures come from my code. If there are any suggestions with regard to the structure, I am very happy to adapt it.

PR #2617, opened 2022-08-11
Small extension of Thermal displacement matrices

JaGeo:
## Summary

I have now also added an implementation of a quality criterion for computed thermal displacement matrices as defined in https://doi.org/10.1039/C9CE00794F 
Once can now also initialize the class from thermal displacement matrices as defined in the cif only. 
Very tiny fixes to the documentation have been added as well.

No further extensions are planned for now. Maybe in the upcoming months, some extensions will follow. I would especially be happy to have functionality to read thermal displacements from cif files and be able to connect this to this class.


JaGeo:
Messed something up with git. If the tests pass, everything should be fine from my side.

JaGeo:
Not sure what is going on with the other tests at the moment.

PR #2620, opened 2022-08-16
Added 'get_orbit_and_generators'-method to SpaceGroup to class.

nheinsdorf:
## SUMMARY

I added a method to the SpaceGroup class that returns both the orbit (like the 'get_orbit'-method does) as well as the symmetry operations that generate it. As opposed to 'get_orbit' the method makes sure that the first element of the orbit is the point that has been passed to the function. Therefore the first element in the list of generators that are returned is always the identity operation. 

A test making sure that this is always true has been implemented. 

The above functionality is useful if a physical property (e.g. angular momentum) that needs to obey the crystal symmetry is assigned to a given site. The user can then generate the corresponding  properties of the other sites of the orbit by using the generators of the orbit. 

## TODO 

It would be probably nice to implement this for PointGroup as well

### DIFF
--- a/pymatgen/symmetry/groups.py
+++ b/pymatgen/symmetry/groups.py
@@ -190,6 +190,7 @@ class PointGroup(SymmetryGroup):
                 orbit.append(pp)
         return orbit
 
+
 @cached_class
 class SpaceGroup(SymmetryGroup):
     """
--- a/pymatgen/symmetry/groups.py
+++ b/pymatgen/symmetry/groups.py
@@ -190,7 +190,6 @@ class PointGroup(SymmetryGroup):
                 orbit.append(pp)
         return orbit
 
-
 @cached_class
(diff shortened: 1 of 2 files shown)


shyuep:
Great. Thanks.

mkhorton:
Missed this PR when it was opened, also thanks! Super useful

On a test note, there's a subtle point about whether tests should be fully deterministic or not. Having randomness in a test can be useful (e.g., see [hypothesis](https://hypothesis.readthedocs.io/) for a formalized version of this) but can also make the test failures difficult to re-produce if there is a bug.

If using this approach in a future PR, one useful thing to do would be to add additional info to the `assert` statement, e.g. here:

`self.assertLessEqual(len(orbit), sg.order, f"{p=}")`

This way, if this `assert` fails, we still see the value of `p` so we can go back to reproduce the bug.

PR #2622, opened 2022-08-17
Integrate `pymatgen-ext-matproj` addon package

munrojm:
## Summary

This PR removes the `MPRester` code contained in `pymatgen.ext.matproj` and replaces it with an updated client in the external `pymatgen-ext-matproj` addon package. This addon provides access to both the updated and legacy Materials Project API. 

`pymatgen-ext-matproj` is added to the dependency list which itself is dependent on the new `mp-api` package. This introduces a few other dependencies: `emmet-core`, `pydantic`, `maggma`, `mpcontribs-client`. These additions do not cause any issues.

## Checklist

in the pull request title.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -5,6 +5,9 @@ name: Testing Linux
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   test:
     # prevent this action from running on forks


shyuep:
Please do not merge until I personally review this PR. @mkhorton

munrojm:
@shyuep @mkhorton happy to make whatever changes you both feel are necessary.

mkhorton:
Many thanks @munrojm! I'll pull + test soon. Let me know if you'd like a TestPyPI release too, for extra verification.

munrojm:
Thanks @mkhorton . I think a TestPyPI release is probably a good idea.

mkhorton:
Hi @munrojm, this looks good! Switching between new/legacy APIs works well.

Not officially approving this yet, have a few comments first, then we can get a TestPyPI release out and do some extra testing.

For this PR:

* I would add a mention of `pymatgen-ext-matproj` to [`addons.rst`](https://github.com/materialsproject/pymatgen/blob/master/docs_rst/addons.rst)

For `pymatgen-ext-matproj` package:

* Perhaps put all the legacy code into a separate file (say, `legacy.py`), to make the `client.py` cleaner and easier to parse.
* I have some suggestions for the docstrings:
  * Replace "old and new" with "new and legacy"
  * Replace "At the current moment, full parity between old and new API MPRester has not been implemented. This will be resolved in the near future. Power users of the old API should not use the new API until all such issues are resolved." with "Consult the Materials Project documentation at ... for advice on which API to use." since (1) I'm not expecting parity (does this mean full backwards compatibility, etc.?) and (2) it's easier to update the docs and better to have a single source-of-truth for this advice.
  * Slight edit to "You are using the legacy MPRester, which is deprecated. If you are not a power user, ie., you do not have a lot of legacy code that uses many different MPRester methods, it is recommended you get a new API key from the new Materials Project front end. Once you use get your new API key, using the new API key will automatically route you to using the new MPRester.", perhaps, "You are using the legacy MPRester. This version of the MPRester will no longer be updated. To access the latest data with the new MPRester, obtain a new API key from ... and consult the docs at ... for more information."
  * I would remove "It is not recommended, but if you would like to select the specific version of the MPRester, you can initialize either _MPResterNew or _MPResterLegacy directly." If you moved the legacy MPRester into a separate file, it wouldn't need the `_`, and people could import it from, say, `pymatgen.ext.matproj.legacy` as they see fit.
* Consider using `__new__` instead of `__init__` to return the correct class depending on API key, e.g. the legacy or new API. This will ensure that `type(MPRester)` will give more useful information (e.g. `pymatgen.ext.matproj.legacy.MPRester` or `mp_api.client.MPRester` accordingly).

With these suggestions in mind, maybe something like this could work?

```
class MPRester:
    """
    A convenience class to switch between new and legacy API.
    
    The new API requires an API key from ... and will return an
    instance of mp_api.client.MPRester
    
    More information on the new API is available at: 
    
    The legacy API requires an API key from ... and will return 
    an instance of pymatgen.ext.matproj.legacy.MPRester
    
    More information on the legacy API is available at: ...
    
    For advice on which API to choose, consult ...
    """
    
    def __new__(cls, api_key, *args, **kwargs):
        
        if not api_key:
            raise ValueError("Please supply an API key. You can retrieve an API key from...")
        
        # new API
        if len(api_key) == 32:
            return MPResterNew(api_key, *args, **kwargs)
        # legacy API
        elif len(api_key) == 16:
            return MPResterLegacy(api_key, *args, **kwargs)
        
        raise ValueError(f"API key {api_key} is in an unknown format. Please check your API key at ...")
```

Let me know what you think!

munrojm:
Thanks @mkhorton! No disagreements on my end. Ill make the suggested changes.

mkhorton:
I should add that once it's on TestPyPI I want to do a few rounds of installation testing via different installation methods, just to fully satisfy myself there's no strange dependency issues we haven't accounted for -- I think we should be fine.

From a quick audit, `robocrys` is the only major dep formally outside of MP and requires `matminer` so this may actually be an issue (it has very permissive dependencies itself otherwise though, and nothing else exceptional). We might also want to un-pin specific versions of `pydantic` and `pybtex` from `emmet-core` (the former is in `monty` too, and the latter is in `pymatgen`) and specify minimum versions for these instead.

I might suggest, in this PR, that we put "pymatgen-ext-matproj>=0.0.2,<1" in the `setup.py` too, just in case we need to break something in future.

shyuep:
A few critical issues:
1. Why do we need pymatgen to depend on pymatgen-ext-matproj, which then depends on mp-api? This kind of structure is rather convoluted. Either mp-api is integrated into pymatgen-ext-matproj or pymatgen-ext-matproj is integrated into pymatgen itself. There shouldn't be three layers of dependencies, which in turn have more dependencies.
2. Why is pymatgen-ext-matproj not a public repo?
3. I don't seem to see mp-api package within the materialsproject org. So I have no way to review the dependencies.
4. In general, I do not like too many hard dependencies. Only things that are absolutely critical for 80% functionality should be within the dependency tree. I am not sure how robocys/matminer fits into a API package. Again, I have no way of seeing how it is structured.

munrojm:
> Why do we need pymatgen to depend on pymatgen-ext-matproj, which then depends on mp-api? This kind of structure is rather convoluted. Either mp-api is integrated into pymatgen-ext-matproj or pymatgen-ext-matproj is integrated into pymatgen itself. There shouldn't be three layers of dependencies, which in turn have more dependencies.

The issue is that I don't think it makes sense for the de-facto MP API client to live as an addon package to pymatgen. If we don't want to have that kind of structure, then I would vote to integrate the legacy client code into mp-api and make it an explicit dependency of pymatgen. I recognize others may disagree with me here.

> Why is pymatgen-ext-matproj not a public repo?

My intention was to keep it private while we work out integration details to avoid confusing people who might stumble upon it.

> I don't seem to see mp-api package within the materialsproject org. So I have no way to review the dependencies.

https://github.com/materialsproject/api

> In general, I do not like too many hard dependencies. Only things that are absolutely critical for 80% functionality should be within the dependency tree. I am not sure how robocys/matminer fits into a API package. Again, I have no way of seeing how it is structured.

Agreed, the problem dependencies are introduced via emmet-core (https://github.com/materialsproject/emmet/blob/main/emmet-core/setup.py). I vote we make robocrys and the pymatgen analysis addon packages optional there to address this.

PR #2623, opened 2022-08-18
Structure Graphs from Lobster Data - Update

JaGeo:
## Summary

I have extended the structure graphs that can be derived by Lobster. 

To do so, I have extended the method `with_local_env_strategy` to have the additional parameter edge_properties. 

I am now also setting the node properties in get_bonded_structure to arrive at a networkx graph object that includes the information of the nodes as well. So far, this information is only present in the additional structure object but not in the graph itself. I hope this does not lead to side effects in another package. 

I have added a test case as well and further documenation. 

I hope this is okay. I would be happy about further feedback.


JaGeo:
I think the precommit leads to these errors. It reformats the files in such a way that the lines are too long. I guess I have to go into the pre-commit config and see how to fix this.

JaGeo:
Fixed everything. Introduced a side effect that I hadn't anticipated. All tests are passing now.

JaGeo:
This was the last change from my side.

shyuep:
Great. Thanks.

PR #2630, opened 2022-08-22
Handle disordered structures in `CrystalNN.get_bonded_structure()` and `get_cn()`

### LINKED ISSUE #2070: Make CrystalNN work with fractional occupation
When the CrystalNN neighbor-finding method is applied to a structure with fractional occupation (in the test case from a CIF file), it exits with an attribute error:

  File "pymatgen/pymatgen/analysis/local_env.py", line 4010, in get_cn
    return super().get_cn(structure, n, use_weights)
  File "pymatgen/pymatgen/analysis/local_env.py", line 265, in get_cn
    siw = self.get_nn_info(structure, n)
  File "pymatgen/pymatgen/analysis/local_env.py", line 3842, in get_nn_info
    nndata = self.get_nn_data(structure, n)
  File "pymatgen/pymatgen/analysis/local_env.py", line 3905, in get_nn_data
    X1 = structure[n].specie.X
  File "pymatgen/pymatgen/core/sites.py", line 80, in __getattr__
    raise AttributeError(a)
AttributeError: specie

Removing the fractional occupation in the underlying CIF file results in no problems.

Maybe the fractional occupation should simply be ignored for determining the neighbors and coordination numbers. The only issue that remains is then when fractional site a close to each other so that they might be deemed coordinated when the question might be: is either site 1 or site 2 present, not both so that they cannot be coordinated.

janosh:
Closes #2070.

This PR fixes the `AttributeError` currently raised when passing disordered structures to methods like `get_cn()` and `get_bonded_structure()` of `CrystalNN` and other `NearNeighbors` subclasses.

Several open questions:
1. I added a helper function `_handle_disorder(structure: Structure, on_disorder: Literal["take majority", "error"])` that's reused in multiple places since there doesn't appear to be one central place where disordered structures can be ordered by taking majority species from each site for all downstream processing like @mkhorton [suggested](https://github.com/materialsproject/pymatgen/issues/2070#issuecomment-782239047).
2. What should the options for the `on_disorder` kwarg be? Is `"take majority"` and `"error"` ok?
3. Perhaps most important @shyuep @mkhorton: What should the default for `on_disorder` be? I chose `'take majority'` to avoid new people encountering issues like #2070 but safer and less backward incompatible change would be `'error'`.

I added tests that check both the error case and `'take majority'` case to make sure we no longer raise a generic `AttributeError` without a clear explanation of why it's not working.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,16 +6,10 @@ ci:
 
 repos:
 
-  - repo: https://github.com/myint/autoflake
+  - repo: https://github.com/PyCQA/autoflake
     rev: v1.4
     hooks:
       - id: autoflake
-        args:
-          - --in-place
-          - --remove-unused-variables
-          - --remove-all-unused-imports
-          - --expand-star-imports
-          - --ignore-init-module-imports
 
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.37.3
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,16 +51,16 @@ Welcome to new contributors @naveensrinivasan, @xivh, @dgaines2, @yang-ruoxi, @c
 * Enhancement: Remove not converged warning for VASP AIMD runs, PR #2571 by @mjwen
 * Fix: generation of continuous line-mode band structures, PR #2533 by @munrojm
 * Fix: duplicate site properties for magnetic moments hwen using `AseAtomsAdaptor`, PR #2545 by @arosen93
-* Fix: bug in Grüneisen parameter calculation, PR #2543 by by @ab5424
+* Fix: bug in Grüneisen parameter calculation, PR #2543 by @ab5424
 * Fix: allow a comment on final line of KPOINTS file, PR #2549 by @xivh
 * Fix: for `Composition.replace` with complex mappings, PR #2555 by @jacksund
 * Fix: Implement equality method and fix __iter__ for InputSet, PR #2575 by @rkingsbury
 * Fix: use negative charge convention for electron in "update_charge_from_potcar", PR #2577 by @jmmshn
 * Fix: ensure charge is applied to initial and final structures parsed from vasprun.xml, PR #2579 by @jmmshn
 * Chore: Set permissions for GitHub actions, PR #2547 by @naveensrinivasan
-* Chore: Included GitHub actions in the Dependabot config, PR #2548 by by @naveensrinivasan
+* Chore: Included GitHub actions in the Dependabot config, PR #2548 by @naveensrinivasan
 * Documentation: fix typos in pymatgen.symmetry.analyzer docstrings, PR #2561 by @dgaines2
-* Documentation: clarification about usage of InputFile, PR #2570 by by @orionarcher
+* Documentation: clarification about usage of InputFile, PR #2570 by @orionarcher
 * Documentation: Improve messages and warnings, PR #2572 and PR #2573 by @cajfisher
 * Documentation: fix typo, PR #2580 by @janosh
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -123,7 +123,7 @@ v2022.3.22
 * Support kwargs for ASE adaptor. (@arosen93)
 * Fix for cation error in Lobster analysis. (@JaGeo)
 * Major revampt of Abstract interface for Input classes in IO. (@rkingsbury)
-* Orbital-projected band center, band filling, band ceneter, skewness, kurtosis, etc. (@arosen93)
+* Orbital-projected band center, band filling, band center, skewness, kurtosis, etc. (@arosen93)
 * Misc cleanups. (@janosh)
 
 v2022.3.7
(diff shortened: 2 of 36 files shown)


mkhorton:
Thanks for having a go at this @janosh 

> What should the default for on_disorder be? I chose 'take majority' to avoid new people encountering issues like https://github.com/materialsproject/pymatgen/issues/2070 but safer and less backward incompatible change would be 'error'.

I think the default should be `take majority`; in terms of breaking backwards compatibility, I'm not sure anyone would have been relying upon the fact that CrystalNN raises when encountering a disordered structure, so it seems quite safe to add this new functionality.

One caution with bonding with disordered structures is sometimes, say, a light element like H might live on, say, four very near-by sites with 0.25 occupancy each. This kind of strategy will suggest the structure contains "H-H" bonds when they do not. In this sense, perhaps `take majority` should require that the occupancy really is the majority (i.e., > 0.5)?

PR looks good otherwise, thank you! Unfortunate the easiest implementation requires making a copy of the structure, but a lot of the code in local_env.py is pretty slow anyway right now, so I'm not too concerned by this.

janosh:
> I'm not sure anyone would have been relying upon the fact that CrystalNN raises when encountering a disordered structure, so it seems quite safe to add this new functionality.

I am currently relying on the error being an `AttributeError` over at https://github.com/janosh/pymatviz/pull/50/commits/99bd2cd. Changing to `ValueError` is strictly speaking breaking. But yeah, probably not many cases of this.

> One caution with bonding with disordered structures is sometimes, say, a light element like H might live on, say, four very near-by sites with 0.25 occupancy each. This kind of strategy will suggest the structure contains "H-H" bonds when they do not. In this sense, perhaps `take majority` should require that the occupancy really is the majority (i.e., > 0.5)?

Excellent point. As I was submitting this, I realized maybe there should be 3rd option `'take max species'` which does what `'take majority'` currently does, and then `'take majority'` should raise `ValueError` on disordered sites that don't have a majority species. Happy to implement this if you think it's sensible.

mkhorton:
> I am currently relying on the error being an AttributeError over at https://github.com/janosh/pymatviz/commit/99bd2cd.

Fair enough, I stand corrected :) Shows I should not make assumptions!

> Excellent point. As I was submitting this, I realized maybe there should be 3rd option 'take max species' which does what 'take majority' currently does, and then 'take majority' should raise ValueError on disordered sites that don't have a majority species. Happy to implement this if you think it's sensible.

This would be great, thanks. One option here is instead of raising, that site could simply be omitted when you make the ordered copy. This would allow `take_majority` to still work fine for the rest of the structure, e.g. the case I mentioned with the disordered H sites, all the other sites could still be bonded.

janosh:
> This would allow `take_majority` to still work fine for the rest of the structure, e.g. the case I mentioned with the disordered H sites, all the other sites could still be bonded.

You mean keep the existing behavior of `take_majority`  as a 4th option, call it `take_majority_strict` which raises `ValueError` if no majority present, and make a new option `take_majority_drop` or whatever that drops all sites without majorities?

Btw, do you prefer underscores over spaces/ Happy to rename the options `take max species` $\to$ `take_max_species`

mkhorton:
> You mean keep the existing behavior of take_majority as a 4th option, call it take_majority_strict which raises ValueError if no majority present, and make a new option take_majority_drop or whatever that drops all sites without majorities?

I was proposing that the behavior of `take_majority` is the latter, i.e., what you suggest as `take_majority_drop`. However, having both options, `take_majority_strict` and `take_majority_drop` would also work (and be better). It's difficult to imagine all the situations in which someone might want to use this.

> Btw, do you prefer underscores over spaces? Happy to rename the options take max species → take_max_species

No strong feelings! I default to underscores since it makes it a valid identifier, but there's no reason spaces should be a problem. I'm not aware on any style guides for this sort of thing.

janosh:
> However, having both options, `take_majority_strict` and `take_majority_drop` would also work (and be better).

Alright, that's done.

janosh:
@mkhorton The linter errors here are unrelated (also present [on `master`](https://github.com/materialsproject/pymatgen/runs/8079752678?check_suite_focus=true))

```txt
pymatgen/analysis/nmr.py:59:19: E1121: Too many positional arguments for classmethod call (too-many-function-args)
pymatgen/analysis/nmr.py:61:19: E1121: Too many positional arguments for classmethod call (too-many-function-args)
pymatgen/analysis/nmr.py:155:19: E1121: Too many positional arguments for classmethod call (too-many-function-args)
pymatgen/analysis/nmr.py:157:19: E1121: Too many positional arguments for classmethod call (too-many-function-args)
```

I didn't touch `pymatgen/analysis/nmr.py`. The weird thing is I can't even reproduce locally:

```sh
pylint pymatgen/analysis/nmr.py                        

------------------------------------
Your code has been rated at 10.00/10
```

JaGeo:
This linting problem suddently showed up in my PR. All tests were passing before on the master branch and I really only changed the documentation: https://github.com/materialsproject/pymatgen/pull/2639

janosh:
Then maybe it's coming from a version change since stuff in `requirements-dev.txt` is unpinned? I just upgraded my local `pylint` but still can't reproduce. We should prob remove `--quiet` here to see what versions are being installed. Makes debugging easier.

https://github.com/materialsproject/pymatgen/blob/b1f8c168dedb0ff2381b7da2437eac8da2f48249/.github/workflows/lint.yml#L28

mkhorton:
Thanks for the PR @janosh! Merged

Have implemented the suggested change for the linter to see if we can track it down, but I won't discuss further in this specific PR just because it's unrelated to these changes.

PR #2636, opened 2022-08-25
[WIP] Fix InsertionElectrode.get_summary_dict()

acrutt:
Update InsertionElectrode.get_summary_dict() to replace entry.entry_id with entry.data["material_id"] to correctly pull the material_id given changes in the entry


mkhorton:
Thanks for this PR! Noting we're having some offline discussion about this.

acrutt:
It was decided that this issue would be resolved through a change in emmet which is being addressed in a separate PR (https://github.com/materialsproject/emmet/pull/520) so that `.get_summary_dict()` method would not need to be changed in pymatgen.

PR #2644, opened 2022-09-02
Integrate `mp-api.client.MPRester` into `pymatgen.ext.matproj`

munrojm:
## Summary

- `mp-api` made an explicit dependency 
- `MPRester` now detects the API key type (new or legacy) and instantiates the appropriate version of `MPRester` 

`msgpack` is the only additional dependency added to enable BS and DOS data retrieval with the new API.

## TODO (if any)

Finalize clean up of `mp-api` dependencies and client code (https://github.com/materialsproject/api/pull/666) before merging.

## Checklist

in the pull request title.


munrojm:
@mkhorton @shyuep I have realized that `msgpack` is required as an install to properly pull BS and DOS data. It only depends on cython (which is already required through scipy), and is therefore lightweight. Let me know your thoughts and if this is okay.

mkhorton:
Interesting, I have no objections (given that it's lightweight, as you say). It could be added to the `mp-api` requirements.

munrojm:
> Interesting, I have no objections (given that it's lightweight, as you say). It could be added to the `mp-api` requirements.

Indeed, that is where I have added it. I believe it used to get automatically installed with maggma, which was removed as a required dependency.

shyuep:
msgpack is fine. Monty has it as an optional dependency too

mkhorton:
Many thanks @munrojm! This looks good. Noting here that I'd like to make a test (TestPyPI) release prior to the next real release to double-check nothing unexpected happens.

PR #2646, opened 2022-09-06
improve outcar parsing

gpetretto:
## Summary

While parsing a large OUTCAR file coming from an MD simulation I realized that a lot of time was taken by the `read_table_pattern` to extract the number of plane waves for kpoints. The quantitiy appears early in the file, but for very long simulations the execution of the regex on the rest of the file can be very slow and pointless. 

I added the option to stop the parsing of the table patterns at the first occurence and used it for the case of the plane waves.


shyuep:
Please add tests for the new functionality...

shyuep:
thanks.

PR #2649, opened 2022-09-10
Tweak  .github/workflows/release.yml

janosh:
I just noticed the `release` job GitHub action runs on a matrix strategy. That seems very odd since you can't release a package version to PyPI more than once. I might be missing sth?

I also changed the trigger to

```diff
on:
  release:
-    types: [created]
+    types: [published]
```

based on [this SO heuristic](https://stackoverflow.com/a/61066906).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   release:
-    types: [created]
+    types: [published]
   workflow_dispatch:
 
 permissions:
@@ -71,18 +71,12 @@ jobs:
           cache-dependency-path: |
             requirements.txt
             setup.py
-      - name: Install dependencies
-        run: |
-          pip install --upgrade pip wheel
-          pip install -r requirements.txt
-          pip install -e .
       - name: Release
         env:
           TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
         run: |
-          pip install setuptools wheel twine
+          pip install --upgrade pip wheel setuptools twine
           python setup.py sdist bdist_wheel
-          twine upload --skip-existing dist/*.whl
-          twine upload --skip-existing dist/*.tar.gz
+          twine upload --skip-existing dist/*.{whl,tar.gz}
         continue-on-error: true
--- a/docs_rst/compatibility.rst
+++ b/docs_rst/compatibility.rst
@@ -49,7 +49,7 @@ pulling in Pymatgen versions with breaking changes that may end up breaking
 your tool.
 
 An even more conservative approach is to pin the Pymatgen dependency to a fixed version, for
-example `pymatgen==2021.3.3`. While this will always install the the same version of pymatgen,
+example `pymatgen==2021.3.3`. While this will always install the same version of pymatgen,
 it can lead to unnecessary dependency conflicts with other tools that depend on (a different
 version of) Pymatgen.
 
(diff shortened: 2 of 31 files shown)


shyuep:
You **can** release to pypi more than once. Different wheels for different python versions and OSes.

janosh:
Is that actually happening here? The only thing that looks platform dependent in `setup.py` are the `numpy` header files (which should be wrapped in a list btw, i.e. `include_dirs=[numpy.get_include()]`). Is this the reason for OS -specific releases?

https://github.com/materialsproject/pymatgen/blob/d26d374dabe01686950162132f24c66baaf298fe/setup.py#L192-L193

And also, which files actually compile against `numpy` headers? This `.pyx` file maybe (which is old enough to contain py2 compat code 😄)?

https://github.com/materialsproject/pymatgen/blob/d26d374dabe01686950162132f24c66baaf298fe/pymatgen/optimization/neighbors.pyx#L10

shyuep:
Yes. The cython based modules are compiled against specific numpy.

janosh:
@shyuep Good to know! I restored the release job's matrix strategy in 
f6aaae0 but removed the dependency installation step since running `python setup.py` [will do that anyway](https://stackoverflow.com/a/53069528).

.github/workflows/release.yml
```diff
@@ -71,18 +71,12 @@ jobs:
...
         env:
           TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
         run: |
-          pip install setuptools wheel twine
+          pip install --upgrade pip wheel setuptools twine
           python setup.py sdist bdist_wheel
-          twine upload --skip-existing dist/*.whl
-          twine upload --skip-existing dist/*.tar.gz
+          twine upload --skip-existing dist/*.{whl,tar.gz}
```
janosh, line 81: @shyuep One thing I'm unsure about is if this bash expansion works on windows:

```sh
dist/*.{whl,tar.gz}
```

If not, we could just do

```sh
dist/*
```

shyuep: I wouldn't bother with the expansion. It is not critical to the functioning of the code. Technically, only the tar.gz needs the skip-existing arg. This is the non-compiled version that can get uploaded several times and the workflow will fail if skip-existing is not used. The wheels for each OS and python version should be unique. But we can just leave it as is.

janosh:
I noticed the [PyPI description](https://pypi.org/project/pymatgen) is incorrectly parsed:

Screen Shot 2022-09-12 at 13 09 29

Fixed in this PR.

mkhorton:
Merging (thanks!), but adding a note that I've been meaning to switch over to use this action for a while: https://cibuildwheel.readthedocs.io/en/stable/

I've used this in another project and it works very well for producing cross-platform wheels.

janosh:
@mkhorton I assume cross-platform wheels containing compiled code would be larger than platform-specific? Presumably they would need to bundle the `numpy` header files for Windows, Linux _and_ macOS arm64?

I guess the pymatgen wheel is already quite large so adding that wouldn't make much difference?

mkhorton:
The wheels are uploaded separately to PyPI, you only download the one you need on the platform you're installing on, so that's not a big concern.

For an example, contrast the `numpy` wheels available for the latest version here:

https://pypi.org/project/numpy/#files

to the `pymatgen` wheels here:

https://pypi.org/project/pymatgen/#files

mkhorton:
We do occasionally have user support requests from people with difficulty installing on one system or another (typically, if they install outside `conda`), and that's the main reason this is useful. It also slightly speeds up the install time too which is a nice side benefit.

PR #2659, opened 2022-09-17
Fix ThermalDisplacementMatrices.compute_directionality_quality_criterion()

janosh:
Screen Shot 2022-09-16 at 19 49 13

https://github.com/materialsproject/pymatgen/blob/722ea4648bbcd9e792991e51365fed5b78eba3f4/pymatgen/phonon/thermal_displacements.py#L264-L270


JaGeo:
I will work on an additional test so that this error is fixed for the long term.

PR #2660, opened 2022-09-19
Patch the cube spherical mask

nwinner:
Simple patch for cube parsers spherical mask. It now functions properly and is cleaner.

### DIFF
--- a/pymatgen/io/cube.py
+++ b/pymatgen/io/cube.py
@@ -120,27 +120,22 @@ class Cube:
             radius: (float) of the mask (in Angstroms)
             cx, cy, cz: (float) the fractional coordinates of the center of the sphere
         """
+
+        # the radius along x,y,z in terms of number of indices than angstroms
         dx, dy, dz = (
             np.floor(radius / np.linalg.norm(self.X)).astype(int),
             np.floor(radius / np.linalg.norm(self.Y)).astype(int),
             np.floor(radius / np.linalg.norm(self.Z)).astype(int),
         )
-        gcd = max(np.gcd(dx, dy), np.gcd(dy, dz), np.gcd(dx, dz))
-        sx, sy, sz = dx // gcd, dy // gcd, dz // gcd
         r = min(dx, dy, dz)
 
+        # indices of center of sphere
         x0, y0, z0 = int(np.round(self.NX * cx)), int(np.round(self.NY * cy)), int(np.round(self.NZ * cz))
 
-        centerx, centery, centerz = self.NX // 2, self.NY // 2, self.NZ // 2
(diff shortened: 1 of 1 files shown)


jmmshn:
@nwinner since we already have an internal pymatgen definition of `VolumentricData` it might be helpful to have some of the methods here moved to VolumetricData and that way the VASP folks and get access to them too.
Then we just need a `from_cube_file` class method for `Volumetric data`.  This also avoids the problem discussed here for similar large objects that might need to be serialized at some point down the line:
https://github.com/materialsproject/pymatgen/issues/2661

Some of the masking logic here can also merge with or replace the masking logic here:

https://github.com/materialsproject/pymatgen-analysis-defects/blob/a56874d8b3ed8a411669805be0339d5b65d76e3e/pymatgen/analysis/defects/utils.py#L360-L395

nwinner:
Seems sensible. Wouldn't that require moving VolumetricData to a code agnostic location? If we do what you suggest right now, anyone wanting to parse cube data, which is used by people outside that vasp community, would have to go into the vasp module to find it.

shyuep:
I don't have a problem with moving VolumetricaData.

jmmshn:
Agreed! moving VolumetricData makes sense. 

PS: on a semi related note I have this package as well that's meant to deal with representation of periodic data:
https://github.com/materialsproject/pyrho

If we rip volumetric data out into a code agnostic place that leaves the possibility of intergrating pyrho in the future and you can get some regridding and compression and interpolation features for free.

nwinner:
On a semi-semi-related note. Any possibility to move some of the k-point logic from the VASP module into something code agnostic? Generating k-point grids does not need to be specific to VASP.

jmmshn:
So I took a look at the `Kpoints` code and I think that'll be difficult.
Since the object is already `MSONable` with heavy design influences from VASP I think it might be difficult to strip it out.
Especially with all the tetrahedron stuff.

I think we shelve the idea of stripping out the `Kpoints` until another program needs it and think about a redesign (inheritance pattern) at a later time.

nwinner:
Sure. I'm going to push k-points and bandstructure support for cp2k by next week, and it uses the vasp kpoints module, but its not *crucial* to modify that at the moment since not many people use the code. Something to keep in mind for later though.

nwinner:
Anyway where should the VolumetricData be moved to that would make the most structural sense? A new place in io module?

jmmshn:
I think something like `io.periodic.py` would be goodc then you can put any future common classes for Kpoints and band structure in there.

shyuep:
I would just call it io.common or something.

nwinner:
I've made a new PR for this `io.common` idea and it can be found here:

https://github.com/materialsproject/pymatgen/pull/2667#issue-1386968182

PR #2663, opened 2022-09-22
Additional test for thermal displacement matrices

JaGeo:
## Summary

I have added an additional test to make sure that the error fixed in https://github.com/materialsproject/pymatgen/pull/2659 does not show up again. I have started from @janosh 's branch.

JG


JaGeo:
Some tests seem to faile because phonopy is not present.

JaGeo:
Fixed.

Issue #2661, opened 2022-09-20
Proposed change to waveder


jmmshn:
https://github.com/materialsproject/pymatgen/blob/722ea4648bbcd9e792991e51365fed5b78eba3f4/pymatgen/io/vasp/outputs.py#L5509

Currently, there are two Waveder objects basically accomplish the same thing be just have two different constructor methods.
They also cannot be serialized since their `__init__` requires the filename.
Proposed change:
- Combine the two objects into one `Waveder` dataclass object with the `cder` data.
- Change the two different constructors `from_waveder` and `from_wavederf`.

Let me know if this messes with anyone's stuff (especially the original author @migueldiascosta) and I will PR.

Best

PR #2666, opened 2022-09-24
fixed waveder serialization

jmmshn:
# Merge Waverder and Wavederf

Originally discussed in #2661 

- Merge the two classes into one object with two constructors
- Allow the object to be serializable
- Modified to test files with `LVEL=.TRUE.` in INCAR
- Added comment about running with `LVEL=.TRUE`


mkhorton:
Thanks @jmmshn. Since this is a breaking change, can you add a note to https://github.com/materialsproject/pymatgen/blob/master/docs_rst/compatibility.rst, and then I can merge?

jmmshn:
Ok, added the comment about the breaking change.

jmmshn:
There are some unrelated `mypy` errors in linting, also present on the other PRs.

PR #2668, opened 2022-09-27
Extension ThermalDisplacementMatrices class

JaGeo:
## Summary

I have extended the ThermalDisplacementMatrices class. It can now also read cif files in P1 format, produce structures including Ucif as site_property (handy for sorting structures and thermal displacement properties, for example), and print VESTA files in P1 space group including vectors indicating the directionality of two different sets of computed displacement parameters.

While doing so, I have refactored the classes parsing the cif files a tiny bit. In general, I found it very hard to reuse any of the code in there as this warning state of the class would require me to make instances of classes. While I understand the purpose of saving the warnings like this, it has this drawback. Otherwise, most of the methods in there could be static methods.

### DIFF
--- a/pymatgen/io/cif.py
+++ b/pymatgen/io/cif.py
@@ -615,12 +615,7 @@ class CifParser:
         """
         try:
 
-            lengths = [str2float(data["_cell_length_" + i]) for i in length_strings]
-            angles = [str2float(data["_cell_angle_" + i]) for i in angle_strings]
-            if not lattice_type:
-                return Lattice.from_parameters(*lengths, *angles)
-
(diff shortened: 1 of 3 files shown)


JaGeo:
Currently mypy is failing but I don't see a direct connection to my pull request. I would be very happy about any hint how to potentially solve this issue.

JaGeo:
All linting errors are fixed now.

janosh:
Thanks!

JaGeo:
Thanks for merging and fixing the docu and comments, @janosh !
JG

janosh:
We should probably automate doc style checking with `pre-commit` at some point.

PR #2670, opened 2022-09-29
Optical analysis VASP 

jmmshn:
## More Flexible reproduction of VASP's optical code

Added `DielectricFunctionCalculator` class which will allow you to:
- Have full control of the matrix elements before and after Fermi's golden rule and before and after Kramers-Kronig transformation.
- This allows you to obtain the band-decomposed and kpoint-decomposed versions of the optical properties: Free-carrier absorption + defect absorption.


mkhorton:
This is a fantastic PR, thanks @jmmshn!

Pushed a small fix a test in another module due to the addition of the extra vasprun.xml, and some typos, otherwise looks good.

Regarding printing the symmetry operators in the vasprun.xml, I could ask the VASP devs if this would be helpful? Would this be a straightforward add for them?

jmmshn:
> Regarding printing the symmetry operators in the vasprun.xml, I could ask the VASP devs if this would be helpful? Would this be a straightforward add for them?

That would be super useful and should be an easy add for them since it is store in memory and used  in various places in VASP.  Can you also CC me on that email?

jmmshn:
Also if you are able to turn on ISYM there are other subtle things when the data is printed in VASP that might need to change like one of the indices of the WAVEDER only runs over occupied bands with ISYM != 0 but runs over all bands with ISYM=0.  And there are sum weird interactions with a hidden VASP LVEL flag there were basically all brushed under the rug with ISYM=0.

PR #2672, opened 2022-09-29
Cp2k 2.0

nwinner:
## Summary

PR for major updates to CP2K functionality in pmg, plus a minor edit to `compatibility.py`. PR has been geared towards simplifying certain features, while making others more compatible with the new `atomate2` package, which will use this version.

`inputs.py`
  - New Kpoints section objects
  - New bandstructure objects
  - dataclass representations of basis sets and pseudopotentials
  
`sets.py`
 - Simplified number of sets
 - DftSet now supports new basis/potential scheme
 - DftSet now handles more printing/features
 - DftSet now includes a simple validator for non-compatible features

`outputs.py`:
  - Now supports parsing some of the CP2K properties, such as the hyperfine_coupling tensor. Support for all properties not available
  - New Dos parsing

`utils.py`
  - Now supports fewer functions. Some of the messy basis utility functions are now handled using the new objects in `inputs.py`

`pmg_config.py`
  - In order to make code similar to VASP module, and to allow for improved handling of pseudopotentials, a config option has been added to pmg_config that is similar to the potcar configuration for VASP. This will now replace the settings.yaml file, which is a large file to move through github.


shyuep:
Thanks. based on the changes, it seems only the test_inputs was changed in one line. Should there be more tests for all the new functionality in outputs, sets, etc.?

nwinner:
I can add a few basic tests and some new test data for it. I'll @ you afterwards.

nwinner:
@shyuep I did a minor update to the tests and the test files to reflect changes.

nwinner:
Sorry for the premature @, @shyuep . Those simple tests were bugged. Now it should be. ready for merge if you agree.

pymatgen/io/cp2k/inputs.py
```diff
@@ -27,7 +27,7 @@
 import os
 import re
 import textwrap
-import warnings
+from collections import OrderedDict
```
janosh, line 30: Normal `dict` is guaranteed to be insertion-ordered since py37. Is this needed for order-sensive equality or sth?

nwinner: That's for a very specific case. When basis sets or pseudopotentials are listed raw instead of from another file, then those keywords must maintain their order. This does happen when cp2k generates a restart file; it will put the raw potentials into those files.

I didn't know that dict are insertion ordered. Are there any issues with those dicts getting shuffled around from other operations? If that won't happen, then I agree we can probably revert to dict.

janosh: You can overwrite dict keys but that won't change their order. The only way to change key order is to reassign the variable to a new dict (which of course can happen with `OrderedDict` too).

nwinner: Alright then yeah your changes make sense.

nwinner:
Just as a note, I merged master when I fixed @janosh 's comment and a test is failing there.

nwinner:
@shyuep @janosh (anybody else who should get @ for this?)

After a very long time, I finally finished this PR. Could I get a review?

Two tests failing on split 1 are related to chemenv tests, which are not part of this PR and are an issue elsewhere in pymatgen.

janosh:
Sorry, currently traveling and not in front of a large enough screen to review such a significant PR. But looks like a lot of great work! 🥇

pymatgen/io/cp2k/utils.py
```diff
@@ -2,33 +2,30 @@
...
-from ruamel.yaml import YAML
-
-from pymatgen.core import SETTINGS
 
-MODULE_DIR = Path(__file__).resolve().parent
+from pymatgen.core import Molecule, Structure
 
 
-def _postprocessor(s):
+def _postprocessor(s: str):
```
janosh: The underscore marks the pre- and postprocessor as private (to the module) but their not used in this module. Maybe remove the underscore? Also I changed the error type here to from `OSError` to `ValueError` which is strictly speaking breaking. Let me know if you see any issues with this.

nwinner: I think you're right. I've switched it from the underscore.

janosh:
Thanks @nwinner! This looks like a ton of work! 👍 

@shyuep @mkhorton Merging this as test and lint errors appear unrelated.

PR #2674, opened 2022-10-01
Changes to Q-Chem IO to allow CDFT and coupling calculations

espottesmith:
## Summary

This PR adds to the Q-Chem IO module in pymatgen, allowing for the following types of calculations:
- Calculations where the input is not well defined by a single molecule (e.g. fragment calculations)
- Charge- or spin-constrained DFT calculations (CDFT)
- Direct coupling (DC) calculations, which rely on CDFT
- Calculations involving diabatization of multiple states and the calculation of electronic coupling parameters, using the absolutely localized molecular orbitals (ALMO), projection operator diabatization (POD), and fragment orbital DFT (FODFT) methods

These additions

## TODO (if any)

QCInput allows for CDFT-CI (configuration interaction) multistate calculations (this required only small modifications to the CDFT input section). However, an output parser for CDFT-CI outputs has not been written.

## Checklist

in the pull request title.

- [X] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). The easiest way to handle this
      is to run the following in the **correct sequence*- on your local machine. Start with running
- [X] Doc strings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [X] Type annotations are **highly*- encouraged. Run [mypy](https://github.com/python/mypy) to type check your code.
- [X] All linting and tests pass.

(As usual, I've only checked tests related to files that I've changed)


espottesmith:
Hey @janosh @mkhorton @shyuep, I believe this is ready to be merged.

test_files/molecules/new_qchem_files/almo.out
```diff
@@ -0,0 +1,485 @@
+
```
janosh, line 1: Just a nitpick but maybe we want to rename this test files dir? From my experience `new_<sth>` names don't age that well.

pymatgen/io/qchem/tests/test_inputs.py
```diff
@@ -45,6 +45,51 @@ def test_molecule_template(self):
...
+ 0 1
+ C      0.0000000000      4.0000000000      0.0000000000
+ C      1.3320000000      4.0000000000      0.0000000000
+ H     -0.5743010000      4.0000000000     -0.9287850000
+ H     -0.5743010000      4.0000000000      0.9287850000
+ H      1.9063010000      4.0000000000      0.9287850000
+ H      1.9063010000      4.0000000000     -0.9287850000
+$end"""
+
+        self.assertEqual(molecule_test, molecule_actual)
```
janosh, line 91: Since we use `pytest` to run tests, could simplify to

```py
assert molecule_test == molecule_actual
```

janosh (commented):
Thanks @espottesmith, this is big contribution! 👍 Very good doc strings and type annos too.

Just 2 suggestions.

test_files/molecules/new_qchem_files/almo.out
espottesmith: I'm not opposed to such a change (I personally don't care what the directory name is at all). But is this relevant to the present PR? That name has been in place for years.

pymatgen/io/qchem/tests/test_inputs.py
espottesmith: Sure. Small change, happy to do it.

test_files/molecules/new_qchem_files/almo.out
janosh: Oh, didn't realize. Not too familiar with the QChem side of pymatgen. But sort of makes my point that the name is confusing. `new` is meaningless after a while. A date or a QChem version number would be better. Feel free to ignore though since not specific to this PR. Except maybe now is as good a point as any to fix.

PR looks good! So if no other comments appear, happy to merge.

PR #2679, opened 2022-10-07
Added get_proj_plot() method in phonon.plotter.PhononBSPlotter class

cnncnnzh:
## Summary

A method that allows users to plot the atom-resolved phonon band structures. The contribution from each type (or group) of atoms is calculated based on the magnitude of the eigendisplacements.  Support 2 (rb), 3 (rgb) and 4 (cmyk) unique groups of atoms.

So far I just used the existing NaCl example for testing, let me know if more cases are needed!

matplotlib.pyplot.plt —> essential for the colored plot
matplotlib.collections.LineCollection —> essential for the colored plot
pymatgen.electronic_structure.plotter.BSDOSPlotter —> For creating the color line and color triangle

## TODO (if any)

1. Make a color legend for the 4-color case.
2. Allow projecting band based on eigenvectors instead of eigendisplacement?

I would appreciate any suggestions from you! Thanks.

### DIFF
--- a/pymatgen/phonon/plotter.py
+++ b/pymatgen/phonon/plotter.py
@@ -382,9 +382,9 @@ class PhononBSPlotter:
 
         return plt
 
-    def _get_weight(self, vec: np.array, indices: list[list[int]]) -> np.array:
+    def _get_weight(self, vec: np.ndarray, indices: list[list[int]]) -> np.ndarray:
         """
-        compute the weight for each combintaion of sites according to the
+        compute the weight for each combination of sites according to the
         eigenvector
         """
         num_atom = int(self._nb_bands / 3)
@@ -422,6 +422,7 @@ class PhononBSPlotter:
             g = (1 - colors[1]) * (1 - colors[3])
             b = (1 - colors[2]) * (1 - colors[3])
             return [r, g, b]
+        raise ValueError(f"Expected 2, 3 or 4 colors, got {len(colors)}")
 
     def get_proj_plot(
         self,
@@ -451,22 +452,23 @@ class PhononBSPlotter:
         elements = [e.symbol for e in self._bs.structure.composition.elements]
         if site_comb == "element":
             assert 2 <= len(elements) <= 4, "the compound must have 2, 3 or 4 unique elements"
-            indices = [[] for _ in range(len(elements))]
+            indices: list[list[int]] = [[] for _ in range(len(elements))]
             for i, ele in enumerate(self._bs.structure.species):
                 for j, unique_species in enumerate(self._bs.structure.composition.elements):
                     if ele == unique_species:
                         indices[j].append(i)
         else:
+            assert isinstance(site_comb, list)
             assert 2 <= len(site_comb) <= 4, "the length of site_comb must be 2, 3 or 4"
             all_sites = self._bs.structure.sites
-            all_indices = {i for i in range(len(all_sites))}
+            all_indices = {*range(len(all_sites))}
             for comb in site_comb:
-                for i in comb:
-                    assert 0 <= i < len(all_sites), "one or more indices in site_comb does not exist"
-                    all_indices.remove(i)
+                for idx in comb:
+                    assert 0 <= idx < len(all_sites), "one or more indices in site_comb does not exist"
+                    all_indices.remove(idx)
             if len(all_indices) != 0:
                 raise Exception(f"not all {len(all_sites)} indices are included in site_comb")
-            indices = site_comb
+            indices = site_comb  # type: ignore[assignment]
         assert rgb_labels is None or len(rgb_labels) == len(indices), "wrong number of rgb_labels"
 
         u = freq_units(units)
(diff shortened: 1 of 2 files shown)


pymatgen/phonon/plotter.py
```diff
@@ -739,6 +867,7 @@ def get_plot(self, marker="o", markersize=None, units="thz"):
             units: unit for the plots, accepted units: thz, ev, mev, ha, cm-1, cm^-1
 
         Returns: plot
+
```
janosh, line 870: Could you undo the white space changes?

pymatgen/phonon/plotter.py
```diff
@@ -381,6 +381,134 @@ def get_plot(self, ylim=None, units="thz"):
...
+        elif len(colors) == 3:
+            return colors
+        # if there are four groups, use cyan, magenta, yellow and black
+        elif len(colors) == 4:
+            r = (1 - colors[0]) * (1 - colors[3])
+            g = (1 - colors[1]) * (1 - colors[3])
+            b = (1 - colors[2]) * (1 - colors[3])
+            return [r, g, b]
+
+    def get_proj_plot(self, site_comb="element", ylim=None, units="thz"):
```
janosh, line 425: Would be great if new code had type annotations. Something like:

```py
    def get_proj_plot(self, site_comb: str = "element", ylim: tuple[float | None, float | None] = None, units: str ="thz") -> plt.Axes:
```

janosh (changes requested):
Looks good! 👍 Just a few nitpicks.

pymatgen/phonon/tests/test_plotter.py
```diff
@@ -77,6 +77,13 @@ def test_plot(self):
         rc("text", usetex=False)
         self.plotter.get_plot(units="mev")
 
+    def test_proj_plot(self):
+        # Disabling latex for testing.
+        from matplotlib import rc
+
+        rc("text", usetex=False)
+        self.plotter.get_proj_plot(units="mev")
```
janosh, line 85: Try to test all (or most) code paths, i.e. pass different values for all arguments (incl. `site_comb` and `ylim`) to test edge cases.

cnncnnzh: It's my first-time contributing, so thank you for all the suggestions. Here are some changes I made:

- delete the extra space.
- added type hints.
- added more assertions to catch exceptions (e.g., specify more atoms than what the structure has, not list all atoms, etc.).
- added a new test file SrTiO3_phonon_bandstructure.json to test len(site_comb) = 3 or 4 cases.
- pass different values to get_proj_plot().

Let me know if there is anything else. Thanks!

janosh:
Here's the plot output from `test_proj_plot()` for later reference/discussion.

Is the text extending beyond the legend boundary easy to fix?

test_proj_plot-1
test_proj_plot-2
test_proj_plot-3
test_proj_plot-4
test_proj_plot-5
test_proj_plot-6
test_proj_plot-7

cnncnnzh:
I changed the default legends to "0", "1" and "2" so that they fit in the box. But now the rgb_label variable can be passed to the method to change the legend. How does that sound?

Figure 2022-10-12 222115

janosh:
I think you force-pushed over the commit 709cda1 I added after 702b5bd to fix `mypy` errors. Would be better if you could `git pull` remote changes to your branch first before you force push in the future.

Same thing with the commits pushed by the `pre-commit.ci` bot. Simply pull them in with `git pull`.

> I changed the default legends to "0", "1" and "2" so that they fit in the box. But now the rgb_label variable can be passed to the method to change the legend. How does that sound?

Sounds good. 👍

pymatgen/phonon/plotter.py
```diff
@@ -407,6 +548,46 @@ def save_plot(self, filename, img_format="eps", ylim=None, units="thz"):
...
+        self,
+        filename,
+        img_format="eps",
+        site_comb: str | list[list[int]] = "element",
+        ylim: tuple[None | float, None | float] = None,
+        units: str = "thz",
+        rgb_labels: tuple[str] = None,
+    ):
+        """
+        Save matplotlib projected plot to a file.
```
janosh, line 579: One more question: do we really need `save_proj_plot()`? Seems like it's a wrapper around calling `fig.savefig()` which can just be left to the user.

cnncnnzh: > I think you force-pushed over the commit [709cda1](https://github.com/materialsproject/pymatgen/commit/709cda100218e47dba247ba90aadf649e1d7912b) I added after [702b5bd](https://github.com/materialsproject/pymatgen/commit/702b5bdc8e054b0f269dffadc44440f3098bf16b) to fix `mypy` errors. Would be better if you could `git pull` remote changes to your branch first before you force push in the future.
> 
> Same thing with the commits pushed by the `pre-commit.ci` bot. Simply pull them in with `git pull`.
> 
> > I changed the default legends to "0", "1" and "2" so that they fit in the box. But now the rgb_label variable can be passed to the method to change the legend. How does that sound?
> 
> Sounds good. 👍

Oh sure. I'll do that in the future 👍.

cnncnnzh: > One more question: do we really need `save_proj_plot()`? Seems like it's a wrapper around calling `fig.savefig()` which can just be left to the user.

I intended to write a counterpart for 'save_plot()'. But I think you are right that it's not necessary.

janosh (approved):
Thanks for this contribution! 👍

PR #2683, opened 2022-10-11
Run CI in lowest supported Python version

janosh:
The most important Python version to run CI is the lowest supported (currently 3.8). Running in higher versions means we won't catch some SyntaxErrors e.g. from modern type annotations like `lst: list[int] = [1, 2, 3]` (which [I think was the reason](https://github.com/materialsproject/pymatgen/runs/8157637404?check_suite_focus=true) for bumping to 3.10?). Such errors would now pop up at runtime for people below Python 3.10.

If we want to test later Python versions too (e.g. due to upcoming 3.11 release), we should setup a matrix strategy.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       max-parallel: 20
       matrix:
-        python-version: ['3.9']
+        python-version: ['3.8']
         # This distribution of tests is designed to ensure an approximately even time to finish for parallel jobs.
         include:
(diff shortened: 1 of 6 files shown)


janosh:
Remaining error

```py
pymatgen/ext/tests/test_matproj.py:45: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pymatgen/ext/matproj.py:210: in __init__
    db_version = self.get_database_version()
pymatgen/ext/matproj.py:307: in get_database_version
    d = self._make_request("/api_check")
```

is due to https://matsci.org/t/44750

PR #2685, opened 2022-10-12
Lint dev scripts

janosh:
@shyuep Are the files under `dev_scripts/` still used?


shyuep:
The dev scripts are of course rarely used. But they are used from time to time. E.g., when I need to regenerate the periodic_table.json file when certain properties are added.

PR #2686, opened 2022-10-13
Typos

janosh:


shyuep:
Why would we bother to lint cmd_line? These are not our files. Other people are responsible for them.

PR #2691, opened 2022-10-14
Test methods downstream from `PD.get_decomp_and_e_above_hull_on_error()` can also ignore errors

janosh:
Related #2681.


janosh:
@shyuep I removed `pydocstyle` in 7699c98 from `lint` CI as it was raising a lot of errors with the new config from cf8e0af. I think before it wasn't doing anything. At least I don't remember ever getting an error from it.

PR #2692, opened 2022-10-17
Updating Lammps inputs and inputsets/generators to facilitate integration into atomate2

gbrunin:
## Summary

Based on previous work from @manaslkht and @rkingsbury, I create this PR to integrate the LammpsInputFile object in pymatgen. Basic InputSet and InputGenerator are added as well, based on templates. For more advanced automatic tools and usage, the user is suggested to use atomate2's addon atomate2-lammps (see https://github.com/Matgenix/atomate2-lammps).


pymatgen/io/lammps/inputs.py
```diff
@@ -17,17 +17,293 @@
...
+        Args:
+            input_settings: dictionary of LAMMPS input settings.
+        """
+        self.nstages = 0
+        self.ncomments = 0
+        self.input_settings = input_settings if input_settings else {}
+
+    def add_stage(self, command: str | list | dict, stage_name: str = None):
+        """
+        Adds LAMMPS command(s) and its arguments to LAMMPS input file.
```
JaGeo, line 54: Maybe a simple example for a stage here would help with understanding how this input file works.

JaGeo:
Hey @gbrunin , I am currently on holiday. It might take a bit until I have time review in detail. I am adding @QuantumChemist to the pull request as well.

pymatgen/io/lammps/outputs.py
```diff
@@ -187,3 +116,262 @@ def _parse_thermo(lines):
...
+                    diff = pbc_diff(Xtp, Xtpt)
+                    diff = self.trajectory[t0].box.to_lattice().get_cartesian_coords(diff)
+                    MSD[ityp, time_index, 0:3] += np.average(diff**2, axis=0)
+                    MSD[ityp, time_index, 3] += np.average(norm(diff, axis=1) ** 2)
+
+                MSD[ityp, time_index] = MSD[ityp, time_index] / len(tprime)
+
+        return MSD
+
+    def get_diffusion_coefficient(self, MSD=None, start=None, stop=None):
```
JaGeo, line 335: Type hinting?

pymatgen/io/lammps/data.py
```diff
@@ -863,6 +861,36 @@ def from_structure(cls, structure: Structure, ff_elements=None, atom_style="char
         topo = Topology(boxed_s)
         return cls.from_ff_and_topologies(box=box, ff=ff, topologies=[topo], atom_style=atom_style)
 
+    def set_charge_atom(self, charges):
```
JaGeo, line 864: type hinting

pymatgen/io/lammps/outputs.py
```diff
@@ -187,3 +116,262 @@ def _parse_thermo(lines):
...
         runs.append(_parse_thermo(lines[b + 1 : e]))
     return runs
+
+
+class LammpsDump(MSONable):
+    """
+    Object for representing dump data for a single snapshot.
+    """
+
+    def __init__(self, timestep, natoms, box, data):
```
JaGeo, line 126: Type hinting

pymatgen/io/lammps/inputs.py
```diff
@@ -17,15 +17,439 @@
...
+                else:
+                    args = comment
+            else:
+                args = comment
+            self.add_command(command=command, args=args, stage_name=stage_name)
+
+        else:
+            raise NotImplementedError("If you want to add an inline comment, please specify the stage name.")
+
+    def get_string(self, ignore_comments: bool = False, keep_stages: bool = False) -> str:
```
QuantumChemist, line 266: I wonder if this will or can be written into a file?

pymatgen/io/lammps/outputs.py
```diff
@@ -187,3 +116,262 @@ def _parse_thermo(lines):
...
+            if not (isinstance(timestep, int) or isinstance(timestep, float)):
+                raise ValueError("The provided keys should be timesteps.")
+            if not isinstance(dump, LammpsDump):
+                raise ValueError("The provided values should be LammpsDump objects.")
+
+            trajectory[timestep] = dump
+
+        return LammpsTrajectory(trajectory)
+
+    def get_msd(self, atom_type=None, time_average=True):
```
QuantumChemist, line 261: Type hinting?

pymatgen/io/lammps/inputs.py
gbrunin: Yes, through the write_file method (see InputFile object in pymatgen.io.core, or the write_file method that I have now added here to override the InputFile method, because the signatures differ). This is used for instance by InputSets to write the input files.

QuantumChemist: Super! Thanks for letting me know.

JaGeo:
Hi @gbrunin, 
I had a chance to shortly have a look the classes. Thanks for your work!

I have one question: 
how would one generate an md run with a new structure with the help of these classes? Would I need to start from LAMMPsData? To me, this is not immediately obvious.

gbrunin:
Hi @JaGeo,

Thanks for your comments.
In the end, we decided to have sets and generators in a separate repository (https://github.com/Matgenix/atomate2-lammps) in the same spirit as atomate2. So, this stuff in pymatgen is there to be used as an input analyzer/builder/modifier but not necessarily automatically, while the input generation from a structure will be handled in atomate2-lammps (it is already, through https://github.com/Matgenix/atomate2-lammps/blob/main/src/atomate2/lammps/sets/base.py). I hope this makes sense !

JaGeo:
Thanks! It does make sense but I would still recommend a method/function that allows to easily create inputs for LAMMPS with different structures in pymatgen. After all, this is what I would expect from an IO module for LAMMPS. I would recommend moving the LammpsInputSet back to pymatgen.

pymatgen/io/lammps/sets.py
```diff
@@ -0,0 +1,199 @@
...
+class LammpsInputSet(InputSet):
+    r"""
+    Container class for all LAMMPS inputs. This class is intended to provide
+    general functionality that can be customized to many purposes.
+    InputGenerator classes elsewhere in this module are used to create
+    specific instances of LammpsInputSet that are tailored to specific purposes.
+
+    /!\ This InputSet and InputGenerator implementation is based on templates and is not intended to be very flexible.
+    For instance, pymatgen will not detect whether a given variable should be adapted based on others
+    (e.g., the number of steps from the temperature), it will not check for convergence not will it actually run LAMMPS.
```
QuantumChemist, line 39: Just a small typo "...nor will it..." (likewise in the other warnings)

pymatgen/io/lammps/inputs.py
```diff
@@ -222,6 +253,61 @@ def add_stage(self, command: str | list | dict, stage_name: str | None = None):
...
+
+        merged_stages = [merge_name, merged_commands]
+        stages_names.append(merge_name)
+        list_of_commands.append(merged_stages)
+
+        for i_stage, stage in enumerate(self.list_of_commands):
+            if i_stage > indices_stages_to_merge[0] and i_stage not in indices_stages_to_merge:
+                list_of_commands.append(stage)
+
+        self.list_of_commands = list_of_commands
```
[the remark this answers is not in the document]
gbrunin: What do you think would be more useful: that the method returns the stage(s) that has/have been removed, the entire initial LammpsInputFile, or something else ? Or that they are stored as an attribute of the object, like LammpsInputFile.backup_list_of_commands ? Just trying to find what would be the best solution for the user.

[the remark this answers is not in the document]
JaGeo: Wouldn't they just start from scratch in such a case? I am currently thinking about a possible scenario where one would like to continue after an error in a script. Sorry, if I did not get it correctly.

pymatgen/io/lammps/inputs.py
```diff
@@ -14,18 +17,669 @@
...
+
+        Args:
+            command (str): String with the command to find in the input file (e.g., "units").
+            stage_name (str): String giving the stage name where the change should take place.
+
+        Returns:
+            True if the command is present, False is not.
+        """
+        args = self.get_args(command, stage_name)
+        return True if args else False
```
ml-evs, line 136:
```suggestion
        return bool(self.get_args(command, stage_name))
```

pymatgen/io/lammps/inputs.py
```diff
@@ -14,18 +17,669 @@
...
+
+    def set_args(self, command: str, argument: str, stage_name: str | None = None, how: str | int | list = "all"):
+        """
+        Set the argument to a given command.
```
ml-evs, line 140:
```suggestion
        Sets the arguments for the given command to the given string.
```

ml-evs (commented):
Hi @gbrunin, sorry for the delay, I ended up tunneling into readability of (I think the older bits of) code a bit too much in this review (haven't even checked all files yet). If you'd rather, I'm happy to fork your fork and commit these changes myself (and make sure the tests continue passing as they are).

pymatgen/io/lammps/data.py
```diff
@@ -865,6 +863,36 @@ def from_structure(cls, structure: Structure, ff_elements=None, atom_style="char
...
+    def set_charge_atom(self, charges: dict):
+        """
+        Add or modify charges of atoms of the data.
+
+        Args:
+            charges: Dict containing the charges for the atoms to set.
+                     The dict should contain atom indexes and charges.
+                     Example: change the charge of the third atom of the list to -2:
+                     charges={3: -2}
+        """
```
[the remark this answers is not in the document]
gbrunin: This doesn't actually add a command to the LAMMPS input, but it rather changes the data file directly. The "set charge" command is used in the input when the data does not contain information (or not the right information) on the charge. 
 But indeed, the 3rd index is not necessarily the 3rd atom in the list, it was the case when I was including some other changes that I removed since. 
The type hints from typing are not that much used in pymatgen, so I don't know if we should start using them or not?

pymatgen/io/lammps/inputs.py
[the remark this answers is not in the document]
QuantumChemist: @JaGeo not continue after the error, but go back one step before the error occurred and re-start (with different settings) from there. Then, I think it's easier to have a direct access to what has been changed instead of having to look at the files before and after removing/merging.
@gbrunin I think it's the easiest to return the stages that were removed and store the information somehow. 
But I'm not a LAMMPS user at all, if you see no necessity for any kind of stage backup, it's completely fine.

pymatgen/io/lammps/inputs.py
```diff
@@ -14,18 +17,669 @@
...
+    input where the simulation box is set up, a set of variables are declared or
+    quantities are computed.
+
+    The LammpsInputFile is defined by its list_of_commands,
+    i.e. a list of stages (strings) together with the corresponding LAMMPS input settings (strings).
+    The structure is the following:
+    list_of_commands = [
+        ["Stage 1", [[cmd1, args1], [cmd2, args2]]],
+        ["Stage 2", [[cmd3, args3]]]
+    ]
```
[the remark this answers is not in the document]
gbrunin: I can use this format and we'll see once the PR is reviewed.

pymatgen/io/lammps/inputs.py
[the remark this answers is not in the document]
gbrunin: Thinking more about it, I think this is something that should be left to the user or to third-party codes that will use it (e.g., atomate2). In case of error, you can always restart the process of creating the LammpsInputFile. In other cases (e.g., workflow creation), you can always save the LammpsInputFile in another variable to use it again later. 
If we decide to save previous states of the LammpsInputFile, it might become uneasy to handle and not very readable.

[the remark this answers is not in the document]
JaGeo: I agree. This should not be part of the io module.

pymatgen/io/lammps/sets.py
```diff
@@ -0,0 +1,219 @@
...
+    You can create a template for your own task following those present in pymatgen/io/lammps/templates.
+    The parameters are then replaced based on the values found
+    in the settings dictionary that you provide, e.g., `{"nsteps": 1000}`.
+
+    Parameters:
+        template: Path (string) to the template file used to create the input file for LAMMPS.
+        calc_type: String used to shortly describe the type of computations performed by LAMMPS.
+    Args:
+        settings: Dictionary containing the values of the parameters to replace in the template.
+        keep_stages: Whether to keep the stage structure of the LammpsInputFile or not.
```
rkingsbury, line 127: What is the difference between Parameters and Args? Also, it does not appear that `template` or `calc_type` can be set anywhere by the user (apologies if I'm missing it).

pymatgen/io/lammps/sets.py
```diff
@@ -0,0 +1,219 @@
...
+        self.read_data = read_data
+        self.force_field = force_field
+        self.settings = {
+            "units": units,
+            "atom_style": atom_style,
+            "dimension": dimension,
+            "boundary": boundary,
+            "read_structure": read_data,
+            "force_field": force_field,
+        }
```
rkingsbury, line 217: Is it necessary to store these settings as individual attributes _and_ in `self.settings`? It seems redundant to me. You could instead define, for example, `self.units` as a @property that returns `self.settings["units"]`. That way there is no possibility of a mismatch between `self.settings` and `self.units`

pymatgen/io/lammps/sets.py
```diff
@@ -0,0 +1,219 @@
...
+    For instance, pymatgen will not detect whether a given variable should be adapted based on others
+    (e.g., the number of steps from the temperature), it will not check for convergence nor will it actually run LAMMPS.
+    For additional flexibility and automation, use the atomate2-lammps implementation
+    (https://github.com/Matgenix/atomate2-lammps).
+    """
+
+    template = os.path.join(template_dir, "minimization.template")
+    calc_type = "minimization"
+
+    def __init__(
```
rkingsbury, line 194: Please add a docstring describing the arguments here, as well as a type annotation so that users will understand they're going to a get a `LammpsInputSet`

pymatgen/io/vasp/outputs.py
```diff
@@ -1985,6 +1985,9 @@ def __init__(self, filename):
         run_stats["cores"] = None
         with zopen(filename, "rt") as f:
             for line in f:
+                if "serial" in line:
+                    run_stats["cores"] = 1
+                    break
                 if "running" in line:
```
rkingsbury, line 1991: Why is this VASP-related file being modified? Was this just leftover from a `git pull`?

rkingsbury (commented):
Thank you for the great work on this @gbrunin . I reviewed the code and made a few comments regarding the interface and docstrings, but overall it's looking good to me. Please let me know if you want me to clarify any of my comments.

pymatgen/io/lammps/inputs.py
```diff
@@ -14,18 +17,682 @@
...
+            ```
+            command = [
+                "thermo 1",
+                "thermo_style custom step lx ly lz press pxx pyy pzz pe",
+                "dump dmp all atom 5 run.dump",
+                "min_style cg",
+                "fix 1 all box/relax iso 0.0 vmax 0.001",
+                "minimize 1.0e-16 1.0e-16 5000 10000",
+                "write_data run.data"
+            ]
```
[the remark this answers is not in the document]
gbrunin: The example is correct. Using this with the add_stage, we get e.g.
```
In [19]: lmp_input = LammpsInputFile()
In [20]: lmp_input.add_stage(command = [
    ...:                 "thermo 1",
    ...:                 "thermo_style custom step lx ly lz press pxx pyy pzz pe",
    ...:                 "dump dmp all atom 5 run.dump",
    ...:                 "min_style cg",
    ...:                 "fix 1 all box/relax iso 0.0 vmax 0.001",
    ...:                 "minimize 1.0e-16 1.0e-16 5000 10000",
    ...:                 "write_data run.data"
    ...:             ])
In [21]: lmp_input.list_of_commands
Out[21]: 
[['Stage 1',
  [['thermo', '1'],
   ['thermo_style', 'custom step lx ly lz press pxx pyy pzz pe'],
   ['dump', 'dmp all atom 5 run.dump'],
   ['min_style', 'cg'],
   ['fix', '1 all box/relax iso 0.0 vmax 0.001'],
   ['minimize', '1.0e-16 1.0e-16 5000 10000'],
   ['write_data', 'run.data']]]]
```

I think there is a confusion between "command" in add_stage (and other places in the code) and "list_of_commands" (LammpsInputFile attribute) because it's not very clear in the documentation. Would it be better to rename list_of_commands? Any suggestion? It was initially input_settings, and we changed it for clarity but maybe it had the opposite effect...

pymatgen/io/lammps/inputs.py
```diff
@@ -14,18 +17,682 @@
...
+                        if i in how:
+                            self.list_of_commands[i_stage][1][i_cmd][1] = argument
+                        i += 1
+
+    def add_stage(
+        self,
+        command: str | list[str] | dict[str, str | float],
+        stage_name: str | None = None,
+        after_stage: str | None = None,
+    ):
```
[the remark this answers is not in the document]
gbrunin: It was a dict in the initial implementation. We decided to get rid of dicts because we might have different stages with the same name (by mistake or not). For instance, a LammpsInputFile could consist of multiple steps including multiple thermalizations, that could all be called "Thermalization". Writing this, I realize that I should check the behavior of the methods where a stage_name can be provided by the user, I'm pretty sure I did not pay attention to that.

About your first comment, I think this may be related to the one on the example of command. I believe this method makes things easy to add a stage defined by a list of commands as a user would write them in an input file. Playing with the list_of_commands can be a bit painful because of its structure (list of lists of lists).

pymatgen/io/core.py
```diff
@@ -68,7 +68,6 @@ def write_file(self, filename: str | Path) -> None:
 
         Args:
             filename: The filename to output to, including path.
-            kwargs: Keyword arguments passed to get_string()
```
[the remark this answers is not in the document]
gbrunin: Yes, I thought about it indeed. But mypy was really not happy about it, so in the end I decided to do it like this to avoid future developers having problems with mypy.

pymatgen/io/lammps/sets.py
```diff
@@ -0,0 +1,219 @@
...
+        super().__init__(inputs={"in.lammps": self.inputfile, "system.data": self.data})
+
+    def get_inputs(self) -> dict[str, LammpsInputFile | LammpsData | CombinedData]:  # pylint: disable=E1131
+        """
+        Generate a dictionary of one or more input files to be written. Keys
+        are filenames, values are the contents of each file.
+        This method is called by write_input(), which performs the actual file
+        write operations.
+        """
+        return {"in.lammps": self.inputfile, "system.data": self.data}
```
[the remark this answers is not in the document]
gbrunin: Indeed, this was useless

pymatgen/io/lammps/sets.py
gbrunin: I did that initially only to make the as_dict() work, but you're right that using @property is better to avoid mismatches.

pymatgen/io/vasp/outputs.py
gbrunin: No it's not from a git pull, it's actually a little thing that I have added because I once compiled a serial version of VASP (no mpi enabled) and this related pymatgen code broke in such a case. It's so small that I thought I could include it here, but maybe a separate PR is better?

pymatgen/io/lammps/sets.py
gbrunin: I have refactored this part, thanks for pointing out that is was not consistent. Hopefully it's better now.

gbrunin:
Thanks a lot @rkingsbury ! This is very helpful. I have made a few changes according to your comments. Just a few points left open for discussion.

pymatgen/io/vasp/outputs.py
rkingsbury: Ah I see. Well, thank you for the enhancement! Yes, for clarity I think it would be best to make a separate small PR for this so it doesn't get lost. I know it's a bit of extra work but it will help the maintainers I think (and likely be merged quickly)

pymatgen/io/lammps/sets.py
rkingsbury: Looks great, thank you!

rkingsbury:
> Thanks a lot @rkingsbury ! This is very helpful. I have made a few changes according to your comments. Just a few points left open for discussion.

I'm glad they were helpful! Let me think more about the two comments that are still open and I'll reply later today.

pymatgen/io/lammps/inputs.py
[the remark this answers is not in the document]
rkingsbury: > It was a dict in the initial implementation. We decided to get rid of dicts because we might have different stages with the same name (by mistake or not). For instance, a LammpsInputFile could consist of multiple steps including multiple thermalizations, that could all be called "Thermalization". 

I see. I would advocate for enforcing unique stage names (whether you stick with the list input format or not). I say this for two reasons;

1. When we do automated output parsing, it would be good to organize the results by stage name. If the names are not unique, this will be harder to do and harder to disambiguate results
2. When actually executing the calculation, I could foresee wanting to organize the output files in subdirectories named according to stage (for example, maybe we want every stage to be contained in a folder and have an output called "output.dump". This isn't possible if the stage names are not unique.


> About your first comment, I think this may be related to the one on the example of command. I believe this method makes things easy to add a stage defined by a list of commands as a user would write them in an input file. Playing with the list_of_commands can be a bit painful because of its structure (list of lists of lists).

Yes I see what you mean. You might consider using `Tuple` for each line of commands rather than `List`, since they always have exactly 2 values (e.g., command and args, each as a `str`). Then your input would look like

```
list_of_commands = [
        ["Stage 1", [(cmd1, args1), (cmd2, args2)],
        ["Stage 2", [(cmd3, args3)]
    ]
```

I personally still feel that a `dict` might make more sense for `list_of_commands`, but you are the ones who will be using this most so I'll defer to your judgment. Semantically (see also other comment), I do think this argument name is confusing, because it's really a list of _stages_ (i.e., each element in the top-level list is the stage name plus the associated commands. So I'd suggest a name like `.stages` instead.

An in-between option, which could allow you to leverage built-in python semantics a little bit more, is to structure `.stages` as a list of dict. Each dict could explicitly set the `stage_name` and the `commands`, e.g.

```
stages = [
        {"stage_name": "Stage 1", "commands": [(cmd1, args1), (cmd2, args2)]},
        {"stage_name:" "Stage 2", "commands":  [(cmd3, args3)]}
    ]
```

This `list` of `dict` structure parallels MongoDB documents and is widely utilized elsewhere in our data processing architecture. It also has an advantage of being a little more human-readable on its own. By using this format you can

1. Still access stages by positional indexing
2. Access stages by name using a list comprehension, as you already do
3. Add stages using `.stages.append({}` or at an explicit position using `.stages[1]={"stage_name":XX, "commands":[]}

You could take this a step further by making each element of `commands` a dict too, with keys `cmd` and `args` or something, but that might be overcomplicating things.

Just some food for thought!

pymatgen/io/lammps/inputs.py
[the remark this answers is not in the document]
rkingsbury: Yes, I see that you're right. As I said in the other comment, I agree that `list_of_commands` is a confusing name; I think it should be `.stages`.

Semantically, I also feel that whatever is passed to `add_stage` should have the same syntax as an element of `.stages` (that's what led to my confusion about the example - since `list_of_commands` takes `
list_of_commands = [  ["Stage 1", [(cmd1, args1), (cmd2, args2)] ]`, I was expecting the `command` here to have the same format.

pymatgen/io/lammps/inputs.py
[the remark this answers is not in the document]
gbrunin: I really like the "in-between" solution, indeed it could make things more clear. I have to think about it a bit more, because of how LAMMPS works : for instance, you can have a given command defined multiple times with different or with the same arguments, in the same stage or not. For this reason, having a dict with commands as keys would not work (this is what was initially done as well by Manas), and why we went for lists. We have also thought about tuples, but I feel like the future way of dealing with force-field parameters might be easier with lists (maybe this is totally wrong, actually we don't know yet how we will handle that properly).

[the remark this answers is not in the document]
gbrunin: Hi @rkingsbury, thanks again for your comments on this. I implemented this last solution in my last commit. 

About the add_stage method: one could indeed argue about its utility now. I personally like having a helper method to construct a stage using a stage name and a list of commands (multiple formats are accepted), instead of having the (slight) trouble of formatting this on my own each time I want to add a stage to an InputFile. Both are currently possible, so that offers a bit of flexibility. However, I understand the semantic issue of having an àdd_stage` that does not take a `stage` as input. Maybe I could change the method so that it accepts either a formatted stage (it would basically be the same as .append) or a stage name and list of commands: what do you think? 

Alternatively, removing the add_stage could maybe lead to errors on the user side. Let's consider a user who wants to add a command to a given stage. The user can use the add_command for this purpose. This method would have to change to allow the user to provide a stage name or not, and if the stage name that is provided is not in the list of stage names of the current InputFile, a new stage is created. Having a small typo in the provided stage name could in this case change the behavior of LAMMPS executing the input file, because a new stage would be created at the end with the provided command instead of inserting the command to the desired stage. With the add_command as is now, this error is not possible: if the stage name is not already present, an error is thrown. The stage has to be created first. 

Another possible solution would be to have add_stage only initializing the stage (empty), and add_command being able to add different formats of commands. In this case, it could be renamed initialize_stage so that the input (stage name) is not confusing. 

Let me know what you think of these possibilities.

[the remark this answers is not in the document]
rkingsbury: > About the `add_stage` method: one could indeed argue about its utility now. I personally like having a helper method to construct a stage using a `stage_name` and `commands` (multiple formats are accepted), instead of having the (slight) trouble of formatting this on my own each time I want to add a stage to an `InputFile`. Both are currently possible, so that offers a bit of flexibility (even though the append does not make sure that all stage names are unique). However, I understand the semantic issue of having an `add_stage` that does not take a `stage` as input. Maybe I could change the method so that it accepts either a formatted stage (it would basically be the same as `.append`) or a `stage_name` and `commands`: what do you think?

I see your points about the utility here, and I hadn't thought about the fact that `.append` won't enforce unique stage names. If you think this will be useful to you I think you should leave it. And with the clarifications to the attribute name (`.stages`) and more expressive syntax, I'm personally less concerned about the semantic confusion. I think it's OK as-is.


> Alternatively, removing the `add_stage` could maybe lead to errors on the user side. Let's consider a user who wants to add a command to a given stage. The user can use the `add_command` for this purpose. This method would have to be modified to allow the user to provide a `stage_name` or not, and if the `stage_name` that is provided is not in the list of stage names of the current `InputFile`, a new stage is created. Having a small typo in the provided stage name could in this case change the behavior of LAMMPS executing the input file, because a new stage would be created at the end with the provided command instead of inserting the command to the desired stage (not sure this is super clear). With the `add_command` as it is now, this error is not possible: if the `stage_name` is not already present, an error is thrown. The stage has to be created first.

Good point! And anyway as you said, having these methods does not prevent users from directly modifying the underlying `list`/`dict`, but wrapping those modifications in methods clearly does have some benefits from a validation / error checking standpoint.

> Another possible solution would be to have `add_stage` only initializing the stage (empty), and `add_command` being able to add different formats of commands. In this case, `add_stage` could be renamed `initialize_stage` so that the input (`stage_name`) is not confusing.

I'm not in favor of this; I think it will be harder to understand and more complicated than the alternative, without a clear benefit.

gbrunin:
Hi @JaGeo, @QuantumChemist, @ml-evs and @rkingsbury, thanks again to all of you for your suggestions. This is just to let you know that I think this PR can drop its WIP status now. Let me know if you still want to review it before that. 
Cheers !

gbrunin:
Hi @janosh and @shyuep, this PR is now ready to be reviewed. Thanks in advance!

janosh (approved):
Looks great and like a lot of work! Thanks @gbrunin. 👍

PR #2694, opened 2022-10-19
VASP Optics Masking, and plotting

jmmshn:
## Vasp optics masking and plotting features

Summary:

- Allow masking of matrix elements without recreating the object
- Allow users to access a filtered version of the matrix element data for analysis

You can now pass a `mask` kwarg in the calculation and get the portion of the dielectric function contributed by the transitions that you selected.


jmmshn:
@janosh, ready to merge this is failing due to something unrelated.

janosh:
> @janosh, ready to merge this is failing due to something unrelated.

Not sure where the error on master crept in. I was just investigating and working on a fix in #2695. It seems to be related to the pinned version of `spglib`:

https://github.com/materialsproject/pymatgen/blob/e01d044986d6b321ee0254a6c9c89a5edd693935/requirements.txt#L10

pymatgen/io/vasp/optics.py
```diff
@@ -171,6 +176,69 @@ def _use_default(param, default):
...
+
+        x_val = []
+        y_val = []
+        text = []
+        (
+            nb,
+            _,
+            nk,
+            nspin,
+        ) = cderm.shape[:4]
```
janosh, line 221: This could prob fit on 1 line and be more readable?

pymatgen/io/vasp/optics.py
```diff
@@ -301,14 +373,43 @@ def epsilon_imag(
...
+            return egrid, np.zeros_like(egrid, dtype=np.complex_)
+        raise e
+
+    (
+        _,
+        _,
+        nk,
+        nspin,
+    ) = cder.shape[:4]
+    iter_idx = [
```
janosh, line 398: Same here.

pymatgen/io/vasp/optics.py
```diff
@@ -171,6 +176,69 @@ def _use_default(param, default):
             eps += 1.0 + 0.0j
         return egrid, eps
 
+    def plot_me_data(self, idir: int, jdir, mask: npt.NDArray | None = None, min_val: float = 0.0):
```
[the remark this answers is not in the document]
jmmshn: changed to `plot_weighted_transition_data`.

pymatgen/io/vasp/optics.py
jmmshn: `black` decided to do this.

pymatgen/io/vasp/optics.py
jmmshn: Same here.

pymatgen/io/vasp/optics.py
```diff
@@ -110,7 +112,8 @@ def from_directory(cls, directory: Path | str):
         """Construct a DielectricFunction from a directory containing vasprun.xml and WAVEDER files."""
         d_ = Path(directory)
         vrun = Vasprun(d_ / "vasprun.xml")
-        waveder = Waveder.from_binary(d_ / "WAVEDER")
+        is_gamma = "gamma" in vrun.generator["subversion"]
```
jmmshn: This is kinda hacky but the "gamma-only" is defined in `src/version.F` so I think we are OK.

pymatgen/io/vasp/optics.py
janosh: `black` will leave the following as is:

```py
nb, _, nk, nspin = cderm.shape[:4]
```

jmmshn: Oh wow... the comma!

jmmshn:
OK, so due to the fact that VASP can be compiled with different dtypes I'm just gonna set `data_type="complex64"` and let the users pick when they are running for the `Waverder` object, `gamma_only` is just not enough information.

In the optics module, I'm trying to read the larger dtypes first to force a value error during reshaping. then trying smaller dtypes.

janosh:
Thanks @jmmshn! 👍

PR #2698, opened 2022-10-21
Let env vars override `.pmgrc.yaml` in `_load_pmg_settings()` and make `~/.config` new default location for `.pmgrc.yaml`

janosh:
@shyuep @mkhorton `_load_pmg_settings()` currently reads env vars first and then overrides them with `.pmgrc.yaml`. It's usually the other way around.

This gives users no way to override on a case by case basis. E.g.

```sh
PMG_MAPI_KEY=xxxxxxxxxxxxxxxx python -c 'from pymatgen.ext.matproj import MPRester; MPRester().query("mp-1234", ["formula"])'
```

will be identical to

```py
python -c 'from pymatgen.ext.matproj import MPRester; MPRester().query("mp-1234", ["formula"])'
```

Following [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I think we should swap this around and let env vars override `.pmgrc.yaml`.

### DIFF
--- a/pymatgen/cli/pmg_config.py
+++ b/pymatgen/cli/pmg_config.py
@@ -19,7 +19,7 @@ from urllib.request import urlretrieve
 
 from monty.serialization import dumpfn, loadfn
 
-from pymatgen.core import SETTINGS_FILE
+from pymatgen.core import OLD_SETTINGS_FILE, SETTINGS_FILE
 
 
 def setup_potcars(potcar_dirs: list[str]):
@@ -178,18 +178,27 @@ def install_software(install: Literal["enumlib", "bader"]):
 
 def add_config_var(tokens: list[str], backup_suffix: str) -> None:
     """Add/update keys in .pmgrc.yaml config file."""
-    d = {}
     if os.path.exists(SETTINGS_FILE):
+        # read and write new config file if exists
+        fpath = SETTINGS_FILE
+    elif os.path.exists(OLD_SETTINGS_FILE):
+        # else use old config file if exists
+        fpath = OLD_SETTINGS_FILE
+    else:
+        # if neither exists, create new config file
+        fpath = SETTINGS_FILE
+    d = {}
+    if os.path.exists(fpath):
         if backup_suffix:
-            shutil.copy(SETTINGS_FILE, SETTINGS_FILE + backup_suffix)
-            print(f"Existing {SETTINGS_FILE} backed up to {SETTINGS_FILE}{backup_suffix}")
-        d = loadfn(SETTINGS_FILE)
+            shutil.copy(fpath, fpath + backup_suffix)
+            print(f"Existing {fpath} backed up to {fpath}{backup_suffix}")
+        d = loadfn(fpath)
     if len(tokens) % 2 != 0:
         raise ValueError(f"Uneven number {len(tokens)} of tokens passed to pmg config. Needs a value for every key.")
     for key, val in zip(tokens[0::2], tokens[1::2]):
         d[key] = val
-    dumpfn(d, SETTINGS_FILE)
-    print(f"New {SETTINGS_FILE} written!")
+    dumpfn(d, fpath)
+    print(f"New {fpath} written!")
 
 
 def configure_pmg(args: Namespace):
--- a/pymatgen/core/__init__.py
+++ b/pymatgen/core/__init__.py
@@ -28,29 +28,34 @@ __maintainer_email__ = "shyuep@gmail.com"
 __version__ = "2022.9.21"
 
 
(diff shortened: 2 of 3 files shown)


mkhorton:
If we're going to mess around with `.pmgrc.yaml` I would prefer more comprehensive changes:

1. `pydantic` is now a dependency via `monty`, and has excellent features for [settings management](https://pydantic-docs.helpmanual.io/usage/settings/) -- we use this in other projects, why not here?
2.

shyuep:
I think pydantic is overcomplicated. We are talking about a few settings here. Unless we want to move a lot of pymatgen settings to the config file (e.g., default potentials, etc.), I think the current format is fine.

I am also fine if we want to generate the file in .config by default henceforth, but keep backward compatibility by checking $HOME/.config, followed by $HOME.

mkhorton:
What makes the pydantic `Settings` option over-complicated?

What I like:

1. It implements logic (like deciding between env variables and config files) for us, we don't have to write our own
2. It makes it very obvious what settings are possible and what their types and defaults are (-> it makes it easier to add new settings too), this I think is the main benefit
3. It's already dependency, and seeing increased adoption in community, so I don't see specific downside
4. Other MP codes are using it, so we make it easier for developers to contribute to multiple codes, since we're adopting a common pattern
5. It has additional features, like support for `.env` files, secrets, etc. that we may not wish to implement ourselves, but could be useful to others

shyuep:
@mkhorton I am not saying pydantic does not have benefits. All the benefits you listed are valid. I am saying right now, pymatgen has just the API key and a few config settings like VASP PSP directory. It is not like we have a list of 50 settings to manage. I think the pydantic use case is a lot more useful for something like FireWorks or atomate where you need to specify server settings, and a whole host of config variables. Some of the other benefits are not that useful in pymatgen's case, e.g., knowing what are valid values for certain settings. 

Feel free to implement pydantic if you really feel strongly about it. I don't mind one way or the other. My one and only condition is that whatever implemented must work with existing pymatgen installations.

janosh:
Big fan of option 2 (moving `.pmgrc.yaml` to `$HOME/.config`)! Been thinking of suggesting that as well.

> I am also fine if we want to generate the file in .config by default henceforth, but keep backward compatibility by checking $HOME/.config, followed by $HOME.

We could even move it to `$HOME/.config` ourselves when detecting a `.pmgrc.yaml` in the home directory and print a message to the user about it. Would speed up the transition but might create edge cases that break existing pymatgen installations.

---

Re `pydantic`: haven't used it yet so don't have an opinion on it (yet). One very minor downside is that it's incompatable with certain future type annotations (as I discovered in https://github.com/materialsproject/atomate2/pull/165#issuecomment-1238620700) which pmg makes widespread use of now. But should be fine as long as we stick to old annotations in files parsed by `pydantic`. Just something that contributors will have to be aware of.

shyuep:
I don't think we should automatically move files to .config. That is executing a specific write operation on a user's computer. It is not too much trouble to just support $HOME/.config and $HOME.

I didn't know about the type annotation issue. If that is the case, let's just table the pydantic move for now. I recognize the benefits but at the time, there is no pressing need.

mkhorton:
I believe that issue was fixed: https://github.com/pydantic/pydantic/pull/2761

janosh:
@mkhorton That was a partial fix. Some future annotations syntax like the pipe `|` for type unions can't be supported unless the codebase is py3.10+ (pmg is 3.8+) where `|` between types becomes valid syntax outside of annotation contexts. See https://github.com/pydantic/pydantic/issues/3300#issuecomment-1034007897.

mkhorton:
To be specific, at what point would this bug be triggered? Would it only be if we use the `futures` import in the specific file that holds the settings, i.e. as long as we don't use the pipe operator inside settings it would be ok?

janosh:
> Would it only be if we use the futures import in the specific file that holds the settings, i.e. as long as we don't use the pipe operator inside settings it would be ok?

That's correct.

mkhorton:
Ok -- this is something I do feel strongly about. I think a clear list of allowed settings (with types) is beneficial. However, I don't have time to implement this minute, so in this sense I abdicate my opinion for the moment :)

janosh:
> However, I don't have time to implement this minute, so in this sense I abdicate my opinion for the moment :)

I'll go ahead and implement support for `.pmgrc.yaml` under `~/.config` in this PR then and then merge it since prioritizing env vars is independent of using `pydantic`.

rkingsbury:
Thanks for the suggestions @janosh . I am in favor of moving `.pmgrc.yaml` to `$HOME/.config` and of reversing the order such that ENV variables override the .yaml file; I think that behavior will reduce the potential for confusion. I don't know enough about pydantic settings to offer a well-informed opinion.

PR #2706, opened 2022-10-24
Fix parsing of "Atomic configuration" in POTCAR

yuzie007:
First of all, thank you so much for the development of pymatgen!

## Summary

Due to the missing regex option of `(?s)` (https://docs.python.org/3/library/re.html#re.S, which makes the `.` special character match any character at all, including a newline), "Atomic configuration" in POTCAR (52 and 54) was actually not parsed. I think this part is sometimes useful for, e.g., estimating a reasonable NBANDS value. My pull request fixes the issue above. In more detail:
- `re.search` for this is slightly simplified.
- The change above adds "nentries" and "Orbitals" to PSCTR. In `get_potcar_hash`, these two keys are excluded to keep the same hash values as before.
- A new test `test_psctr` is added to check if the parsing is done correctly.

### DIFF
--- a/pymatgen/io/vasp/tests/test_inputs.py
+++ b/pymatgen/io/vasp/tests/test_inputs.py
@@ -881,6 +881,37 @@ class PotcarSingleTest(PymatgenTest):
         }
         self.assertEqual(self.psingle.keywords, data)
 
+    def test_psctr(self):
+        filename = PymatgenTest.TEST_FILES_DIR / "POT_GGA_PAW_PBE_54" / "POTCAR.Fe.gz"
+
+        with pytest.warns(None):
+            psingle = PotcarSingle.from_file(filename)
+
+        data = {
+            "nentries": 9,
+            "Orbitals": (
+                (1, 0, 0.50, -6993.8440, 2.0000),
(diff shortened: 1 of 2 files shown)


pymatgen/io/vasp/tests/test_inputs.py
```diff
@@ -881,6 +881,37 @@ def test_keywords(self):
         }
         self.assertEqual(self.psingle.keywords, data)
 
+    def test_psctr(self):
+        filename = PymatgenTest.TEST_FILES_DIR / "POT_GGA_PAW_PBE_54" / "POTCAR.Fe.gz"
+
+        with pytest.warns(None):
```
janosh, line 887: I think this has been deprecated:

> E   pytest.PytestRemovedIn8Warning: Passing None to catch any warning has been deprecated, pass no arguments instead:
E    Replace pytest.warns(None) by simply pytest.warns().

yuzie007: Two pytest.warns(None) in test_inputs.py are replaced by pytest.warns() in the new commit.

janosh:
Thanks for the fix and the great summary! 👍

PR #2704, opened 2022-10-24
`pytest-split` in CI

janosh:
https://github.com/jerry-git/pytest-split is a `pytest` plugin that splits the test suite into equally sized splits based on test execution time.

PR #2708, opened 2022-10-25
Run full test suite on Windows

janosh:
Now that we auto-split tests onto 10 parallel runners using `pytest-split` (#2704), we could run all tests on all platforms (`ubuntu`, `mac`, `windows`).

Before, the largest test suite ([don't even think it was the full one](https://github.com/materialsproject/pymatgen/commit/5cb1270310cd25a3ba1f3bbb4584f84466592c31#diff-fb0daef94ea304514bcccf51bdd0e30a073788e28fce22a7303d91203c69db70L23-L32)) was run only on `ubuntu`.

Still a WIP. CI is expected to fail due to install errors from BoltzTrap2 on Windows and [due to commenting out](https://github.com/materialsproject/pymatgen/compare/single-ci-tests-file?expand=1#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R45-R48):

```sh
# for pkg in cmd_line/*;
#   do echo "$(pwd)/cmd_line/$pkg/Linux_64bit" >> "$GITHUB_PATH";
# done
```


janosh:
Maybe we shouldn't run on `macos` actually since it's limited to 5 concurrent runners so would come with significant wait times ([docs on GH action usage limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits)) and it's unlikely to surface errors that wouldn't also show up on `ubuntu`.

janosh:
@shyuep These lines don't work on windows. Is there any point trying to fix that? `Linux_64bit` sounds like the binary wouldn't run anyway?

```sh
for pkg in cmd_line/*;
  do echo "$(pwd)/cmd_line/$pkg/Linux_64bit" >> "$GITHUB_PATH";
done
```

No tests fail without them (presumably because tests are skipped if CLI commands can't be found). So we could just skip those lines on Windows.

shyuep:
There is no need for full tests on windows or macOS.

janosh:
Why not on windows?

mkhorton:
I'd be pro testing on more platforms too (Windows especially, since we have seen some odd errors in the past).

I don't feel the need for this on every PR, although I am not against that either, but on the main branch sure.

janosh:
> Because some CLI tools are not available on windows. If you want to compile all the CLI tools for all platforms and keep them updated, sure. But I only care that those work on Linux at the minimum.

Agreed, I should have been clearer. My proposal was all the tests that can easily be run on Windows should be. For stuff that doesn't install, I agree we should skip it (like I did in f89b31b).

sgbaird:
I was chatting with @janosh a bit. Our group has also had some issues with pymatgen on Windows in the past, I think that was mostly on the installation side when trying to install pymatgen via Anaconda (I generally prefer miniconda environments).

mkhorton:
@sgbaird raises a good point too; in our test suite, we don't actually test our packaging at all. Packaging issues have actually caused several issues in the past (I'm thinking specifically of one issue we had with Linux wheels, another issue we've had a few times with missing files), none of which are caught by our current CI.

shyuep:
I am not sure how packaging can be tested. We already automate most of it and missing files happen rarely only when someone added some new config file but forgot to add it into META or setup.py. Those usually affect a very small number of people.

janosh:
@shyuep Something like this maybe?

```py
from glob import glob
from os.path import dirname, exists

import pytest

ROOT = dirname(dirname(__file__))
package_sources_path = f"{ROOT}/pymatgen.egg-info/SOURCES.txt"


@pytest.mark.skipif(
    not exists(package_sources_path),
    reason="No pymatgen.egg-info/SOURCES.txt file, run pip install . to create",
)
def test_egg_sources():
    """Check that all source and data files under pymatgen/**/* are listed in pymatgen.egg-info/SOURCES.txt."""
    with open(package_sources_path) as file:
        sources = file.read()

    src_files = [x for x in glob(f"{ROOT}/pymatgen/**/*", recursive=True) if "tests" not in x]

    for filepath in src_files:
        rel_path = filepath.split(f"{ROOT}/pymatgen/")[1]
        assert rel_path in sources
```

I added a similar test in https://github.com/CompRhys/aviary/pull/46 after we repeatedly managed to forget to package some JSON data and embedding files.

shyuep:
I wouldn't do it as a unittest. That is a violation of what a unittest means. I would add it as a release check in the invoke tasks.py file.

janosh:
@shyuep Happy to add it to `tasks.py`.

> That is a violation of what a unittest means.

Do you mean we're not testing the code base itself but its deployment? I would say let's test whatever gives us peace of mind.

mkhorton:
Can't one test the packaging also simply by having a workflow that does a `pip install` or similar of the latest version, and run tests against that?

janosh:
@mkhorton We could do that too. Though considering we're at <80% coverage, there's no guarantee we would catch files missing from the package this way.

mkhorton:
Fair. I was really thinking e.g. of the one issue we had with Linux installs a while back. It seems like installation issues are ones that can affect large numbers of people, but are also ones that have slipped through our CI without being noticed. With that said, we usually hear about it fairly quickly if it happens so can fix...

PR #2711, opened 2022-10-27
Add `check_egg_sources_txt_for_completeness()` to `tasks.py`

janosh:
Following discussion initiated by @sgbaird and @mkhorton in https://github.com/materialsproject/pymatgen/pull/2708#issuecomment-1292477362.

@shyuep I'm not familiar with `invoke`. The function I added `check_egg_sources_txt_for_completeness()` is not `@task` decorated. it's simply called by the `release` task. Every other function in `tasks.py` seems to be. Is that a requirement?

### DIFF
--- a/pymatgen/entries/compatibility.py
+++ b/pymatgen/entries/compatibility.py
@@ -61,7 +61,7 @@ class Correction(metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def get_correction(self, entry):
+    def get_correction(self, entry: AnyComputedEntry) -> EnergyAdjustment:
         """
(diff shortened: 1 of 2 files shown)


shyuep:
You need `@task` only if it is a command that is invoked. E.g., `invoke release` means run the function called `release`. Since your new method is only used internally for checking the generated files, you do not need to add `@task`

janosh:
> You need `@task` only if it is a command that is invoked. E.g., `invoke release` means run the function called `release`. Since your new method is only used internally for checking the generated files, you do not need to add `@task`

Ok, that's what I assumed.

PR #2713, opened 2022-10-31
Breaking: swap `fmt` and `filename` in `Structure.to()`

janosh:
@shyuep @mkhorton It's a breaking change but was wondering if we could swap the order of kwargs `fmt` and `filename` in `Structure.to()`.

https://github.com/materialsproject/pymatgen/blob/25768faa7aa0a34ee4357a92bc92ae41ef21bc1d/pymatgen/core/structure.py#L2421-L2427

The annoying thing about having `fmt` first is that the most common use case (imo writing to file) of `to()` raises an error:

```py

structure = Structure(
    lattice=Lattice.cubic(5),
    species=("Fe", "O"),
    coords=((0, 0, 0), (0.5, 0.5, 0.5)),
)

struct.to("struct.json")
>>> ValueError: Invalid format: `struct.json`
```

I did a quick look through the code base and didn't find any cases where `fmt` was passed as positional arg so don't think this would break much.

```py
struct.to("json")  # this would break
struct.to(fmt="json")  # this still works
```

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -406,11 +406,12 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
         return bool(np.min(all_dists) > tol)
 
     @abstractmethod
-    def to(self, fmt: str = None, filename: str = None) -> str | None:
+    def to(self, filename: str = "", fmt: str = "") -> str | None:
         """
-        Generates well-known string representations of SiteCollections (e.g.,
-        molecules / structures). Should return a string type or write to a file.
+        Generates string representations (cif, json, poscar, ....) of SiteCollections (e.g.,
+        molecules / structures). Should return str or None if written to a file.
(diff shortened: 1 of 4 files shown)


shyuep:
My vote is no. In an ideal world, I would prefer these to be kw only args. I almost always write Structure.to(filename="hello.cif"). We are already friendlier than pandas, which requires to_csv, to_excel etc to be specified. Matt can cast the deciding vote. I am fine if it changes.

mkhorton:
I agree, these should remain as keyword arguments regardless. But swapping the order of the keyword arguments as a usability benefit seems worthwhile and harmless enough. People who continue to use these as keyword arguments would be unaffected. I don't feel strongly about this either however!

mkhorton:
I'm equivocating; I say go for it.

janosh:
Roger! :shipit:

PR #2720, opened 2022-11-07
Fix tensor mapping

utf:
A recent change (1bdd1b7c6f5b12079c48cc06a6e6443bec070ed3) broke tensor mapping by making the default tensor and value lists immutable tuples. Tests didn't catch this because there was no tests for creating a tensor mapping starting from an empty mapping.

This PR fixes the bug and adds a new test. If possible, please could you release a new version with this bug fix as currently atomate, atomate2, amset, and other packages are broken.


janosh:
@utf Apologies for the breaking change!

Thanks for the fix and the new tests. Will try to make a new release right now (first time for me). Here goes nothing...

utf:
Thanks very much @janosh!

PR #2721, opened 2022-11-07
VaspInputSet couldn't write cif structures

JiQi535:
## Summary
This pr deals with the failure of writing .cif format structures by VaspInputSet. I encountered an error of `'PosixPath' object has no attribute 'lower'` when using the `write_input` method with `include_cif=True`. After casting the PosixPath to str, this problem is solved.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
         additional_dependencies: [flake8-bugbear]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.981
+    rev: v0.982
     hooks:
       - id: mypy
 
--- a/pymatgen/io/lobster/inputs.py
+++ b/pymatgen/io/lobster/inputs.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import itertools
 import os
 import warnings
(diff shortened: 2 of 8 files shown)


shyuep:
Pls add a test for this.

janosh:
@shyuep There is already a test for this actually.

https://github.com/materialsproject/pymatgen/blob/ffbebceafc3482e1a63008974a6da64e522e78c9/pymatgen/io/vasp/tests/test_sets.py#L471

The tests are just not running because of

https://github.com/materialsproject/pymatgen/blob/ffbebceafc3482e1a63008974a6da64e522e78c9/pymatgen/io/vasp/tests/test_sets.py#L60-L61

We're not setting `PMG_VASP_PSP_DIR`. This is not due to https://github.com/materialsproject/pymatgen/pull/2708 btw, looks like they weren't running before either.

When I run

```py
export PMG_VASP_PSP_DIR=/path/to/vasp/pseudo-potentials
pytest pymatgen/io/vasp/tests/test_sets.py -k test_write_input
```

I can reproduce the error @JiQi535 reported:

```py
>       if fmt == "cif" or fnmatch(filename.lower(), "*.cif*"):
E       AttributeError: 'PosixPath' object has no attribute 'lower'
```

PR #2726, opened 2022-11-08
Release v2022.11.7

janosh:
As requested by @utf in #2720, here's a new release to unblock downstream repos like atomate, atomate2, amset.

@shyuep `invoke release` is giving me a bunch of sphinx errors. Can we release without updating docs since last doc update is just 6 days old?


shyuep:
If you install sphinx, it shouldn't be giving errors.

janosh:
@shyuep Does the large line deletion count mean sth went wrong? After my last comment, I reverted the auto-pushed commit af5fd94 and got `invoke release` to run error-free locally but got the same results. The 22k deleted lines look suspicious.

shyuep:
If the deleted lines are html files, then it is ok.

janosh:
Yes, no Python files changed except `setup.py` and `pymatgen/__init__.py`.

utf:
This did not get released to pypi only GitHub. E.g., https://pypi.org/project/pymatgen/ the latest version is still 2022.11.1

janosh:
@utf Yeah, I saw the [release workflow failed](https://github.com/materialsproject/pymatgen/actions/runs/3416368703/jobs/5686441847) but got drawn away by another problem last night so didn't have time to debug this yet. Let me take a look now.

janosh:
So the problem is I didn't notice `setup.py` imports `numpy` when I did #2707. Now that the CI scripts have single-line installs for (almost) all deps

https://github.com/materialsproject/pymatgen/blob/d17a0f4e42203a460fecae78a5009928cc5e3330/.github/workflows/test.yml#L54

`numpy` won't be installed until after `python setup.py` has run which already tries to import `numpy`. I think the correct way to fix this is to use `setuptools`'s [`fetch_build_eggs`](http://code.nabla.net/doc/setuptools/api/setuptools/dist/setuptools.dist.Distribution.html#setuptools.dist.Distribution.fetch_build_eggs) which "resolves pre-setup requirements".

Issue #2725, opened 2022-11-07
SpacegroupAnalyzer returns duplicate symmetry operations 


lbluque:
**Describe the bug**
The `SpacegroupAnalyzer` class returns a list of symmetry operations, where each unique symop appears twice.

This seems to be related to the upgrade of spglib, since the issue occurs with spglib >= 2, but not with earlier versions.

**To Reproduce**

```python
from pymatgen.core import Structure, Lattice
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer

lattice = Lattice([[3, 3, 0], [0, 3, 3], [3, 0, 3]])
species = 3 * ["A"] + ["D"]
coords = (
    (0.25, 0.25, 0.25), (0.75, 0.75, 0.75), (0.5, 0.5, 0.5),  (0, 0, 0)
)
structure = Structure(lattice, species, coords)
sga = SpacegroupAnalyzer(structure)
symops = sga.get_symmetry_operations()

# each symop appears twice!
for i in range(0, len(symops) - 1, 2):
    assert symops[i] == symops[i + 1]
```

**Expected behavior**
The list of symmetry operations should not contain duplicates.

**Desktop (please complete the following information):**

- OS: Linux
- Version: 2022.11.1

lan496:
`SpacegroupAnalyzer.get_symmetry_operations` calls `spglib.get_symmetry` with cell `self._cell`. The current implementation seems to attach magmoms in `self._cell` even when magmoms are not specified in `Structure`:
https://github.com/materialsproject/pymatgen/blob/ffbebceafc3482e1a63008974a6da64e522e78c9/pymatgen/symmetry/analyzer.py#L79-L80

From spglib v2, the behavior of `spglib.get_symmetry` with zero magmoms is changed. Now `spglib.get_symmetry` returns the same spatial symmetry with time_reversal=True and time-reversal=False. So the size doubles.
(c.f. https://github.com/spglib/spglib/issues/150#issuecomment-1180158502)

I am not sure a reason to attach magmoms in this case. But removing the above line will solve this issue.

lbluque:
Thanks a lot @lan496 !

As for pymatgen, I think removing the line you suggest is the way forward.

Otherwise, adding a `unique` keyword parameter to `get_symmetry_operations`  and following your suggestion in the comment linked (https://github.com/spglib/spglib/issues/150#issuecomment-1180158502) is an option too.

PR #2724, opened 2022-11-07
Fix TypeError in SpacegroupAnalyzer.get_symmetrized_structure() if spglib returns no symmetries

janosh:
Closes #2723.

Maybe there's a smarter fix than fb2fc72f09bbccba12cc4d6b1d47d146840a62e4? Just returning two empty tuples for rotations and translations from `SpacegroupAnalyzer._get_symmetry()` if `spglib.get_symmetry()` returned `None` for now.

PR #2727, opened 2022-11-08
Do not introduce zero magmoms in SpacegroupAnalyzer

lbluque:
## Summary

Do not introduce zero `magmoms` in `SpacegroupAnalyzer` following changes in spglib >= 2. See #2725 

## TODO 

- With this change, the exception tested in the unit test added in #2724 is not raised.

## Checklist

in the pull request title.

- [X] Doc strings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- [X] Type annotations are **highly*- encouraged. Run [`mypy`](https://github.com/python/mypy) `path/to/file.py` to type check your code.

errors prior to submitting the PR. We highly recommended installing `pre-commit` hooks. Simply `pip install -U pre-commit && pre-commit install` in the repo's root directory. Afterwards linters will run before every commit and abort if any issues pop up.

### DIFF
--- a/pymatgen/symmetry/analyzer.py
+++ b/pymatgen/symmetry/analyzer.py
@@ -76,12 +76,14 @@ class SpacegroupAnalyzer:
                 magmoms.append(site.magmom)
             elif site.is_ordered and hasattr(site.specie, "spin"):
                 magmoms.append(site.specie.spin)
-            else:
-                magmoms.append(0)  # needed for spglib
 
         self._unique_species = unique_species
         self._numbers = zs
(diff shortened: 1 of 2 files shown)


janosh:
This is great! Thanks @lbluque. Just tried out your changes on another 5 structures for which `spglib` with 0 magmons gave me no symmetries. This fix works for all of them!

I think it might still be useful to keep the test that raises `ValueError` on the `Co8` structure I added in #2724. Just so we keep the nicer error message in case a similar problem reappears in the future. We can just manually insert `magmoms = [0] * len(struct)` in the test `_cell` to still make it raise the `ValueError`.

PR #2728, opened 2022-11-08
Fix release CI workflow

janosh:
I'm a bit confused after https://github.com/materialsproject/pymatgen/pull/2726#issuecomment-1307409880. As I understand the docs, `pyproject.toml`'s [`[build-system].requires`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#build-system-requirement) is supposed to already handle installing `numpy` and `cython` prior to running `setup.py` so there should be no need for [`fetch_build_eggs`](https://github.com/materialsproject/pymatgen/pull/2726#issuecomment-1307419037).

But the [failed release workflow](https://github.com/materialsproject/pymatgen/actions/runs/3416368703/jobs/5686441847) suggests that's not happening. I think the only reason this hasn't come up in CI before is that the `test.yml` workflow installs `m3gnet` first, then `pymatgen` and `m3gnet` comes with `numpy`. Before #2707, all CI workflows installed `requirements.txt` first before running `pip install .` so unless I'm mistaken, the `build-system.require` hasn't actually setup the build-time requirements for a while?

https://github.com/materialsproject/pymatgen/blob/d17a0f4e42203a460fecae78a5009928cc5e3330/.github/workflows/test.yml#L51-L55


janosh:
@shyuep Maybe we use this hack in `setup.py` for now so we no longer rely on some other `pip` command installing `numpy` first before `pip install .`?

```py
import subprocess

try:
    import numpy
except ImportError:
    subprocess.run(['pip', 'install', 'numpy'])
    import numpy
```

PR #2733, opened 2022-11-10
Markdown readme

janosh:
Convert the readme from rST to markdown:

```sh
pandoc --from rst --to markdown_github --no-highlight README.rst --output README.md
```

Any objections @shyuep? markdown is more readable, much more widely used and understood than rST so should allow more people to contribute updates/corrections.


janosh:
This change allows us to have a single source of truth for GH and PyPI pages since `setup.py` currently contains basically a diverged copy of the rST readme as markdown.

PR #2736, opened 2022-11-16
Sunset module `pymatgen/util/serialization.py`

janosh:
Going by the module doc string

> Most features of this module have been moved to `monty`.

and the few call sites of this module, I think we can sunset `pymatgen/util/serialization.py`. Strictly speaking a breaking change but seems unlikely anyone would use this module externally. It's only used for pickling by `class PymatgenTest`. We could do a deprecation phase though if anyone wants.


shyuep:
Fine with me.

PR #2780, opened 2022-12-21
Fix typo in equation that computes interplanar angels

EurusWei:
## Summary

In the function get_interplnar_angle of module analysis/diffraction/tem.py, there are some typos about the angels and index. 

## Todo (if any)

N/A

## Checklist

Work-in-progress pull requests are encouraged, but please put \[WIP\] in the pull request title.

- [ ] Doc strings have been added in the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [ ] Type annotations are *highly* encouraged. Run [`mypy path/to/file.py`](https://github.com/python/mypy) to type check your code.

```sh
pip install -U pre-commit
pre-commit install
```

in the repo's root directory. Afterwards linters will run before every commit and abort if any issues pop up.

### DIFF
--- a/pymatgen/analysis/diffraction/tem.py
+++ b/pymatgen/analysis/diffraction/tem.py
@@ -404,7 +404,7 @@ class TEMCalculator(AbstractDiffractionPatternCalculator):
             + p1[2] ** 2 * c_star**2
             + 2 * p1[0] * p1[1] * a_star * b_star * cos_gamma_star
             + 2 * p1[0] * p1[2] * a_star * c_star * cos_beta_star
-            + 2 * p1[1] * p1[2] * b_star * c_star * cos_gamma_star
+            + 2 * p1[1] * p1[2] * b_star * c_star * cos_alpha_star
         )
         r2_norm = np.sqrt(
             p2[0] ** 2 * a_star**2
@@ -412,14 +412,14 @@ class TEMCalculator(AbstractDiffractionPatternCalculator):
             + p2[2] ** 2 * c_star**2
             + 2 * p2[0] * p2[1] * a_star * b_star * cos_gamma_star
             + 2 * p2[0] * p2[2] * a_star * c_star * cos_beta_star
-            + 2 * p2[1] * p2[2] * b_star * c_star * cos_gamma_star
+            + 2 * p2[1] * p2[2] * b_star * c_star * cos_alpha_star
         )
         r1_dot_r2 = (
             p1[0] * p2[0] * a_star**2
             + p1[1] * p2[1] * b_star**2
             + p1[2] * p2[2] * c_star**2
             + (p1[0] * p2[1] + p2[0] * p1[1]) * a_star * b_star * cos_gamma_star
-            + (p1[0] * p2[2] + p2[0] * p1[1]) * a_star * c_star * cos_beta_star
+            + (p1[0] * p2[2] + p2[0] * p1[2]) * a_star * c_star * cos_beta_star
             + (p1[1] * p2[2] + p2[1] * p1[2]) * b_star * c_star * cos_alpha_star
         )
         phi = np.arccos(r1_dot_r2 / (r1_norm * r2_norm))


mkhorton:
Thanks @EurusWei, much appreciated for catching this!

Can you suggest a good test for this function, e.g. a known input and output?

EurusWei:
> Thanks @EurusWei, much appreciated for catching this!
> 
> Can you suggest a good test for this function, e.g. a known input and output?

Hi Horton,

Has this catch been confirmed to be true? I am comparing the diffraction patterns (DPs) generated before and after the fix to those on the MaterialsProject website. I found the code before the fix generated DPs matched the MaterialsProject, instead of after the fix. Based on symmetry I am sure the equation should be fixed, but I don't know why the above phenomenon happens (not what I expected). Do you know how the MaterialsProject website generates their DPs?

mkhorton:
Yes, I do -- the MP website uses the `TEMCalculator` exactly from *pymatgen*. At time of writing, this is [pymatgen 2022.11.7](https://materialsproject.org/healthcheck).

We, at Materials Project, are likely switching to use functionality from py4DSTEM in future, and this is being worked on by @sezelt (and is looking really nice!)

EurusWei:
> 
I see, that makes sense. Since I don't have much background in this field, I cannot tell which DP is more accurate, before or after the fix. What about this: I will give you some examples that I find having discrepancies in DPs before and after the fix, together with my calculation results, and your group can help identify if this fix makes sense. After that you can either add some tests or directly merge the fix.

shyuep:
Any update on this?

shyuep (commented):
> **Automated PR review generated by Claude (Opus 4.7)** on behalf of @shyuep. Based purely on the diff and current GitHub Actions state — no code was executed locally.

The fix looks correct. In the reciprocal-lattice norm `r1_norm` / `r2_norm`, the cross term involving `b* c*` is paired with `cos α*`, not `cos γ*`, so changing `cos_gamma_star → cos_alpha_star` on the `p[1] * p[2] * b_star * c_star` lines is right. The `p1[1] → p1[2]` change in the `r1_dot_r2` `a* c*` cross term is also a clean copy/paste fix — by symmetry with the `p[1] * p[2]` `b* c*` term immediately below, both indices on each cross term must reference the same axis.

A couple of follow-ups worth doing in this PR (the diff is 3 lines so the bar is low):
- Add a regression test in `tests/analysis/diffraction/test_tem.py` covering a low-symmetry (triclinic) cell, where `cos α* ≠ cos γ*`. Cubic / orthorhombic test cases would have masked this bug because all the off-diagonal cosines are zero.
- The PR title still has "angels" — please fix to "angles" before merge.
- No CI has run on this branch (likely because it's old and base is far behind). A rebase onto current `master` is needed before merge.

Otherwise LGTM — happy to see this old typo finally fixed.

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)**

- ✅ All three corrections are mathematically right. For the reciprocal-lattice magnitude `r_norm`, the cross-axis cosine on the `b*·c*` term has to be `cos α*` (not `cos γ*`) — the existing code had `cos γ*` in both `r1_norm` and `r2_norm`, which only happens to be correct for orthorhombic+. Likewise, in the `r1·r2` expansion, the `(p1[0]*p2[2] + p2[0]*p1[1])` term mixed `p1[1]` with the `a*·c*` cosine — the symmetric form is `(p1[0]*p2[2] + p2[0]*p1[2])`. Both fixes are necessary for monoclinic/triclinic cells.
- ❌ **No regression test added**, despite the "needs testing" label that's been on this PR since 2022. A two-line test that constructs a triclinic `Structure` and checks `get_interplanar_angle(structure, (1,0,0), (0,1,0))` against a hand-computed value (or against the orthorhombic limit where old and new code agree) would lock in the fix and demonstrate the bug.
- 🔍 The PR still targets the old `pymatgen/analysis/diffraction/tem.py` path; pymatgen has moved to a `src/` layout, so this will need rebasing onto current `main` before merge.
- 🔍 CI hasn't run on this branch — please rebase, push, and confirm a green build.

Substance is correct and should land. Blocking on: rebase to src-layout, add a regression test on a non-orthogonal lattice, get CI green. Once those are in I'm happy to approve.

Issue #2673, opened 2022-09-30
Structure.charge and as_dict() bug in recent versions of pmg


fraricci:
Here what happens when I try to get a dict of a structure after applying a MagOrderingTransformation.

```
enumerator.transformations['afm'].apply_transformation(st).as_dict()

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-3d3739281aeb> in <module>
----> 1 enumerator.transformations['afm'].apply_transformation(st_p).as_dict()

~/python-venvs/main/lib64/python3.9/site-packages/pymatgen/core/structure.py in as_dict(self, verbosity, fmt, **kwargs)
   2206             "@module": self.__class__.__module__,
   2207             "@class": self.__class__.__name__,
-> 2208             "charge": self.charge,
   2209             "lattice": latt_dict,
   2210             "sites": [],

~/python-venvs/main/lib64/python3.9/site-packages/pymatgen/core/structure.py in charge(self)
    966         """
    967         if self._charge is None:
--> 968             return super().charge
    969         return self._charge
    970 

~/python-venvs/main/lib64/python3.9/site-packages/pymatgen/core/structure.py in charge(self)
    344         for site in self:
    345             for specie, amt in site.species.items():
--> 346                 charge += getattr(specie, "oxi_state", 0) * amt
    347         return charge
    348 

TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

I investigated where this comes from and found that, while the input structure has a `_charge` set to `0`, the output structure has a `_charge` set to `None'. This lead to call the charge function in the as_dict(). And,
as specie.oxi_state are None (I think this is general), the charge function is failing.

To note that in previous versions of pmg (at least 2022.0.4) the Structure.as_dict() had this line instead:
"charge": self._charge
This allows to avoid calling the charge function and not to crash for any values of _charge (int or None).

I see different problems here and I do not know what's the best approach:
1. simply go back to self._charge. I noticed that in the structure.py _charge is mostly used to get the charge
2. make MagOrderingTransformation set the `_charge=0` as in the input structure. I haven't check if other transformers have the same issue
3. the function to compute the charge could be improved to handle oxi_state=None

Let me know what you think.

arosen93:
[...]

PR #2791, opened 2023-01-06
Fixed charge function for Structure obj (also fix for #2673)

fraricci:
AFAIK, a site.species in Structure can be either an Element or a Species obj.
The Element obj does not have an "oxi_state" attribute and the current function handle
this case setting the oxi_state value to zero. 
But it fails when a site.species is a Species obj with an oxi_state set to None.
This is the case when a spin is added to a structure (via e.g. add_spin_by_site()) or
when using the MagOrderingTransformation .

The edit to the function proposed here handles this case and should solve the issue #2673

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -341,8 +341,7 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
         charge = 0
         for site in self:
             for specie, amt in site.species.items():
-                if hasattr(specie,"oxi_state") and specie.oxi_state:
-                    charge += specie.oxi_state * amt
+                charge += (getattr(specie, "oxi_state", 0) or 0)* amt
                     
         return charge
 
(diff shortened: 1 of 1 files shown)


pymatgen/core/structure.py
```diff
@@ -341,7 +341,9 @@ def charge(self) -> float:
         charge = 0
         for site in self:
             for specie, amt in site.species.items():
-                charge += getattr(specie, "oxi_state", 0) * amt
+                if hasattr(specie,"oxi_state") and specie.oxi_state:
```
janosh, line 344: Maybe less readable? Shorter in any case.

```suggestion
                charge += (getattr(specie, "oxi_state", 0) or 0)* amt
```

If we go with `hasattr`, I'd say let's be explicit:

```diff
- if hasattr(specie,"oxi_state") and specie.oxi_state:
+ if hasattr(specie,"oxi_state") and specie.oxi_state != 0:
```

fraricci: The first one works correctly. It takes 0 in case specie.oxi_state = None.
The second does not work as it is. It should be specie.oxi_state != None.

The first one is less clear, while the second more explicit.
I have no preference ;-)

shyuep: Pls just use getattr. Using hasattr is very verbose.

fraricci:
yes, that's right. Sorry about that.

Issue #2788, opened 2023-01-02
Many failing tests


JaGeo:
**Describe the bug**
Hi all, 

there are currently many, many failing tests connected to evaluations of structures.  Once the new year vacations are over, it would be great to resolve it so that contributing gets easier again.

janosh:
Hey @JaGeo, some tests started failing when we went from `numpy` 1.23.0 to 1.24.0 in CI. @mkhorton and me fixed some of those. I think there's just 1 of those remaining that requires updating 2 test files.

For other test failures I'm not sure how they arose. There's one in `ReadWriteChemenvTest.test_read_write_structure_environments()` which i discussed with @shyuep here
https://github.com/materialsproject/pymatgen/commit/0eb1e3d72fd894b7ba39a5129fbd8b18aedf4b46#commitcomment-92101479. Tried to fix it a while back but gave up.

Another in `pymatgen/analysis/tests/test_piezo_sensitivity.py`. I hope that's all of them but not sure.

Anyway, plan is to fix them asap but haven't had time recently. PRs welcome of course.

JaGeo:
@janosh Thanks for looking into it.

janosh:
I fixed 7 failing tests. There's just 1 left (I think) that's not straightforward to debug (the one in `ReadWriteChemenvTest.test_read_write_structure_environments` I mentioned). Maybe someone else has better luck than me with that one.

munrojm:
Came here to post about this. Thanks for looking at it @janosh!

JaGeo:
I think it is related to PeriodicNeigbors and as_dict and from_dict for this one. Species is still a dict after from_dict. I am trying to work on it but I am somehow stuck. 

I think it's this line: https://github.com/materialsproject/pymatgen/blob/e3bcad781fb83978fd7394ffc16d880e9ba9ab8d/pymatgen/analysis/chemenv/coordination_environments/structure_environments.py#L2096 
and this line:
https://github.com/materialsproject/pymatgen/blob/e3bcad781fb83978fd7394ffc16d880e9ba9ab8d/pymatgen/analysis/chemenv/coordination_environments/structure_environments.py#L2125

JaGeo:
Thank you very much for looking into this @janosh and thank you for the work.

JaGeo:
I think one problem might be the following:
`from pymatgen.core.composition import Composition

comp=Composition({'O2-': 1.0})

print(comp == comp.as_dict())`

[...]

PR #2792, opened 2023-01-06
One attempt to fix ChemEnv error in tests

JaGeo:
## Summary

As discussed in https://github.com/materialsproject/pymatgen/issues/2788, one way of fixing the ChemEnv error in the tests would be converting the PeriodicNeighbor object in a PeriodicSite object. I have implemented this. It might however affect the execution speed.

If someone else, manages to fix the as_dict from PeriodicNeighbor in a cheaper way, I would prefer this, of course.

### DIFF
--- a/pymatgen/analysis/chemenv/coordination_environments/structure_environments.py
+++ b/pymatgen/analysis/chemenv/coordination_environments/structure_environments.py
@@ -24,7 +24,7 @@ from pymatgen.analysis.chemenv.utils.chemenv_errors import ChemenvError
 from pymatgen.analysis.chemenv.utils.defs_utils import AdditionalConditions
 from pymatgen.core.periodic_table import Element, Species
 from pymatgen.core.sites import PeriodicSite
-from pymatgen.core.structure import Structure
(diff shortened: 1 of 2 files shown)


JaGeo:
Linting still fails but tests are running through. 
Please let me know what you think and if you have any better/faster ideas (@shyuep , @janosh). Both species and coordinates  are influenced when using from_dict and as_dict from PeriodicNeighbors.

janosh:
This doesn't look like much a slowdown anyway.

janosh:
Thanks @JaGeo. Good to have green CI again. Like you said, makes contributing much easier. I'll close #2788 now.

JaGeo:
Just adding this here: https://github.com/materialsproject/pymatgen/blob/c1a7f5d7635f85e7bc490658d52b70867ed24614/pymatgen/core/structure.py#L158 
This name is a bit unfortunate. If as_dict as inherited from PeriodicSites is then used, "coords" (typically fractional coordinates  but now cartesian ones due to the overloading) is used and this might lead to the other problems with from_dict and as_dict for PeriodicNeighbors (besides the dict issue for the species).

PR #2796, opened 2023-01-09
Added a missing multi element POTCAR file 

MichaelWolloch:
and reverted the file loading change in commit e3bcad781fb83978fd7394ffc16d880e9ba9ab8d

## Summary

- Fix 1: Fixed a unit test that failed due to a missing file by actually committing the new file. The unit test was already fixed by @janosh in e3bcad781fb83978fd7394ffc16d880e9ba9ab8d, by taking another POTCAR, but this kind of defeated the purpose of the test which was to check multi-element POTCARs.


janosh:
@MichaelWolloch Thanks for the fix! 👍

MichaelWolloch:
Thank you @janosh for finding the missing file!

PR #2799, opened 2023-01-10
Added 4-color electronic band plot in BSDOSPlotter

cnncnnzh:
## Summary

Added a feature that supports cmyk band structure plot for 4-element structures, and added the corresponding _cmyk_triangle method for legend. 

Figure 2023-01-10 010306


test_files/KPOINTS_SrBa2Sn2O7_bands
```diff
@@ -0,0 +1,16 @@
+K-points for bandstructure G-X-P-N-G-M
```
janosh, line 1: If possible, it would be great to avoid adding large new test files. Would it be possible to reuse an existing `vasprun.xml` in your test? Maybe `lifepo4.xml`?

janosh (changes requested):

test_files/KPOINTS_SrBa2Sn2O7_bands
cnncnnzh: Is the lifepo4.xml generated with LORBIT>=11, and is there a corresponding KPOINTS file? Or I can upload another structure with a smaller .xml file. Does that work? What will be the maximum size you can accept?

janosh: No, `lifepo4.xml ` has `LORBIT` `F`. Since you're testing plotting code and not file parsing, maybe you could serialize the `pymatgen` via their `as_dict` methods and save the output as JSON, then reload it with `from_fict` in the test?

janosh: Looks numpy related. Maybe todo with the 'fixes' we applied after the `numpy` v.1.24.0 breaking changes. This test case should trigger it though:

https://github.com/materialsproject/pymatgen/blob/44f24db50d3b59ad9bf448c523564c114f472f19/pymatgen/electronic_structure/tests/test_bandstructure.py#L215-L221

What `numpy` version are you running?

I just checked out your PR and ran the updated test. No error on my end.

```py

        band_struct = v.get_band_structure(
            kpoints_filename=os.path.join(PymatgenTest.TEST_FILES_DIR, "KPOINTS_SrBa2Sn2O7_bands")
        )
        print(f"{sorted(band_struct.as_dict())=}")
```

prints

```
sorted(band_struct.as_dict())=['@class', '@module', 'band_gap', 'bands', 'branches', 'cbm', 'efermi', 'is_metal', 'is_spin_polarized', 'kpoints', 'labels_dict', 'lattice_rec', 'projections', 'structure', 'vbm']
```

cnncnnzh: I'm using v.1.24.0.
The `as_dict` function works fine for me. The error comes out when I dump the dict to a json file.

janosh: Oh yes, my bad. Sorry it's a bit late here. You could pass a custom `handler` to `json.dump` to convert to normal `int`. Something like (untested)

```py
def handler(obj: Any):
    if isinstance(obj, np.integer):
        return int(obj)
```

Though I'm wondering if maybe that should be part of `BS.as_dict`. Writing to JSON is a common reason far `as_dict`.

cnncnnzh: Hi Janosh, I replaced the .xml file with the smaller .json file.

janosh:
`test_files/SrBa2Sn2O7.json` is just 1 line now which is great but still 13.7 MB. pymatgen is really slow to clone since it's already 830 MB.

pymatgen size

We're planning to bring this down if at all possible but definitely don't want to add to it. Could you check how much of that JSON data is actually necessary for the test? Maybe some keys with large data can be removed?

janosh:
Sorry to drag this out so much...

cnncnnzh:
I totally understand. The removal of unnecessary keys doesn't make any big effect. I rerun the calculation with a shorter k path and a smaller number of k points. Now the .json file is ~4 MB. The plot is not pretty but I think it's enough for the purpose of testing. Does that work? Or do you need an even smaller file?

janosh:
Thanks! 4 MB is still pretty big but acceptable imo. Only remaining suggestion would be to generate the plotting data on the fly with `np.random.rand()` since like you say, we don't care what the plot looks like. But I leave that to you if you want to look into that or call it finished here.

cnncnnzh:
I will try that. Should I just randomize the 'projections' in the data?

janosh:
Yes, that would be great

cnncnnzh:
Hi Janosh, when I randomize the projection from every atom, the lines turn completely black, which makes sense to me. So for each k and each band, I randomly pick up two atoms and assign random contributions to them. Does that sound good to you?
[pband.pdf](https://github.com/materialsproject/pymatgen/files/10442852/pband.pdf)

janosh:
> Does that sound good to you?

Perfect!

cnncnnzh:
Hi Janosh, I saw that this request is still open. Is there any further change that I need to make?

janosh:
Sorry, I should have been more clear. I thought you were still going to remove the large [JSON file](https://github.com/materialsproject/pymatgen/blob/cbaadb808b81fd4a31dee0733b2277d5630a8fbd/test_files/SrBa2Sn2O7_random.json) and replace it with some calls to `np.random.rand` directly in the test code. (so as to keep repo size and clone time down).

cnncnnzh:
Sorry, I misunderstood your meaning. I generated the random data and save them in .json. I'll generate them in the test.py file.

Issue #2801, opened 2023-01-12
`ComputedEntry.__eq__` method not suitable for new MP entry_id format


mattmcdermott:
Materials Project recently introduced new R2SCAN data, and with it, a new `entry_id` naming scheme where entry IDs include the functional type (e.g., _mp-9-R2SCAN_ ).

Because of this, there is now an issue that will likely be more frequently encountered: `ComputedEntry` objects with the same `entry_id` may evaluate as being equal, even when they have different `energy_adjustments` (i.e., for different mixing schemes). 

For example, this currently makes it problematic to create `EntrySet` objects containing entries downloaded from MP, because entries are incorrectly being flagged as duplicates. Here's an example of code that illustrates this directly. This applies with `mp-api==0.30.5` and `pymatgen==2023.1.9`. (Future mp-api releases will not return multiple types of entries, though):

```python
from mp_api.client import MPRester

with MPRester() as mpr:
    ents = mpr.get_entries_in_chemsys("Y-Mn-O")

r2_scan_ymno3 = []
for e in ents:
    if "mp-19385-R2SCAN" in e.entry_id:
        r2_scan_ymno3.append(e)
    
e1, e2 = r2_scan_ymno3  # two entries with the same entry_id

print("Entry objects equal?", e1 == e2)
print("Energies equal?", e1.energy == e2.energy)
```
This prints:
```
Entry objects equal? True
Energies equal? False
```

I suggest we make the `ComputedEntry.__eq__` method a little more encompassing to account for situations like this. I will make a PR shortly with a proposed fix @munrojm @mkhorton

mattmcdermott:
Closing with merged fix PR #2802

PR #2802, opened 2023-01-12
Fix for `ComputedEntry.__eq__` to allow more robust checking with new MP entries

mattmcdermott:
This PR addresses #2801.

In summary, I made the check between `entry_id` values only capable of returning `False`, so that if two entries have the same `entry_id` they will only evaluate as equal if they also have the same energy and composition. I also added a test to check this.

This does slow down certain operations like `EntrySet` creation, but only when there are many entries with duplicate entry_id values. If all entries have unique entry_ids (as was common before the new MP entry naming scheme), then this should not affect performance. 

Please let me know if you can think of any other problems this might cause -- in my experience, bugs with entry equality checking are particularly insidious... @mkhorton @munrojm 

## Checklist
- [x] All linting and tests pass..


shyuep:
I am a bit concerned as to why this is needed. Technically entry_ids should be unique. Why would there be an instance where there are many same entry_ids with different energies?

mattmcdermott:
@shyuep Please see the example in Issue #2801 for why this may be a more common problem now. In short, the new R2SCAN data added to MP can be downloaded with or without the new mixing scheme adjustment; in both cases, the entries have the same `entry_id` and are said to be equal, even when they have different final energies.

shyuep:
I don't think this is a wise way of dealing with mixing adjustments. The way the entry adjustments are done is that the adjustments are stored differently from the raw energies from the calculations themselves. That logical framework should be maintained. I am unclear what kind of situation would require a user to download both types of entries.

mattmcdermott:
I agree that, in principle, entries should represent calculation results independent of whatever corrections are applied. However, in practice, it is often the final (corrected) energy & composition that are most important in analysis... so it becomes problematic when two entry objects are considered equal and behave differently (e.g., in phase diagram construction). 

While this particular situation was an undesired result of `MPRester` returning both entry types (which was fixed yesterday in the last `mp-api` release), I still think the situation could arise again for people querying new MP entry data. Perhaps this Issue/PR could be left open and reevaluated if it becomes a problem again for someone else?

mkhorton:
Yes, the `entry_id` maps to a specific calculation (e.g., combination of total energy, functional/methods, etc.), but a specific calculation may be adjusted or corrected in multiple ways, leading to different final energies with the same entry_id.

shyuep:
There is a difference between whether something can happen and whether something should happen. I get that you can apply different corrections to the same entry. But for all practical purposes, you use one definitive value of the energy for a material. Someone needs to educate me on a situation where differently corrected entries are useful in a scientific application.

janosh:
Isn't this mostly a question of which behavior is least likely to mislead users? Telling them two entries are equal even if they have different energy adjustments seems like it would surprise people. Like the `5 == '5'` debacle in JS. Imagine as a user testing

```py
entry_1 == entry_2
>>> True
```

but then seeing

```py
entry_1.energy_adjustments
>>> [a, b, c]

entry_2.energy_adjustments
>>> [x, y, z]
```

shyuep:
It is a balance between efficiency and use case. If someone can explain to me an actual useful situation where you want multiple entries with different corrections, I am more than happy to sacrifice efficiency to make it happen. I would also point out that corrections can be applied and changed at any time. That is actually the only way to ensure that a set of entries have a consistent set of corrections.

shyuep:
I would also point out that mismanaging entries can have severe consequences for the accuracy of an analysis. Let's say the user intends to use entries that only has one correction scheme (say the normal MP PBE/PBE+U correction scheme). But there are multiple entries with different correction schemes present and one of the correction schemes result in a lower energy than the MP correction scheme. What will happen if you do any phase diagram or thermodynamic analysis based on ground states is that the lower energy entry will prevail. 

In short, the MP API needs to return one and only one entry per entry id, with a well-defined default correction scheme. Return multiple entries per entry id is a recipe for disaster, regardless of whether we check for energy equality. If an advanced computational person wants to explore different correction schemes and their effects, they can do so with post-processing using the correction schemes implemented in pymatgen.

munrojm:
I just want to add to this thread and explain what the API is doing and how the data is structured. Entry data is returned via the thermo endpoint, which has a `thermo_id` as its key. This is a combination of the MPID and a `ThermoType` which indicates the mixing scheme. The entries returned are consistent with the `ThermoType` that is requested. By default, this is set to GGA/GGA+U. 

Now that we have R2SCAN data, we build and offer thermo data using three schemes. Mixed GGA/GGA+U, mixed GGA/GGA+U/R2SCAN, and sole R2SCAN. Since each `ComputedStructureEntry` is associated with a calculation, the `entry_id` now contains the MPID and the run type (e.g. R2SCAN). Because of the multiple schemes, we now have the same `ComputedStructureEntry` associated with an R2SCAN calculation being used both in the GGA/GGA+U/R2SCAN thermo data where it has a correction applied, and in the sole R2SCAN thermo data where it does not.

janosh:
@shyuep During this morning's MP update meeting, @mkhorton and me both voted in favor of merging this PR. Would you like to hold off and discuss this some more?

mkhorton:
Agreed; the logic here is "what is expected behavior?"

Specifically, the example of `entry1 == entry2 == True` even if `entry1.energy != entry2.energy` or `entry1.energy_adjustments != entry2.energy_adjustments`.

Regardless of what Materials Project is doing with its entry_ids, this behavior seems unexpected and incorrect.

mattmcdermott:
One more quick comment regarding the concern about efficiency. The new scheme is this:
```python
if self.entry_id != other.entry_id:
    return False
```
This means that the code will still perform just as fast as before for collections of entry objects where all entry IDs are unique. The next steps in equality checking only occur for entries that share the same `entry_id`.

shyuep:
@mkhorton @janosh I agree only partly. By default, all objects do not implement equals and you only check for equality based on the id. The question here is - what should we encourage the user to do? Mess around with multiple entries with the same entry_id but different corrections, or just have one entry with a single entry_id. I doubt you will say let the user go ahead and play around with entries as they see fit.

janosh:
@shyuep Isn't what users do with the code beside the point? All we care about is that the equality operation between entries doesn't return misleading results no matter what users decide to do.

shyuep:
@janosh By that argument, this code should never exist in Python:

```
In [3]: class A:
   ...:     def __init__(self, a):
   ...:         self.a = a
   ...: 

In [4]: A(1) == A(2)
Out[4]: False
```

It does. And it is the *default* unless you implement something else.

shyuep:
Anyway, since both @janosh and @mkhorton already voted, just go ahead and merge it for now. I will revert only if there are complaints about speed, despite my misgivings about allowing users to have two entries with different corrections.

janosh:
Thanks @mattmcdermott and @shyuep for green lighting!

mkhorton:
Thanks for merge.

Regarding the "this code should never exist in Python [...] it is the *default*" argument, I would offer a counter-example:

```
In [1]: from dataclasses import dataclass

In [2]: @dataclass
   ...: class A:
   ...:     a: int
   ...:

In [3]: A(1) == A(2)
Out[3]: False
```

Semantically, I think the purpose of something like an `Entry` is fairly well-aligned with what dataclasses have been designed for, and there the default behavior is to return `False`. The default behavior of a minimal subclass of object is different, because there, speed is the absolute concern since everything is an object, and any default behavior will therefore have a penalty associated with it.

That's all I have to offer here; not relevant now we're merging, but perhaps in case a similar instance occurs in future.

shyuep:
@mkhorton Actually a ComputedEntry is deliberately **not** a dataclass... anyways...

mkhorton:
Perhaps it should be; its purpose is to store data. I always thought these did not use dataclasses simply because they predate dataclasses in the standard library. I mention this not to create an argument, but sincerely think they're very useful and appropriate -- I know @jmmshn has opinions on this too.

jmmshn:
2cents: I guess we also didn't use dataclasses because they have difficult with handling inheritance even after things were "[fixed](https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5)" in  python 3.10
But they do save a tremendous amount of dev time when it comes to building the APIs so I think they should be used wherever some objects gets very close to the what we serve the users.

shyuep:
It is true that entries predate dataclasses. But that is not the only reason why entries do not use dataclasses. Don't get me wrong, I think dataclasses are great for a lot of purposes. But in the case of entries, I see the use of dataclasses as hindering rather than improving readability. Inheritance was one big factor. People may not realize this now, but the idea of having a GrandPotEntry being an inherited version of Entry and using essentially the same phase diagram code otherwise was something that had major implications for the rest of the design of the code.

Also, I still prefer my entries defined by structure/composition uniqueness rather than internal property uniqueness.

janosh:
This PR addresses a related issue reported here https://github.com/CompRhys/pymatgen/issues/1: unequal hash for equal entries. Before, pymatgen broke the only requirement placed on a hash function, which is that objects which [compare equal have the same hash value](https://docs.python.org/3/reference/datamodel.html#object.__hash__). Thanks for pointing that out @CompRhys!

```py
>>> from pymatgen.entries.computed_entries import ComputedEntry
>>> from pymatgen.analysis.phase_diagram import PDEntry
>>> pdentry = PDEntry("NaCl", -1.3)
>>> centry = ComputedEntry("NaCl", -1.3)
>>> hash(pdentry) == hash(centry)
False
>>> pdentry == centry
# before this PR
True
# after
False
>>> 
```

PR #2803, opened 2023-01-13
Add some weight-fraction-based functions

amkrajewski:
## Summary

When working with metallic alloys, weight-fraction-based notations such as Ti64 / Ti-6V-4Al or NiTiNOL60 / Ni-40Ti are commonly employed in both industrial specifications and scientific literature. Regardless of the numerous downsides of this situation, including errors in scientific experiments or NLP-parsing when they are mistaken for atomic fractions or chemical formulas, being able to create a Composition object from them (under correct interpretation) would be a useful pymatgen feature.

- Composition class method to initialize it from a dictionary of weight fractions
- Composition property giving a dictionary of weight fractions
- concise tests for the two above were added


shyuep:
Thanks for the useful contribution!

PR #2816, opened 2023-01-24
Bug in get_plotly method in the WulffShape

CifLord:
## Summary

Minor bug fix to the plotly interface for the WulffShape class. 

- Plotly WulffShape was trying to normalize the value of each surface energy by zero whenever only one facet existed on the Wulff Shape. The normalized surface energy of only one facet on the shape will be changed to 1 by default to avoid division by zero.


janosh:
@CifLord Thanks! Could you add or modify an existing test that would have triggered the error and now passes thanks to the `if max(e_surf_on_wulff) == min(e_surf_on_wulff):` case?

PR #2818, opened 2023-01-26
Minor fix to `get_element_profile`

jmmshn:
# Minor fixes to the return types of `get_element_profile` and electrode

`get_element_profile` fix:
The entries returned is given as `dict_keys` but that's rarely what people want.
- Return a list of entries instead.
- Also return the critical composition, which is usually useful information.

electrode fixes:
- Linting on some docs
- require framework formula for the `from_steps` constructor.


janosh:
Thanks @jmmshn! 👍

PR #2830, opened 2023-02-04
Avoid adjusting input entries in mixing schemes.

peikai:
Here to submit changes to avoid change input entries in place.

In both of GGA/GGA+U and GGA/GGA+U/R2SCAN mixing processes, the methods below not only return adjusted entries, but also change input entries in place. Is it necessary? I think this may cause bugs in programming.

MaterialsProjectDFTMixingScheme.process_entries()
entries.compatibility.process_entries()

Thanks!

### DIFF
--- a/pymatgen/entries/compatibility.py
+++ b/pymatgen/entries/compatibility.py
@@ -582,7 +582,10 @@ class Compatibility(MSONable, metaclass=abc.ABCMeta):
 
         processed_entry_list: list[AnyComputedEntry] = []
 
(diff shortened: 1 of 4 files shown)


peikai:
I found that the usage of process_entries() method needs attention. It only returns the entries that have been processed, not always the complete set of adjusted entries. If input entries have contained energy adjustments, they would not be processed again, and even would not be shown in the results that are returned. I would close this PR, to avoid changing this usage.

peikai:
This PR is renewed in #2841.

PR #2853, opened 2023-02-20
Fix OPTIMADE client generated filters and URLs

### LINKED ISSUE #2852: OPTIMADE client generates malformed queries
**Describe the bug**
At some point, the OPTIMADE client in this repo was linted (presumably) and `f'?response_fields="{response_fields}"...'`  was replaced with `f'?response_fields={response_fields!r}'`, which generates query parameters with the wrong (according to the OPTIMADE spec, at least) kind of quotes:

https://github.com/materialsproject/pymatgen/blob/ebec8497c05055db3b78fec4c4bba75aaefb7839/pymatgen/ext/optimade.py#L316

This renders the client useless, as every request has such malformed queries.

I'm happy to fix this (will make a PR straightaway), but please let me know if this could be released before the end of the week --- if not I will have to adjust the pymatgen OPTIMADE tutorials that are being delivered at a conference on the 28th Feb.

ml-evs:
Closes #2852 by properly quoting OPTIMADE query URLs without using f-strings.

Also fixes an issue where `chemical_formula_hill` was not queryable.

### DIFF
--- a/pymatgen/ext/optimade.py
+++ b/pymatgen/ext/optimade.py
@@ -172,7 +172,7 @@ class OptimadeRester:
         if elements:
             if isinstance(elements, str):
                 elements = [elements]
-            elements_str = ", ".join(f"{el!r}" for el in elements)
+            elements_str = ", ".join(f'"{el}"' for el in elements)
             filters.append(f"(elements HAS ALL {elements_str})")
 
         if nsites:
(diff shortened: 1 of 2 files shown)


ml-evs:
Sorry, took a few attempts to get around the linter... this should be good to go.

PR #2861, opened 2023-02-27
Remove module pymatgen/command_line/gulp_caller.py

janosh:
Don't want to step on anyone's feet so I'll leave this PR open for a while but looks like this might be dead code (based on not knowing anyone who still uses `gulp`).

@shyuep Let me know if this is a mistake.


shyuep:
Why is this considered dead code? Though I don't use it, I was under the impression that it is working and some people actually use it.

janosh:
My line of inquiry was following [this docs sentence](https://github.com/materialsproject/pymatgen/blob/40cbf1d7c47bb2eb75509780e356c9877ae9c2f7/docs_rst/installation.rst#optional-non-python-programs) saying the pymatgen `gulp_caller` module is "used extensively" by `pymatgen-analysis-defects`. I found no mention of gulp or `GulpIO` or `TersoffPotential` anywhere in `pymatgen-analysis-defects`.

And also because [interest in gulp has steadily dropped](https://trends.google.com/trends/explore?date=all&geo=US&q=%2Fg%2F11byyl8mbg) over the years.

**Google Trends**

Screenshot 2023-02-27 at 10 32 23

shyuep:
I agree that GULP interest has decreased. But I think it is useful to just leave this unless it bothers someone majorly. It is still useful if you want to do some benchmarks with GULP based simulations.

PR #2864, opened 2023-02-28
fix reduced formula in Ion

yang-ruoxi:
Fixed the handling of O-H containing compositions, where the string 'HO' was replaced by 'OH' directly in the reduced formula. This will cause certain materials, e.g. HSbO2, to have the wrong reduced formula of SbOH2.

This fix then adds a condition that the composition must have an equal number of H and O to be handled that way.


mkhorton:
Thanks for catching this!

PR #2876, opened 2023-03-05
IO: add passthrough kwargs to write / from_file

rkingsbury:
Adds passthrough keyword arguments in `write_file` and `from_file`. This allows developers to add keyword arguments to `get_string` and `from_string` methods when implementing `InputFile` classes and still rely on the base `write_file` and `from_file`.

Prompted by my review of https://github.com/materialsproject/pymatgen/pull/2692


rkingsbury:
Nevermind; still impossible to do this properly and satisfy `mypy`.

PR #2878, opened 2023-03-06
New class to handle `NcICOBILIST.lobster` files

QuantumChemist:
## Summary

Implemented a class to handle multi-center ICOBIs via NcICBILIST.lobster files with pymatgen in `pymatgen/io/lobster/outputs.py` and updated some comments. `Ncicobilist` is similar to `Icohplist` but not identical because of the different format styles of NcICOBILIST.lobster and ICO**LIST.lobster output files.

Added the respective NcICOBI test files to `test/files/cohp` for non-spinpolarized, non-orbitalwise case etc.

## Todo (if any)

Checking LOBSTER version (done)
Adapt for orbitalwise case. (will leave that out for now)


janosh:
I think the errors here should (mostly) disappear if you pull in `master` again.

QuantumChemist:
Hi @janosh thx for looking into this :)
Might be that people cannot use it right away because we (+ @JaGeo ) found several problems in how LOBSTER outputs the NcICOBILIST.lobster file (but my test files are not affected as I created them mostly manually).

QuantumChemist:
> Hi @janosh thx for looking into this :) Might be that people cannot use it right away because in the meantime we (+ @JaGeo ) found several problems in how LOBSTER outputs the NcICOBILIST.lobster file (but my test files are not affected as I created them mostly manually).

The failing checks are caused by my implementation of the wrong format for the pytests. I will correct that within the next weeks and then I hope this can be merged, even though for now certain NcICOBILIST.lobster outputs should be handled with some caution.

QuantumChemist:
Hi @janosh  :D 
could you help me to figure out what is making the tests fail and how we can make this PR successful?

JaGeo:
Could you do an empty commit to get another run of the tests, @QuantumChemist ?

`git commit --allow-empty -m "Empty-Commit"`

QuantumChemist:
> Could you do an empty commit to get another run of the tests, @QuantumChemist ?
> 
> `git commit --allow-empty -m "Empty-Commit"`

I had to commit some changes anyways :)

pymatgen/io/lobster/outputs.py
```diff
@@ -408,6 +409,113 @@ def icohpcollection(self):
         return self._icohpcollection
 
 
+class Ncicobilist:
```
janosh, line 412: Why not pascal case `NciCobiList` here? Easier to read imo.

.attach_pid50466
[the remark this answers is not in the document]
QuantumChemist: oopsie, sorry that was accidental. I created a mess when I merged the two versions

pymatgen/io/lobster/outputs.py
QuantumChemist: Here I wanted to stick to only capitalizing the first letter like in  `class Icohplist` 
For me `NciCobilist` is not very readable as the method referring to is NcICOBI, so maybe we can go with `NcIcobilist` or `ncIcobilist` ?

test_files/cohp/NcICOBILIST.lobster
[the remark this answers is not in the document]
QuantumChemist: The files mentioned here are not the ones I need. Seems that they haven't been fetched in the merge process. I gonna upload them quickly.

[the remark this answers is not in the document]
QuantumChemist: Oh now I see with the merging process they have just been added to test_files/cohp (so like it was structured before). I gonna fix that now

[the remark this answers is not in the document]
QuantumChemist: so the files in question are 
```
$ ls tests/files/** | grep NcICOBI
NcICOBILIST.lobster
NcICOBILIST.lobster.gz
NcICOBILIST.lobster.nospin
NcICOBILIST.lobster.nospin.withoutorbitals
NcICOBILIST.lobster.withoutorbitals
```

please let me know if you see something unnecessary. From my POV this is the least user cases that should be checked. 
Thank you for helping me with this PR :)

.attach_pid50466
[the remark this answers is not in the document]
QuantumChemist: I removed it now :)

pymatgen/io/lobster/outputs.py
QuantumChemist: Hi @janosh, please let me know what you think about this. If it's not ok, I'm gonna adopt your suggestion.

janosh: I still prefer consistency with the rest of `pymatgen` which uses PascalCase for all class names. E.g. we have `PotcarSingle` and not `POTCARSingle` for that reason. But maybe another maintainer wants to chime in here.

JaGeo: (It does not have to be consistent with the older Lobster class names.)

QuantumChemist: @janosh, ok it's done :)))

pymatgen/io/lobster/outputs.py
```diff
@@ -408,6 +409,109 @@ def icohpcollection(self):
...
+    """
+
+    def __init__(self, filename: str | None = None):
+        """
+        Args:
+            filename: Name of the NcICOBILIST file.
+        """
+
+        if filename is None:
+            warnings.warn("Please consider using the newest LOBSTER version (4.1.0+). See http://www.cohp.de/.")
```
janosh, line 431: Is this warning correctly indented? Why only warn is `filename is None`?

pymatgen/io/lobster/outputs.py
```diff
@@ -408,6 +409,109 @@ def icohpcollection(self):
...
+
+        # If the calculation is spin-polarized, the line in the middle
+        # of the file will be another header line.
+        self.is_spin_polarized = "spin" in data[len(data) // 2]  # TODO: adapt this for orbitalwise case
+
+        # check if orbitalwise NcICOBILIST
+        # include case when there is only one NcICOBI
+        for entry in data:  # NcICOBIs orbitalwise and non-orbitalwise can be mixed
+            if len(data) > 2 and "s]" in str(entry.split()[3:]):
+                self.orbitalwise = True
```
janosh, line 449: Is `self.orbitalwise` missing a default value of `False`? Tests are currently failing with

```
AttributeError: 'NciCobiList' object has no attribute 'orbitalwise'
```

pymatgen/io/lobster/outputs.py
```diff
@@ -408,6 +409,109 @@ def icohpcollection(self):
...
+                warnings.warn(
+                    "This is an orbitalwise NcICOBILIST.lobster file. "
+                    "Currently, the orbitalwise information is not read!"
+                )
+                break  # condition has only to be met once
+
+        if self.orbitalwise:
```
janosh, line 456: Maybe rename to `self.orbital_wise` for readability?

pymatgen/io/lobster/outputs.py
```diff
@@ -408,6 +409,109 @@ def icohpcollection(self):
...
+            self.list_labels.append(label)
+            self.list_n_atoms.append(n_atoms)
+            self.list_ncicobi.append(ncicobi)
+            self.list_interaction_type.append(interaction_type)
+            self.list_num.append(num)
+
+        # TODO: add functions to get orbital resolved NcICOBIs
+
+    @property
+    def ncicobilist(self) -> dict[Any, dict[str, Any]]:
```
janosh, line 500: Maybe rename to `def ncicobi_list()`?

pymatgen/io/lobster/outputs.py
```diff
@@ -433,27 +433,23 @@ def __init__(self, filename: str | None = None):
...
 
         # check if orbitalwise NcICOBILIST
         # include case when there is only one NcICOBI
         for entry in data:  # NcICOBIs orbitalwise and non-orbitalwise can be mixed
             if len(data) > 2 and "s]" in str(entry.split()[3:]):
                 self.orbitalwise = True
                 warnings.warn(
-                    "This is an orbitalwise NcICOBILIST.lobster file. Currently, the orbitalwise information is not "
-                    "read!"
+                    "This is an orbitalwise NcICOBILIST.lobster file. "
```
JaGeo, line 451: Is it possible to get orbital-wise interactions for multicenter ICOBIs?

janosh (changes requested):
Thanks @QuantumChemist. Left some minor comments about naming and failing test.

pymatgen/io/lobster/outputs.py
JaGeo: It looks like a leftover from the icohplist. I remember implementing this (with another if clause). I guess this can be removed.

pymatgen/io/lobster/outputs.py
QuantumChemist: Yes, it is and the output format can be very complex with more than 3 centers or mix of n center interactions.

JaGeo: Thanks 😃

pymatgen/io/lobster/outputs.py
QuantumChemist: Oh, I have to look more into this... :O

pymatgen/io/lobster/outputs.py
QuantumChemist: Also when Peter and I implemented the COBI, we never really thought of a high-throughput application, so the NcICOBIs are ordered as specified in the lobsterin... mixing different n center interactions up, therefore results in a very messed up NcICOBILIST.lobster file, especially when some are orbitalwise and other are not 🙈 need to check if and how that changed for lobster version 5 before implementing an orbitalwise feature here.

JaGeo: 😅

pymatgen/io/lobster/outputs.py
JaGeo: Wouldn't you still use the default naming option in such a case and then then the file wouldn't be there? Thus, shouldn't it be the warning if the default file name does not exist rather than when it is set to `None`?

pymatgen/io/lobster/outputs.py
```diff
@@ -408,6 +409,109 @@ def icohpcollection(self):
...
+
+    Attributes:
+        is_spin_polarized (bool): Boolean to indicate if the calculation is spin polarized.
+        NciCobiList (dict): Dict containing the listfile data of the form:
+        {bond: "number_of_atoms": number of atoms involved in the multi-center interaction,
+                "ncicobi": {Spin.up: Nc-ICOBI(Ef) spin up, Spin.down: ...}},
+                "interaction_type": type of the multi-center interaction
+    """
+
+    def __init__(self, filename: str | None = None):
```
JaGeo, line 424:
```suggestion
    def __init__(self, filename: str | None = "NCICOBILIST.lobster"):
```
I think the use case is that someone has computed multicenter COBIs and then wants to read the file. Thus, the default name should be there. In a larger code, I would check if the file existed, before I use the class.

QuantumChemist: I just commented this from my phone as I randomly remembered why I did this at all while doing something else xD. I will work on this properly tomorrow :)

QuantumChemist: But yes, this is a much better way to put the filename. 👍🏻

QuantumChemist:
Hi @janosh :D
I think I addressed all issues.

janosh (approved):

QuantumChemist:
yay 🎉 thank you very much for your help @janosh!

PR #2881, opened 2023-03-07
Handle serial version of VASP when reading the OUTCAR

gbrunin:
## Summary

When compiling VASP without MPI support, the OUTCAR has a few differences in some places compared to a compilation including MPI. 
- In the preamble, the number of cores is not indicated with `running on   16 total cores` but `serial version`
- When reporting the number of plane waves for each k-point, in the serial version, only the `maximum number of plane-waves` is reported while in the parallel version, the `maximum and minimum number of plane-waves per node` is reported before and used as a footer_pattern to read the table-like data.

Of course, compiling the serial version of VASP must be a pretty rare thing to do, and I did it for testing only. 
In this small PR, I simply accounts for these changes and add a small test. 

I did not check whether other output files were affected by this.


janosh:
Thanks! Even if it's an edge cases, always good to handle things correctly.

Could you compress `test_files/OUTCAR.serial` and then load the zipped file in your test? Important to minimize file sizes for people cloning the repo.

gbrunin:
Sure! Done now.

janosh:
Thanks! 👍

PR #2886, opened 2023-03-10
MPAqueousCompatiblity: compute hydrate correction using reduced rather than full composition

rkingsbury:
## Summary

Small change to the way `MaterialsProjectAqueousCompatibility` corrects energies for materials containing `H2O`.

As [described by a user on MatSci](https://matsci.org/t/mpaqueouscompatibility-applies-hydrate-correction-to-non-hydrate-materials/47360/2), the current way of detecting whether a material contains 'H2O' can lead to false positives for formulas like `Li2O2H2`, causing this material to be corrected as if it were a hydrate when in fact it is not.

Here, I change the detection of hydrates to be based on `reduced_composition` instead of `composition` to avoid such false positives.

Tests appear to pass, but flagging @montoyjh and @as2362 in case there are concerns about how this will impact Pourbaix Diagrams. This should cause the energies of some -OH phases to change a bit because they won't receive the hydrate correction.

### DIFF
--- a/pymatgen/entries/compatibility.py
+++ b/pymatgen/entries/compatibility.py
@@ -1357,8 +1357,8 @@ class MaterialsProjectAqueousCompatibility(Compatibility):
         # TODO - detection of embedded water molecules is not very sophisticated
         # Should be replaced with some kind of actual structure detection
 
-        # For any compound except water, check to see if it is a hydrate (contains)
-        # H2O in its structure. If so, adjust the energy to remove MU_H2O ev per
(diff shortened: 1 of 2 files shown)


rkingsbury:
Since no concerns / comments have been raised, I'm going to move this out of WIP status.

arosen93:
@janosh soubds like this is ready to go.

janosh:
@arosen93 Thanks for the ping!

@rkingsbury Thanks for adding tests! 👍 Merging now.

PR #2891, opened 2023-03-13
Read input parameters for both internal and cartesian coordinates

xjf729:
If there is only one blank line between coord_lines and input_lines, input_paras will be ignored. So set "spaces" as 1 to avoid this case.


janosh:
This is causing some test failures. Could you take a look?

```py
                        ind = toks.pop(0)
                        data = toks.pop(0)
                        try:
                            nn.append(int(ind))
                        except ValueError:
                            nn.append(species.index(ind) + 1)
                        try:
>                           val = float(data)
E                           ValueError: could not convert string to float: 'B1'

pymatgen/io/gaussian.py:228: ValueError

During handling of the above exception, another exception occurred:

self = <tests.test_gaussian.GaussianInputTest testMethod=test_from_file>

    def test_from_file(self):
        filepath = os.path.join(test_dir, "MethylPyrrolidine_drawn.gjf")
>       gau = GaussianInput.from_file(filepath)

pymatgen/io/tests/test_gaussian.py:[79]
```

janosh:
Thanks! 👍

PR #2896, opened 2023-03-18
Enable Structure relaxations with TrajectoryObserver

janosh:
813019427e add `return_trajectory` option to `Structure.relax()`
 64dee2a8ba test new option in `test_relax_with_observer()`
 db0871a327 convert vars to snake case

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ ci:
 
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.255
+    rev: v0.0.257
     hooks:
       - id: ruff
         args: [--fix, --ignore, D]
--- a/pymatgen/analysis/chemenv/coordination_environments/tests/test_weights.py
+++ b/pymatgen/analysis/chemenv/coordination_environments/tests/test_weights.py
@@ -5,6 +5,7 @@ import os
 import unittest
 
 import pytest
+from pytest import approx
 
 from pymatgen.analysis.chemenv.coordination_environments.chemenv_strategies import (
     AngleNbSetWeight,
@@ -225,7 +226,7 @@ class StrategyWeightsTest(PymatgenTest):
             {"type": "linearly_equidistant", "weight_cn1": 2.0, "weight_cn13": 26.0}
         )
         for cn in range(1, 14):
-            assert round(abs(bias_weight4.cn_weights[cn] - 2.0 * cn), 7) == 0
+            assert bias_weight4.cn_weights[cn] == approx(2.0 * cn)
 
         bias_weight5 = CNBiasNbSetWeight.from_description(
             {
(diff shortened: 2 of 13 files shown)


arosen93:
FYI @janosh, there was no need to use m3gnet's custom `TrajectoryObserver`. ASE's native `Trajectory` object stores all the desired properties (e.g. positions, energies, forces, cells, stresses). I'm going to use ASE's `Trajectory` object in #3044 so that the user doesn't have to install `m3gnet` just to have a trajectory returned.

janosh:
@arosen93 That's interesting. Are the two fully compatible? I think I checked back then that [`TrajectoryObserver` doesn't inherit](https://github.com/materialsvirtuallab/m3gnet/blob/d4ea965150bbbf2edb5239ee9dffe90635cffc5c/m3gnet/models/_dynamics.py#L182-L186) from ASE's `Trajectory` class. Of course, could be that they share enough API anyway. Didn't test that.

This might also interest @matthewkuner in the context of https://github.com/materialsproject/atomate2/pull/322. I think we discussed on Slack how best to record and expose the frames of an ML relaxation to `atomate2`.

arosen93:
As far as I was able to tell, they share the same properties. However, they
are formatted slightly differently, so that's worth noting.

I wonder if TrajectoryObserver exists so ASE doesn't write out a trajectory
file every step, which may lead to large file I/O times for large
structures. That said, keeping it all in memory could be problematic too.
Probably worth figuring it out in my linked PR.

matthewkuner:
@arosen93 In case it is relevant, the [matgl repository](https://github.com/materialsvirtuallab/matgl) just pushed a pretrained m3gnet. That is meant to eventually replace the only [m3gnet repo](https://github.com/materialsvirtuallab/m3gnet)

arosen93:
Thanks! I'll look into replacing it in my open PR.

mkhorton:
Apologies if missing something obvious, but why not use/adapt pymatgen’s Trajectory?

If one wants to use ASE’s Trajectory, it would be better to just use ASE. Mixing and matching data types doesn’t seem ideal.

mkhorton:
I want to make the distinction between internal use of another library’s data type (within the body of a function) and the function return type too.

arosen93:
I agree that it would probably be better to return a Pymatgen Trajectory since the user presumably wants to predominantly stay in the Pymatgen ecosystem if they are calling `Structure.relax`.

Hopefully someone will open a PR to address it because that one won't be me :)

arosen93:
Actually, what should perhaps really be done here is add an ASE `Trajectory` to Pymatgen `Trajectory` (and vice versa) option to the `ASEAtomsAdaptor`. Then it would be extra trivial to simply return the transformed trajectory object in `Structure.relax`.

janosh:
> Hopefully someone will open a PR to address it because that one won't be me :)

I'd be happy to do that at some point. No firm commitment though.



> Actually, what should perhaps really be done here is add an ASE `Trajectory`/m3gnet `TrajectoryObserver` to Pymatgen `Trajectory` (and vice versa) option to the `pymatgen.io.ASEAtomsAdaptor`. Then it would be extra trivial to simply return the transformed trajectory object in `Structure.relax()`.

Good idea! Considering @shyuep is a significant stakeholder in this, I'd be interested in his take. Either way, should probably become a new issue instead of closed PR discussion. 😄

shyuep:
I would say I support the suggestion but it will probably take some while before someone spends the effort doing it.

PR #2899, opened 2023-03-19
fix AbinitHeader functionality broken during linting

gmatteo:


janosh:
Good catch and thanks for the new test! 👍

PR #2901, opened 2023-03-20
Cp2k patch

nwinner:
## Summary

Patch for cp2k module.

- Fix 1: Postprocessor typecasts Element "F" to "False". Add a workaround. 
- Fix 2: Site properties are set to "None" by default, which can cause problems if user doesn't set magnetization or ghost properly. Add a check for this. Long term it might be nice to have a pydantic (or equivalent) typecasting system.
- Fix 3: Add "." for os.path.join as default. If environment isn't set right, this can cause an argument error, making debug difficult.


janosh:
Thanks! 👍

svandenhaute:
Why protect only `ELEMENT` and not `KIND`? When defining basis sets and pseudopotentials per element, wouldn't `SUBSYS` include something like this?

```
      &KIND F
         BASIS_SET some_basis_set
         POTENTIAL some_ecp
      &END KIND
```

nwinner:
@svandenhaute It never came up because the input sets generate kind names with numerical suffixes, so the set will produce 

```
      &KIND F_1
         BASIS_SET some_basis_set
         POTENTIAL some_ecp
      &END KIND
```

However, if the input was generated in some other way, then you are right that there will be an issue. I don't have write access to the master branch, so it would be best if you could open a PR for a fix.

PR #2910, opened 2023-03-25
bug fix for potcar parsing

jmmshn:
## bug fix for potcar parsing

2023.3.10 rewrote the `PotcarSingle.from_file` function slightly.
This caused a parsing error in the tests for `pymatgen-analysis-defects`.

I think this small change gives the intended behavior.

Otherwise, the POTCAR in this folder cannot be parsed.
https://github.com/materialsproject/pymatgen-analysis-defects/tree/main/tests/test_files/v_Ga/ccd_0_-1/optics

Without the fix:
```python
pc = Potcar.from_file("./POTCAR")
{d.header for d in pc}
```
Gives: 
```
{'PAW_PBE Ga_d 06Jul2010', ''}
```

with the fix, you get:
```
{'PAW_PBE Ga_d 06Jul2010', 'PAW_PBE N 08Apr2002'}
```

### DIFF
--- a/pymatgen/io/vasp/inputs.py
+++ b/pymatgen/io/vasp/inputs.py
@@ -302,7 +302,7 @@ class Poscar(MSONable):
 
         vasp5_symbols = False
         try:
-            natoms = [int(i) for i in lines[5].split()]
+            n_atoms = [int(i) for i in lines[5].split()]
             ipos = 6
         except ValueError:
             vasp5_symbols = True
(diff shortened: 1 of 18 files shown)


jmmshn:
@janosh @shyuep, I think I have tests are failing due to some hash mismatches.
I'm not familiar with the hashing setup but since I'm only removing blank lines I think it should have been OK.

janosh:
> I'm not familiar with the hashing setup but since I'm only removing blank lines I think it should have been OK.

Agreed.

jmmshn:
So this is still failing, I really hope this is somewhat connected to the massive amount of warnings I always get during POTCAR reading so w/e fixes this will fix that as well.  Maybe we have to redo the hashes and be more consistent about trimming? Again I'm not too familiar with that part of the code.

janosh:
> Unless someone gives me a good reason why we are spending so much effort on this, I am disabling the tests and in future, I will disable the hash checks for good as well.

I don't see a reason. Just to be clear, you mean keep all the Potcar tests and just disable those that relate to the hashes? I thank that would be good.

jmmshn:
@shyuep I think people only rely on the headers.  I recall there were "secret" custom PAWs with the same headers as standard one that people used (which later got released under the different header).  But I feel like if you are using those then it's on the user to sort that out.  I assume the hashing was added to get to make sure the build process is a bit more careful (and might be used by the builders) so you might need to ping @munrojm before any major changes.

shyuep:
I agree. So why don't you comment out the offending tests in the PR to let it pass. We will then merge. I think the hash checks can throw warnings like they do currently, but I am really tired of these tests failing because of some minor missing new line or space somewhere.

munrojm:
As long as the hashing functionality remains in the code, we should be okay. We currently use hash checking as part of the calculation validation pipeline.

pymatgen/io/vasp/inputs.py
```diff
@@ -2275,19 +2275,19 @@ def from_file(filename: str):
 
         functionals = []
         for p in fdata.split("End of Dataset"):
-            if p.strip():
-                single = PotcarSingle(p + "End of Dataset\n")
+            if p_strip := p.strip():
```
jmmshn, line 2278: Ooo walrus is in 3.8?  nice!
image

shyuep:
@munrojm I will leave the hash generation in. But if we are disabling the tests, I make no guarantees that future hashes generated will be valid. In other words, the hash is just being generated and we do not test for correctness. As I said, this is far too much effort to maintain for a functionality that very few people care about.

munrojm:
@shyuep, that is fair.

jmmshn:
Can this get merged, please? I think the failing tests are from an unrelated issue.

jmmshn:
@janosh I modified to test and now it just attempts parsing withouth any checks.  I actually wasn't sure if it was proper etiquette for someon other than the maintainers to delete tests which is why I didn't do it originally.

janosh:
@jmmshn Thanks! The remaining errors here come from breaking changes in `pandas` v2 should be fixed by #2935. Merging now.

PR #2913, opened 2023-03-27
BUG FIX: BSDOSPlotter.get_plot

m3lab-zzl:
Line 2288 

```py
if dos:
```

change to

```py
if isinstance(dos, CompleteDos): # Dos obj has no structure
```

The DOS object has no structure property, which causes error if parsing DOS data without projection. 
This modification allows user to plot BS and TDOS at the same time, while keeping the BS+PDOS plotting not affected.

Cheers.

### DIFF
--- a/pymatgen/electronic_structure/tests/test_plotter.py
+++ b/pymatgen/electronic_structure/tests/test_plotter.py
@@ -206,39 +206,39 @@ class BSDOSPlotterTest(unittest.TestCase):
     # Minimal baseline testing for get_plot. not a true test. Just checks that
     # it can actually execute.
     def test_methods(self):
-        v = Vasprun(os.path.join(PymatgenTest.TEST_FILES_DIR, "vasprun_Si_bands.xml"))
-        p = BSDOSPlotter()
-        plt = p.get_plot(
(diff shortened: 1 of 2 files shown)


pymatgen/electronic_structure/tests/test_plotter.py
```diff
@@ -206,39 +206,39 @@ def tearDown(self):
...
                     # c = np.random.randint(0,7)
-                    d[i][j][k][a] = np.random.rand()
-                    d[i][j][k][b] = np.random.rand()
+                    projections[i][j][k][a] = np.random.rand()
+                    projections[i][j][k][b] = np.random.rand()
                     # d[i][j][k][c] = np.random.rand()
-        bandstr = BandStructureSymmLine.from_dict(bandstr_dict)
-        plt = p.get_plot(bandstr)
+        band_struct = BandStructureSymmLine.from_dict(band_struct_dict)
+        plt = plotter.get_plot(band_struct)
```
janosh, line 241: Could you add a line here that calls `plotter.get_plot()` with a `Dos` object? Just to cover the new code path?

PR #2925, opened 2023-03-30
auto version

jmmshn:
## Allow proper versioning for forked repos

I'm not 100% sure how the current version numbering is updated for `pymatgen` but it's not able to update the number properly for forks.  The forks seem to keep a much older version number (from before the forking)
This creates problems if you have the forked repo as a dependency in a requirement.txt file.

This should fix that issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,10 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.versioningit.vcs]
+method = "git"
+default-tag = "0.0.1"
+
 [tool.black]
 line-length = 120
 


shyuep:
I think forks just need to merge from the main master branch for versioning to be correct? The versions are updated when the release happens.

jmmshn:
So merging from the upstream master did no work.
The following setting in the `pyproject.toml` breaks because the setuptools thinks the `pymatgen` version is too old despite merging from upstream. 
 
```toml
dependencies = [
  'pymatgen@git+https://github.com/jmmshn/pymatgen#egg=master',
  'jobflow@git+https://github.com/jmmshn/jobflow@main#egg=main',
  'atomate2@git+https://github.com/jmmshn/atomate2@main#egg=main'
]
```

Without `pymatgen`, the other two codes required the tag to be pushed up to the forked repo to resolve dependencies properly.

shyuep:
I am not sure how the pyproject fixes this problem? You need to pull from origin/master for the version to be updated.

jmmshn:
So the problem was that if I forked `master` in 2022, and keep pulling from `materialsproject/master`, I still get a version name `2022...post370+gff40e46 ` even when I have the most up-to-date code.  This is usually not a problem if you just clone all the repos and install the newest code in edit mode.

However, if you have that branch as a dependency, it will cause pip resolve issues since the version number/name is wrong despite the code being correct.
So if I have the following three dependencies in a repo and would like to install them on some system, I could not do that
```
dependencies = [
  'pymatgen@git+https://github.com/jmmshn/pymatgen#egg=master',
  'jobflow@git+https://github.com/jmmshn/jobflow@main#egg=main',
  'atomate2@git+https://github.com/jmmshn/atomate2@main#egg=main'
]
```
 
This morning all three branches were up-to-date in code but not in tags.  For `jobflow` and `atomate2` simply pushing the new tags to `origin` fixed the problem because they had this stanza in their respective `pyproject.toml` but `pymatgen` did not, so I still cannot install these three branches.  After putting this stanza in and pushing the recently released tag to `origin`, it looks like the versioning is permanently changed for all commits past the release tag and now I'm able to install all three packages.

jmmshn:
This is a pretty niche issue since most people aren't doing production runs while changing multiple packages. 
However, I think the setup of automatically pulling from the git tags for version information is generally useful and takes care of edge cases like mine.

PR #2927, opened 2023-04-01
add kwarg to MoleculeGraph method and fix PackmolSet bug

orionarcher:
## Summary

The get_disconnected_fragments method returns a new graph with reordered indices but provides no way to get the mapping from new to original indices. This PR reconstructs that map and adds an optional keyword argument to return it. This will not affect current functionality.

Additionally, I found an error in the `pymatgen.io.packmol` code where a non-abstract class had an abstract method, preventing it from being instantiated. This slipped under the radar because the tests are skipped due to a difficult dependency (`packmol`). The tests are now passing locally for me.

A downstream code I am working on requires this addition.

- a new `return_index_map` kwarg for MoleculeGraph.get_disconnected_fragments
- remove `abc.abstractmethod` decorator from `PackmolBoxGen.get_input_set()`

## Checklist


orionarcher:
Thank you @janosh!

PR #2937, opened 2023-04-10
fix GaussianOutput bug with multiple route lines

xjf729:
When route lines in gaussian output file like:

```plaintext
 # B3LYP/6-311++G(2d,p) EmpiricalDispersion=GD3BJ SCRF=(SMD,READ) freq 
 opt=(loose,maxcyc=400)
```

If we use `GaussianOutput` to parse route lines, `route_parameters` would be looked like as follows, in which, `freq` and `opt` are concatenate.

```python
{'EmpiricalDispersion': 'GD3BJ',
  'SCRF': '(SMD,READ)',
  'freqopt': {'loose': None, 'maxcyc': '400'}}
```

Now, I fix the method to parse route lines, to give the right `route_parameters`:

```python
{'EmpiricalDispersion': 'GD3BJ',
  'SCRF': '(SMD,READ)',
  'freq': None,
  'opt': {'loose': None, 'maxcyc': '400'}}
```


janosh:
Could you zip the log file and load it as `test_files/molecules/EC.log.gz` using e.g. `monty`?

janosh (approved):
Thanks for the fix @xjf729! 👍

PR #2939, opened 2023-04-12
fix on reading multiple route in Gaussian input file

Ameyanagi:
## Summary

This fixes the issue that the Gaussian input file fails to read multiple lines. 
- When the route input has multiple lines as below, GaussianInput.from_string() fails around the parsing of the route. This was due to inappropriate line counting when the route is separated in multiple lines.
- few lines of code were added to parse the route and route_index.
- test files for routes with multiple lines.

Example of Gaussian input (header only) with multiple lines of the route that failed.
```
%nprocshared=28
%mem=60GB
%chk=test.chk
# opt freq wb97xd/gen nosymm scf=(qc,maxcycle=1024) geom=connectivity
pseudo=read
```

## Checklist

- [ ] Type annotations are *highly* encouraged. Run [`mypy path/to/file.py`](https://github.com/python/mypy) to type check your code. Should I also rewrite the other part of the codes?

```sh
pip install -U pre-commit
pre-commit install
```

in the repo's root directory. Afterwards linters will run before every commit and abort if any issues pop up.

### DIFF
--- a/pymatgen/io/gaussian.py
+++ b/pymatgen/io/gaussian.py
@@ -314,7 +314,7 @@ class GaussianInput:
             elif (line == "" or line.isspace()) and route_index:
                 break
             if route_index:
-                route += " " + line
+                route += f" {line}"
                 route_index = idx
         functional, basis_set, route_paras, dieze_tag = read_route_line(route)
         ind = 2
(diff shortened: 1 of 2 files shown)


janosh:
Thanks! 👍

Ameyanagi:
Thanks for merging!

PR #2951, opened 2023-04-24
Fix ValueError when passing `selective_dynamics` to `Poscar`

chiang-yuan:
Before this fix, users could not directly add a `selective_dynamics` array to create a `Poscar` object:

```python
poscar = Poscar(
    structure,
    selective_dynamics=boolean_array
)
```
and the code would produce an error:

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

A similar [issue](https://matsci.org/t/got-problems-in-generating-vasp-input-files-for-structure-with-changed-selective-dynamics/38978) was reported previously about manually using `add_site_property`.

The minimal example reproducing the error and the bug fix is demonstrated in the attached notebook: [poscar_debug.md](https://github.com/materialsproject/pymatgen/files/11305634/poscar_debug.md)

## Summary

- Fixed the error where `selective_dynamics` boolean array is given for creating a `Poscar` object
- Applied similar bug fixes for `velocities` and `predictor_corrector`


janosh:
@chiang-yuan Thanks! The test should go into [`pymatgen/io/vasp/tests/test_inputs.py`](https://github.com/materialsproject/pymatgen/blob/51280b98a60b0dc12e1cc97d755813302a46e3bb/pymatgen/io/vasp/tests/test_inputs.py). There are some examples in there to learn from. Also, please check if you can reuse some of the existing POSCAR test files rather than adding new ones.

chiang-yuan:
Thanks @janosh ! I followed the original style to test by comparing two primitive python lists and converted numpy array back to list to assign variables. Since the system can be large and comparing two nested python lists is somewhat tricky, (For example, `all([[True, False],[True, True]])` will give `True` rather than `False`), I suggest in the future assigning and testing all these variable as numpy array. This also aligns with `frac_coords`, `cart_coords` in the `core.structure` module.

pymatgen/io/vasp/inputs.py
```diff
@@ -122,12 +122,22 @@ def __init__(
...
-                site_properties["selective_dynamics"] = selective_dynamics
-            if velocities:
-                site_properties["velocities"] = velocities
-            if predictor_corrector:
-                site_properties["predictor_corrector"] = predictor_corrector
+
+            if selective_dynamics is not None:
+                selective_dynamics = np.array(selective_dynamics)
+                if not selective_dynamics.all():
+                    site_properties["selective_dynamics"] = selective_dynamics
```
janosh, line 129: @chiang-yuan I simplified the code some. Was there a reason to convert `np.array` to list before assigning to `site_props`?

chiang-yuan: @janosh I converted np.array back to list because the other unit test in `io/vasp/tests/test_inputs.py` compare `selective_dyanmics` with list, which will yield the error in pytest like the one after your revise. 

```
>           assert poscar.selective_dynamics == [[True, True, True], [False, False, False]]
E           ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

/home/runner/work/pymatgen/pymatgen/pymatgen/io/vasp/tests/test_inputs.py:87: ValueError
```

I bypass them by converting it back to list (which is what the original code intended to do) and adding another unit test comparing by list as well. Do you think it is better to revise the previous unit test and compare all of them in the format of `nd.array` like I suggested earlier?

janosh: I think we should make the test datatype agnostic so that it handles both arrays and lists. It's better for tests to care only about semantics and not equivalent ways of encoding the same meaning.

chiang-yuan: Agree. Thanks @janosh !

PR #2969, opened 2023-05-02
Change Yb pseudo-potential on all VASP input sets: Yb_(2->3)

### LINKED ISSUE #2968: Entry sets use wrong Yb pseudo-potential
[The docs](https://docs.materialsproject.org/methodology/mof-methodology/calculation-parameters/psuedopotentials) are incorrect on what pseudo-potential the `EntrySet`s use for `Yb`. Docs say `Yb_3`, actual is `Yb_2`. Just to list a few:

https://github.com/materialsproject/pymatgen/blob/a7ee5adf02e0ac8c0b9e150d10f631bd61f6b5de/pymatgen/io/vasp/MPRelaxSet.yaml#L171

https://github.com/materialsproject/pymatgen/blob/a7ee5adf02e0ac8c0b9e150d10f631bd61f6b5de/pymatgen/io/vasp/MPSCANRelaxSet.yaml#L123

https://github.com/materialsproject/pymatgen/blob/a7ee5adf02e0ac8c0b9e150d10f631bd61f6b5de/pymatgen/io/vasp/MVLRelax52Set.yaml#L177

https://github.com/materialsproject/pymatgen/blob/a7ee5adf02e0ac8c0b9e150d10f631bd61f6b5de/pymatgen/io/vasp/MPAbsorptionSet.yaml#L113

The A-lab revealed that as a result of using `Yb_2` the energy on Yb compounds is off by a lot, resulting in supposedly stable things being unsynthesizable. While an unfortunate mistake, it's also great to see how experiment can help surface simulation errors. (@rekumar) (Although if tales are correct, @arosen93 originally brought this up with @mkhorton in Oct 2021?)

Planned solution is to update all input sets to actually use Yb_3. Why not `Yb` so as not to force an oxidation state? Apparently `Yb` has stability issues. Additional context in
- https://github.com/materialsproject/atomate2/pull/27#issuecomment-986306566 
- https://www.vasp.at/forum/viewtopic.php?t=18367

@shyuep Any concerns on your end? We plan to highlight this change in the pymatgen release notes and in the next MP database release. Any additional venues?

Unfortunately, this comes too late for the large data set Google generated which used the current `MPRelaxSet` and `MPSCANRelaxSet` but better fix this now before another large effort gets this wrong.

janosh:
Closes #2968.

### DIFF
--- a/pymatgen/io/vasp/MPAbsorptionSet.yaml
+++ b/pymatgen/io/vasp/MPAbsorptionSet.yaml
@@ -110,6 +110,6 @@ POTCAR:
   W: W_pv
   Xe: Xe
   Y: Y_sv
-  Yb: Yb_2
+  Yb: Yb_3
   Zn: Zn
   Zr: Zr_sv
--- a/pymatgen/io/vasp/MPHSERelaxSet.yaml
+++ b/pymatgen/io/vasp/MPHSERelaxSet.yaml
@@ -109,6 +109,6 @@ POTCAR:
   W: W_pv
   Xe: Xe
   Y: Y_sv
-  Yb: Yb_2
+  Yb: Yb_3
   Zn: Zn
   Zr: Zr_sv
--- a/pymatgen/io/vasp/MPRelaxSet.yaml
+++ b/pymatgen/io/vasp/MPRelaxSet.yaml
@@ -168,6 +168,6 @@ POTCAR:
   W: W_pv
   Xe: Xe
   Y: Y_sv
-  Yb: Yb_2
+  Yb: Yb_3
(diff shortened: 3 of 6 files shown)


janosh:
Yes, this change will be highlighted in the next release notes. I'll also change the PR title to make this clearer.

shyuep:
I would suggest you document the change to Yb_3 in the input sets yaml themselves. E.g., `# May 2 2023: Changed to Yb_3 from Yb_2 as the latter gives incorrect thermodynamics for most systems with Yb3+`

janosh:
@shyuep Excellent suggestion!

PR #2972, opened 2023-05-04
Revert to `Yb_2` on pre-`PBE_54` input sets

### LINKED ISSUE #2970: Bug in `DictSet.kpoints` due to operator precedence
@shyuep I think there's a subtle operator precedence bug here:

https://github.com/materialsproject/pymatgen/blob/ed66c977d8c2c1168a9945d64f5d043d9263fc43/pymatgen/io/vasp/sets.py#L633-L638

`and` has higher precedence than `or` but unless I'm mistaken we actually want to evaluate the `or` first. The thing that throws me off is why this wasn't caught before.

```py
if (
    self.user_incar_settings.get("KSPACING") or self._config_dict["INCAR"].get("KSPACING")
) and self.user_kpoints_settings == {}:
```

### LINKED ISSUE #2971: `pytest.raises() as exc_info` has no `.exception`
We're misusing `pytest.raises()` in at least 4 places in 2 files. 2 Examples below where it should be

```diff
- str(context.exception)
+ str(context.value)
```

https://github.com/materialsproject/pymatgen/blob/ed66c977d8c2c1168a9945d64f5d043d9263fc43/pymatgen/io/vasp/tests/test_sets.py#L505-L516

janosh:
Follow up to #2969.

Instead of changing from `Yb_2` to `Yb_3` everywhere, we issue a warning in pre-`PBE_54` input sets that `Yb_2` will give very bad results for most systems since Yb is usually Yb3+.

Reason: The better fix `Yb_3` only became available in the `PBE_54` set of PSPs so using it in pre-`PBE_54` input sets effectively means you can't run Yb compounds.

Closes #2970 , closes #2971.


pymatgen/io/vasp/sets.py
```diff
@@ -335,6 +335,15 @@ def __init__(
...
         """
+        struct_has_Yb = any(specie.symbol == "Yb" for site in structure for specie in site.species)
+        uses_Yb_2_psp = self.CONFIG["POTCAR"]["Yb"].lower() == "yb_2"
+        if struct_has_Yb and uses_Yb_2_psp:
+            warnings.warn(
+                "The structure contains Ytterbium (Yb) and this InputSet uses the Yb_2 PSP.\n"
+                "Yb_2 is known to often give bad results since Yb has oxidation state 3+ in most compounds.\n"
+                "See https://github.com/materialsproject/pymatgen/issues/2968 for details.",
+                BadInputSetWarning,
+            )
```
janosh, line 346: @shyuep As discussed, I'm reverting the change from Yb_2 to Yb_3 on pre-PBE_54 input sets and issuing a strong warning instead.

janosh:
@jmmshn asked for a new release which I think we're ready for once this is merged.

PR #2973, opened 2023-05-05
fix recursion error by adding copy and deepcopy dunder methods

orionarcher:
## Summary

Both copy and deepcopy are currently broken for the `InputSet` object. This PR implements dunder methods to fix both. There very well may be a more elegant solution but I couldn't find it!

- Feature 1: `__copy__` and `__deepcopy__` dunder methods

## Checklist


janosh (approved):
Thanks for the fix @orionarcher! 👍

PR #2983, opened 2023-05-09
Fix for memory leak in pbc_shortest_vectors

stichri:
## Summary

The `pbc_shortest_vectors` function leaks all memory allocated to store the fractional coordinates of surrounding periodic sites. As a result, memory consumption keeps creeping up when (heavily) calling the function, see [this memory profile](https://github.com/materialsproject/pymatgen/files/11433457/v2023.5.8.pdf) of a script of mine which made me aware of the problem. The suggested modifications completely eliminate the issue, [curtailing the memory consumption](https://github.com/materialsproject/pymatgen/files/11433643/fixed.pdf) of the same script.

With the suggested minimal modifications, additional memory is only allocated (and freed accordingly) when needed - that is, when periodic boundary conditions are applied selectively.

### Before

v2023 5 8

### After

fixed

### DIFF
--- a/pymatgen/util/coord_cython.pyx
+++ b/pymatgen/util/coord_cython.pyx
@@ -93,7 +93,7 @@ def pbc_shortest_vectors(lattice, fcoords1, fcoords2, mask=None, return_d2=False
     cdef int n_pbc_im = 3 ** n_pbc
     
     cdef np.float_t[:, ::1] frac_im
-    cdef int i, j, k, l, I, J, bestK
+    cdef int i, j, k, l, I, J
 
     if n_pbc == 3:
         fcoords1 = lattice.get_lll_frac_coords(fcoords1)
--- a/pymatgen/util/coord_cython.pyx
+++ b/pymatgen/util/coord_cython.pyx
@@ -91,9 +91,9 @@ def pbc_shortest_vectors(lattice, fcoords1, fcoords2, mask=None, return_d2=False
     pbc = lattice.pbc
     cdef int n_pbc = sum(pbc)
     cdef int n_pbc_im = 3 ** n_pbc
-    cdef np.float_t[:, ::1] frac_im = <np.float_t[:n_pbc_im, :3]> malloc(3 * n_pbc_im * sizeof(np.float_t))
-
-    cdef int i, j, k, l, I, J
+    
+    cdef np.float_t[:, ::1] frac_im
+    cdef int i, j, k, l, I, J, bestK
 
     if n_pbc == 3:
         fcoords1 = lattice.get_lll_frac_coords(fcoords1)
@@ -101,6 +101,7 @@ def pbc_shortest_vectors(lattice, fcoords1, fcoords2, mask=None, return_d2=False
         matrix = lattice.lll_matrix
         frac_im = images_view
     else:
+        frac_im = <np.float_t[:n_pbc_im, :3]> malloc(3 * n_pbc_im * sizeof(np.float_t))
         matrix = lattice.matrix.copy()
         k = 0
         for i in range(27):
@@ -178,6 +179,8 @@ def pbc_shortest_vectors(lattice, fcoords1, fcoords2, mask=None, return_d2=False
                 for l in range(3):
                     vs[i, j, l] = 1e20
 
+    if not (n_pbc == 3):
+        free(&frac_im[0,0])
     free(&cart_f1[0,0])
     free(&cart_f2[0,0])
     free(&cart_im[0,0])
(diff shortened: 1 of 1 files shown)


shyuep:
Great. Thanks! I have a small request. Can you provide a simple benchmark of the difference in speed (if any) between old and new version? This is a rather core part of the code. I would want to make sure that the efficiency is not compromised.

pymatgen/util/coord_cython.pyx
```diff
@@ -178,6 +179,8 @@ def pbc_shortest_vectors(lattice, fcoords1, fcoords2, mask=None, return_d2=False
                 for l in range(3):
                     vs[i, j, l] = 1e20
 
+    if not (n_pbc == 3):
+        free(&frac_im[0,0])
```
janosh, line 183: These changes don't look like they should affect performance. Not sure how fast the `free` operation is but considering we're currently leaking memory, _some_ slowdown is definitely preferable.

janosh:
I've never written a test for `cython` code. @stichri Is there a way we can ensure no regressions on mem leaks?

stichri:
As request by @shyuep, I wrote [this very simple benchmark](https://github.com/materialsproject/pymatgen/files/11439571/benchmark.py.txt).
The idea is to assess the overall runtime impact, which I think will play out along two parameters: The relative amount of work done per call - either lots or little - and the applied periodic boundary conditions (PBC) - either in all directions or just in one direction. Generally, I'd expect the _highest impact_ from _avoided calls of `malloc`_ when doing _little work per call_ with _PBC in all directions_ and from _additional calls of `free`_ when doing _little work per call_ with _PBC in just one direction_.
Running the benchmark on my mobile AMD Ryzen 7 using **pymatgen v2023.5.8** in python 3.11 on Fedora38 in WSL2 running on Windows 11 results in:
```
PBC in ALL directions:
fastest out of 10 runs with 100000 calls each, LITTLE WORK per call: 1.76117e+00s
fastest out of 10 runs with 100 calls each, LOTS OF WORK per call: 3.22905e+00s
PBC in ONE directions:
fastest out of 10 runs with 100000 calls each, LITTLE WORK per call: 7.77626e-01s
fastest out of 10 runs with 100 calls each, LOTS OF WORK per call: 6.76336e-01s
```
The same benchmark on the same machine **with the modifications applied** results in:
```
PBC in ALL directions:
fastest out of 10 runs with 100000 calls each, LITTLE WORK per call: 1.68359e+00s
fastest out of 10 runs with 100 calls each, LOTS OF WORK per call: 3.28771e+00s
PBC in ONE directions:
fastest out of 10 runs with 100000 calls each, LITTLE WORK per call: 7.95001e-01s
fastest out of 10 runs with 100 calls each, LOTS OF WORK per call: 7.19025e-01s
```
Judging from this very rough benchmarking, I would assess the _impact on runtimes_ to be _typically less than 10%_.
Moreover, should that turn out to be an issue, I think this additional memory managment could be eliminated altogether when considering additional modifications; The PBC application is ultimately [controlled by a bool triplet in the passed lattice object](https://github.com/materialsproject/pymatgen/blob/c2f2d724c89fa8b73044824b2e4ae3c26938280f/pymatgen/util/coord_cython.pyx#L91), thus there should be just 7 different cases to which the `images_view` variable would be needed to be assigned to, [just as it currently is](https://github.com/materialsproject/pymatgen/blob/c2f2d724c89fa8b73044824b2e4ae3c26938280f/pymatgen/util/coord_cython.pyx#LL31C45-L31C45) for the special case of full PBCs. All the additional 6 cases could thus be hardcoded into constant module-level objects as well.

Regarding additional tests for memory leaks within `cython` code as @janosh mentioned, unfortunately I wouldn't be aware of a particularly convenient way of doing it 'from the python side of things'.
That is, if I were to systematically check for these kind of errors, I would probably use (dynamical memory analysis tools like) [valgrind](https://valgrind.org/). After all, the CPython implementation is conveniently shipped with a [suppression file for false positives](https://github.com/python/cpython/blob/main/Misc/valgrind-python.supp) when running python. I guess it's always possible to script some tests that automatically do just that at the cost of pulling in potentially new dependencies like valgrind.

janosh:
Thanks for the benchmark!

shyuep:
Thanks. Much appreciated!

pymatgen/util/coord_cython.pyx
shyuep: Just a minor pt - find_pbc_vector is very very performance critical (this is why it is in cython). I am willing to lose < 10% performance for mem, but if it is a 10x diff (this is not difficult to imagine), we need to consider the tradeoff. Basically all operations involving bond distances will slow down. Since the performance loss seems minimal, this is ok.

Issue #2985, opened 2023-05-11
CompleteCohp.cohp and CompleteCohp.icohp produce the same output


QuantumChemist:
When trying to access the cohp and the icohp feature of CompleteCohp for a COHPCAR.lobster file, one gets the same output. In the following minimal code, the sum is 0 (and also the individual entries for each energy point is 0).

## Minimal code to reproduce the issue

cohp_test = CompleteCohp.from_file(filename="path_to_files/COHPCAR.lobster", fmt = 'LOBSTER', structure_file="path_to_files/POSCAR")

cohp = cohp_test.cohp[Spin.up]
icohp = cohp_test.icohp[Spin.up]

print(sum(icohp-cohp))

QuantumChemist:
Totally forgot to tag @JaGeo  :D

JaGeo:
Thanks for writing it down! I will work on a fix as soon as I get to it. I never use the average COHP/ICOHP. 😅

JaGeo:
Can be closed then, @janosh or @QuantumChemist

janosh:
@JaGeo Thanks. Btw, if you include "closes #issue-number" or "fixes #issue-number" in the PR description or any commit message, the issue will be auto-closed on merge.

JaGeo:
@janosh I think I actually did this but it did not auto-close. Do you see what I did wrong?

janosh:
I edited the PR description. 😄

JaGeo:
> I edited the PR description. 😄

🙃 I see.

YuanbinLiu:
> 

Brilliant! Thanks!

PR #2986, opened 2023-05-12
Fix average error

JaGeo:
## Summary

This will fix the issue mentioned in https://github.com/materialsproject/pymatgen/issues/2985. There was a simple typo. As I usually never use the average cohp, it never even occurred to me to test it. I adapted all tests and included a test to now also test the average cohp.

I also gzipped some COHPCARs to start reducing test file size for the lobster/cohp modules.


naik-aakash:
Thanks!! @JaGeo 😃

janosh (approved):
This is great, thank you!

PR #2989, opened 2023-05-14
Suspected Typo Fix in `pymatgen.io.vasp.optics`

kavanase:
Hi!
Firstly, thanks to the developers for the recent addition of the `pymatgen.io.vasp.optics` module, it's been very very useful! 🙌

This is just a small fix for a suspected typo in handling `ismear` in the `delta_func()` and `step_func()` functions. `ISMEAR` is zero for Gaussian smearing, and indeed in the docstrings for `get_epsilon()` it states: `ismear: Smearing method (only has 0:gaussian, >0:Methfessel-Paxton)`. However, in the current version of the code:
```python
def delta_func(x, ismear):
    """Replication of VASP's delta function"""
    if ismear < -1:
        raise ValueError("Delta function not implemented for ismear < -1")
    if ismear == -1:
        return step_func(x, -1) * (1 - step_func(x, -1))
    if ismear < 0:
        return np.exp(-(x * x)) / np.sqrt(np.pi)
    return delta_methfessel_paxton(x, ismear)
```

`ismear = 0` ends up returning the Methfessel-Paxton smearing (`delta_methfessel_paxton`), however the function under `if ismear < 0` is the Gaussian smearing. `ismear` should be an integer anyway, so having `ismear < -1` raise an error and `ismear == -1` return something else, it means that the `ismear < 0` line is always skipped. This is also the case for the `step_func()` function.

I believe this is likely a small typo, and should instead by `if ismear == 0:`. I have changed it to this in this PR, and tests are passing.

### DIFF
--- a/pymatgen/io/vasp/optics.py
+++ b/pymatgen/io/vasp/optics.py
@@ -48,9 +48,9 @@ class DielectricFunctionCalculator(MSONable):
         - Perform symmetry operations (this is not implemented here)
         - Calculate the real part
 
-    Currently, this Calculator only works for ``ISYM=0`` calculations since we cannot gauranttee that our
-    externally defined symmetry operations are the same as VASP's.  This can be fixed by printing the
-    symmetry operators into the vasprun.xml file.  If this happens in future versions of VASP,
+    Currently, this Calculator only works for ``ISYM=0`` calculations since we cannot guarantee that our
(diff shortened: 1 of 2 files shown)


janosh:
Thanks for the fix! 👍 This sounds sensible. We should definitely have tests for these functions.

PR #2991, opened 2023-05-16
Add type hints for `pymatgen.io.ase` module

arosen93:
Pretty self-explanatory. I added type hints to `pymatgen.io.ase`.

### DIFF
--- a/pymatgen/io/ase.py
+++ b/pymatgen/io/ase.py
@@ -20,6 +20,7 @@ except ImportError:
     ase_loaded = False
 
 if ase_loaded:
+    from ase.atoms import Atoms
     from ase.calculators.singlepoint import SinglePointDFTCalculator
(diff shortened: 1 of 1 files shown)


arosen93:
Welp, sorry @janosh. I thought this would be a trivial add. But I'm not sure how to make the `mypy` overlord happy with this...

pymatgen/io/ase.py
```diff
@@ -13,13 +13,12 @@
 from pymatgen.core.structure import Molecule, Structure
 
 try:
-    from ase import Atoms
```
janosh, line 16: I think this line has to stay here?

janosh:
Oh and I think here you meant

```diff
- cls: Structure = None
cls: type[Structure] = None
```

arosen93:
Thanks!!

PR #3004, opened 2023-05-21
Lookup `MPRester` API key in settings if `None` provided as arg

ml-evs:
## Summary

Both the new and legacy versions of `MPRester` accepted `api_key=None` to be passed as an argument, triggering a fallback to the settings version of the key. The new behavior of the dynamic `MPRester` is to attempt to use the `None` value directly, i.e., without checking for an externally configured value.

This PR simply adjusts the behavior so that `api_key=None` will still trigger a settings lookup.

### DIFF
--- a/pymatgen/ext/matproj.py
+++ b/pymatgen/ext/matproj.py
@@ -1761,9 +1761,11 @@ class MPRester:
            *args: Pass through to either legacy or new MPRester.
            **kwargs: Pass through to either legacy or new MPRester.
         """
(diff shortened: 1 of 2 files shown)


janosh:
For the new Rester, reading the API key from `.pmgrc.yaml` or env variable required passing no args, i.e.

```diff
- MPRester(None)
+ MPRester()
```

Thanks for making the upper case work as well @ml-evs!

PR #3021, opened 2023-05-30
Don't use `np.nan` as kwarg default, prefer `None`

### LINKED ISSUE #2107: NaN should be removed from any init functions
An example is shown here:

https://github.com/materialsproject/pymatgen/blob/04beca9af8e575005a40117f1779f1d464c0a697/pymatgen/entries/computed_entries.py#L49

I think these should be removed from any MSONable init function since they don't play well with database and model validation.
I'm currently just replacing these with 0's in any emmet model that contains `ComputedEntry`s 

https://github.com/materialsproject/emmet/blob/ebcc50bc47a8b65d661877f9edeb7f4087a7b4e5/emmet-core/emmet/core/utils.py#L117

but if the NaN's are physically meaningful I think we need to replace them with `None`.

@mkhorton @rkingsbury @shyamd

janosh:
Closes #2107.


janosh:
Closing with reasoning in https://github.com/materialsproject/pymatgen/issues/2107#issuecomment-1569239404.

PR #3031, opened 2023-06-01
Update `DictSet` to allow direct initialisation

kavanase:
This is a quick fix to allow direct initialisation of the `DictSet` class, which was possible before but broke in https://github.com/materialsproject/pymatgen/pull/2972 due to the `Yb_2` check querying `self.CONFIG`, which is only defined if `DictSet` was being initialised from a subclass and not directly. 

This fixes this behaviour by checking the `config_dict` which is a required input parameter so always defined (same as `self.CONFIG` if initialising from a subclass), and also updates this with the `user_potcar_settings` before checking, in case the user has explicitly specified to avoid the `Yb_2` `POTCAR`. 

Demo in the attached notebook and PDF!

[pymatgen_DictSet_fix.zip](https://github.com/materialsproject/pymatgen/files/11628196/pymatgen_DictSet_fix.zip)


shyuep:
Thanks!

PR #3035, opened 2023-06-02
Breaking: rename VaspInputSet.(potcar_functional->user_potcar_functional) and start testing `DictSet`

janosh:
Follow up to #3031 adding a test to make sure we don't regress again. I just realized we had 0 tests for `DictSet` before. While writing some, I noticed `DictSet.as_dict()` is also broken. Not only since https://github.com/materialsproject/pymatgen/pull/2972, even before due to not storing the `validate_magmons` init kwarg on instances.

### The breaking part

`VaspInputSet.potcar_functional` was renamed to `VaspInputSet.user_potcar_functional` to fix `DictSet.as_dict()` which otherwise errors

```py
NotImplementedError: Unable to automatically determine as_dict format from class. MSONAble requires all args to be present as either self.argname or self._argname
```

@shyuep Is this justifiable? `DictSet.as_dict()` is not very important and renaming `potcar_functional` -> `user_potcar_functional` might incur a lot of breaking changes in downstream code. (Auto-merge disabled until further discussion)

### DIFF
--- a/pymatgen/core/tests/test_structure.py
+++ b/pymatgen/core/tests/test_structure.py
@@ -1229,6 +1229,9 @@ class StructureTest(PymatgenTest):
         assert sites[-1].specie.symbol == "C"
         self.structure.add_oxidation_state_by_element({"Si": 4, "C": 2})
         assert self.structure.charge == 62
+        species = self.structure.species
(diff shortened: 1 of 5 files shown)


shyuep:
Agreed.

PR #3036, opened 2023-06-02
Don't do a hard `bandgap==0` check in the r2SCAN workflow

### LINKED ISSUE #3023: `if self.bandgap == 0` is probably not the best in `MPScanRelaxSet`
https://github.com/materialsproject/pymatgen/blob/3c91e69838500120cf62032037324f88defcb1a4/pymatgen/io/vasp/sets.py#L949-L952

This checks if `bandgap` is strictly equal to zero, but there are likely going to be edge cases where `bandgap` is some very small value (say, 1e-6 eV) that should also be flagged as metallic. It'd probably be better to do `if self.bandgap < tol` where `tol` is something like `1e-4`.

CC @janosh, @rkingsbury

arosen93:
Closes #3023. CC'ing @janosh and @munrojm.

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -948,7 +948,6 @@ class MPScanRelaxSet(DictSet):
         self.bandgap = bandgap
         self.kwargs = kwargs
 
-        # self.kwargs.get("user_incar_settings", {
         updates = {}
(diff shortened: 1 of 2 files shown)


arosen93:
Thanks for adding those unit tests! Greatly appreciated!!

PR #3041, opened 2023-06-04
Fix LMAXMIX default in VASP INCAR now set based on element blocks in structure

### LINKED ISSUE #3040: `LMAXMIX` should be set based on the presence of s, p, d, f electrons not based on Z
Currently, `LMAXMIX` is set as follows:

https://github.com/materialsproject/pymatgen/blob/67ba063073b468cb33b93f2dc3841fd6ae3941d5/pymatgen/io/vasp/sets.py#L529-L543

However, this is based on Z value and not based on the presence or lack of s, p, d, or f electrons (see [VASP manual](https://www.vasp.at/wiki/index.php/LMAXMIX)). For instance, Sr has Z > 20 but does not have d electrons. La has Z > 56 but does not have f electrons. And so on. We should be checking this based on the "block" of the electrons in the `Structure` object.

See https://github.com/materialsproject/atomate2/issues/364 for a mirror of this comment and the linked PR for the fix.

janosh:
Closes #3040.

@arosen93 Do you have a good test case for this by any chance?

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -526,20 +526,19 @@ class DictSet(VaspInputSet):
                 if key.startswith("LDAU"):
                     del incar[key]
 
-        # Modify LMAXMIX if you have d or f electrons present.
-        # Note that if the user explicitly sets LMAXMIX in settings it will
-        # override this logic.
-        # Previously, this was only set if Hubbard U was enabled as per the
-        # VASP manual but following an investigation it was determined that
-        # this would lead to a significant difference between SCF -> NonSCF
-        # even without Hubbard U enabled. Thanks to Andrew Rosen for
-        # investigating and reporting.
+        # Modify LMAXMIX if you have d or f electrons present. Note that if the user
+        # explicitly sets LMAXMIX in settings it will override this logic.
+        # Previously, this was only set if Hubbard U was enabled as per the VASP manual
+        # but following an investigation it was determined that this would lead to a
+        # significant difference between SCF -> NonSCF even without Hubbard U enabled.
+        # Thanks to Andrew Rosen for investigating and reporting.
         if "LMAXMIX" not in settings:
+            blocks = {site.specie.block for site in structure}
             # contains f-electrons
-            if any(el.Z > 56 for el in structure.composition):
+            if "d" in blocks:
                 incar["LMAXMIX"] = 6
             # contains d-electrons
-            elif any(el.Z > 20 for el in structure.composition):
+            elif "p" in blocks:
                 incar["LMAXMIX"] = 4
 
         # Warn user about LASPH for +U, meta-GGAs, hybrids, and vdW-DF
@@ -555,8 +554,8 @@ class DictSet(VaspInputSet):
             )
 
         if self.constrain_total_magmom:
-            nupdown = sum(mag if abs(mag) > 0.6 else 0 for mag in incar["MAGMOM"])
-            if abs(nupdown - round(nupdown)) > 1e-5:
+            n_up_down = sum(mag if abs(mag) > 0.6 else 0 for mag in incar["MAGMOM"])
+            if abs(n_up_down - round(n_up_down)) > 1e-5:
                 warnings.warn(
                     "constrain_total_magmom was set to True, but the sum of MAGMOM "
                     "values is not an integer. NUPDOWN is meant to set the spin "
@@ -565,7 +564,7 @@ class DictSet(VaspInputSet):
                     "NUPDOWN directly in your INCAR settings.",
                     UserWarning,
                 )
-            incar["NUPDOWN"] = nupdown
+            incar["NUPDOWN"] = n_up_down
 
         if self.use_structure_charge:
             incar["NELECT"] = self.nelect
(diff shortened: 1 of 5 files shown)


pymatgen/io/vasp/sets.py
```diff
@@ -526,20 +526,19 @@ def incar(self) -> Incar:
...
+        # Modify LMAXMIX if you have d or f electrons present. Note that if the user
+        # explicitly sets LMAXMIX in settings it will override this logic.
+        # Previously, this was only set if Hubbard U was enabled as per the VASP manual
+        # but following an investigation it was determined that this would lead to a
+        # significant difference between SCF -> NonSCF even without Hubbard U enabled.
+        # Thanks to Andrew Rosen for investigating and reporting.
         if "LMAXMIX" not in settings:
+            blocks = {site.specie.block for site in structure}
             # contains f-electrons
-            if any(el.Z > 56 for el in structure.composition):
```
arosen93, line 539: Should be `if "f" in blocks`. I made a mistake in the WIP Atomate2 PR (fixed now).

pymatgen/io/vasp/sets.py
```diff
@@ -526,20 +526,19 @@ def incar(self) -> Incar:
...
+            if "d" in blocks:
                 incar["LMAXMIX"] = 6
             # contains d-electrons
-            elif any(el.Z > 20 for el in structure.composition):
```
arosen93, line 542: Should be `if "d" in blocks`. I made a mistake in the WIP Atomate2 PR (fixed now).

arosen93:
I don't have a unit test for this one yet. We can copy it from the Atomate2 PR once I add it there. I have a unit test in my [Quacc code](https://github.com/Quantum-Accelerators/quacc/blob/3e2e6c2281c214ada86fcba44d43750017844a2a/tests/calculators/vasp/test_vasp.py#L63-L75), but it's not quite the same.

pymatgen/io/vasp/sets.py
janosh: Ah, I thought it might be swapped :)

janosh:
@arosen93 How's this test?


```py
    def test_incar_lmaxmix(self):
        # structure containing neither f- nor d-electrons
        structure_f = self.get_structure("Si")
        assert "LMAXMIX" not in MPRelaxSet(structure_f).incar

        # structure containing d-electrons but no f-electrons
        structure_d = self.get_structure("LiFePO4")
        assert MPRelaxSet(structure_d).incar["LMAXMIX"] == 4

        # structure containing f-electrons but no d-electrons
        structure_f = structure_d.copy()
        structure_f.replace_species({"Fe": "La"})
        assert MPRelaxSet(structure_f).incar["LMAXMIX"] == 6

        # explicit LMAXMIX in settings overrides automatic selection
        structure_override = self.get_structure("Si")  # Iron is in the d-block
        set_override = MPRelaxSet(structure_override, user_incar_settings={"LMAXMIX": 3})
        assert set_override.incar["LMAXMIX"] == 3
```

arosen93:
Looks almost right, but La doesn't actually have f electrons. Try Ce instead.

Also, it might be worth adding a test for a structure containing d and f electrons (e.g. Fe and Ce). It should set LMAXMIX = 6.

janosh:
Are you sure?

```py
from pymatgen.core import Element

Element("La").block
>>> 'f'
```

arosen93:
Ugh, it's an f-block element but has no f electrons... lol. https://en.wikipedia.org/wiki/Lanthanum. It's probably fine to let LMAXMIX = 6 here regardless.

arosen93:
> Among the lanthanides, lanthanum is exceptional as it has no 4f electrons as a single gas-phase atom. Thus it is only very weakly [paramagnetic](https://en.wikipedia.org/wiki/Paramagnetic), unlike the strongly paramagnetic later lanthanides (with the exceptions of the last two, [ytterbium](https://en.wikipedia.org/wiki/Ytterbium) and [lutetium](https://en.wikipedia.org/wiki/Lutetium), where the 4f shell is completely full).[[17]](https://en.wikipedia.org/wiki/Lanthanum#cite_note-17) However, the 4f shell of lanthanum can become partially occupied in chemical environments and participate in chemical bonding

Ah okay, so testing for block is the right thing to do. Carry on with La.

janosh:
> Also also, your final test has Si in place of Fe.

Good catch.

> Also, it might be worth adding a test for a structure containing d and f electrons (e.g. Fe and Ce). It should set LMAXMIX = 6.

Will do.

matthewkuner:
@arosen93 @janosh would this set LMAXMIX = 4 for Gallium (a p-block element)? Because it should according to https://www.vasp.at/forum/viewtopic.php?p=24392#p24374

arosen93:
This is an interesting point worth solidifying. In practice, I think the answer is that it depends on the pseudopotential. For instance, if one were to use `Ge`, there are no d electrons in the valence. However, if you use the (recommended) `Ge_d` then of course they are. There is never harm in "overdoing it" with regards to LMAXMIX, so I would agree that something like Ge should have LMAXMIX = 4.

What this all boils down to is that:

1. The VASP manual was ambiguous, per usual.
2. This merged PR should probably be undone to go back to the old way of checking based on Z.

matthewkuner:
We could also just parse the electrons included in the particular pseudopotential and set LMAXMIX according to what types of electrons are present

arosen93:
That might be smarter, albeit slightly more complex.

matthewkuner:
This change should be reverted, no? @janosh

janosh:
>  We could also just parse the electrons included in the particular pseudopotential and set LMAXMIX according to what types of electrons are present

@matthewkuner are you able to submit a PR for this?

matthewkuner:
@janosh I may be able to do it eventually, but my bandwidth is a bit low rn. I think it is best to revert this change for the time being.

arosen93:
Agreed. I think simplest option is to revert the LMAXMIX-related change. No need to overengineer a solution.

janosh:
Just to confirm, you'd like to revert to this logic?

```py
        if "LMAXMIX" not in settings:
            # contains f-electrons
            if any(el.Z > 56 for el in structure.composition):
                incar["LMAXMIX"] = 6
            # contains d-electrons
            elif any(el.Z > 20 for el in structure.composition):
                incar["LMAXMIX"] = 4
```

matthewkuner:
@janosh that seems good to me

PR #3044, opened 2023-06-05
Add `(Structure|Molecule).calculate()` + `Molecule.relax()`, improve `Structure.relax()`

### LINKED ISSUE #3042: Wrong typehint in `Structure.relax`
https://github.com/materialsproject/pymatgen/blob/f90929f75b27fe1d06faa4e2a5f44654009b98b0/pymatgen/core/structure.py#L3989

Should be type `str  | Calculator` where `Calculator` is `from ase.calculators import calculator`

### LINKED ISSUE #3043: Add a `Structure.calculate()` feature analagous to `Structure.relax`
Currently, there is a method to do `Structure.relax` which calls an ASE calculation with the ASE Optimizer class. There should be a similar method called `Structure.run` which calls ASE to run `atoms.get_potential_energy()`. This will allow for the use of the calculator's internal optimizers if the user wishes.

I can add this.

### LINKED ISSUE #3047: `Structure.relax` only works if m3gnet is installed even if it is not used
Currently, if you want to use `Structure.relax` with an ASE `Calculator` object, you still need to install `m3gnet` even though it is not strictly needed for the relaxation procedure. The imports should be moved around to provide additional user flexibility.

arosen93:
Closes #3042. Closes #3043. Closes #3047.

Links:
- #3042
- #3043 
- #3047

In addition to the prior issues, I also increased the range of optimizers that can be passed to `Structure.relax` so the user doesn't have to rely solely on FIRE.

For molecules, I added a `Molecule.relax()` feature that defaults to GFN2-xTB and an analogous `Molecule.calculate()`.


shyuep:
I suggest:
1. We have a `run_calculation` method that replaces and extends most of the code in `relax` to be more flexible.
2. We have `relax` call run_calculator with a specific set of parameters to do a relaxation.

`Structure.run` as a naming sounds a bit strange to me.

arosen93:
Agreed. That's a much better name. That way it doesn't sound like the structure is running away...

arosen93:
@janosh --- whenever you have some time, I could use a bit of help with `mypy`, the bane of my existence. Somehow unrelated tests have `mypy` errors.

arosen93:
@shyuep, @janosh: This is ready for either of you to review.

A couple of notes:
- I did not see any clean way of using `Structure.run_calculation` as the base logic for `Structure.relax`. They both do fairly different things, and the code between the two reflects that. Nonetheless, I did use `.run_calculation` instead of `run` as requested.
- I can add in the analogous functions in `Molecule` once things look good for `Structure`. I imagine we should not use m3gnet as the default, so I'm open to suggestions.
- For some reason, m3gnet gives different lattice constants on Mac/Windows than on Linux for one of the tests.
- Instead of using the m3gnet custom `TrajectoryObserver`, I have switched to using ASE's native `Trajectory` object when `return_trajectory` is `True`. The latter has all the desired properties (e.g. energies, forces, cells, stresses, etc.) without requiring the m3gnet dependency for all ASE calculations. 
- I will need a bit of help with the `mypy` test failures.

pymatgen/core/structure.py
```diff
@@ -3980,67 +3984,174 @@ def set_charge(self, new_charge: float = 0.0) -> None:
 
     def relax(
         self,
-        calculator: str = "m3gnet",
+        calculator: str | Calculator = "m3gnet",
         relax_cell: bool = True,
+        optimizer: str = "FIRE",
```
janosh, line 3989: How about we make this `str | Optimizer` to match `calculator: str | Calculator`?

pymatgen/core/structure.py
```diff
@@ -3980,67 +3984,174 @@ def set_charge(self, new_charge: float = 0.0) -> None:
...
+        elif optimizer.lower() == "lbfgslinesearch":
+            opt_class = LBFGSLineSearch
+        elif optimizer.lower() == "gpmin":
+            opt_class = GPMin
+        elif optimizer.lower() == "mdmin":
+            opt_class = MDMin
+        elif optimizer.lower() == "fire":
+            opt_class = FIRE
+        else:
+            raise ValueError(f"Unknown optimizer: {optimizer}")
```
janosh, line 4045: Didn't test this but I would prob enforce users use the right casing in which case you could simplify to:

```suggestion
        from ase import optimize
        try:
            opt_class = getattr(optimize, optimizer)
        except AttributeError:
            raise ValueError(f"Unknown {optimizer=}, must be one of {list(dir(optimize))}")
```

pymatgen/core/structure.py
```diff
@@ -3980,67 +3984,174 @@ def set_charge(self, new_charge: float = 0.0) -> None:
...
 
+        # Get Atoms object
         adaptor = AseAtomsAdaptor()
         atoms = adaptor.get_atoms(self)
 
-        if return_trajectory:
-            # monitor the relaxation
-            traj_observer = TrajectoryObserver(atoms)
+        # Write out a trajectory file
```
janosh, line 4051: Writing out a file might be unexpected. I know I sometimes run 1000s of relaxations in which case 1000s of new log files can really slow down VS Code Remote server.

janosh (commented):
Just left some general feedback. Excited to try this out later!

pymatgen/core/structure.py
arosen93: I see. Perhaps that's the original motivation for the observer.

I think a good middleground would be to keep the TrajectoryObserver for m3gnet and use the Trajectory file for regular ASE calculators. This has the benefit of making it so m3gnet is not a strict dependency but doesn't require file I/O for the very fast ML potential. The file I/O shouldn't be an issue for standard ASE calculators because they will always do file I/O (e.g. VASP, Gaussian).

arosen93:
Thanks, @janosh! I'll address your comments.

Regarding the naming, I actually assign the ASE calculator to `Structure.calc`, which is worth doing because this object contains all the calculation results and parameters (presumably highly desirable for the end user to be able to access).

Maybe `Structure.calculate` would be a better name for the function. Open to other ideas.

shyuep:
calculate sounds good to me.

shyuep:
I would point out one thing - the Structure.relax / whatever are generally meant to be convenience methods for simple needs. It is for the user who just wants an answer without having to mess around with lots of differnet options. 

If you are planning to properly use ASE calculators and/or M3Gnet etc., it is best to use the API in those packages instead.

Also, one more thing - matgl now has the M3Gnet universal potential implemented. We would probably need to replace the M3Gnet in this function shortly.

arosen93:
> I would point out one thing - the Structure.relax / whatever are generally meant to be convenience methods for simple needs. It is for the user who just wants an answer without having to mess around with lots of differnet options.

Definitely understood. In case it's of interest, I will note that from speaking with folks who use both ASE and Pymatgen, there is a strong interest in building workflows centered around Pymatgen `Structure`/`Molecule` objects while still taking advantage of the convenience associated with ASE calculators. I also feel like many people are simply not aware of the `AseAtomsAdaptor` and so would not think to use it to construct similar functions on their own. With the `.relax` and `.calculate` methods, this should hopefully reduce the need for people to recreate the same boilerplate functions while still maximizing convenience. Of course, we shouldn't cover every edge case in Pymatgen, but I think the ability to pass objects like the `Calculator` or `Optimizer` class keeps things pretty general.

Hopefully with this PR we'll hit a good middleground! Always open to suggestions.

pymatgen/core/structure.py
arosen93: Brilliant. I knew there was a cleaner way to do this but didn't know how!

arosen93:
@janosh: Okay, this ready for review again! I'm done! :)

Some examples you might consider trying:

## Structure Relax with M3GNet

```python

structure = Structure(
    lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],
    species=["Cu", "O"],
)
structure.relax()
```

## Structure Relax with EMT

```python
from ase.calculators.emt import EMT

structure = Structure(
    lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],
    species=["Cu", "O"],
)
structure.relax(calculator=EMT())
```

## Molecule Relax with GFN2-xTB

```python
from pymatgen.core import Molecule

coords = [
    [0.000000, 0.000000, 0.000000],
    [0.000000, 0.000000, 1.089000],
    [1.026719, 0.000000, -0.363000],
    [-0.513360, -0.889165, -0.363000],
    [-0.513360, 0.889165, -0.363000],
]
mol = Molecule(["C", "H", "H", "H", "H"], coords)
mol.relax()
```

## Molecule Relax with Lennard-Jones

```python
from ase.calculators.lj import LennardJones

coords = [
    [0.000000, 0.000000, 0.000000],
    [0.000000, 0.000000, 1.089000],
    [1.026719, 0.000000, -0.363000],
    [-0.513360, -0.889165, -0.363000],
    [-0.513360, 0.889165, -0.363000],
]
mol = Molecule(["C", "H", "H", "H", "H"], coords)
mol.relax(calculator=LennardJones())
```

setup.py
```diff
@@ -81,6 +82,7 @@
             "netCDF4>=1.5.8",
             "phonopy>=2.4.2",
             "seekpath>=1.9.4",
+            "tblite[ase]>=0.3.0; platform_system=='Linux'",
```
janosh, line 85: I just tried installing `tblite` on macOS and got the error 

```txt
      ../../subprojects/tblite/config/meson.build:46:12: ERROR: Dependency "openmp" not found, tried system
```

I assume that's why `platform_system=='Linux'`. So is Molecule relax currently a Linux-only feature?

pymatgen/core/structure.py
```diff
@@ -3980,68 +3986,67 @@ def set_charge(self, new_charge: float = 0.0) -> None:
 
     def relax(
         self,
-        calculator: str = "m3gnet",
+        calculator: str | Calculator = "m3gnet",
```
janosh, line 3989: I think we should add CHGNet as another `str` option here and even make it the default considering it's easier to install on account of not needing TF (PyTorch causes me less trouble anyways).

janosh (commented):
@arosen93 Sorry for the slow reply! This is looking great! I left 2 questions/comments. Also, I saw that `_prep_calculator` and `_relax` are marked as private but since they have user-facing features, i.e. raise error messages, I think it would be good to add a test for that behavior. Can try to do that later.

pymatgen/core/structure.py
```diff
@@ -4498,6 +4504,266 @@ def substitute(self, index: int, func_group: IMolecule | Molecule | str, bond_or
...
+                Defaults to 'm3gnet', i.e. the M3GNet universal potential.
+            verbose (bool): whether to print stdout. Defaults to False.
+
+        Returns:
+            Molecule: Molecule following ASE calculation.
+        """
+        return _calculate(self, calculator, verbose=verbose)
+
+
+def _calculate(
```
janosh, line 4566: Would it maybe make more sense to add these private functions as methods to `class SiteCollection`?

setup.py
arosen93: @janosh: Could you try installing from [conda-forge](https://anaconda.org/conda-forge/tblite)? I don't have a Mac so can't test. Maybe that will give you better luck. I'm not 100% sure how to build it with the ASE support though from Conda.

pymatgen/core/structure.py
arosen93: Happy to add support for CHGNet. Would have to get @shyuep's approval to change the default, as that'd be a breaking change (and I imagine he has personal preference about M3GNet).

shyuep: Yes on supporting CHGNet. No on changing default. In any case, we will replace the m3gnet calculator with the Matgl version soon. That will be Pytorch.

arosen93: Actually, it seems like CHGNet by default takes a `Structure` object in as input. Obviously that's not a problem, but it'd require some modification of the logic here which converts between ASE `Atoms` and `Structure`. I think that's best left to a future PR.

setup.py
janosh: I don't have `conda`. Not hard to install of course but would like to avoid. Thanks for opening the tblite issue though! I don't think lack of Mac support is a deal breaker. We could not make it the default if others disagree. But i think the `Molecule.relax`  will take some time to attract attention anyway and by then the issue you opened might be fixed.

pymatgen/core/structure.py
janosh: > Actually, it seems like CHGNet by default takes a Structure object in as input.

Ah yes, forget to mention that since I already started implementing and noticed it would take some reworking.

> I think that's best left to a future PR.

Agreed.

setup.py
arosen93: Makes sense. I also would be pretty surprised if we found a better default that worked for most of the periodic table with reasonable accuracy. Hard to beat semi-empirical tight binding... until someone trains an M3GNet-equivalent on Molecules data on MP.

pymatgen/core/structure.py
arosen93: Ah, I was looking for something that was inherited by both Structure and Molecule. Didn't know about this. I think it makes more sense to add there.

arosen93:
@janosh --- ready for either re-review or merging. Some of the Mac tests died but seems to be a problem for everyone.

janosh:
Thanks @arosen93! Took a few iterations but this was no small PR and I think it turned out really well! 😎  Thanks for the excellent test coverage! 🧪

PR #3052, opened 2023-06-07
`IcohpValue.__str__` don't `return None`

janosh:
Also use modern f-string syntax.


janosh:
CI failure caused by legacy `MPRester`, unrelated to PR.

Issue #3049, opened 2023-06-06
Generate random ordered structure from disordered structure


exenGT:
#### Problem

For molecular dynamics simulation and machine learning data generation of random structures, one would like to have a routine that transforms a given disordered structure $S$ to a set of random ordered structures ${S'_i}$, where each random ordered structure $S'_i$ has the same overall composition as the disordered structure $S$.

Currently for this purpose, there exists `OrderDisorderedStructureTransformation` in transformations/standard_transformations.py. However, this function over-complicates the problem by using the Ewald energy to rank the generated random ordered structures. When the structure in question is very large (e.g. thousands of atoms) in a typical MD simulation, this approach simply doesn't work due to the combinatorially large search space. Instead, what we actually want is very simple: a set of random ordered structures. The energy rankings is not of concern here.

#### Proposed Solution

We have implemented a new standard transformation routine, named `RandomStructureTransformation`. Given a disordered structure `structure`, and a number `num_copies`, its method `apply_transformation` returns a set of `num_copies` copies of random ordered structures. This routine distinguishes between inequivalent sublattices of the crystal.

#### Additional Info

The following example shows the usage of this routine:
```
from mp_api.client import MPRester

from pymatgen.transformations.standard_transformations \
  import RandomStructureTransformation

# put your api_key here
api_key = "*****"

with MPRester(api_key) as mpr:
    struct = mpr.get_structure_by_material_id("mp-2534", conventional_unit_cell=True)

struct[0:4] = {"Ga": 0.5, "In": 0.5}
struct[4:] = {"P": 0.5, "As": 0.5}

struct.make_supercell([3, 3, 3])

## random structure transformation
trans = RandomStructureTransformation()
new_structs = trans.apply_transformation(structure = struct,
                                         num_copies = 5)

## output to files
for i_struct, new_struct in enumerate(new_structs):
    new_struct.to("random_structure_{}.cif".format(i_struct))
```

Below are the screenshots of the generated random structures:
<img width="310" alt="Screenshot 2023-06-06 at 1 37 52 AM" src="https://github.com/materialsproject/pymatgen/assets/13700125/35133031-3d5b-4959-8f91-93717dbedf09">
<img width="310" alt="Screenshot 2023-06-06 at 1 37 48 AM" src="https://github.com/materialsproject/pymatgen/assets/13700125/29ae578b-cbd8-4a65-889f-23a267218e83">
[...]

PR #3057, opened 2023-06-10
Adding property `sublattices` to pymatgen/core/structure.py; adding class `RandomStructureTransformation` to pymatgen/transformations/standard_transformations.py

exenGT:
This pull request is on issue #3049

### DIFF
--- a/pymatgen/transformations/standard_transformations.py
+++ b/pymatgen/transformations/standard_transformations.py
@@ -494,6 +494,86 @@ class PartialRemoveSpecieTransformation(AbstractTransformation):
         return
 
 
+class RandomStructureTransformation(AbstractTransformation):
+
+    def __init__(self):
+        pass
+
+    def apply_transformation(self, structure: Structure, num_copies: int):
+        """
+        For this transformation, the apply_transformation method will return
+        a random ordered structure from the given disordered structure.
+        """
+        subl = structure.sublattices
+
+        # fill the sublattice sites with pure-element atoms
+        self.all_structures = []
+
+        for _ in range(num_copies):
+
+            new_structure = structure.copy()
+
+            for subl_comp, subl_indices in subl.items():
+    
+                # convert composition into a dictionary
+                subl_comp_dict = subl_comp.as_dict()
+
+                el_list = list(subl_comp_dict.keys())
+                el_concs = np.array(list(subl_comp_dict.values()))
+
+                # randomly choose site indices for each element present in the sublattice
+
+                el_indices = self.random_assign(sequence = subl_indices,
+                                                lengths = (el_concs * len(subl_indices)).astype(int))
+
+                for i_el, el in enumerate(el_list):
+
+                    new_structure[el_indices[i_el]] = el
+
+            self.all_structures.append(new_structure)
+
+        return self.all_structures
+
+    @property
+    def inverse(self):
+        """
+        Returns: None
+        """
+        return None
+
+    @property
+    def is_one_to_many(self) -> bool:
+        """
+        Returns: True
+        """
+        return True
+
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @exenGT, I added a question to the associated issue.

pymatgen/transformations/standard_transformations.py
```diff
@@ -494,6 +498,82 @@ def inverse(self):
...
+        for length in lengths:
+            end_pos = min(start_pos + length, len(sequence))
+
+            ## check if end_pos is greater than start_pos
+            if end_pos > start_pos:
+                assignments.append(sequence[start_pos:end_pos])
+                start_pos = end_pos
+
+            else:
+                print("Sum of lengths must be equal to the length of the sequence!")
```
mkhorton, line 571: This should `raise`, instead of print/return. Using -1 as a return code isn’t typically used in Python.

exenGT: I have corrected this issue, along with several other linting issues.

janosh (changes requested):

exenGT:
@janosh All your suggestions have been addressed in my new commits.

pymatgen/transformations/standard_transformations.py
```diff
@@ -494,6 +495,111 @@ def inverse(self):
         return
 
 
+class RandomStructureTransformation(AbstractTransformation):
+    """
+    Transform a disordered structure into a given number of ordered random structures.
+    """
+
+    def __init__(self):
+        pass
```
janosh, line 504: delete

pymatgen/transformations/standard_transformations.py
```diff
@@ -494,6 +495,111 @@ def inverse(self):
...
+            structure: input structure
+            n_copies (int): number of copies of ordered random structures to be returned
+
+        Returns:
+            A list of ordered random structures based on the input disordered structure.
+        """
+        subl = structure.sublattices
+
+        # fill the sublattice sites with pure-element atoms
+        self.all_structures = []
```
janosh, line 521: is it necessary to store the structures on the transformation instance? might balloon mem usage if people create many of these `RandomStructureTransformation`s.

janosh (changes requested):
@exenGT Looking good! Could you add a test in [`test_standard_transformations.py`](https://github.com/materialsproject/pymatgen/blob/6e7e737d1feefd2ff2764188a90e665e979a8375/pymatgen/transformations/tests/test_standard_transformations.py)?

pymatgen/transformations/standard_transformations.py
exenGT: I used this syntax because there is an `is_one_to_many` property in the class; should it be `False` by default? (I see it is `True` for other classes)

janosh: I think there's no connection between `is_one_to_many` (that part is good) and storing the transformed structures on the transformation instance?

exenGT: Could you elaborate more on this? I feel I'm not totally understanding. For example, the `apply_transformation` method of `OrderDisorderedTransformation` returns a list of structures as well (`self._all_structures`). What is the issue with this approach?

janosh: Could you explain the need to store all structures for later? Why not just return them and let them be garbage collected if they go out of scope in the user's code? Currently they live as long as the `Transformation` instance which is not great given structures are notoriously mem hungry.

exenGT: I see. This feature is here simply because I'm copying what's been written for the other transformations. If you think it's not a good approach, then I can modify it in my code accordingly.

shyuep:

PR #3060, opened 2023-06-12
Skip `Molecule.relax()` tests

Andrew-S-Rosen:
See below.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -788,8 +788,8 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
                 from matgl.ext.ase import TrajectoryObserver
 
                 traj_observer = TrajectoryObserver(atoms)
-            elif "trajectory" not in opt_kwargs:
-                opt_kwargs["trajectory"] = "opt.traj"
+            else:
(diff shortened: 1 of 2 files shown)


arosen93:
@janosh: I can confirm after running tests locally that everything works fine with `Molecule.calculate()` when `matgl` is not installed, but when you install `matgl` in the same environment as `tblite[ase]`, `tblite` to run into convergence errors and therefore fails the test suite.

The root cause is `torch`. If you uninstall `torch`, everything works fine with `tblite`. Related error: https://github.com/tblite/tblite/issues/110

How should we best proceed? Should we split the tests up based on dependencies a bit until it's resolved upstream? Note that there was one import change I had to make in this PR regardless (and the `tmpdir` did not fix anything but I kept it).

arosen93:
Pinging @shyuep to let him know the (rather disappointing) cause.

shyuep:
I think the simple way to deal with this would be to have the tests in one single method, run the tblite first, and then import matgl and run the other tests? I agree this is a nutty bug.

shyuep:
The e alternative is to comment out the tblite tests first. I think that is less used than Structure.relax.

arosen93:
Yup, that's exactly what I did here right now was comment them out.

Issue #3065, opened 2023-06-13
Error in writing Structure to cif file, when site species have "spin" property


kamronald:
Structure arose when enumerating magnetic structures, related to #3006 . When saving the enumerated structures, cif files are very different from the actual Structure objects generated by MagneticStructureEnumerator.
Writing to cif is fixed when the structure is modified such that spins are not defined, but replaced by defining a "magmom" site property.

mkhorton:
Be aware that disordered magnetic materials are possible. I've never been entirely satisfied with the distinction between Species.spin and the magmom site property, but the former does allow a disordered magnetic material to be defined, while the latter does not.

shyuep:
I would distinguish spin as an theoretical construct attached to an element. Eg a low spin Co3+ in octahedral has a spin of 0. Magmom is an actual calculated property. Eg an actual Co3+ in LiCoO2 can have a non zero magmom.

mkhorton:
I remember this comment from you last time we discussed @shyuep :)

I think the reason I find it unsatisfying is simply that the site properties are not typed, and the use of "magmom" is mostly by our own convention.

janosh:
Is this technically a bug? Not sure if `pymatgen` guarantees CIF round-trip identity?

@mkhorton Are you mostly concerned about not following a standard or site props being untyped? We could start typing by defining a `SiteProps` `TypedDict`:

```py
from typing import TypedDict

class SiteProps(TypedDict, total=true):
    magmom: list[float]
    charge: list[float]
```

`total=True` allows any key to be present in the dictionary, even if not specified in the type hint.

@kamronald Could you post a minimal repro [similar to this](https://github.com/materialsproject/pymatgen/issues/3006#issuecomment-1563093693)?

kamronald:
Yes here is the repro. I have attached the CONTCAR and output structures below.

```
from pymatgen.core import Structure, PeriodicSite
from pymatgen.analysis.magnetism import MagneticStructureEnumerator

struc = Structure.from_file('CONTCAR')
mag_enum = MagneticStructureEnumerator(structure=struc, automatic=False, default_magmoms={'Li': 0.0, 'Mn': 4.0, 'O': 0.0},
                                       strategies=['ferromagnetic', 'antiferromagnetic'])
for i, struc in enumerate(mag_enum.ordered_structures):
    struc.to(f'enum_{i}.cif')

def replace_spins_w_magmom(structure):
    sites = []
[...]

PR #3071, opened 2023-06-15
Bug fix: change CifWriter to save only the element symbol, and not the entire species string

kamronald:
Change the "atom site type symbol", such that the element symbol is saved instead of the full species string (which includes site properties). This fixes Issue #3065, allowing for accurate writing of structures containing species properties to CIF and MCIF files.

## Summary

Major updates:

Change the "atom site type symbol", such that the element symbol is saved instead of the full species string (which includes site properties).

## Checklist

Ensure:

- [ ] All tests and linting pass.

### DIFF
--- a/pymatgen/io/cif.py
+++ b/pymatgen/io/cif.py
@@ -1326,7 +1326,7 @@ class CifWriter:
         if symprec is None:
             for site in struct:
                 for sp, occu in sorted(site.species.items()):
-                    atom_site_type_symbol.append(str(sp))
+                    atom_site_type_symbol.append(str(sp.symbol))
                     atom_site_symmetry_multiplicity.append("1")
                     atom_site_fract_x.append(format_str.format(site.a))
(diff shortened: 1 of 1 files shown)


mkhorton:
@kamronald does this still handle species with oxidation state correctly?

kamronald:
@mkhorton good point, I just checked and it does not handle them properly, I'll fix that

kamronald:
@mkhorton oxidation states are now properly saved

janosh:
Thanks @kamronald! 👍 

IIUC @mkhorton in https://github.com/materialsproject/pymatgen/issues/3065#issuecomment-1593845643, he's suggesting we continue saving scalar magmoms to regular CIF using the local data name `mp` registered here https://www.iucr.org/cgi-bin/cifreserve.pl.

Also, would be great if you could refactor your repro https://github.com/materialsproject/pymatgen/issues/3065#issuecomment-1593523071 into a new test case.

shyuep:

PR #3073, opened 2023-06-17
Update test_kpath_lm.py, remove and simplify tests

JaGeo:
HI @janosh ,

I removed and simplified the tests. I am still not sure why there is an "or" for the other tests. In my opinion, the result should be always the same for one structure.

It's also redundant to test if g has a certain k-point and then later on tests if point g has the same kpoint or point xyz has a certain k-point.

### DIFF
--- a/pymatgen/symmetry/tests/test_kpath_lm.py
+++ b/pymatgen/symmetry/tests/test_kpath_lm.py
@@ -79,21 +79,9 @@ class KPathLatimerMunroTest(PymatgenTest):
 
         assert kpoints["e"] == approx([0.0, 0.49999999999999956, 0.5000000000000002])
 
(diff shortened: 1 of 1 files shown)


JaGeo:
Openbabel seems to be missing. Not sure why. Should be unrelated to thus change.

janosh:
Yeah I think that's caused by a the latest `mp-api` release. Thank for the cleanup!

janosh (approved):

JaGeo:
Thanks!

 Maybe @munrojm could also shortly check if it should be an "and" rather than an "or" in the other lines and the one that I removed. I assume some of the tests should be related to the time-reversal symmetry for magnetic space groups.

PR #3077, opened 2023-06-18
Fix `ase` tests not running due to `pymatgen/io/ase.py` shadowing `ase` package

janosh:
Tests were set to skip if `ase` not importable (now removed so we catch this in the future).

@arosen93 `sys.path.insert(0, getsitepackages()[0])` is admittedly an ugly hack but, like you said, preferable over a breaking module rename.

### DIFF
--- a/pymatgen/io/ase.py
+++ b/pymatgen/io/ase.py
@@ -21,6 +21,8 @@ if TYPE_CHECKING:
     from pymatgen.core.structure import SiteCollection
 
 
+# ensure site packages have higher precedence than local modules so that this file
+# can't shadow the ase package
 sys.path.insert(0, getsitepackages()[0])
 try:
     from ase import Atoms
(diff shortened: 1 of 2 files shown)


arosen93:
Thank you!! 🙏

arosen93:
If the user doesn't have ASE installed though, the test still skips right?

janosh:
No, running _all_ tests now requires `ase` to be installed. I think that's fine. Running tests locally usually involves specifying a path to a test module so you only run the tests you care about in which case `ase` is not needed.

shyuep:
@janosh @arosen93 I think this is where the problem started.

shyuep:
Let's not do ugly hacks in main code for things like tests, linting, etc.! It is more important that features work properly. If tests or lintinig require hacks, they should be done in test code that don't affect deployed code.

janosh:
Oops. Don't remember why I put this in `pymatgen/io/ase.py`. Definitely seems like a bad idea in retrospect. I think I meant to put this in `pymatgen/io/tests/test_ase.py` (presumably that didn't work, hence this bad alternative).

PR #3079, opened 2023-06-19
Improvements to the documentation of the Lobsterenv module

JaGeo:
I improved the documentation of the LobsterEnv module which was triggered by our updates to LobsterPy (https://github.com/JaGeo/LobsterPy/pull/81) and suggestions by @ajjackson

@janosh : happy to hear additional thoughts from you.

### DIFF
--- a/pymatgen/io/lobster/lobsterenv.py
+++ b/pymatgen/io/lobster/lobsterenv.py
@@ -43,43 +43,42 @@ class LobsterNeighbors(NearNeighbors):
 
     def __init__(
         self,
-        are_coops=False,
-        filename_ICOHP=None,
-        valences=None,
-        limits=None,
-        structure=None,
-        additional_condition=0,
-        only_bonds_to=None,
-        perc_strength_ICOHP=0.15,
-        valences_from_charges=False,
-        filename_CHARGE=None,
-        which_charge="Mulliken",
-        adapt_extremum_to_add_cond=False,
-        add_additional_data_sg=False,
-        filename_blist_sg1=None,
-        filename_blist_sg2=None,
-        id_blist_sg1="ICOOP",
-        id_blist_sg2="ICOBI",
+        are_coops: bool = False,
+        filename_ICOHP: str = None,
+        valences: list[int | float] = None,
+        limits: tuple[float, float] | None = None,
+        structure: Structure = None,
+        additional_condition: int = 0,
+        only_bonds_to: list[str] | None = None,
+        perc_strength_ICOHP: float = 0.15,
+        valences_from_charges: bool = False,
+        filename_CHARGE: str = None,
+        which_charge: str = "Mulliken",
+        adapt_extremum_to_add_cond: bool = False,
+        add_additional_data_sg: bool = False,
+        filename_blist_sg1: str = None,
+        filename_blist_sg2: str = None,
+        id_blist_sg1: str = "ICOOP",
+        id_blist_sg2: str = "ICOBI",
     ) -> None:
         """
 
         Args:
             are_coops: (bool) if True, the file is a ICOOPLIST.lobster and not a ICOHPLIST.lobster; only tested for
-            ICOHPLIST.lobster so far
+                ICOHPLIST.lobster so far
             filename_ICOHP: (str) Path to ICOHPLIST.lobster
-            valences: (list of integers/floats) gives valence/charge for each element
-            limits: limit to decide which ICOHPs should be considered
-            structure: (Structure Object) typically constructed by: Structure.from_file("POSCAR") (Structure object
-            from pymatgen.core.structure)
-            additional_condition:   Additional condition that decides which kind of bonds will be considered
-                                    NO_ADDITIONAL_CONDITION = 0
-                                    ONLY_ANION_CATION_BONDS = 1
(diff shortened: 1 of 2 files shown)


JaGeo:
@janosh, I tried to add more type hinting but mypy was raising very weird errors.

pymatgen/io/lobster/lobsterenv.py
```diff
@@ -354,19 +354,21 @@ def get_light_structure_environment(self, only_cation_environments=False, only_i
...
-            these bonds
-            [the latter is useful for plotting summed COHP plots],
-            list of the central isite for each label
+        Returns: tuple including
+            sum of icohp values of neighbors to the selected sites [given by the id in structure] (float),
+            list of summed icohp values for all identified interactions with neighbors (list),
+            number of identified bonds to the selected sites (int),
+            labels (from ICOHPLIST) for all identified bonds returned as a list of site names, e.g.,
+            ['Ag3', 'O5'] [the latter is useful for plotting summed COHP plots] (list),
+            list of the central isite for each identified interaction (list)
```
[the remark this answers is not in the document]
JaGeo: Yes, you are correct. I somehow merged labels and atoms in the description.

ajjackson:
Where functions are returning a large tuple containing various types and lengths of data, it might be helpful to wrap them up as a NamedTuple. This would be backward-compatible as you still _can_ access the elements by index, but also allows:

- clearer type-hinting than a dict would have, taking advantage of the immutability
- named attribute access for more intelligible code and lower chance of accessing the wrong element by mistake
- easier interactive exploration e.g. in a Jupyter notebook or IPython terminal; we get something a bit more self-describing than a bare list of tuple, but not a complicated new class that might have surprising behaviours

Be aware that while a lot of the old docs and tutorials recommend to use the `namedtuple` function to produce them, we can now build them with `class` for clearer type-hinting. See examples at https://docs.python.org/3/library/typing.html#typing.NamedTuple

So we could do something like

```python
class ICOHPNeighboursInfo(NamedTuple):
    total: float
    interaction_sums: List[float]
    n_bonds: int
    labels: List[str]
    atoms: List[SomeAtomTypeOrOther]
    central_isites: List[int]

class LobsterNeighbors(NearNeighbors):
...
    def get_info_icohps_to_neighbors(
        self,
        isites: Optional[List[int]] = None,
        onlycation_isites: bool = True
            ) -> ICOHPNeighboursInfo:
    ...
    return ICOHPNeigboursInfo(summed_icohps, list_icohps, number_bonds, labels, atoms, final_isites)
```

JaGeo:
Thanks, @ajjackson , this is a great suggestion. I will implement it as it is much cleaner.

JaGeo:
@ajjackson , I implemented the named tuple

pymatgen/io/lobster/lobsterenv.py
```diff
@@ -43,43 +43,42 @@ class LobsterNeighbors(NearNeighbors):
...
-        add_additional_data_sg=False,
-        filename_blist_sg1=None,
-        filename_blist_sg2=None,
-        id_blist_sg1="ICOOP",
-        id_blist_sg2="ICOBI",
+        are_coops: bool = False,
+        filename_ICOHP: str = None,
+        valences: list[int | float] = None,
+        limits: tuple[float, float] | None = None,
+        structure: Structure = None,
```
janosh, line 50: > @janosh, I tried to add more type hinting but mypy was raising very weird errors.

Is the `structure` arg really optional here? What happens if you don't supply a structure?

pymatgen/io/lobster/lobsterenv.py
```diff
@@ -214,8 +213,8 @@ def molecules_allowed(self):
 
     def get_anion_types(self):
```
janosh, line 214: Maybe this could be a property?
```suggestion
    @property
    def anion_types(self):
```

JaGeo: Yes. It will just break some code in lobsterpy.

pymatgen/io/lobster/lobsterenv.py
JaGeo: You are correct. Same is true for "filename_ICOHP".

pymatgen/io/lobster/lobsterenv.py
JaGeo: I have now implemented this but the `get_anion_types` method will still exit until we have made the changes in lobsterpy.

pymatgen/io/lobster/lobsterenv.py
```diff
@@ -227,6 +227,17 @@ def get_anion_types(self):
...
         return set(anion_species)
 
+    def get_anion_types(self):
+        """
+        Return the types of anions present in crystal structure as a set
+        Returns: set of Element describing anions in the crystal structure
+        """
+        from warnings import warn
+
+        warn("get_anion_types will be deprecated. Please use anion_types instead.")
```
janosh, line 237: monty has a [decorator `deprecated`](https://pythonhosted.org/monty/monty.html#monty.dev.deprecated) for this:

```suggestion
        @deprecated(replacement=self.anion_types, message="use anion_types instead")
        def get_anion_types(self):
```

janosh (approved):
Thanks @JaGeo! 👍

PR #3080, opened 2023-06-19
Add a cache to speed up get_el_sp

v1kko:
Uses functools.cache to speed up the conversion of arbitrary objects into Element | Species | DummySpecies.

- One thing to keep in mind is that returned objects are not necessarily unique anymore, they could be returned again through the cache.
- Arguments to `get_el_sp` must be `Hashable`, luckily this seems to be the case

## Summary

Major updates:

- feature 1: massive speed up for get_el_sp

## Checklist

Ensure:

- [x] All tests and linting pass.

Tip: Install `pre-commit` hooks to auto-check before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/core/periodic_table.py
+++ b/pymatgen/core/periodic_table.py
@@ -1562,7 +1562,7 @@ class DummySpecie(DummySpecies):
     to maintain backwards compatibility.
     """
 
-
+@functools.cache
 def get_el_sp(obj) -> Element | Species | DummySpecies:
     """
     Utility method to get an Element or Species from an input obj.
--- a/pymatgen/core/periodic_table.py
+++ b/pymatgen/core/periodic_table.py
@@ -1562,6 +1562,7 @@ class DummySpecie(DummySpecies):
     to maintain backwards compatibility.
     """
 
+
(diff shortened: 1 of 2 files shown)


janosh:
> One thing to keep in mind is that returned objects are not necessarily unique anymore, they could be returned again through the cache.

This sounds like a foot gun we should stay away from. Could you describe your motivation for this change? Maybe there's a safer way to go about this.

janosh:
Closing this as stale for now. Feel free to report back and we can reopen/discuss further.

shyuep:
Element and Species are immutables. If this has a clear speed up, it should be fine.

janosh:
Species are immutable?

```py
from pymatgen.core import Species

sp = Species("Fe2+", properties={"spin": 5})
sp.spin
>>> 5

sp.spin = 4
sp.spin
>>> 4
```

shyuep:
They are *supposed* to be. That is the reason why this change has zero failures across pymatgen. If you really want to enforce strict immutability, then Species should be changed to a dataclass.

PR #3086, opened 2023-06-21
Fix hill_formula Composition property

amkrajewski:
## Summary

This PR fixes an error in the **hill_formula** property of the Composition object, which occurs in cases where C, H, and other element/s with symbol ordering alphabetically before H are present at the same time. In such a case, H should be placed directly after C, but this is not implemented nor tested in the current version.

- Fixed the property to match the description.
- Implemented additional tests.

Sidenote: @ml-evs This fix should help eliminate compatibility issues with the OPTIMADE _chemical_formula_hill_ field (except for the presence/lack of empty spaces).


janosh (approved):
Nice work! Thanks for the detailed description and new tests! 👍

PR #3095, opened 2023-06-23
Move `get_zmatrix` from `GaussianInput` to `Molecule`

### LINKED ISSUE #3090: Z-matrix representation
I am currently using some pymatgen parts that I hadn't used before. As far as I see, we currently have an option for a z-matrix representation as part of the Gaussian output generation: https://github.com/materialsproject/pymatgen/blob/f90929f75b27fe1d06faa4e2a5f44654009b98b0/pymatgen/io/gaussian.py#L383-L413

I wonder if this wouldn't be a good convenience method for molecule as this is also used in other quantum-chemical codes. 
I am leaving this here as a suggestion for refactoring and in case pymatgen plans to support more molecular qm codes.

JaGeo:
Closes #3090 

I really just moved the method. It's not sophisticated yet but maybe someone can build on this implementation. I therefore left a todo.

### DIFF
--- a/pymatgen/analysis/molecule_matcher.py
+++ b/pymatgen/analysis/molecule_matcher.py
@@ -858,7 +858,7 @@ class BruteForceOrderMatcher(KabschMatcher):
         of atoms from the same species.
     """
 
-    def match(self, p: Molecule, ignore_warning=False):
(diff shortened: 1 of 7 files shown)


JaGeo:
@janosh I don't understand why this test would be failing. Hints are appreciated.

JaGeo:
Ah, they have been failing already before. Okay. I was just confused that it was related to molecules.

janosh (approved):

PR #3097, opened 2023-06-24
Add CHGNet to `structure.relax` and `structure.calculate`

janosh:
Following hot (or not?) on the heels of #3044, this PR adds `calculator="chgnet"` as a recognized option to `Structure.relax()` and `Structure.calculate()` for running [CHGNet](https://github.com/CederGroupHub/chgnet)-powered structure relaxations and static calculations.

@BowenD-UCB This is a 1st draft. Will require further tweaks but would be good to get your feedback.

Also @arosen93 feel free to chime in if you feel inclined.


pymatgen/core/structure.py
```diff
@@ -644,31 +644,40 @@ def extract_cluster(self, target_sites: list[Site], **kwargs) -> list[Site]:
...
+                result of ASE calculation.
         """
+        if isinstance(calculator, str) and calculator.lower() == "chgnet":
+            from chgnet.model import CHGNet
+
+            chgnet = CHGNet.load()
+            preds = chgnet.predict_structure(self)
+
+            out = self.copy()  # type: ignore[attr-defined]
+            out.calc = preds
```
janosh, line 668: @arosen93 @shyuep Setting `out.calc = preds` is inconsistent with the other `calculator` options which store a `ase.Calculator` instance as the `calc` attribute on the calculated structure. But not sure this is the optimal API design and feels like fitting a round peg in a square hole to try and conform CHGNet to this which just returns a dict with predicted `energies`, `forces`, `stresses` and `magmoms`.

arosen93: In general, I think the mechanism to return calculated results (beyond the updated `Structure`) could/should be improved, so I wouldn't worry about trying to match what exists currently if you have a better idea. My thinking at the time was that, originally, the updated `Structure` object was the only thing returned. But the user also calculated forces, stresses, magmoms, etc. so it seemed worthwhile to return those too. Rather than just sticking the calculator (which contains the results in `calc.results`) onto the `Structure` object, it might be cleaner to add a kwarg that would enable the results to be returned as a separate `return` variable. Or maybe there is another clean way to do it. But the tl;dr here is that if there's a way to make it more consistent/cleaner, I would suggest doing so because that was bothering me when I wrote it.

arosen93:
@janosh: A small note. If you do 
```python
from chgnet.model.dynamics import CHGNetCalculator
```

then you have imported an ASE calculator. You can provide the calculator as a kwarg via `calculator=CHGNetCalculator` just like any other calculator. I'm not sure if this would reduce any redundancy in what you've written but just sharing that as well.

pymatgen/core/structure.py
JaGeo: But couldn't one then just use atomate2? There, you get everything and it also only takes a few lines of code. Import the job, run it with a memory db and then access the output.

arosen93: That's true (although I don't necessarily view the users of Pymatgen to necessarily be the same as those who would open up Atomate2).

Since this is just meant to be a "quick cleanup" kind of thing, one could quite reasonably argue that you're just asking for the `Structure` or `Molecule` to update if you're doing `.relax()`.

In summary, I'm sure you can come to a clean resolution, @janosh!

pymatgen/core/structure.py
```diff
@@ -644,31 +644,40 @@ def extract_cluster(self, target_sites: list[Site], **kwargs) -> list[Site]:
...
+            Structure | Molecule: Structure or Molecule with new calc attribute containing
+                result of ASE calculation.
         """
+        if isinstance(calculator, str) and calculator.lower() == "chgnet":
+            from chgnet.model import CHGNet
+
+            chgnet = CHGNet.load()
+            preds = chgnet.predict_structure(self)
+
+            out = self.copy()  # type: ignore[attr-defined]
```
arosen93, line 667: Could you use a more descriptive variable name than `preds` and `out`?

pymatgen/core/structure.py
```diff
@@ -644,31 +644,40 @@ def extract_cluster(self, target_sites: list[Site], **kwargs) -> list[Site]:
...
+            out.calc = preds
+            return out
+
         from pymatgen.io.ase import AseAtomsAdaptor
 
-        is_molecule = isinstance(struct, Molecule)
-        if is_molecule and calculator == "m3gnet":
+        is_molecule = isinstance(self, Molecule)
+        if is_molecule and isinstance(calculator, str) and calculator.lower() == "m3gnet":
             raise ValueError(f"Can't use {calculator=} for a Molecule.")
```
arosen93, line 675: Can you use CHGNet for a molecule? I don't believe you can...

pymatgen/core/structure.py
JaGeo: You can also always use the potentials itself in combination with ase. Thus, I don't really see why anyone would use the pymatgen implementation to access the forces etc.

I would see it as a convenient method to clean up a structure before creating a quantum chemistry input, for example.

arosen93: I agree. Sometimes less is more.

arosen93:
@janosh: Following up the conversation with @JaGeo, could you remove `out.calc = preds` in addition to the block below:

```python
# Attach important ASE results
system.calc = atoms.calc
system.dynamics = dyn.todict()
```

Both of these are solving a problem that doesn't really exist. Thanks!

pymatgen/core/structure.py
BowenD-UCB: I agree here. The applicability domain should be consistent to M3GNet

pymatgen/core/structure.py
janosh: Got rid of this whole section entirely.

pymatgen/core/structure.py
janosh: Done. `ValueError` now raises when passing `calculator='chgnet'` on a `Molecule`.

janosh:
Thanks for the feedback and esp. the discsussion helped see things more clearly. I think we're in better shape now with the changes I just pushed.

Seems to me that there's no need to return the structure or molecule when calling `calculate` on them. They don't change during the calculation so the user already has them. Instead, we're now returning the `ase` `Calculator` instance. Feels more natural to do this:

```py
calculator = structure.calculate(calculator="chgnet")
{*calculator.results}
>>> {"stress", "energy", "free_energy", "magmoms", "forces"}
```

arosen93:
@janosh: I agree that's more natural and likely preferred. Additionally, the calculator object already stores the `Atoms` object in it via `calc.atoms` (I think), so the user could still get the structure if they want.

PR #3108, opened 2023-06-26
BUG: force contiguous calling find_points_in_spheres

lbluque:
## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -1522,10 +1522,10 @@ class IStructure(SiteCollection, MSONable):
         else:
             if sites is None:
                 sites = self.sites
-            site_coords = np.array([site.coords for site in sites], dtype=float)
-            cart_coords = np.array(self.cart_coords, dtype=float)
-            lattice_matrix = np.array(self.lattice.matrix, dtype=float)
-            pbc = np.array(self.pbc, dtype=int)
+            site_coords = np.ascontiguousarray([site.coords for site in sites], dtype=float)
+            cart_coords = np.ascontiguousarray(self.cart_coords, dtype=float)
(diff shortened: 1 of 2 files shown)


shyuep:
Quick qn: what is the purpose of this? Speed up?

lbluque:
Having the arrays contiguous makes it faster., which I had already added in the previous PR.

However, we noticed that when called internally by a structure where the lattice matrix or frac_coords had been changed in such a way (such as applying strain) that the arrays were not contiguous then the call to the cython function would raise errors.

janosh:
@shyuep Just for context, this came up in https://github.com/CederGroupHub/chgnet/pull/39 where tests were failing with

```py
chgnet/graph/converter.py:142: in get_neighbors
    center_index, neighbor_index, image, distance = structure.get_neighbor_list(
/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/site-packages/pymatgen/core/structure.py:1533: in get_neighbor_list
    center_indices, points_indices, images, distances = find_points_in_spheres(
pymatgen/optimization/neighbors.pyx:49: in pymatgen.optimization.neighbors.find_points_in_spheres
    ???
stringsource:660: in View.MemoryView.memoryview_cwrapper
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   ValueError: ndarray is not C-contiguous
```

@lbluque Thanks for the really fast fix and unit test. 👍  Makes me sleep better every time we add a new test. 😄

shyuep:
Thanks.

PR #3111, opened 2023-06-27
fix AttributeError in MoleculeGraph.draw_graph_to_file

xjf729:
In line 2602 and 2603, `g` is `MultiDiGraph` and has no attribute `node`.

## Summary

Major changes:

- fix 1: change g.node to g.nodes

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


mkhorton:
Thanks @xjf729, I believe this change was the result of an API change in networkx, good catch!

PR #3136, opened 2023-07-06
Read site labels from cif file

stefsmeets:
## Summary

Major changes:

- feature 1: Read site labels from cif file and store these under `Structure.site_properties`.

Closes #3135

## Checklist

- [x] Google format doc strings added. Check with `ruff`.
- [x] Type annotations included. Check with `mypy`.
- [x] Tests added for new features/fixes.
- [ ] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

PR #3137, opened 2023-07-10
Read site labels from cif file (2)

### LINKED ISSUE #3135: Get site labels from cif file
#### Problem

I'm interested in having access to the site labels in a cif file when reading a structure ("_atom_site_label"). I need this to classify different groups of sites with the same symbol.

#### Proposed Solution

I think these data can be made available via `Structure.site_properties`. 

In [`pymatgen.io.cif:CifBlock._get_structure`](https://github.com/materialsproject/pymatgen/blob/a5596534548abc1f29cb1d5e9cd9e995e185a4eb/pymatgen/io/cif.py#L874) the site label is already being used [to identify the sites](https://github.com/materialsproject/pymatgen/blob/a5596534548abc1f29cb1d5e9cd9e995e185a4eb/pymatgen/io/cif.py#L911).

I think it would be quite straightforward to store the label in `site_properties`. This will expand with the symmetry.

This solution would be good for my use case, without too much impact on the rest of the code base.

#### Alternatives

I think the ideal solution is if the site label would be accessible attribute under `PeriodicSite`, i.e. `PeriodicSite.label`. This would require far more significant changes to the code.

stefsmeets:
## Summary

This PR follows up from #3136 which was reverted because of CI test failures, see https://github.com/materialsproject/pymatgen/pull/3136#issuecomment-1624198814 . This PR addresses these issues.

Major changes:

- feature 1: Read site labels from cif file and store these under `Structure.site_properties`.

Closes #3135

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/core/sites.py
+++ b/pymatgen/core/sites.py
@@ -36,6 +36,7 @@ class Site(collections.abc.Hashable, MSONable):
         species: SpeciesLike | CompositionLike,
         coords: ArrayLike,
         properties: dict | None = None,
+        label: str | None = None,
         skip_checks: bool = False,
     ) -> None:
         """
@@ -52,6 +53,7 @@ class Site(collections.abc.Hashable, MSONable):
         :param coords: Cartesian coordinates of site.
         :param properties: Properties associated with the site as a dict, e.g.
             {"magmom": 5}. Defaults to None.
+        :param label: Label for the site. Defaults to None.
         :param skip_checks: Whether to ignore all the usual checks and just
             create the site. Use this if the Site is created in a controlled
             manner and speed is desired.
@@ -69,6 +71,7 @@ class Site(collections.abc.Hashable, MSONable):
         self._species: Composition = species  # type: ignore
         self.coords: np.ndarray = coords  # type: ignore
         self.properties: dict = properties or {}
+        self.label = label
 
     def __getattr__(self, attr):
         # overriding getattr doesn't play nicely with pickle, so we can't use self._properties
@@ -208,7 +211,12 @@ class Site(collections.abc.Hashable, MSONable):
         return el in self.species
 
     def __repr__(self):
-        return f"Site: {self.species_string} ({self.coords[0]:.4f}, {self.coords[1]:.4f}, {self.coords[2]:.4f})"
+        name = self.species_string
+
+        if self.label:
+            name = f"{self.label} ({name})"
+
+        return f"Site: {name} ({self.coords[0]:.4f}, {self.coords[1]:.4f}, {self.coords[2]:.4f})"
 
     def __lt__(self, other):
         """
@@ -283,6 +291,7 @@ class PeriodicSite(Site, MSONable):
         to_unit_cell: bool = False,
         coords_are_cartesian: bool = False,
         properties: dict | None = None,
+        label: str | None = None,
         skip_checks: bool = False,
     ):
         """
@@ -305,6 +314,7 @@ class PeriodicSite(Site, MSONable):
             Cartesian coordinates. Defaults to False.
         :param properties: Properties associated with the site as a dict, e.g.
             {"magmom": 5}. Defaults to None.
+        :param label: Label for the site. Defaults to None.
         :param skip_checks: Whether to ignore all the usual checks and just
             create the site. Use this if the PeriodicSite is created in a
             controlled manner and speed is desired.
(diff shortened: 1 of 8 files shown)


pymatgen/core/sites.py
```diff
@@ -467,11 +467,14 @@ def __eq__(self, other: object) -> bool:
         if not isinstance(other, Site):
             return NotImplemented
 
+        self_props = {k: v for k, v in self.properties.items() if k != "label"}
+        other_props = {k: v for k, v in other.properties.items() if k != "label"}
+
```
janosh, line 472: I don't think we can ignore the label when comparing sites. Would be quite unexpected behavior for most users:

```
from pymatgen.core.sites import Site

s1 = Site("Fe", [0.25, 0.35, 0.45], properties={"label": 'foo'})
s2 = Site("Fe", [0.25, 0.35, 0.45], properties={"label": 'bar'})
s1 == s2
>>> True
```

shyuep: I am unsure what the site labels will be used for. But I would refrain from adding it to properties. When we talk about properties, we usually mean real physical quantities. E.g., magmom. Then two sites with different magmom must be unequal. I would instead use an attribute. E.g., just site.label = "Fe1". People who care will know to use it. Most people don't really care?

mkhorton: Maybe creating an issue for a larger discussion would be useful, since I think this ultimately relates to `site_properties` not being typed, which caused a lot of downstream problems.

For example, imagine we had explicit types for possible site properties: `ScalarProperty`, `PolarVectorProperty`, `AxialVectorProperty`, etc, `Categorical`, `Comment`. Something physical like `Magmom` could be handled under this system, and also something like a label could be a `Comment` type, which could then be gracefully excluded from equality comparisons. Just a thought.

janosh: I like @shyuep's suggestion of storing label directly as a site attribute instead of in properties. Perhaps the most sensible short-term solution.

We can open an issue to track building a `properties` ontology as suggested by @mkhorton later so we don't forget.

shyuep: This is the reason why I switched Species.properties to simply Species.spin in the recent commits (with some backward compatibility code). In the old days, I was thinking that we may want to put other properties on Species. It turns out no one wanted to do that. In the end, that may be how we go. Instead of Site.properties, we simply have proper attributes for everything. So Site.magmom, Site.whatever, Site.label. We then decide which of these attributes are checked for equality.

stefsmeets: Thanks for the suggestions. I have implemented this now as `site.label` as suggested. This change turned out to be a bit more substantial than I was hoping for.

Let me know if this approach works for you. If so, I'm willing to spend some time to complete the implementation for all methods on Structure that work with the sites list.

janosh: Did you forget to push? Nothing for us to review at this point.

shyuep: I suggest you push your changes to this PR. We will look at it and give suggestions (or just simply merge). In the end, a label attribute is simply something that can be used by people who really want the site label. As I said, most users don't really care about the site labels in a CIF file. So this will be invisible to most people.

pymatgen/core/structure.py
```diff
@@ -897,13 +907,18 @@ def __init__(
             if site_properties:
                 prop = {key: val[idx] for key, val in site_properties.items()}
 
+            label = None
+            if labels:
```
janosh, line 911: This can be a ternary.
```suggestion
label = labels[idx] if labels else None
```

pymatgen/io/tests/test_cif.py
```diff
@@ -316,6 +316,17 @@ def test_implicit_hydrogen(self):
...
 
+    def test_site_labels(self):
+        parser = CifParser(f"{self.TEST_FILES_DIR}/garnet.cif")
+        struct = parser.get_structures(primitive=True)[0]
+
+        assert struct.labels
+        assert len(struct.labels) == len(struct) == 80, "Mismatch between number of labels and sites."
+        assert len(set(struct.labels)) == 4, "Expecting only 4 unique labels"
+
+        for site in struct.sites:
```
janosh, line 327:
```suggestion
        for site in struct:
```

janosh (changes requested):
@stefsmeets This looks great! Thanks for being so thorough. I think the only thing missing are some unit tests in `test_sites.py`, e.g. for `__eq__`

https://github.com/materialsproject/pymatgen/blob/dd953dbe2c163607342dcf88f00ab5d1aaaa3660/pymatgen/core/tests/test_sites.py#L157-L160

stefsmeets:
Thanks for the feedback! I added a bunch of tests for the new code. Let me know if there is anything else.

janosh (approved):
Damn son, tests are top notch. 💯  Thanks a lot! 👍

Issue #3142, opened 2023-07-11
`MaterialsProjectDFTMixingScheme.process_entries()` should have a kwarg to avoid POTCAR checks


arosen93:
When using `MaterialsProjectDFTMixingScheme.process_entries()`, if the user does not provide `ComputedStructureEntry` objects with POTCAR symbols in `.parameters`, the mixing scheme will crash. While it's understandable to do a validity check by default, there should be a kwarg to disable this.

If a user is trying to use the `MaterialsProjectDFTMixingScheme` on calculations they ran manually (i.e. without Atomate), the procedure is something like:

- Run VASP calculation with MP-compatible settings
- Make a `Structure` and get the energy from the outputs
- From that, make a `ComputedStructureEntry`
- Run the `ComputedStructureEntry` objects through `MaterialsProjectDFTMixingScheme` alongside `ComputedStructureEntry` objects pulled from MP for that chemical system.

By default, this `ComputedStructureEntry` does not have (nor does it require) any information about the POTCARs used. When passing these `ComputedStructureEntry` objects to `MaterialsProjectDFTMixingScheme`, one would naively expect the mixing scheme to "just work," but instead it raises an error because the `ComputedStructureEntry` doesn't have the properties needed for a validity check. Pymatgen shouldn't expect the user to know to parse the POTCAR hash from the `vasprun.xml` or `POTCAR` and add it to the `properties` of the `ComputedStructureEntry`, especially when it's only needed for a validity check. Alternatively, there should be clearer documentation at the least.
 
On a related note, there should be some additional documentation that the user should be specifying `.parameters["run_type"]` and the related Hubbard parameters when constructing their `ComputedStructureEntry` objects. The way the mixing scheme currently works, it kind of assumes the user is familiar with the Atomate/MP entry format. This is very closely tied to #2500.

 Edit: I realized this is basically the same issue as #2499 from a while ago. I'll keep this one open since it has a bit more context.
 
----------------------------------------------------------------------------
```python
      1 scheme = MaterialsProjectDFTMixingScheme()
----> 2 corrected_entries = scheme.process_entries(entries)
      3 rosen_corrected_entries = [
      4     entry for entry in corrected_entries if "rosen" in entry.entry_id
      5 ]

[...]

PR #3143, opened 2023-07-11
Add `check_potcar: bool = True` to  `MaterialsProject2020Compatibility`, `MaterialsProjectDFTMixingScheme` and `PotcarCorrection`

### LINKED ISSUE #2499: Add a kwarg to avoid POTCAR checking in correction scheme
This mainly concerns the `MaterialsProject2020Compatibility` scheme, so I'm CC'ing @rkingsbury here.

https://github.com/materialsproject/pymatgen/blob/dcc4576fb8fb89b7adc2f31158654392348e6498/pymatgen/entries/compatibility.py#L941-L944

The above codeblock checks the POTCAR symbols when getting a correction. While this is a very useful tool, I don't know that it should be _required_ to obtain a correction. For instance, my entry objects don't have any data about POTCARs in them, but I can't get the corrections without commenting out this codeblock.

@rkingsbury: Are you opposed to me adding a kwarg `check_potcar` to `get_adjustments()` or to `__init__` to optionally bypass this check? I feel like the difference between this and `check_potcar_hash=False` is potentially confusing though. In the case of `check_potcar_hash`, the POTCARs are still check but just their symbols and not the hash itself. Any suggestions?

For additional context, see #3142.

### LINKED ISSUE #2952: MP2020Compability
**Issue**: [MP2020Compatibility](https://github.com/materialsproject/pymatgen/blob/83a87bb5039414aeb68f25162b5c1403478028d8/pymatgen/entries/compatibility.py#L821-L1097) (and all other compatibility modules) currently requires a ComputedEntry object containing VASP settings and POTCAR information to operate. This makes it inconvenient to apply corrections when only the structure and energy are available. This issue is becoming more common when dealing with large datasets (e.g., directly downloading structures from MP) and machine learning methods like M3GNet.

**Example**: M3GNet is trained on MPRelaxSet entries, but it's challenging to apply MP2020 compatibility to M3GNet predicted energies since there are no VASP settings available.

**Proposed Solution**: Add an extra key under the compatibility modules (or ComputedEntry) to simplify the process. For instance, `VASPsettings = "MPRelaxSet"` could be added. With this key, energy correction can be applied without explicitly including all VASP settings and creating a ComputedEntry.

janosh:
Closes #2499, closes #2952 (and #3142).

07f9730c2 add check_potcar: bool = True to PotcarCorrection
c0087761e add check_potcar: bool = True to MaterialsProject2020Compatibility
19f88c07e add check_potcar: bool = True to MaterialsProjectDFTMixingScheme
0209a5ab1 fix broken test due to modified error msg


arosen93:
You are the hero we don't deserve.

arosen93:
Closes #2952 as well?

shyuep:
Pls do not merge this. I have a different fix.

janosh:
Maybe, not quite sure since there was talk of a `ComputedEntry.from_settings` method in that issue. Not sure if anyone still wants that if you can set `MaterialsProject2020Compatibility(check_potcar=False)`

shyuep:
I just pushed a fix. A global .pmgrc.yaml setting called DISABLE_POTCAR_CHECKS would skip all potcar checking in compatibility.

janosh:
I don't think this needs to be closed. Having the option to disable `check_potcar` on a case-by-case basis is still useful. Plus not every user knows about `.pmgrc.yaml`.

BowenD-UCB:
Thanks a lot for implementing this!

arosen93:
I agree that there is still value for a kwarg.

janosh:
@shyuep Forgot to mention this morning I think it'd be better to call the `.pmgrc` flag `POTCAR_CHECKS: true | false` rather than `DISABLE_POTCAR_CHECKS: false | true`. Easier to read:

```diff
- if SETTINGS.get("DISABLE_POTCAR_CHECKS", False):
+ if not SETTINGS.get("POTCAR_CHECKS"):
    return ufloat(0.0, 0.0)

**Comment by shyuep** [comment]
Sure. Go ahead and make the change, including the new documentation on global config variables that I added this morning. I don't have a strong opinion on this.

**Comment by janosh** [comment]
Will do. Just added tests for `check_potcar` in `MaterialsProjectDFTMixingScheme` and `MaterialsProjectCompatibility2020`. I think this is ready to go.

**Comment by janosh** [comment]
@shyuep One more thing (bringing this up just to make sure I'm not missing something). It seems you can't actually set booleans (or any type other than `str` like `floats`, `ints`, `None` for that matter) using the `pmg` CLI atm.

```sh
pmg config --add PMG_POTCAR_CHECKS False
```

will be stored as

```
PMG_POTCAR_CHECKS: 'False'
```

which is truthy and so the opposite of expected behavior.

PR #3146, opened 2023-07-11
read `mag` from OSZICAR

chiang-yuan:
I tried to read `mag` from OSZICAR and found `mag` is not read from `ionic_MD_pattern`. Not sure if `mag` is available in all vasp md calculation. 

@janosh should I add unit test for this? The OSZICAR in `test_files/OSZICAR` doesn't have all the available output fields like

```
2 T=  1239. E= 0.11712678E+03 F= -.20873815E+03 E0= -.20873814E+03  EK= 0.10092E+02 SP= 0.32E+03 SK= 0.10E-02 mag=     -0.000
```

## Checklist


janosh:
@chiang-yuan Yes, we definitely need a test for this.

Do I understand correctly that your PR adds a `mag` key to any parsed `OSZICAR` even if that data is not present in the file? Will the value be `None` in that case?

chiang-yuan:
Haven't tested different OSZICAR yet. Looks like OSZICAR test in test_ouptut.py doesn't test different combinations of available field. I am going to add a test considering if the field is missing. https://github.com/materialsproject/pymatgen/blob/1b04a0a626dae8ae42324ac1d99d151591519c6e/pymatgen/io/vasp/tests/test_outputs.py#L1422-L1428

chiang-yuan:
Hi @janosh I think this PR is ready to be merged. Let me know if I can do anything to make it better!

pymatgen/io/vasp/tests/test_outputs.py
```diff
@@ -1420,12 +1420,26 @@ def test_get_band_structure(self):
...
+        assert len(self.oszicar_mag.electronic_steps) == len(self.oszicar_mag.ionic_steps)
+        assert len(self.oszicar_mag.all_energies) == 60
+        assert self.oszicar_mag.final_energy == approx(-526.63928)
+
+        assert len(self.oszicar_md.electronic_steps) == len(self.oszicar_md.ionic_steps)
+        assert len(self.oszicar_md.all_energies) == 3
+        assert self.oszicar_md.final_energy == approx(-209.86273)
+
+    def test_fields(self):
+        assert set(self.oszicar_mag.ionic_steps[-1].keys()) == {"F", "E0", "dE", "mag"}
```
janosh, line 1441: We don't need the `.keys()` here.

test_files/OSZICAR_MD
```diff
@@ -0,0 +1,45 @@
+       N       E                     dE             d eps       ncg     rms          rms(c)
```
janosh, line 1: Do we actually need the new test file? Looks like `test_files/OSZICAR` also has a `mag` column.

pymatgen/io/vasp/outputs.py
```diff
@@ -3913,22 +3914,7 @@ def __init__(self, filename):
...
-        ionic_MD_pattern = re.compile(
-            r"(\d+)\s+T=\s*([\d\-\.E\+]+)\s+"
-            r"E=\s*([\d\-\.E\+]+)\s+"
-            r"F=\s*([\d\-\.E\+]+)\s+"
-            r"E0=\s*([\d\-\.E\+]+)\s+"
-            r"EK=\s*([\d\-\.E\+]+)\s+"
-            r"SP=\s*([\d\-\.E\+]+)\s+"
-            r"SK=\s*([\d\-\.E\+]+)"
-        )
+        ionic_general_pattern = re.compile(r"(\w+)=\s*(\S+)")
```
janosh, line 3917: I like the improved readability of this new regex but it's also less specific. Just wondering if there can be any parsing edge cases from not matching column headers like `E0` exactly. Are there any unit tests we could here for added safety?

janosh (changes requested):

test_files/OSZICAR_MD
chiang-yuan: That's the beauty of this new parsing method! Actually we don't, but the old test doesn't consider other fields like `SP`, `SK` from MD simulation. I vote for removing one of these test files.

janosh: >  I vote for removing one of these test files.

Sure, go right ahead. Fewer files is better.

pymatgen/io/vasp/outputs.py
chiang-yuan: I think \w+ matches one or more word characters (same as [a-zA-Z0-9_]+ ) so it should not be a problem? I haven't seen any other strange OSZICAR pattern before (based on my novice experience of vasp 😂)

pymatgen/io/vasp/tests/test_outputs.py
```diff
@@ -1420,12 +1420,18 @@ def test_get_band_structure(self):
 
 
 class OszicarTest(PymatgenTest):
+    @classmethod
+    def setUpClass(cls):
+        filepath = f"{cls.TEST_FILES_DIR}/OSZICAR"
+        cls.oszicar_mag = Oszicar(filepath)
```
janosh, line 1426: We have two `OSZICAR` test files that we could loop over to make maximum use of them. The other is at `test_files/static_silicon/OSZICAR`.

janosh (changes requested):

pymatgen/io/vasp/tests/test_outputs.py
chiang-yuan: Added! 😃

janosh: Thanks! 👍

janosh (approved):

Issue #3148, opened 2023-07-12
Incorrect formula for `estimate_nbands` method


matthewkuner:
https://github.com/materialsproject/pymatgen/blob/71864db310569946703bfa8c3f42eaf69c533744/pymatgen/io/vasp/sets.py#L689

According to my version of the VASP source code, in `main.F`, the actual formula for NBANDS should be:

`max((valid_nelect+2)/2 + max(nions/2,3),   valid_nelect*0.6)) + nmag`
with `nmag=0` always set for ISPIN=1 calcs 

Also, `nmag` should be calculated as:
```
nmag = len([m for m in self.incar["MAGMOM"] if not np.allclose(m, 0)])
nmag =(nmag+1)/2    # this step is currently excluded from the code
```

It can later be increased (to be divisible by the number of cores, etc.), but that can't be accounted for without additional context for a given calc.

mkhorton:
This might well be true, but I expect this was written with reference to the [VASP docs](https://www.vasp.at/wiki/index.php/NBANDS). I'm not sure the appropriateness of referring to the source code here.

matthewkuner:
@mkhorton what do you mean? I thought that using the formula from the source code (which is distinct from the docs) would be desirable, as it would closer to the actual NBANDS used by VASP.

mkhorton:
VASP is not an open source code (meaning, a code released under an open source license).

I'm not a lawyer, but generally it's not ok to reference a closed-source code to re-implement a feature into an open-source code, since it violates the terms of the original license.

I don't know how much of that is context dependent; this is something trivial, and is essentially a single line to implement an alternative API to enable use of the original code (anyone using pymatgen for VASP would also need to hold a license for VASP themselves), but nevertheless ...

matthewkuner:
Closed by #3149

PR #3149, opened 2023-07-12
fix `estimate_nbands` function

matthewkuner:
Fixes the `estimate_nbands` function. See #3148 for more context.


janosh:
@matthewkuner Thanks for spotting this!

Would be great if @aaron-kaplan or someone else more qualified than me could sign off on this.

Also, @matthewkuner we need a unit test in `pymatgen/io/vasp/tests/test_sets.py` for this change in behavior. It should pass with and fail without this change. Maybe in `test_nelect`.

https://github.com/materialsproject/pymatgen/blob/71864db310569946703bfa8c3f42eaf69c533744/pymatgen/io/vasp/tests/test_sets.py#L233

aaron-kaplan:
> Would be great if @aaron-kaplan or someone else more qualified than me could sign off on this.

You tagged the wrong person. I'm not involved with this project.

janosh:
@aaron-kaplan Apologies.

I meant @esoteric-ephemera.

esoteric-ephemera:
@matthewkuner, lines 694-695 seem wrong to me per our discussion:
>  nmag = len([m for m in self.incar["MAGMOM"] if not np.allclose(m, 0)])
> nmag = np.ceil((nmag+1)/2)

The line of code in VASP's main.F is:
> NMAG=SUM(T_INFO%ATOMOM(1:T_INFO%NIONS))
> NMAG = (NMAG+1)/2
which to me looks like VASP is simply summing the initial magnetic moments (ATOMOM), not counting the number of atoms with nonzero magnetic moments

The weird thing is that NMAG is an integer but ATOMOM is declared to be a float in poscar_struc.F. Fortran seems to do int + float = int + floor(float) when assigning to a variable. You probably want to double check your code against a few sets of INCAR's and OUTCAR's.

Also, @matthewkuner, if the user specifies NPAR (if 'NPAR' is in self.incar), you probably want to add a line as in main.F:
> NBANDS=((NBANDS+NPAR-1)/NPAR)*NPAR

@janosh, my name is unfortunately too common to get it as a username 😂

arosen93:
@matthewkuner: Copying over our brief convo from Slack, it would perhaps be nice (if you're willing to do so) to move the `estimate_nbands` definition out of the `DictSet` class. It would be nice to be able to call `estimate_nbands` for a given `Structure` without instantiating a `DictSet`, e.g. to call it from Custodian during error checking.

matthewkuner:
Please see comments in [#3148](https://github.com/materialsproject/pymatgen/issues/3148#issuecomment-1635345159)  by @mkhorton

janosh:
The license concerns? Perhaps we can loop in someone from VASP and get their ok. Haven't looked into this myself, but taking @matthewkuner's word for it, I doubt VASP minds us fixing downstream repercussions arising from disparities between their docs and their actual implementation. Given it's a paid product, this should not have happened in the 1st place.

arosen93:
Perhaps @MichaelWolloch could help clarify or get us in touch with folks who can.

mkhorton:
Yes, agreed--just asking the VASP devs is the best course of action here. I also agree hopefully it's a non-issue, but good to check.

matthewkuner:
@janosh I've updated the tests and verified them to quick VASP calcs that I ran.

janosh:
> Also, @matthewkuner, if the user specifies NPAR (if 'NPAR' is in self.incar), you probably want to add a line as in `main.F`:
> `NBANDS=((NBANDS+NPAR-1)/NPAR)*NPAR`

Thanks @matthewkuner! Did you address this comment raised by @esoteric-ephemera?

matthewkuner:
@janosh i addressed everything from Aaron’s comment except for the NPAR stuff (reason being that many people won’t edit the MP input set directly). I could add it anyway I guess, tho I question how often it would be used.

mkhorton:
I’m assuming since work is proceeding that VASP devs ok’d using `main.F` as a reference? Would it be possible to include a copy of the response here for the record?

janosh:
No response yet from VASP. Not sure who to ping. @sudarshanv01 Do you have any suggestions?

matthewkuner:
@janosh @mkhorton I asked a similar question about NGX/Y/Z, and they basically just said that as long as it isn't a direct copy-paste, it should be fine. See: https://www.vasp.at/forum/viewtopic.php?p=24351#p24351
Let me know if you think my interpretation is too biased/generous to our efforts.

mkhorton:
I think that’s reasonable. I just wanted to make sure there was a record. Perhaps this thread can be linked on that forum post as well just to be completely above board and leave it there if there are no objections.

janosh (approved):
Thanks @matthewkuner! Great catch. 👍

MichaelWolloch:
Hi, and apologies for the late reply. I am on holidays this week and did not check my emails for a while. I have never dealt with this type of code base related issue, and am hesitant to comment either way. Maybe @henriquemiranda is in-office and knows more about VASPs policy here? Or can quickly ask someone more senior?
Cheers, M

shyuep:
This code has no tests whatsoever. Can someone pls add a unittest for it? @janosh Pls make sure that this is always done for any new code before merging.

matthewkuner:
@shyuep there are indeed new tests for this PR, check the second file changed

shyuep:
@matthewkuner Ah yes. However, it says skip if no PSP directory. That basically means it is not running at all on Github Actions. Can the test be modified to at change check a few POTCAR files? Since the code is not running on the CI, we can't detect any bugs until someone runs it on their computer.

PR #3151, opened 2023-07-12
`AseAtomsAdaptor`: Retain `tags` property when interconverting `Atoms` and `Structure`/`Molecule`

arosen93:
Currently, the `.tags` property of the ASE `Atoms` object is lost when interconverting between ASE `Atoms` and Pymatgen `Structure`/`Molecule`. I adopted the solution by @brookwander in https://github.com/ulissigroup/catapalt/issues/2 to ensure the tags do not get lost.


brookwander:
thanks @arosen93 !

arosen93:
@janosh --- why does ruff hate me? :(

janosh:
Looks like `ruff` just had a new release that finds new issues on `master`. So nothing to do with your changes. Does this still need testing?

arosen93:
Ah I see. No, this should be done!

PR #3172, opened 2023-07-21
Fix a bug in pwscf.py. The proc_val function modifies string values.

pablogalaviz:
## Summary

The proc_val function modifies the value of string variables (changes 'd' for 'e'). For example, smearing = 'cold' becomes smearing = 'cole'. It is solved by using a local variable.

Major changes:

Added  a local variable
```
    val_replace = val.replace("d", "e")
    return smart_int_or_float(val_replace)
```
## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/io/pwscf.py
+++ b/pymatgen/io/pwscf.py
@@ -371,7 +371,7 @@ class PWInput:
             "conv_thr",
             "Hubbard_U",
             "Hubbard_J0",
-            "defauss",
+            "degauss",
             "starting_magnetization",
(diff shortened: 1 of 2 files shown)


janosh (changes requested):
Thanks @pablogalaviz! We need a new test for this that fails without this fix.

janosh (approved):

janosh:
Nice test, thanks @pablogalaviz! 👍

PR #3180, opened 2023-07-24
Prefer `pymatviz` interactive plotly version of periodic table heatmap if available

janosh:
This PR tries to `from pymatviz import ptable_heatmap_plotly` in `periodic_table_heatmap()`. If successful, all arguments are passed to that function along with deprecation messages for args that have different names in `pymatviz.ptable_heatmap_plotly` vs `pymatgen.util.plotting.periodic_table_heatmap`.

It explicitly doesn't add `pymatviz` as a new dependency of `pymatgen` to avoid circularity.

**Before**

pmg-ptable-heatmap

**After**
ptable-heatmap-plotly-more-hover-data

**Code**

```py
import pandas as pd
from pymatgen.util.plotting import periodic_table_heatmap

df_ptable = pd.read_csv(
    "https://github.com/janosh/pymatviz/raw/main/pymatviz/elements.csv", comment="#"
).set_index("symbol")

periodic_table_heatmap(df_ptable.atomic_mass.to_dict(), cmap="viridis")
```

### DIFF
--- a/pymatgen/symmetry/kpath.py
+++ b/pymatgen/symmetry/kpath.py
@@ -1824,7 +1824,7 @@ class KPathLatimerMunro(KPathBase):
 
         g = np.dot(W.T, W)  # just using change of basis matrix rather than
         # Lattice.get_cartesian_coordinates for conciseness
-        ginv = np.linalg.inv(g)
+        g_inv = np.linalg.inv(g)
         D = np.linalg.det(W)
 
         primary_orientation = secondary_orientation = tertiary_orientation = None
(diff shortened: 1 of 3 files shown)


janosh:
@mkhorton That's the long-term plan. This is meant as the 1st step in a deprecation/migration phase. But happy to discuss other options.

shyuep:
I would just note this - there will always be newer better packages coming up. There are two outcomes: 1. pymatgen depends on the new package for some functionality. or 2. that new package gets merged into pymatgen. 

For (1) to be accepted by me, the package has to deliver major functionality, have reasonable long-term maintenance, and does not depend on pymatgen itself. Something like numpy satisfies all three criteria. mp-api fails the last criterion. 

Right now, pymatviz fails at least one, if not all three criteria (it depends on pymatgen, and while I have confidence in Janosh's coding, pymatviz is not part of some long-term funded infrastructure). 

If (1) or (2) are not acceptable to either pymatgen maintainers or the author of the package in question, then the only other outcome is that it is an optional functionality that is used when present. That is what this PR is.

janosh:
@mkhorton Should I leave this PR open until further brainstorming?

shyuep:
You can merge.

Issue #2994, opened 2023-05-17
Compress all test files


janosh:
See https://github.com/materialsproject/foundation/pull/3#issuecomment-1544290657 and surrounding discussion for motivation.

rkingsbury:
I'm in favor of this! If someone is going to try this, I would also suggest doing some light reorganization of the test files so that it's clearer which ones go with which parts of the code. 

For example, move a lot of the test files in the root directory that are associated with VASP IO into a `vasp/` subdirectory, etc. Also I believe there are some old Q-Chem test files under `qchem/` but the more relevant ones are under `molecules/new_qchem_files`. Perhaps these can be consolidated.

chiang-yuan:
In PR #3146, I added a new `OSZCIAR_MD` to the `test_files`. @janosh you may want to include that into the compiled test files? But since now `Oszicar` can read fields dynamically in principle we can shrink two OSZICAR test files into one? I feel like there are many other duplicate test files can be shrunk into one too

janosh:
> I feel like there are many other duplicate test files can be shrunk into one too

I'm sure there are. Would be some effort to identify them but PRs welcome!

DanielYang59:
Can we close this now or there is more test files to be compressed or relocated I could help with? @janosh 

I guess it would be good to collect all `cif` (not sure about `json`) files together and I might do this in the near future 😃 

What about the `compress-test-files` branch?

janosh:
i think there are a lot of uncompressed output files left. try running `compress_test_files.py` from https://github.com/materialsproject/pymatgen/pull/3132 and it should show which ones i mean.

`compress-test-files` branch is super outdated but might help to keep it around for reference until we're all done

PR #3192, opened 2023-07-29
speed up `bader_caller` and `chargemol_caller`

### LINKED ISSUE #2487: Bader analysis via Pymatgen is extremely slow
**Describe the bug**
When using the `bader_caller` in Pymatgen (specifically `bader_analysis_from_path`), it can take many times longer to run a Bader analysis than just using the standalone scripts provided by the Henkelman group. I suspect this is related to excessive file I/O related to reading in the CHGCAR, AECCAR0, and AECCAR2 every time and also writing out the CHGCAR again every time (in addition to the summed AECCARs, but that is necessary). I've run Bader analyses on systems with hundreds of atoms, and it never takes more than a minute or so, but with Pymatgen the same analysis can take tens of minutes in comparison.

We should track down the bottleneck here and then figure out how to fix this. I think having an option to do the following would be ideal:
- Copy the AECCAR0, AECCAR2, CHGCAR, and POTCAR files to scratch (is this even necessary? I don't think `bader` will overwrite anything).
- Use `chgsum.pl AECCAR0 AECCAR2.pl` directly. Don't read in the AECCARs via Pymatgen. This would require `chgsum.pl` to be in the PATH in addition to just the `bader` executable.
- Run `bader CHGCAR -ref CHGCAR_sum`. Repeat this step but yield the spin densities if the calculation is spin-polarized.

At no point in the above should the CHGCAR be read or written, in contrast with the current approach.

Tagging @mkhorton since we briefly discussed this earlier today.

chiang-yuan:
In the original implementation, CHGCAR is copied into temporary folder by `monty.zopen`. Since CHGCAR is usually very large, this makes bader analysis calling from pymatgen extremely slow. 

this might close #2487 

- fix 1: use `monty.tempfiie.ScratchdDr` context manager
- fix 2: do not copy large files like CHGCAR into temporary folder but read them in place instead
- fix 3: use `os.symlink` to link larges files for `chargemol_caller`

## Checklist


arosen93:
Thank you so much for doing this! It has been on my backlog for a very long time.

arosen93:
You don't need to add it in this PR, but it's likely that an analogous fix will solve similar issues with the `chargemol_caller`, shown here: https://github.com/materialsproject/pymatgen/blob/master/pymatgen/command_line/chargemol_caller.py

chiang-yuan:
@arosen93 thanks for pointing out chagemol as well! It actually requires little modification so I guess I can try to solve them both.

arosen93:
@chiang-yuan: I would certainly be very grateful if you did!!! 🙏

chiang-yuan:
Chargemol needs to see all the input files in the same folder so the way around here is using symbolic link to point to large files . I also port the error message by bader and chagmol back to python RuntimeError.

I have used `pytest` to test `test_bader_caller.py` and `test_chargemol_caller.py` locally on Perlmutter and they all passed. If this change cannot pass pytest on github workflow, I don't know where is the problem then... Perhaps somehow bader and chagemol are not correctly installed on the pytest environment.

chiang-yuan:
Looks like `CHGCAR` passed but `cube` failed. Maybe @janosh @arosen93 you will be interested in looking into this?

https://github.com/materialsproject/pymatgen/actions/runs/5704593631/job/15458352230?pr=3192#step:7:183

chiang-yuan:
I found the problem. The code was reading compressed test files but somehow didn't raise error on Perlmutter. Modify the code to handle compressed files and decompress them to temp directory if needed. Not sure if this is the ideal solution since this falls back to copying data to temp folder again (but only for compressed file and folder, the uncompressed files are still read at their original place). 

This will also affect when we want to compress all the test files. #2994

pymatgen/command_line/bader_caller.py
```diff
@@ -15,19 +15,19 @@
...
-import shutil
 import subprocess
 import warnings
 from glob import glob
 from shutil import which
 from tempfile import TemporaryDirectory
 
 import numpy as np
-from monty.io import zopen
+from monty.tempfile import ScratchDir
```
janosh, line 25: I actually don't understand why the files need to be copied at all? Does `bader` modify them in place? Seems unlikely and even if, they should provide an option to disable that?

pymatgen/command_line/chargemol_caller.py
```diff
@@ -179,14 +177,13 @@ def _execute_chargemol(self, **jobcontrol_kwargs):
...
-                shutil.copyfileobj(f_in, f_out)
-            with zopen(self._aeccar2path, "rt") as f_in, open("AECCAR2", "w") as f_out:
-                shutil.copyfileobj(f_in, f_out)
+            try:
+                os.symlink(self._chgcarpath, "./CHGCAR")
+                os.symlink(self._potcarpath, "./POTCAR")
+                os.symlink(self._aeccar0path, "./AECCAR0")
+                os.symlink(self._aeccar2path, "./AECCAR2")
+            except OSError as e:
+                print(f"Error creating symbolic link: {e}")
```
janosh, line 186: Using `os.symlink` here seems strange. Why not read the files from where they are?

pymatgen/command_line/tests/test_bader_caller.py
```diff
@@ -78,6 +78,8 @@ def test_from_path(self):
         assert np.allclose(charge, charge0)
         if os.path.exists("CHGREF"):
             os.remove("CHGREF")
+        if os.path.exists(os.path.join(test_dir, "CHGREF")):
+            os.remove(os.path.join(test_dir, "CHGREF"))
```
janosh, line 82: Would probably be best to use the `PymatgenTest.pmt_path` fixture so we don't have to worry about manually cleaning up test-generated files.

pymatgen/util/io_utils.py
```diff
@@ -17,6 +17,29 @@
 __date__ = "Sep 23, 2011"
 
 
+def decompress_file_to_path(fin_path, fout_path=None):
```
janosh, line 20: Isn't this quite similar to `monty.io.zopen`?

janosh (changes requested):
@chiang-yuan Thanks for taking this on! Very excited about fully reviving the `BaderAnalysis` in `pymatgen`. Left a few questions.

pymatgen/command_line/bader_caller.py
chiang-yuan: They don't! So what I did here is creating a temporary scratchdir to store the output files from bader. However, bader cannot read compressed file so at the bottom we need to decompress file if it is compressed and store it temporarily somewhere (I just put it in ScratchDir as well)

pymatgen/command_line/chargemol_caller.py
chiang-yuan: `chargemol` is different. It seems like here is no input arguments to indicate where the input files (e.g. CHGCAR AECCAR0) stored, so we need to all the input file in ScratchDir. To avoid copying, just use symbolic link can "pretend" there are in the same directory but actually they are just the shortcut linked to the file in the original place.

pymatgen/command_line/tests/test_bader_caller.py
chiang-yuan: Is `PymatgenTest.tmp_path` already implemented? If yes, I need some time to look into how to refactor this part. If no, I need more time 😂

pymatgen/util/io_utils.py
chiang-yuan: Actually is more similar to `decompress_file`  in `monty` but the function won't return the path and delete the original compressed file. Created a helper function there as a fix

janosh: I assume @shyuep wouldn't mind modifying `monty.shutil.decompress_file`, esp. since the doc string is currently wrong: arg `compression` doesn't actually exist.

https://github.com/materialsvirtuallab/monty/blob/2b91561d2483f2bebfc185c4945dfae85008c439/monty/shutil.py#L98

We can make a PR to monty to have the function return the path.

pymatgen/command_line/tests/test_bader_caller.py
janosh: Yes, right here

https://github.com/materialsproject/pymatgen/blob/71b7aec2b6199b8f5b00464ee98e6a30cee441e1/pymatgen/util/testing.py#L50-L54

chiang-yuan: Thanks I am looking into it!

janosh:
Sorry, looks like this was waiting on me. I missed that. Just fixed the merge conflict. I'll submit the PR to `monty` to return `filepath` from `decompress_file`.

@chiang-yuan Any additional tests we might want to add?

pymatgen/util/io_utils.py
janosh: @chiang-yuan Let's delete this function and use the changes in https://github.com/materialsvirtuallab/monty/pull/536 instead.

@shyuep Can we have a `monty` release whenever we do the next `pymatgen` release?

janosh (changes requested):

chiang-yuan:
@janosh I made two modifications, 1) deleted ` decompress_file_to_path` and use `monty` instead 2) use `PymatgenTest.tmp_path` fixture. 

The `monty` version of the test is still older version so the test is unsuccessful.

janosh:
Thanks @chiang-yuan! 👍 this is a big improvement. required a lot of tweaking but i'm glad we got it merged.

mkhorton:
Also offering thanks for this :)

PR #3193, opened 2023-07-29
[WIP] Add I/O interface with Metalwalls

chiang-yuan:
> MetalWalls (MW) is a molecular dynamics code dedicated to the modelling of electrochemical systems. Its main originality is the inclusion of a series of methods allowing to apply a constant potential within the electrode materials.

This is the same PR but without dirty commit from other branch. The input classes are mostly finished.

- [ ] Need to add string representation for classes and file export method to print out `runtime.inpt` file
- [ ] Data input class. Should be convenient to convert `Structure` to metalwalls `data.inpt` format
- [ ] Output classes for post-processing
- [ ] Add tests

## Checklist

### DIFF
--- a/pymatgen/io/mw/__init__.py
+++ b/pymatgen/io/mw/__init__.py
@@ -1,0 +1,5 @@
+from __future__ import annotations
+
+from .inputs import MWInput
+
+__all__ = ["MWInput"]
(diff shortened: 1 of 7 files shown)


pymatgen/io/mw/__init__.py
janosh, line 1: I think better to be explicit here and write out the name: `pymatgen/io/metalwalls/__init__.py`

chiang-yuan: Thanks @janosh! Will change that when I get back to this. It has been put off in my TODOs and is going to stale for a while.

shyuep:

PR #3179, opened 2023-07-23
BUG: fix setting zero magmoms

lbluque:
## Summary

With the new changes that set the default `Sepcies.spin` to `None`, I fixed the logic for setting zero magmoms in `SpacegroupAnalyzer` to check for `None` values as well.

## Checklist

- [X] Google format doc strings added. Check with `ruff`.
- [X] Type annotations included. Check with `mypy`.
- [X] Tests added for new features/fixes.
- [X] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

PR #3203, opened 2023-08-02
[BUG] Allow None for Magmom values

jmmshn:
A possible follow-up to #3179 

Recent changes resulted in some of the structures in my atomate2 workflow that had `[None, ...]` for all the `Magmom` values.  I'm still looking into where these `[None, ...]` values came from (not 100% sure it's related to #3179) but this fix:
- Skips writing Magmom completely if all values are `None`
- Change `None` -> 0 if only some of the values are `None`


shyuep:
Pls add a test for this.

jmmshn:
Actually, I was able to find the bug in atomate2 that caused this.
https://github.com/materialsproject/atomate2/pull/460

Issue #3068, opened 2023-06-14
Reading and writing POTCAR files is broken


henriquemiranda:
#### Description

Reading and writing POTCAR file with the Potcar class from pymatgen.io.vasp seems broken.

#### Repro

To reproduce the problem, I start with a file:

POTCAR
```
 PAW Sb 04Feb1998
  5.00000000000000
 [...]
  0.000000000000E+00  0.000000000000E+00  0.000000000000E+00
 End of Dataset
  PAW Te 03Oct2001
   6.00000000000000
 [...]
  0.000000000000E+00  0.000000000000E+00  0.000000000000E+00 0.000000000000E+00  0.000000000000E+00
 End of Dataset
```

I use the following script to copy the POTCAR:
```
from pymatgen.io.vasp import Potcar
potcar = Potcar.from_file('POTCAR')
potcar.write_file('POTCAR_NEW')
```

the resulting file becomes:
POTCAR_NEW
```
 PAW Sb 04Feb1998
  5.00000000000000
 [...]
  0.000000000000E+00  0.000000000000E+00  0.000000000000E+00End of Dataset
  PAW Te 03Oct2001
   6.00000000000000
 [...]
  0.000000000000E+00  0.000000000000E+00  0.000000000000E+00 0.000000000000E+00  0.000000000000E+00End of Dataset
```

#### Expected behavior

I would expect the files in POTCAR_NEW to be the same as the ones in POTCAR

#### Problem

The problem originated after #2910 was merged.
Perhaps @jmmshn and @janosh have some insight.
From the discussion, it looks like the relevant tests were broken by these changes and then deactivated.

I also mention @MichaelWolloch since he knows about the handling of the hashes of the POTCAR files.

janosh:
Just to confirm, the problem is `"End of Dataset"` not being on a new line?

henriquemiranda:
Yes, exactly

jmmshn:
Thanks for reporting this!
I think the last fix on parsing fixed one white space problem and created another. I'm currently on family leave until the end of next week I can fix it when I come back.

MichaelWolloch:
I think the problem is in reading the POTCAR from file. The `.stip()` method is used when constructing the PotcarSingle objects when it should not be used.

This should work (in the Potcar class):
~~~
    @staticmethod
    def from_file(filename: str):
        """
        Reads Potcar from file.

        :param filename: Filename
        :return: Potcar
        """
[...]

PR #3204, opened 2023-08-02
Bugfix for Vasprun reading with POTCAR for chemical shift calculations

MichaelWolloch:
## Summary

Major changes:

- Fixes a bug updating the charges from POTCAR while reading a vasprun.xml file in the case of a chemical shift calculation
- Slight update to the splitting of POTCARs when reading from file.
- Added a test for the chemical shift vasprun.xml parsing with a POTCAR
- Re-added a test for internal POTCAR hashes.

This work is a result of updating atomate to work with the newest pymatgen and custodian releases and figuring out a hard to spot bug that was present there since May. Please see https://github.com/hackingmaterials/atomate/pull/778#issuecomment-1662336368 for details.

Another discussion related to this was taking place in  #3068. There the hashing system was (hotly) debated and the initial problem was fixed (afaik). I still think that this PR has a nicer version, which is easy to read and preserves the internal hashes. This is why I re-enabled that unit-test. However, I am fine with commenting that out again or use the new `PMG_POTCAR_CHECKS` environmental variable to make it conditional?


shyuep:

PR #3237, opened 2023-08-13
Breaking: Have plot methods return `plt.Axes` object, not `matplotlib` module

### LINKED ISSUE #3138: returning matplotlib.pyplot module in plotting methods
#### Description

This may be more a clarification than a bug necessarily.

I have noticed that many of the plotting methods using the `matplotlib` backend return the `matplotlib.pyplot` module.
For example in the [phase diagram plotter](https://github.com/materialsproject/pymatgen/blob/9f0c7c28fee6681058dadbaafd2b2b52a0f9e251/pymatgen/analysis/phase_diagram.py#L3719) or this general [pretty_plot](https://github.com/materialsproject/pymatgen/blob/9f0c7c28fee6681058dadbaafd2b2b52a0f9e251/pymatgen/util/plotting.py#L63) function.

#### Expected behavior

 I am confused by this design choice since it is not particularly helpful to return the `plt` module.

Would it be more appropriate to to return the `Figure` object? Though users can access get it by using `plt.gca` it makes more sense to me, and seems to parallel the `go.Figure` object returned when using the `plotly` backend in plotting methods.

janosh:
Closes #3138.

`pymatgen` plotting methods using the `matplotlib` backend return the `matplotlib.pyplot` module (`plt`), rather than an object related to the plot itself, such as the `Figure` or `Axes`. This is suboptimal API design and inconsistent with the `plotly` backend, where a `go.Figure` object is returned.

This PR changes all plotting methods to return a `plt.Axes` object (same as [`pymatviz`](https://github.com/janosh/pymatviz)) instead of `plt`.

Reasons:
- **More Intuitive** as `Axes` actually contains the plot attributes (lines, legend, labels, etc.), and is more in line with what users may expect and what they typically interact with when customizing plots.
- **Consistency** with `pymatviz` and `pymatgen`'s own `plotly` backend.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.284
+    rev: v0.0.285
     hooks:
       - id: ruff
         args: [--fix]
@@ -27,7 +27,7 @@ repos:
       - id: black
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.4.1
+    rev: v1.5.1
     hooks:
       - id: mypy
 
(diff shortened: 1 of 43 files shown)


janosh:
@fraricci Thanks for reporting and troubleshooting!

PR #3254, opened 2023-08-19
Add VASP input set `MatPESStaticSet`

SophiaRuan:
## Summary

Major changes:

- feature 1: add MatPESStaticSet class

## Todos

1. make sure all the input set default settings are right.
2. write unit tests

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


pymatgen/io/vasp/MatPESStaticSet.yaml
```diff
@@ -0,0 +1,160 @@
+# Default VASP settings for static calculations in the MatPES Project
+# using the Regularized-Restored Strongly Constrained and Appropriately
+# Normed functional (PBE, PBE+U or r2SCAN).
+INCAR:
+  ALGO: ALL
+  EDIFF: 1.e-05
+  EDIFFG: -0.02
```
shyuep, line 7: Remove EDIFFG, ISIF, IBRION. When NSW=0, all these variables have no meaning.

pymatgen/io/vasp/MatPESStaticSet.yaml
```diff
@@ -0,0 +1,160 @@
...
+  LASPH: True
+  LCHARG: True
+  LELF: False # LELF = True restricts calculation to KPAR = 1
+  LMIXTAU: True
+  LORBIT: 11
+  LREAL: Auto
+  LVTOT: True
+  LWAVE: False
+  METAGGA: PBE # change R2SCAN to PBE as default
+  NELM: 200
```
shyuep, line 25: We may possibly want to increase this.

pymatgen/io/vasp/MatPESStaticSet.yaml
```diff
@@ -0,0 +1,160 @@
+# Default VASP settings for static calculations in the MatPES Project
+# using the Regularized-Restored Strongly Constrained and Appropriately
+# Normed functional (PBE, PBE+U or r2SCAN).
+INCAR:
+  ALGO: ALL
```
shyuep, line 5: Default ALGO should be Normal for PBE. R2SCAN uses All. But our input set is written with PBE as a default.

pymatgen/io/vasp/MatPESStaticSet.yaml
```diff
@@ -0,0 +1,160 @@
...
+  LAECHG: True
+  LASPH: True
+  LCHARG: True
+  LELF: False # LELF = True restricts calculation to KPAR = 1
+  LMIXTAU: True
+  LORBIT: 11
+  LREAL: Auto
+  LVTOT: True
+  LWAVE: False
+  METAGGA: PBE # change R2SCAN to PBE as default
```
shyuep, line 24: I don't think METAGGA=PBE is actually a valid value (see the VASP wiki). This should not be set in the YAML.

pymatgen/io/vasp/MatPESStaticSet.yaml
```diff
@@ -0,0 +1,160 @@
...
+    Pm3+: 4
+    Pr3+: 2
+    Sm3+: 5
+    Tb3+: 6
+    Tm3+: 2
+    V: 5
+    W: 5
+    Yb3+: 1
+POTCAR_FUNCTIONAL: PBE_54
+POTCAR:
```
shyuep, line 61: I don't think we should be hard coding POTCARs or U values in this input set. The VASP input sets support inheritance. E.g., VASPIncarBase defines U values. We should write a base for PBE54 that incorporates things we feel would be common across all calculations. E.g., POTCAR_FUNCTIONAL, POTCAR and maybe the U values.

pymatgen/io/vasp/sets.py
```diff
@@ -1215,6 +1215,73 @@ def from_prev_calc(cls, prev_calc_dir, **kwargs):
...
+        """
+        Args:
+            structure (Structure): Structure from previous run.
+            functional (str): We can set the VASP functional to R2SCAN or R2SCAN+U or PBE+U, or it will
+                generate input sets PBE functional.
+            **kwargs: kwargs supported by MPRelaxSet.
+        """
+        super().__init__(structure, MatPESStaticSet.CONFIG, **kwargs)
+        kwargs.setdefault("user_potcar_functional", "PBE_54")
+        if isinstance(functional, str):
```
shyuep, line 1238: What is this isinstance for? I don't think functional should ever be None. Just set the functional default to PBE.

pymatgen/io/vasp/sets.py
```diff
@@ -1215,6 +1215,73 @@ def from_prev_calc(cls, prev_calc_dir, **kwargs):
...
+    ):
+        """
+        Args:
+            structure (Structure): Structure from previous run.
+            functional (str): We can set the VASP functional to R2SCAN or R2SCAN+U or PBE+U, or it will
+                generate input sets PBE functional.
+            **kwargs: kwargs supported by MPRelaxSet.
+        """
+        super().__init__(structure, MatPESStaticSet.CONFIG, **kwargs)
+        kwargs.setdefault("user_potcar_functional", "PBE_54")
```
shyuep, line 1237: Isn't this already specified in the input set?

pymatgen/io/vasp/sets.py
```diff
@@ -1215,6 +1215,73 @@ def from_prev_calc(cls, prev_calc_dir, **kwargs):
...
+        self.kwargs = kwargs
+        self.functional = functional
+
+    @property
+    def incar(self):
+        """Incar"""
+        parent_incar = super().incar
+        incar = Incar(self.prev_incar) if self.prev_incar is not None else Incar(parent_incar)
+
+        incar.update(
```
shyuep, line 1255: I am confused why this section is needed? What is the reason to specify the settings in the YAML file, but still have to hard code this in the input set? If anything, it should be incar.update(parent_incar) to override all these settings from prev_incar if there is any. Ultimately, all these will need to have tests in order for us to know if the code is correctly working.

pymatgen/io/vasp/MatPESStaticSet.yaml
SophiaRuan: I just saw on the VASP wiki that there was a GGA:PE. Can I put this in the INCAR part in the YAML? Then I can change the value of GGA and METAGGA for different functionals.

pymatgen/io/vasp/MatPESStaticSet.yaml
SophiaRuan: Should I keep the value of NELM:200 for now?

pymatgen/io/vasp/MatPESStaticSet.yaml
shyuep: Yes, that would be fine.

pymatgen/io/vasp/MatPESStaticSet.yaml
shyuep: This is fine.

pymatgen/io/vasp/MatPESStaticSet.yaml
SophiaRuan: I felt confused about how we make sure all the MatPES static calculations have the same PSP, so that's why I incorporate the POTCAR values here. When you mentioned write a base for PBE54, do you mean to include another` PBE54Base.yaml` used for all the PBE54 POTCARs, POTCAR_FUNCTIONAL, and U values?

shyuep: Yes.

tests/io/vasp/test_sets.py
```diff
@@ -750,6 +751,116 @@ def test_grid_size_from_struct(self):
             assert matched
 
 
+class TestMatPESStaticSet(PymatgenTest):
+    def setUp(self):
+        self.tmp = tempfile.mkdtemp()
```
janosh, line 756: Not needed, just use `self.tmp_path` inherited from `PymatgenTest`.

tests/io/vasp/test_sets.py
```diff
@@ -750,6 +751,116 @@ def test_grid_size_from_struct(self):
...
+        assert vis.incar["NELM"] == 200
+        assert vis.incar["ENCUT"] == 680
+        assert vis.incar["ENAUG"] == 1360
+        assert vis.incar["EDIFF"] == 1.0e-05
+        assert vis.incar["LREAL"] is False
+        assert vis.incar["LORBIT"] == 11
+        assert vis.incar["LVHAR"]
+        assert vis.incar["ISMEAR"] == 0
+        assert vis.incar["KSPACING"] == 0.22
+        assert vis.incar["GGA"] == "PE"
```
janosh, line 773: No need to assert keys one by one. If you want to make sure they match the values in `pymatgen/io/vasp/MatPESStaticSet.yaml`, you can load that file into a dict and compare `vis.incar` against that dict.

janosh (changes requested):

shyuep:
@SophiaRuan This looks good. But I would like @JiQi535 to go through it carefully and make sure everything is correct.

esoteric-ephemera:
A few remarks given the updates to MP's GGA and r2SCAN WFs:
- LREAL should be False for statics (consistent with how we do MP statics)
- It says this is set is intended to r2SCAN, but the default is PBE? Otherwise, GGA = None and METAGGA = R2SCAN is needed (the GGA = None is needed for the more recent versions of VASP)
- LAECHG should be False unless you plan on using the AECCAR* files
- Same with LCHARG and the CHG/CHGCAR files
- Same with LVTOT and the LOCPOT / POT files

If you want to contribute the statics to MP-core, then unfortunately you need LAECHG, LCHARG, and LOCPOT set to true for consistency with the r2SCAN WF. It'll just slow things down slightly with the file I/O

We're working on independently testing the differences in energies / forces / stresses for ISMEAR = 0 and -5. We may ultimately stay with ISMEAR = -5 for DOS / bandstructures / thermo properties, but that does mean incorrect forces for metals

shyuep:
@esoteric-ephemera this input set is for MatPES (as the name implies). While we are happy to contribute to MPstatics if MP wants them, forces are a priority here. Perhaps more so than even energies. 

I am unsure why MP would want to stick to ISMEAR -5 for all statics. The evidence is quite conclusive than 0 works as well for insulators and certainly better for metals. So what is the advantage of sticking with -5?

esoteric-ephemera:
@shyuep I saw the work from your group for these differences, my only concern is that the average absolute energy differences between the two smearing methods is ~1 eV, which can be large on the scale of energy differences relevant for thermoprops (~5-10 meV/atom). Would be good to see these differences in eV/atom instead of eV

I'm also not sure what the effect of ISMEAR is on the DOS / KS eigenvalues / bandgap. We usually assume ISMEAR = -5 gives a better description of the actual electronic structure, but we're testing this now

shyuep:
@esoteric-ephemera pk. Would be good to have independent validation.

JaGeo:
-5 should give a much better band gap based on DOS.

shyuep:
@esoteric-ephemera This is just my intuition. But if you use a small enough SIGMA, the DOS effect is small. After all, a smearing of 0.02 can't lead to more than a 0.04 eV error in band edges. In the grand scheme of DFT errors in band gaps, it is practically nothing.

JiQi535:
> @SophiaRuan This looks good. But I would like @JiQi535 to go through it carefully and make sure everything is correct.

I have checked @SophiaRuan's implementation of MatPESStaticSet, and it looks nice to me. I have below questions.

1. It seems to me that +U is by default not added. Isn’t this conflicting with the default in MPRelaxSet and MPStaticSet?
2. In MPRelaxSet, there are 89 elements, same as the current coverage by MP. While for the new set, we have 97 elements. Some of the extra elements are Fr, Ra, Rn. @janosh  I’m curious if and why they are added 😃 
3. LVTOT=True. We don’t have this in MPStaticSet. Is MP storing this data? @janosh

JaGeo:
Just adding this paper here for my comment on the DOS computations (a smearing of 0.05 was used):
https://www.sciencedirect.com/science/article/pii/S277294942200002X

janosh:
> 2 In MPRelaxSet, there are 89 elements, same as the current coverage by MP. While for the new set, we have 97 elements. Some of the extra elements are Fr, Ra, Rn. @janosh  I’m curious if and why they are added 😃

@JiQi535 They were copied from [`MPSCANRelaxSet`](https://github.com/materialsproject/pymatgen/blob/b874c73851d133e542c5a13183ce6f123147f13b/pymatgen/io/vasp/MPSCANRelaxSet.yaml#L28) IIRC.

> 3 LVTOT=True. We don’t have this in MPStaticSet. Is MP storing this data? @janosh

No.

pymatgen/io/vasp/MatPESStaticSet.yaml
```diff
@@ -0,0 +1,26 @@
...
+  ISPIN: 2
+  KSPACING: 0.22 # included to have some reasonable default
+  LAECHG: True
+  LASPH: True
+  LCHARG: True
+  LELF: False # LELF = True restricts calculation to KPAR = 1
+  LMIXTAU: True
+  LORBIT: 11
+  LREAL: Auto
+  LVTOT: True
```
JiQi535, line 20: This line should be removed. The LOCPOT file is going to consume storage space.

pymatgen/io/vasp/sets.py
```diff
@@ -1215,6 +1215,43 @@ def from_prev_calc(cls, prev_calc_dir, **kwargs):
...
 
+class MatPESStaticSet(MPStaticSet):
+    """Creates input files for a MatPES static calculation."""
+
+    CONFIG = _load_yaml_config("MatPESStaticSet")
+
+    def __init__(
+        self,
+        structure: str,
+        functional: Literal["R2SCAN", "R2SCAN+U", "PBE", "PBE+U"] = "PBE",
```
JiQi535, line 1226: @SophiaRuan  Here, I think we should have PBE+U as default, inline with default of MPRelaxSet and MPStaticSet. Also, I don't think we should distinguish PBE and PBE+U as two functionals. Instead, +U can be disabled in the same way as in MPRelaxSet and MPStaticSet with `user_incar_settings={"LDAU": False}`.

shyuep:
No. PBE default. U and r2scan as options. This is not MP.

shyuep:
The lint and test failures are unrelated to the PR. I am merging.

janosh:
They're not unrelated. The 1st `TestMatPESStaticSet ` `assert` fails:

```sh
pytest tests/io/vasp/test_sets.py -k TestMatPESStaticSet -vv
```

```py
_____________________________ TestMatPESStaticSet.test_init ______________________________

self = <tests.io.vasp.test_sets.TestMatPESStaticSet testMethod=test_init>

    def test_init(self):
        prev_run = f"{TEST_FILES_DIR}/relaxation"
    
        vis = MatPESStaticSet.from_prev_calc(prev_calc_dir=prev_run)
        # Check that INCAR settings were load from yaml, check the default functional is PBE.
        CONFIG = _load_yaml_config("MatPESStaticSet")
>       assert vis.incar == CONFIG["INCAR"]
E       AssertionError: assert {'PREC': 'acc...ORBIT': False} == {'ALGO': 'Nor...5, 'Yb3+': 1}}
E         Omitting 3 identical items, use -vv to show
E         Differing items:
E         {'PREC': 'accurate'} != {'PREC': 'Accurate'}
E         {'ISMEAR': -5} != {'ISMEAR': 0}
E         {'MAGMOM': [0.6]} != {'MAGMOM': {'Ce': 5, 'Ce3+': 1, 'Co': 0.6, 'Co3+': 0.6, 'Co4+': 1, 'Cr': 5, 'Dy3+': 5, 'Er3+': 3, 'Eu': 10, 'Eu2+': 7,...4+': 3, 'Mo': 5, 'Nd3+': 3, 'Ni': 5, 'Pm3+': 4, 'Pr3+': 2, 'Sm3+': 5, 'Tb3+': 6, 'Tm3+': 2, 'V': 5, 'W': 5, 'Yb3+': 1}}
E         {'SIGMA': 0.2} != {'SIGMA': 0.05}
E         {'LORBIT': False} != {'LORBIT': 11}...
```

shyuep:
I just noticed one problem. Why is the INCAR written as `return Incar(self.prev_incar or parent_incar)`? This is wrong. If the prev_incar is a relaxation, this will make it a relaxation calculator too.

PR #3264, opened 2023-08-23
Add `properties` to Structure and Molecule

### LINKED ISSUE #2884: Add structure_property attribute to `Structure`
I'd like to add a `structure_property` attribute to the `Structure` class. Sometimes it is useful to have structure-wide properties that can be stored alongside the structure. One example is the `Poscar.predictor_corrector_preamble` which is necessary to write structures to continue MD simulations. Currently, only site-specific properties can be stored (e.g., the velocities and predictor-corrector coordinates).

With the present `Structure` class, it is not possible to store structures that can be used to continue MD simulations without storing this data separately which is quite unwieldy. I'm sure there are other instances where I've wanted to do this.

@shyuep would you support me adding this feature? The API would be the same as for adding site properties. E.g., 

```python
structure.add_structure_property("property_name", property_value)
```

This is related to https://github.com/materialsproject/atomate2/issues/185

gpetretto:
Closes #2884.

This PR implements a `properties` attribute for `Structure` and `Molecule`. I also modified the `Poscar` object to store the `predictor_corrector_preamble` in the structure properties.


janosh:
Thanks @gpetretto for taking this on! Looking promising. 👍 

I think we need better file IO support before merging. Right now, the properties are written to JSON

```py
json_str = struct_with_props.to(fmt="json")
assert '"test_property": 42' in json_str
```

but they're not read back in `Structure.from_str`, i.e. this fails:

```py
struct = Structure.from_str(json_str, fmt="json")
assert struct.properties == props
```

As I mentioned in https://github.com/materialsproject/pymatgen/issues/2884#issuecomment-1577522801 it would be nice to retain properties when reading and writing to as many file formats as possible. JSON at least is easy to do.

janosh:
@arosen93 @utf Would you mind taking a look at this to ensure it addresses your use cases?

arosen93:
At a very quick glance, this seems reasonable. Once merged, I will need to update the ASE atoms adapter to ensure this is handled nicely. But that's for me to deal with.

janosh:
Thanks @gpetretto! 👍

arosen93:
@gpetretto: Any thoughts about (de)serialization with the new `.properties` flag? In principle, a user can define `.properties` such that it contains data that is not jsonable or that makes the following fail. Is this something where we put the onus on the user, or should we be incorporating some sort of check?

```python
d = jsanitize(structure, strict=True, enum_values=True)
MontyDecoder().process_decoded(d)
```

janosh:
Good question! Not obvious what's the best solution here. We could issue a runtime warning when users set non-JSONable `properties` and/or add a config option to raise immediately when detecting offending `properties`.

gpetretto:
This is a good point. I actually did not think much about this kind of problems since in my view `properties` is somewhat equivalent to the `site_properties`, which suffer from the same limitation. 
In general, adding non serializable properties could still be a valid use case, as long as the structure in not serialized. So, I would not strictly prevent a user from doing that. In any case one would get an error when trying to dump the structure to a file.

PR #3270, opened 2023-08-24
Breaking: Update `AseAtomsAdaptor` to handle `Structure.properties`/`Molecule.properties`

arosen93:
Ensure `Structure.properties` and `Molecule.properties` do not get lost with ASE `Atoms` interconversion.

Recent PR that added `properties`: #3264 

Breaking because `IMolecule.__eq__` and `IStructure.__eq__` were changed to check `self.properties == other.properties`. See https://github.com/materialsproject/pymatgen/pull/3270#discussion_r1311253937.


arosen93:
I see you @shyuep! 👀😅

I plan to finish this PR later today.

arosen93:
@janosh: When you're back on GitHub, would you mind helping me out with the `mypy` issue? I think it might be because `.properties` is not a `@property` attribute of `Structure` or `Molecule`, but I'm not 100% sure. Thank you 🙏

tests/io/test_ase.py
```diff
@@ -292,10 +297,14 @@ def test_back_forth(self):
...
         molecule = Molecule.from_file(TEST_FILES_DIR / "acetylene.xyz")
         molecule.set_charge_and_spin(-2, spin_multiplicity=3)
-        molecule.info = {"test": "hi"}
+        molecule.properties = {"test": "hi"}
         atoms = aio.AseAtomsAdaptor.get_atoms(molecule)
         molecule_back = aio.AseAtomsAdaptor.get_molecule(atoms)
         atoms_back = aio.AseAtomsAdaptor.get_atoms(molecule_back)
         for k, v in atoms.todict().items():
             assert str(atoms_back.todict()[k]) == str(v)
         assert molecule_back == molecule
```
janosh, line 306: This doesn't currently check that the `properties` were restored since `Molecule.__eq__` doesn't currently look at `properties`. I think that'll be unexpected behavior for must users so we should change `IMolecule.__eq__` and `IStructure.__eq__` even though it's strictly speaking a breaking change.

janosh (commented):
Nice work! Thanks @arosen93! 👍

pymatgen/io/ase.py
```diff
@@ -216,12 +217,19 @@ def get_structure(atoms: Atoms, cls: type[Structure] = Structure, **cls_kwargs)
         else:
             sel_dyn = None
 
+        # Atoms.info <---> Structure.properties (excluding properties["calc"])
+        properties = jsanitize(getattr(atoms, "info", {}))
```
arosen93, line 221: @janosh: Just wanted to point out one subtlety to you. Currently, we have the mapping: `Atoms.info <--> Structure.properties`, since they serve the same purpose between ASE and Pymtatgen, respectively. However, I `jsanitize`-d the `atoms.info` before assigning it to `.properties` to ensure the `Structure`/`Molecule` object can be (de)serialized since this is often an implicit assumption of `Structure`/`Molecule` objects but is not necessarily one for ASE `Atoms` objects. In practice, this means that there could be some scenarios where a user interconverts between ASE `Atoms` and Pymatgen `Structure`/`Molecule` with some loss of information; namely, if the `atoms.info` is modified by `jsanitize()` then it can't be restored in its original form. There might be a better solution than this, but I implemented it because oftentimes ASE will dump un-serializable class objects (namely, a `Spacegroup` class) into `atoms.info`, such as when reading in a CIF with symmetry flags.

janosh: Thanks for highlighting! Seems like a good way to go about this for now. I'll keep this in mind if we run into any issues.

PR #3278, opened 2023-08-28
[WIP] revise MatPESStaticSet, inherit from DictSet 

SophiaRuan:
## Summary
Related PR: #3254 
Major changes:

- feature 1: Revise MatPESStaticSet, inherit from DictSet 

## Todos

Check if the previous Incar will override the default MatPESStaticSet Incar.
Check if the new settings added by Custodian will be updated.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/io/vasp/sets.py
+++ b/pymatgen/io/vasp/sets.py
@@ -1215,15 +1215,16 @@ class MPStaticSet(MPRelaxSet):
         return input_set.override_from_prev_calc(prev_calc_dir=prev_calc_dir)
 
 
-class MatPESStaticSet(MPStaticSet):
+class MatPESStaticSet(DictSet):
     """Creates input files for a MatPES static calculation."""
 
     CONFIG = _load_yaml_config("MatPESStaticSet")
 
     def __init__(
         self,
-        structure: str,
+        structure: Structure,
         functional: Literal["R2SCAN", "R2SCAN+U", "PBE", "PBE+U"] = "PBE",
+        prev_incar=None,
         **kwargs: Any,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


pymatgen/io/vasp/sets.py
```diff
@@ -1215,41 +1215,47 @@ def from_prev_calc(cls, prev_calc_dir, **kwargs):
         return input_set.override_from_prev_calc(prev_calc_dir=prev_calc_dir)
 
 
-class MatPESStaticSet(MPStaticSet):
+class MatPESStaticSet(DictSet):
```
janosh, line 1218: Inheriting from `DictSet` means `MatPESStaticSet.from_prev_calc` is no longer defined so the tests need to change.

janosh (changes requested):
Thanks for the quick followup to https://github.com/materialsproject/pymatgen/pull/3254! 👍 

Perhaps you're already aware but the tests need updating.

pymatgen/io/vasp/sets.py
janosh: The tests are still failing. You can run them using

```py
pytest tests/io/vasp/test_sets.py -k TestMatPESStaticSet -vv
```

Requires `pip install pytest` if you haven't installed it yet. Should print sth like

```txt
___________ TestMatPESStaticSet.test_init ___________

self = <tests.io.vasp.test_sets.TestMatPESStaticSet testMethod=test_init>

    def test_init(self):
        # check if default INCAR settings are loaded
        default = MPStaticSet(self.struct)
        incar = default.incar
        # assert incar["GGA"] == "Pe"
        assert incar["ALGO"] == "Normal"
>       assert incar["EDIFF"] == 1.0e-05
E       assert 0.0012000000000000001 == 1e-05

tests/io/vasp/test_sets.py:747
```

JiQi535:
I was not able to directly commit to this PR, so I opened #3280

PR #3280, opened 2023-08-29
 Corrections for MatPESStaticSet and its tests

JiQi535:
Related PR: #3278 #3254 

Tests now covers all MatPESStaticSet parameters. 

Remaining question: While the default setting of MatPESStaticSet is without Hubbard U, when +U is enabled, the Hubbard U related parameters will be the same as MPRelaxSet. As I have noticed that MITRelaxSet and MPRelaxSet have different default parameters, I'm not sure if setting the defaults same as MPRelaxSet is the most correct.

### DIFF
--- a/tests/io/vasp/test_sets.py
+++ b/tests/io/vasp/test_sets.py
@@ -165,8 +165,8 @@ class TestMITMPRelaxSet(PymatgenTest):
         structure = Structure.from_spacegroup("Fm-3m", Lattice.cubic(3), ["Cu"], [[0, 0, 0]])
 
         with pytest.warns(
-                BadInputSetWarning,
-                match="Relaxation of likely metal with ISMEAR < 1 detected. See VASP recommendations on ISMEAR for metals.",
+            BadInputSetWarning,
+            match="Relaxation of likely metal with ISMEAR < 1 detected. See VASP recommendations on ISMEAR for metals.",
         ) as warns:
             vis = self.set(structure)
             _ = vis.incar
@@ -213,9 +213,9 @@ class TestMITMPRelaxSet(PymatgenTest):
                     structure=struct, user_potcar_functional="PBE_54", user_potcar_settings=user_potcar_settings
                 )
                 expected = {  # noqa: SIM222
-                               **({"W": "W_sv"} if "W" in struct.symbol_set else {}),
-                               **(user_potcar_settings or {}),
-                           } or None
+                    **({"W": "W_sv"} if "W" in struct.symbol_set else {}),
+                    **(user_potcar_settings or {}),
+                } or None
                 assert relax_set.user_potcar_settings == expected
 
     @skip_if_no_psp_dir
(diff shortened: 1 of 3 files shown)


shyuep:
Linting and test workflows are failing. When you add a new input set, the hash test needs to be updated. The linting errors are easy to fix. Fix both first.

tests/io/vasp/test_sets.py
```diff
@@ -734,68 +732,61 @@ def test_grid_size_from_struct(self):
...
+        assert incar["SIGMA"] == 0.05
 
-        # test with changed user_incar_settings
-        non_prev_vis = MatPESStaticSet(vis.structure, user_incar_settings={"ENCUT": 800, "LORBIT": 12, "LWAVE": True})
-        # Check that the ENCUT and Kpoints style has NOT been inherited.
-        assert non_prev_vis.incar["ENCUT"] == 800
-        # Check that user incar settings are applied.
-        assert non_prev_vis.incar["LORBIT"] == 12
-        assert non_prev_vis.incar["LWAVE"]
+    def test_init_with_prev_incar(self):
```
shyuep, line 762: This test is incomplete. Other than checking that the parameters for static are ok, I would want to know that some of the parameters from prev INCAR are moved. Also, an input set also generates KPOINTS and POTCAR files. These need to be checked too, for ALL tests.

tests/io/vasp/test_sets.py
```diff
@@ -734,68 +732,61 @@ def test_grid_size_from_struct(self):
...
+        assert incar_scan.get("GGA") is None
+        assert incar_scan["ALGO"] == "All"
+        assert incar_scan.get("LDAU") is None
 
-        non_prev_vis = MatPESStaticSet.from_dict(non_prev_vis.as_dict())
-        assert non_prev_vis.incar["ENCUT"] == 800
-        # Check that user incar settings are applied.
-        assert non_prev_vis.incar["LORBIT"] == 12
-        assert non_prev_vis.incar["LWAVE"]
+    def test_potcar(self):
```
shyuep, line 777: This should be part of each kind of test.

shyuep (commented):
There are also no tests for overriding parameters, e.g., POTCAR types, etc.

SophiaRuan:
New PR #3281  was opened as the superset of the PR #3280 , so this #3280 can be closed.

tests/io/vasp/test_sets.py
JiQi535: @shyuep Thanks for the comments. I just saw it because of email address transfer at UCSD. 

I previously have removed all tests for KPOINTS. My thought was that our MatPESStaticSet inherits the KPOINTS related functions directly from DictSet, so there is no need to check KPOINTS here in our tests for MatPESStaticSet. 

I now tends to agree that there is no harm to test more comprehensively, especially as we have KSPACING in INCAR settings.

PR #3285, opened 2023-08-31
Slightly relax the constraint satisfy condition of get_primitive_structure()

fyalcin:
## Summary

Context [here](https://matsci.org/t/varying-results-with-get-primitive-structure/41663). I finally took some time to figure out why this was happening and it turned out to be the check that compares the lattice parameters of the primitive structures found against the constraints. The equivalency check in the current implementation is too strict to account for small changes in the constraints, which is why I went with `np.isclose()` with its default `atol` and `rtol` parameters which should still be strict enough.

I also include two unit tests, one for a general `get_primitive_structure()` call with `constrain_latt` which was missing, and  #another to ensure two very similar constraints result in the same primitive structure.


janosh (approved):
Looks good, thanks @fyalcin! 👍 

I added the test file you linked in https://matsci.org/t/41663 but forgot to upload and improved the tests slightly.

PR #3286, opened 2023-08-31
`np.(arange->linspace)` in `io/vasp/optics.py` `get_delta`, `get_setp` and `epsilon_imag`

LucasGVerga:
`np.linspace` ensures the array size is numerically stable. See the [`numpy.arange` warning](https://numpy.org/doc/stable/reference/generated/numpy.arange.html).


janosh:
Thanks @LucasGVerga!

PR #3299, opened 2023-09-03
Add an input arg check for `Kpoints.automatic_density_by_lengths`

arosen93:
## Summary

The function `Kpoints.automatic_density_by_length()` takes in a sequence of values, e.g. `[50, 50, 1]`, to set the k-point density for each dimension. However, the function does not raise any kind of error if a sequence of dimensions != 3 is supplied.

This PR adds a `ValueError` if the `len` is != 3, along with a test for it.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


janosh (approved):
Thanks @arosen93!

PR #3307, opened 2023-09-09
Move `py.typed` to package root

Andrew-S-Rosen:
When using `mypy` on packages that rely on Pymatgen, it often complains like follows:

>  Skipping analyzing "pymatgen.io.vasp.inputs": module is installed, but missing library stubs or py.typed marker

Since `pymatgen.io` is such a commonly used module, let's add a `py.typed` there unless there's a reason to avoid doing so.

Is there a reason why it wasn't included? And should one be included in other directories (e.g. `pymatgen.symmetry`)?


janosh:
Since the goal is to add type hints all across `pymatgen` (although some areas still fall short), I moved the `py.typed` into the package root.

PR #3343, opened 2023-09-20
Fix `TypeError`: can only join an iterable with AECCAR in `VolumetricData.write_file`

chiang-yuan:
Context: I was trying to download `AECCAR`s from aws and save them locally as vasp compatible format. `CHGCAR`s can pass successfully but `AECCAR` will yield error when `aeccar.write_file`  like following

```python
mp_id = 'mp-1523380'
prefix = 'aeccar0s'
suffix = prefix.rstrip("s")

fpath = f'{mp_id}-{suffix}.json.gz'

with zopen(fpath) as f:
    json_string = f.read()
    data = json.loads(json_string, cls=MontyDecoder)['data']
    data.write_file(f'{mp_id}.{suffix.upper()}')
```
Modify `io/vasp/outputs.py` to do the Iterable check first and only write to file when `self.data_aug.get(data_type, [])`  is not empty

# Checklist

- [x] ruff
- [x] mypy
- [x] pre-commit


chiang-yuan:
tag @tschaume you might be interested to comment since I mentioned this bug to you 😄

janosh (changes requested):
Thanks @chiang-yuan! Could you mention the error type and message you're seeing? We'll also need a test for this. You can maybe use `tests/files/bader/AECCAR0.gz`

chiang-yuan:
The error I received is 
```
-> 3585 f.write("".join(self.data_aug.get(data_type, [])))

TypeError: can only join an iterable
```
@janosh this is a file output bug, so do you want me to check if the file looks identical to itself after read in and output?

janosh:
> @janosh this is a file output bug, so do you want me to check if the file looks identical to itself after read in and output?

Sounds good! Just writing to disk would be enough to test this fix but checking for equality after re-read is even better.

chiang-yuan:
Test added :) @janosh

pymatgen/io/vasp/outputs.py
```diff
@@ -3410,7 +3411,10 @@ def write_spin(data_type):
                         lines.append(" ")
                 if count % 5 != 0:
                     f.write(" " + "".join(lines) + " \n")
-                f.write("".join(self.data_aug.get(data_type, [])))
+
+                data = self.data_aug.get(data_type, [])
+                if isinstance(data, Iterable):
```
janosh, line 3416: Why is it ok here not to write to file if `data` is not `Iterable`?

tests/io/vasp/test_outputs.py
```diff
@@ -1558,6 +1558,20 @@ def test_as_dict_and_from_dict(self):
         )
 
 
+class TestAECCARS(PymatgenTest):
+    def test_write_file(self):
+        aeccar0_test = Chgcar.from_file(f"{TEST_FILES_DIR}/bader/AECCAR0.gz")
+        aeccar0_test.write_file("AECCAR0_test")
+        aeccar0_read = Chgcar.from_file("AECCAR0_test")
+        np.allclose(aeccar0_test.data["total"], aeccar0_read.data["total"])
```
janosh, line 1566: This needs to be `assert`ed. Just calling `np.allclose` doesn't fail the test if the check fails.

chiang-yuan: Oh!!! Thanks for the good catch. I missed that.

pymatgen/io/vasp/outputs.py
chiang-yuan: If `data_aug` is empty and not `Iterable`, it will `join` will give error.

chiang-yuan (commented):
Thanks for the help @janosh! All look good to me 😃

pymatgen/io/vasp/outputs.py
janosh: But do we want to write an empty file in that case? Having no file after calling `write_file` sounds like unexpected behavior.

chiang-yuan: Only the partial of the file (`data_aug`) will not be written if it is empty. The rest of the file will still be there.

janosh (approved):
Thanks for the test! 👍

PR #3346, opened 2023-09-21
One small Q-Chem update [WIP]

samblau:
Small change to sets.py to better handle initial hessian for new geometry optimizer

Still need to add a test to go along with it


janosh:
was this accidentally left in draft mode?

shyuep:

PR #3350, opened 2023-09-26
Ion: default hydrates=False in reduced_formula

rkingsbury:
## Summary

Major changes:

- default to `hydrates=False` in `Ion.get_reduced_formula_and_factor()`
- explicitly use `Ion.get_reduced_formula_and_factor()` rather than `super().get_reduced_formula_and_factor()` in `reduced_formula

## Checklist


janosh:
@rkingsbury Is this intentionally in draft mode?

rkingsbury:
@janosh it was intentional, but I think it's good to go. All tests pass, but I wanted to make sure that this didn't mess up anything on the API side. When we re-did the [Pourbaix ion data](https://next-gen.materialsproject.org/contribs/projects/ion_ref_data?_limit=75), I used `Ion.reduced_formula` as the database keys (the MPContribs identifiers). However, it doesn't appear that any of the ions in the dataset were formatted as hydrates (for example, `Be3(OH)3[3+]` is listed there, but could be written `Be3OH2.1H2O`).

In the API client, the only call is `Ion.from_formula()` ([here](https://github.com/materialsproject/api/blob/ffd9713ec0847466eae858a802d6692dfc476159/mp_api/client/mprester.py#L1032C13-L1032C17)) followed by `Ion.composition` so I don't think this change will cause any problems.

janosh (approved):

Issue #2895, opened 2023-03-15
Mulit element POTCAR loading broken.


MichaelWolloch:
**Describe the bug**
Since commit e4bde8e400d28095425d0b1e7f0b1c8923ddc13a  it seems that loading is broken if any but the first POTCAR in the file has a sha256 hash.

I really appreciate the effort by @shyuep to clean up the code and get rid of the regex stuff and unicode handling. However, a small mistake appears to have slipped in in line 2277 of `pymatgen/io/vasp/inputs.py`. There should be a unicode end of line character still:
~~~
potcar_strings = fdata.split("End of Dataset\n")
~~~
instead of 
~~~
potcar_strings = fdata.split("End of Dataset")
~~~

If that is included, also the test that got modified in the commit can be changed back it seems.

However, since this all seems a bit fragile, I suggest to actually also downgrade the ValueError that is raised in line 1829 in  `pymatgen/io/vasp/inputs.py` to a warning. 

**To Reproduce**
load the attached POTCAR (containing O and Fe_pv_with_hash from `pymatgen/test_files/POT_GGA_PAW_PBE_54` with the Fe Potcar following the O) into a Potcar object:
[POTCAR_test.gz](https://github.com/materialsproject/pymatgen/files/10982996/POTCAR_test.gz)
~~~
from pymatgen.io.vasp.inputs import Potcar
pot = Potcar.from_file('POTCAR_test.gz')
~~~
results in:
~~~
ValueError: POTCAR with symbol Fe_pv and functional
PBE has a SHA256 hash defined,
but the computed hash differs.
YOUR POTCAR FILE HAS BEEN CORRUPTED AND SHOULD NOT BE USED!
~~~

**Expected behavior**
There should be no error for the SHA256, since the POTCAR is good.

**Environment (please supply relevant versions and platform info):**

- OS: Linux
- Version 2023.3.10

**Additional context**
I am happy to make a pull request if this is wanted.

e-kwsm:
This is not fixed yet (2023.9.25).

```
~/.local/lib/python3.11/site-packages/pymatgen/io/vasp/inputs.py:1806: UnknownPotcarWarning: POTCAR data with symbol O does not match any VASP POTCAR known to pymatgen. There is a possibility your POTCAR is corrupted or that the pymatgen database is incomplete.
  warnings.warn(
[...]

PR #3351, opened 2023-09-26
Breaking: New method of POTCAR validation

esoteric-ephemera:
## Summary

**Issues:** 
current method of POTCAR validation is too brittle to work well in high-throughput calculations. For example, hashes are sensitive to subtle changes in spaces (addition/deletion). However FORTRAN is space-insensitive, thus two POTCARs that differ by spaces are read identically by VASP.

**Goals**:
Develop less brittle method for validating POTCARs that reflects how POTCAR metadata and numeric values are read by FORTRAN.

Major changes:

- *is_valid* function in *pymatgen.io.vasp.inputs.PotcarSingle*, which checks POTCAR metadata and values against a database of these
- To respect copyright, only statistics on the numerical values are stored separately for the PSCTR / POTCAR header and pseudopotential numeric values
- Keywords found within the PSCTR header and pseudopotential body are stored as well for comparison. For example, older POTCARs may have lacked the kinetic energy density (needed for meta-GGA calcs), and were later updated to include this. The check distinguishes the older and newer version of the same POTCAR
- Tests have been updated (*TestPotcarSingle* class in *tests/io/vasp/test_inputs.py*)
 
## Todos

- While I generated POTCAR metadata from those available to our group, might be beneficial to expand the database of POTCAR metadata to older / newer releases


janosh:
@esoteric-ephemera Could you resolve the merge conflicts?

mkhorton:
Thanks for taking a fresh look at this problem @esoteric-ephemera, it's been unsatisfactory for a while. These warnings frequently confuse or alarm people unnecessarily, but the underlying motivation of rigorously checking pseudopotential compatibility is of course very important.

> However FORTRAN is space-insensitive, thus two POTCARs that differ by spaces are read identically by VASP.

I imagine there's a way around this, to normalize the POTCAR file, but it might be slow and perhaps there'd still be brittleness in other ways so I like your suggestions.

When the original POTCAR hash database was assembled we tried to find as many examples as we could, but I'm still surprised there are files out there that might differ by e.g. spaces within the file itself, since I can't imagine why these edits would have been made.

pymatgen/io/vasp/inputs.py
```diff
@@ -2175,6 +2170,124 @@ def get_potcar_hash(self):
...
+            if "." in input_str:
+                return float(input_str)
+            return int(input_str)
+        try:
+            return float(input_str)
+        except ValueError:
+            return input_str
+
+    def is_valid(self, tol: float = 1.0e-6) -> bool:
+        """Check that POTCAR matches reference metadata."""
```
mkhorton, line 2196: Could you expand this docstring to specify what is being checked (e.g., what attributes of `PotcarSingle`), and also what the `tol` is doing, and why this can be changed by the user? It seems like a fixed tolerance might be preferred.

esoteric-ephemera: Yes! Might have gone a little too verbose in explaining what's going on in is_valid(), but there's a better explanation there now.

For "tol", I left that adjustable for debugging - it's frozen now at the same value

esoteric-ephemera:
Thanks both @janosh  and @mkhorton! Tried to address your comments. 

It is possible just to completely strip a POTCAR of spaces and hash it that way, but I see one issue with that: in the PSCTR / header section of the POTCAR, there are comments. Normally, in fortran, in-line comments would be indicated by a leading "!", but these are omitted in the proprietary POTCAR format. One could add comments without changing any numeric values, and get the same POTCAR

There was actually a test of PotcarSingle in PMG that did just that - it basically added comments to a PSCTR tag. A hash would report these as different, but the numeric values are identical to vasp

tests/io/vasp/test_inputs.py
```diff
@@ -1039,47 +1037,45 @@ def test_identify_potcar(self):
         filename = f"{TEST_FILES_DIR}/POT_GGA_PAW_PBE_54/POTCAR.Fe.gz"
 
         psingle = PotcarSingle.from_file(filename)
-        assert "PBE_54" in psingle.identify_potcar()[0]
-        assert "Fe" in psingle.identify_potcar()[1]
+        matched_funcs = [refpsp["POTCAR_FUNCTIONAL"] for refpsp in psingle._matched_meta]
+        assert "PBE_54" in matched_funcs
+        assert "Fe" in psingle._matched_meta[0]["TITEL"]
```
janosh, line 1042: Tests generally shouldn't test an implementation detail like this private `_matched_meta` list. We need a test called `test_is_valid` that only calls the public `PotcarSingle.is_valid` property both for a case where it's `True` and `False`.

janosh (changes requested):
Nice work @esoteric-ephemera! 👍 

I think we just need to rename `test_identify_potcar` and call `is_valid` in there.

Andrew-S-Rosen:
Thank you, @esoteric-ephemera, for helping me sleep better at night with this PR.

tests/io/vasp/test_inputs.py
esoteric-ephemera: Done! And thanks for the cleaning up my messy python. The new `test_is_valid` operates on a working Fe POTCAR, and one where one variable is changed by 0.001. Should highlight the that the check truly is sensitive to individual numeric values.

tests/io/vasp/test_inputs.py
```diff
@@ -1035,51 +1033,58 @@ def test_functional_types(self):
...
+        assert not psingle.is_valid
 
-    def test_potcar_hash_warning(self):
+        psingle.keywords.pop("RCORE")
+        assert not psingle.is_valid
+
+        psingle.data = psingle.data.replace("RCORE  =    2.300", "RCORE  =    2.200")
+        assert not psingle.is_valid
+
+    def test_unknown_potcar_warning(self):
```
janosh, line 1053: I think rather than storing a new POTCAR file with just 1 changed line, it should be easy to invalidate a `PotcarSingle` instance on the fly. But actually this turned out to be pretty hard. You might say this is a contrived example since people shouldn't change `PotcarSingle` objects once they've been read from a file. But fwiw, I can currently reassign

```
psingle.keywords["RCORE"] = 2.2
```

remove it altogether

```
psingle.keywords.pop("RCORE")
```

and modify the underlying string data

```
psingle.data = psingle.data.replace("RCORE  =    2.300", "RCORE  =    2.200")
```

all without invalidating the `PotcarSingle` instance. See the new `assert`s I added. I think it would make sense for all of them to pass but they currently fail.

esoteric-ephemera: This is because the header info is taken from `self.PSCTR` and not the `parse_functions` attributes of PotcarSingle. For each such attribute like RCORE (the ones in `self.parse_functions`), the key/value are stored both in `self.keywords` and `self.PSCTR`

So if you do instead `self.PSCTR['RCORE'] = 2.2`, the second `psingle.is_valid` will fail. Or, if you did 
`psingle.data = psingle.data.replace('local part','local port')`
the second `psingle.is_valid` will also fail.

Hesitant to change how this is done (storing the same data in three different attributes of PotcarSingle) because it might break something that relies on it

janosh: > Hesitant to change how this is done (storing the same data in three different attributes of PotcarSingle) because it might break something that relies on it

I don't think we need to be hesitant here. having redundant data in `PotcarSingle.keywords` and `PotcarSingle.PSCTR` is needlessly complicated, goes against the SSOT coding principle and now would be a great time to clean this up imo.
@shyuep @mkhorton Any concerns about breaking reworking `PotcarSingle` to e.g. remove `PSCTR` and turn `keywords` into a `@property` so they're computed on the fly and always in sync with the underlying `data` string?

mkhorton:
At the risk of asking a stupid question, I saw the MPRelaxSet defaults changed Yb_2 to Yb_3 recently (see #2968).

However, as far as I can tell, only `Yb_2` is in the POTCAR database for the `PBE` functional (i.e., the MPRelaxSet default). The `Yb_3` only exists for the `PBE_52`/`PBE_54`.

So my questions are, (1) is this correct?, and (2), if so, is it even possible to run MPRelaxSet with Yb-containing materials at present?

tests/io/vasp/test_inputs.py
mkhorton: I agree, having redundant data doesn't make sense for the reasons you state. My only hesitation here is that I wouldn't want to impose scope creep on this PR for @esoteric-ephemera, since it's probably not strictly necessary to change it for this PR to achieve its goal of better validation. While the `PotcarSingle` should be immutable, I think the risk of someone mutating it in practice are quite low.

More broadly, I've never had a good understanding of the internals of the POTCAR file. If anyone did want to spend more time on the `PotcarSingle` class, I'd love to see more documented `@property` etc. for some of these keywords so that people can better understand what the POTCAR files they're using are actually specifying.

janosh:
> So my questions are, (1) is this correct?, and (2), if so, is it even possible to run MPRelaxSet with Yb-containing materials at present?

@mkhorton I think you forgot about https://github.com/materialsproject/pymatgen/pull/2972 which reverted the `Yb_2` -> `Yb_3` migration for `MPRelaxSet`:

https://github.com/materialsproject/pymatgen/blob/8e0b09967d4135e33675bc7615c2b4d5e5613162/pymatgen/io/vasp/MPRelaxSet.yaml#L175

mkhorton:
Indeed I did, thanks @janosh! This was not reverted in [atomate2](https://github.com/materialsproject/atomate2/blob/8c151790ec1db7b6e7a55e15f2d5a1fd34235e97/src/atomate2/vasp/sets/BaseMPGGASet.yaml#L202-L204) so the two sets are currently inconsistent (I realise this is an active topic of discussion right now).

tests/io/vasp/test_inputs.py
janosh: I merged `PotcarSingle.keywords` and `PotcarSingle.PSCTR`, now just called `PotcarSingle.keywords` since `PSCTR` has no obvious meaning. Maybe we should rename `PotcarSingle.keywords` to `PotcarSingle.header_keywords` for added clarity?

pymatgen/io/vasp/inputs.py
```diff
@@ -1773,55 +1756,45 @@ def __init__(self, data, symbol=None):
...
             )
-        elif has_sh256 and not hash_check_passed:
+        elif has_sha256 and not sha256_pass:
             warnings.warn(
                 f"POTCAR with symbol {self.symbol} and functional\n"
                 f"{self.functional} has a SHA256 hash defined,\n"
                 "but the computed hash differs.\n"
-                "YOUR POTCAR FILE HAS BEEN CORRUPTED AND SHOULD NOT BE USED!"
+                "YOUR POTCAR FILE HAS BEEN CORRUPTED AND SHOULD NOT BE USED!",
+                UnknownPotcarWarning,
```
janosh, line 1790: @esoteric-ephemera @mkhorton @Andrew-S-Rosen Are we sure we want to keep the old unknown POTCAR warning when the SHA256 hash provided by VASP doesn't match the one computed from the file? Isn't that what has been giving so many false positive warnings?

mkhorton: Does this PR not make this more robust, reducing false positives?

The warning is definitely badly phrased if we do keep it, it’s inscrutable unless someone is very familiar with VASP and pymatgen. We could link to a docs page for explanation.

janosh: > Does this PR not make this more robust, reducing false positives?

The new warning that's supposed to be more robust comes just above these lines:

https://github.com/materialsproject/pymatgen/blob/c329ff56e65be1c4f77d5e5a6de168e683902133/pymatgen/io/vasp/inputs.py#L1778-L1783

Currently, when `not self.is_valid` is `False`, i.e. the POTCAR is valid according to the new validation, then we still check whether computed and in-file hash match and trigger the old warning if not (which has many false positives, I think):
```
elif has_sha256 and not sha256_pass:
    warnings.warn(
                f"POTCAR with symbol {self.symbol} and functional\n"
                ...
```

esoteric-ephemera: The SHA56 warning is not what's been giving us so many `UnknownPotcarWarnings`, that was the internal PMG hashing which has been removed. Some versions of VASP POTCARs have a SHA56 tag in the header that VASP computed on their side. We don't have any SHA56-tagged POTCARs for me to test that the current validator works well in their place

esoteric-ephemera: Though honestly, given the comments on issue [#2895](https://github.com/materialsproject/pymatgen/issues/2895), it seems like the SHA56 has similar fragility as PMG's hashes. I would lean towards removing the SHA56 check

janosh: > given the comments on issue https://github.com/materialsproject/pymatgen/issues/2895, it seems like the SHA56 has similar fragility as PMG's hashes

that was my understanding as well

>  I would lean towards removing the SHA56 check

same!

janosh (approved):
I think this is great! 2 things in this PR:
1. less brittle way of checking for invalid POTCARs
2. cleaning up the `PotcarSingle` class to be less confusing due to
   - less duplication between `keywords` and (the no longer existent) `PSCTR`
   - using `@property` methods for hashes instead of direct class attributes

Thanks @esoteric-ephemera!

PR #3354, opened 2023-09-27
[WIP] Adding alias `.to_file()` for `.to()` method of site collections, structures and molecules

QuantumChemist:
## Summary

I have noticed that many beginners (including myself a while ago) find it very unintuitive to write e.g. structures to a file by using `.to()`, especially with the `.from_file()` method counterpart. That's why I refactored each instance/occurrence using the [pycharm](https://www.jetbrains.com/pycharm/) built-in refactoring feature to `.write_to_file()` and suggest this change in this PR.  Of course I made sure not to change any of the unit conversion `.to()` instances.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -213,7 +213,7 @@ structure = pmg.Structure(lattice, ["Cs", "Cl"], ...[[0, 0, 0], [0.5, 0.5, 0.5]]
 
 # You can create a Structure using spacegroup symmetry as well.
 li2o = pmg.Structure.from_spacegroup(
-    "Fm-3m", pmg.Lattice.cubic(3), ["Li", "O"], [[0.25, 0.25, 0.25], [0, 0, 0]]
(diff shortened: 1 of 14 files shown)


shyuep:
Sorry but we cannot do this. Whatever the merits of the renaming, this will cause all other packages that depends on pymatgen to break. Writing out files is probably one of the main uses. At best, we can allow write_file as an alias for to so that we preserve backwards compatibility.

QuantumChemist:
> Sorry but we cannot do this. Whatever the merits of the renaming, this will cause all other packages that depends on pymatgen to break. Writing out files is probably one of the main uses. At best, we can allow write_file as an alias for to so that we preserve backwards compatibility.

Sure, I can totally understand that! Then I will rewrite it that it is simply added as alias!

QuantumChemist:
I have to close this PR and will open a new one....

PR #3356, opened 2023-09-27
Add alias `.to_file()` for `.to()` method of structures and molecules

QuantumChemist:
## Summary

I have noticed that many beginners (including myself a while ago) find it very unintuitive to write e.g. structures to a file by using .to(), especially with the .from_file() method counterpart. That's why I added aliases `to_file` for structures and molecules.

PS: This is the second attempt of #3354 as something went wrong there...


pymatgen/core/structure.py
```diff
@@ -2713,6 +2713,10 @@ def to(self, filename: str = "", fmt: str = "", **kwargs) -> str:
             writer.write_file(filename)
         return str(writer)
 
+    def to_file(self, filename: str = "", fmt: str = "", **kwargs) -> str:
+        """An alias for .to() for more user-friendliness."""
+        return self.to(filename, fmt, **kwargs)
```
janosh, line 2718: I think we can move this to `SiteCollection` so only need a single alias and forward both `*args` and `**kwargs`?

shyuep:
You don't actually need to write it this way. All that needs to be done is to add `to_file = to` and `write_file = to` in SiteCollection. That will create the aliases without so much code.

QuantumChemist:
> You don't actually need to write it this way. All that needs to be done is to add `to_file = to` and `write_file = to` in SiteCollection. That will create the aliases without so much code.

I know, that's how I started, but when I tried to call the alias in pycharm, it wouldn't give me the input hints and at least for me they are really important. If you think that's not necessary, then I can change it to the simple alias.

janosh (changes requested):
We definitely need a test for the new alias.

QuantumChemist:
> We definitely need a test for the new alias.

Aye aye! 🫡

QuantumChemist:
hi @janosh I hope the tests are ok like that?

janosh (approved):
Thanks @QuantumChemist! 👍

PR #3359, opened 2023-09-28
Don't rely on `jsanitize` in `Atoms` <--> `Structure` object interconversion 

### LINKED ISSUE #3358: `jsanitize` has downstream issue when converting ase `Atoms` to pymatgen `Structure` in `multiprocessing`
I just merged this branch to my local and found that `jsanitize` breaks my code where I have `torch.multiprocessing` to convert `ase.Atoms` into `pymatgen.core.Strucuture`.

```
  File "/global/homes/c/cyrusyc/.conda/envs/chace/lib/python3.9/site-packages/monty/json.py", line 585, in jsanitize
    return [jsanitize(i, strict=strict, allow_bson=allow_bson, enum_values=enum_values) for i in obj.tolist()]
TypeError: 'float' object is not iterable
```

The info in my ase.Atoms is empty so it might be a general problem.

The detailed information below

              @janosh: Just wanted to point out one subtlety to you. Currently, we have the mapping: `Atoms.info <--> Structure.properties`, since they serve the same purpose between ASE and Pymtatgen, respectively. However, I `jsanitize`-d the `atoms.info` before assigning it to `.properties` to ensure the `Structure`/`Molecule` object can be (de)serialized since this is often an implicit assumption of `Structure`/`Molecule` objects but is not necessarily one for ASE `Atoms` objects. In practice, this means that there could be some scenarios where a user interconverts between ASE `Atoms` and Pymatgen `Structure`/`Molecule` with some loss of information; namely, if the `atoms.info` is modified by `jsanitize()` then it can't be restored in its original form. There might be a better solution than this, but I implemented it because oftentimes ASE will dump un-serializable class objects (namely, a `Spacegroup` class) into `atoms.info`, such as when reading in a CIF with symmetry flags.

_Originally posted by @Andrew-S-Rosen in https://github.com/materialsproject/pymatgen/pull/3270#discussion_r1312508843_

Andrew-S-Rosen:
Closes #3358.

As raised in #3270, I stated:

> Currently, we have the mapping: `Atoms.info <--> Structure.properties`, since they serve the same purpose between ASE and Pymtatgen, respectively. However, I `jsanitize`-d the `atoms.info` before assigning it to `.properties` to ensure the `Structure`/`Molecule` object can be (de)serialized since this is often an implicit assumption of `Structure`/`Molecule` objects but is not necessarily one for ASE `Atoms` objects. In practice, this means that there could be some scenarios where a user interconverts between ASE `Atoms` and Pymatgen `Structure`/`Molecule` with some loss of information; namely, if the `atoms.info` is modified by `jsanitize()` then it can't be restored in its original form. There might be a better solution than this, but I implemented it because oftentimes ASE will dump un-serializable class objects (namely, a `Spacegroup` class) into `atoms.info`, such as when reading in a CIF with symmetry flags.

I now have avoided the needed to call `jsanitize` so that we don't lose any info between interconversion. To do this, I have "manually" JSON-ified and reconstructed the ASE `Spacegroup` object if present. I also did a much-needed general cleanup of both the class and test suite.


Andrew-S-Rosen:
@chiang-yuan, @janosh: This is ready for your review.

tests/io/test_ase.py
Andrew-S-Rosen, line 1: This looks like a lot of changes, but all I did was:

- Remove the unnecessary class
- Add a `pytest.importorskip` line.
- Make use of `pytest.mark.parameterize` to get rid of `for` loops.

chiang-yuan:
Just want to offer thanks for this! @Andrew-S-Rosen you are so swift on this! (even on PTO 🫡

Andrew-S-Rosen:
Had a spare moment and couldn't get it off my mind. I can go back to the beach life now. 😎🏖️

PR #3362, opened 2023-09-30
Fix `TestMPResterNewBasic` + `AseAtomsAdaptor` test errors and `TransformedStructure.from_snl` overwriting `hist` variable

janosh:
d919b591a fix typos
c1c9a1041 fix for loop overwriting hist variable in TransformedStructure.from_snl
596a56eea fix TestVasprun.test_update_potcar using wrong PotcarSingle hash: md5_computed_file_hash should be md5_header_hash

### DIFF
--- a/pymatgen/io/ase.py
+++ b/pymatgen/io/ase.py
@@ -8,10 +8,10 @@ from __future__ import annotations
 
 import warnings
 from collections.abc import Iterable
+from importlib.metadata import PackageNotFoundError
 from typing import TYPE_CHECKING
 
 import numpy as np
-from monty.dev import requires
 
 from pymatgen.core.structure import Molecule, Structure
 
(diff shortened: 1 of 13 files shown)


pymatgen/io/ase.py
```diff
@@ -40,7 +40,6 @@
 
 # NOTE: If making notable changes to this class, please ping @Andrew-S-Rosen on GitHub.
 # There are some subtleties in here, particularly related to spins/charges.
-@requires(ase_loaded, "ASE needs to be installed.")
```
janosh, line 43: @Andrew-S-Rosen @shyuep I had to revert the `@requires` decorator added to `AseAtomsAdaptorin` in #3359 because it breaks the `staticmethod` namespace resolution below Python 3.10.

Issue #3342, opened 2023-09-20
Support VASP 64 pseudopotentials


Andrew-S-Rosen:
### Problem

VASP has now released v64 pseudopotentials (PBE and LDA). This is an issue tracker to monitor progress related to ensuring that Pymatgen plays nicely with them.

### Proposed Solution

N/A

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

janosh:
Besides adding new POTCAR hashes (or whatever future validation technique we migrate to), are we expecting additional TODOs?

Andrew-S-Rosen:
Here are a few questions:

- Everything works as-expected with the `pmg config` command-line utility? 
- Parsing of the POTCAR spec in the `Potcar` class fetches all the metadata appropriately?

It's very possible that there are no changes needed. Part of this issue was also to decide if that's the case :)

mkhorton:
Is there a link to release notes from VASP on these pseudopotentials? I was hoping to find a news item or a wiki page or similar.

janosh:
If you find one, let me know. This is all i could find. The `README.UPDATES` file they mention offers no further details either.

![Screenshot 2023-09-26 at 10 37 33 AM](https://github.com/materialsproject/pymatgen/assets/30958850/47162a6c-d2d4-4ed4-b301-acefabba7e61)

mkhorton:
Thanks @janosh, I haven't found any mention myself but if I do I'll make sure to share.

mkhorton:
Seems the most recent update of https://www.vasp.at/wiki/index.php?title=Available_PAW_potentials&type=revision&diff=14415&oldid=12650 was the addition of the Yb_3 line in Jan 2022.

JaGeo:
Are the changes from the new "Cs_sv" on structural optimizations drastic? I guess we would expect the largest impact here.

mkhorton:
Why would you expect Cs_sv specifically to be largest impact @JaGeo? (I have no intuition here.)

JaGeo:
@mkhorton should have phrased it more carefully. "Cs_pv" and potentially all lanthanides but I assume that the "Cs_pv" potential is used more frequently.

esoteric-ephemera:
Is this closed by [PR #3370](https://github.com/materialsproject/pymatgen/pull/3370)?

Andrew-S-Rosen:
Sounds like it. Thanks!

janosh:
[...]

PR #3370, opened 2023-10-02
Update POTCAR summary stats to include 6.4 POTCARs and add `dev_script` utils for future updates

esoteric-ephemera:
### Major changes:

- Follow-up to issue [#3342](https://github.com/materialsproject/pymatgen/issues/3342) and closed PR [#3351](https://github.com/materialsproject/pymatgen/pull/3351)
- Adds support for version 64 POTCARs
- Adds hidden function for PMG devs to continuously update potcar_summary_stats.json.gz used to validate potcars, `pymatgen.io.vasp.inputs._gen_potcar_summary_stats`
- Added functionality to generate fake POTCARs from real POTCARs (by randomizing numeric and boolean values in them) for tests in `dev_scripts/potcar_scrambler.py`


pymatgen/io/vasp/inputs.py
```diff
@@ -2280,6 +2283,43 @@ def __repr__(self) -> str:
         return f"{cls_name}({symbol=}, {functional=}, {TITEL=}, {VRHFIN=}, {n_valence_elec=:.0f})"
 
 
+def _gen_potcar_summary_stats():
+    """
+    This function solely intended to be used for PMG development to regenerate the
+    potcar_summary_stats.json.gz file used to validate POTCARs
+    """
```
janosh, line 2290: Perfect! This is great to have. Even though it's marked as private, we should add a test for this.

janosh (changes requested):

pymatgen/io/vasp/inputs.py
```diff
@@ -2280,6 +2283,43 @@ def __repr__(self) -> str:
         return f"{cls_name}({symbol=}, {functional=}, {TITEL=}, {VRHFIN=}, {n_valence_elec=:.0f})"
 
 
+def _gen_potcar_summary_stats():
```
mkhorton, line 2286: This is fine here, but noting there is also a top-level `dev_scripts` folder (largely undocumented). On balance it's probably better here for discoverability for anyone curious how this file was generated.

Minor comments: the bare `return` is unnecessary, and `monty.serialization.dumpfn` can be used directly with `dumpfn(new_summary_stats, "filename.json.gz")` which is a convenience function we use a lot.

esoteric-ephemera: Thanks! Much better way of doing this. For placement, you and @janosh probably have a better sense of where this belongs. `_gen_potcar_summary_stats` is a "destructive" function, in that it overwrites the file that's distributed with PMG. `dev_scripts` might then be a better place for it

pymatgen/io/vasp/inputs.py
esoteric-ephemera: Added a `Test_gen_potcar_summary_stats` to `tests/io/vasp/test_inputs.py`. Requires generating a library of fake POTCARs, which are in `tests/fake_POTCAR_library`. Code to generate this library from a real library of POTCARs is in `dev_scripts/potcar_scrambler.py`

janosh (approved):
Really nice work, thanks @esoteric-ephemera! 👍

lan496:
Hi. I think `pymatgen/io/vasp/potcar_summary_stats.json.gz` should be added to `setup.package_data` to use it via pip-installed version.

https://github.com/materialsproject/pymatgen/blob/44bf38318f1ff4265385eb6587bb713780732ff7/setup.py#L95-L119

janosh:
Ah crap!

mkhorton:
Not the first time don’t worry! I once added a catch-all to package all .json, .yaml files by default but this was vetoed for reasons I forget, but perhaps worth reconsidering

janosh:
> I once added a catch-all to package all .json, .yaml files by default but this was vetoed for reasons I forget, but perhaps worth reconsidering

I think it's worth a try. @shyuep Any concerns?

If we get false positives, we can always revert the change.

shyuep:
My concern is that data files should be few and far between and EXPLICIT.

janosh:
> My concern is that data files should be few and far between and EXPLICIT.

Exactly, which is why the ones we deliberately place in the pymatgen pkg dir should be there for a good reason and hence also be included in the PyPI release. At least that's the case right now as you can see from [d8151e5](https://github.com/materialsproject/pymatgen/pull/3374/commits/d8151e5de2c0c90ed5b28375d21bc573dd778dfd) passing.

shyuep:
Yes, that is why it should be explicitly specified. For the same reason, you do not usually do `from package import *`.

janosh:
The biggest reason not to do `from package import *` imo is that it breaks tooling. IDEs can trace where things come from so no doc string glancing, no type hints and worse code completions. But none of those apply here, there's no tooling for pkg data anyway. So far, there's been a 1-to-1 map from data files in `pymatgen` to things included in `SOURCES.txt` so we might as well make the map automatic imo.

shyuep:
There are reasons. Say someone decides to add a `very_big_file.yaml` for some internal testing purposes. That file will automatically be added to package data, even if it has no real purpose in actual function of the main code. Like I said, it is better to be explicit what files are needed and what are not.

janosh:
I don't merge very big files into `master`. Not even in test files, let alone in pkg dir. Test files should never be placed in pkg dir.

PR #3381, opened 2023-10-06
Update `incar_parameters.json` to allow `ISIF=8`

matthewkuner:
simple change to include new ISIF=8 tag in list of valid tags


janosh (approved):
Thanks @matthewkuner. 👍

PR #3389, opened 2023-10-08
Add `SiteCollection.to_ase_atoms()`

### LINKED ISSUE #3385: Make a `structure.to_ase_atoms()` method
> The ASEAdaptor was written a long time ago. The class is purely for organizational purposes. How about we do this - go ahead and move all relevant code to the Structure, Molecule or SiteCollection class as methods from_ase_atoms and to_ase_atoms. We will keep ASEAdaptor and it will be modified simply to be `return structure.to_ase_atoms()` for example. This is just for backwards compatibility since there are codes that already use ASEAdaptor (many ML codes being one example). 
It is a purely copy and paste operation, with some minor refactoring.

_Originally posted by @shyuep in https://github.com/materialsproject/pymatgen/issues/3358#issuecomment-1738393284_

janosh:
Closes #3385.

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -828,7 +828,7 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
         raise ValueError(f"Unknown {calculator=}.")
 
     def to_ase_atoms(self, **kwargs) -> Atoms:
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
Thank you, @janosh!

PR #3393, opened 2023-10-10
Add type annotations to phonon

ab5424:
## Summary

Major changes:

- Added some type annotations, mainly to `phonon/*.py`

## Checklist


janosh (approved):
Thanks a lot! 👍

pymatgen/phonon/dos.py
```diff
@@ -57,7 +63,7 @@ def get_smeared_densities(self, sigma):
 
         return gaussian_filter1d(self.densities, sigma / avgdiff)
 
-    def __add__(self, other):
+    def __add__(self, other: Self) -> Self:
```
janosh, line 66: Sorry, I missed that you're using `Self` in many places. [PEP 673](https://peps.python.org/pep-0673) is Python 3.11 only afaik so we can't use that syntax for another 2 years or so. Could you revert or use the actual class as the type hint instead?

ab5424: Sorry, I didn't realise. Will fix it.

ab5424: I created a new pull request (#3401), I didn't know if it is possible to reopen here...

Would it make sense to also run the linting wf on 3.9 then? To prevent this?

janosh: > Would it make sense to also run the linting wf on 3.9 then? To prevent this?

Yes, feel free to change that in #3401.

Issue #3322, opened 2023-09-13
Changing MP Input Sets


matthewkuner:
There have been internal talks with @munrojm @Janosh @esoteric-ephemera, among others, about changing the MP Input Sets. We want to open the discussion up before moving forward with such changes. 

Currently, we are in favor of the following changes:

For **all** MP sets:
- Set LMAXMIX = 6 for all structures. This is based on a benchmarking study @esoteric-ephemera performed (attached below) which showed that the VASP manual recommendations for setting LMAXMIX based on the element block is actually _not_ sufficient in all cases.

For MPRelaxSet (and the MPMetalRelaxSet):
- Use a force-based convergence criterion.` EDIFFG = -0.05` is standard. We could also use `-0.02` to match the value from the MPScanRelaxSet
- Get rid of EDIFF_PER_ATOM and instead replace it with a flat value. `EDIFF = 1e-5` is a reasonable starting point for discussion.
- Include additional functionality similar to the MPSCANRelaxSet wherein certain parameters are set according to bandgap, if known (see https://github.com/materialsproject/pymatgen/blob/f90929f75b27fe1d06faa4e2a5f44654009b98b0/pymatgen/io/vasp/sets.py#L950). This would likely involve merging the MPMetalRelaxSet and MPRelaxSet.
- Change the ENCUT and ENAUG of MPRelaxSet to match the r2SCAN values of 680 and 1360, respectively.
- Set `LREAL=False`, as `LREAL=Auto` seems to be less reliable for structures further from equilibrium (regardless of # of sites)

Here is the current draft of @esoteric-ephemera's benchmarking study: [bench_vasp_pars.docx](https://github.com/materialsproject/pymatgen/files/12600987/bench_vasp_pars.docx)

@shyuep see the final section of the attached doc regarding LREAL (which we discussed earlier today)

shyuep:
Thanks. I have no problems with most of this, but I would point out a few issues. 

[...]

PR #3405, opened 2023-10-12
[WIP] Change MP input sets

matthewkuner:
Follow-up to #3322.

- Change `LMAXMIX` to 6 for all MP input sets
- Change `LREAL` to `False` for MPRelaxSet and MPScanRelaxSet
- add logic to set `EDIFF = min(1e-4, NSITES * EDIFF_PER_ATOM)` as per [@esoteric-ephemera ](https://github.com/materialsproject/pymatgen/issues/3322#issuecomment-1723907332)

TODO:
- add logic for setting `ISMEAR`, `SIGMA`, and `KPOINTS`/`KSPACING` based on bandgap.  Are we considering changing from KPOINTS to KSPACING like in the r2SCANRelaxSet?
- Add a `bandgap.setter` like in https://github.com/materialsproject/pymatgen/blob/15dc77da508c6824a2e7cc41e983f797c64815c4/pymatgen/io/vasp/sets.py#L397 This would enable bandgap to be set later in the same way as Structures (UPDATE: @utf said to leave this detail out of this PR, as he will address it in a future PR.)

@janosh @munrojm @esoteric-ephemera

### DIFF
--- a/pymatgen/io/vasp/inputs.py
+++ b/pymatgen/io/vasp/inputs.py
@@ -584,6 +584,8 @@ class Poscar(MSONable):
 
         return "\n".join(lines) + "\n"
 
+    get_string = get_str
+
     def __repr__(self):
         return self.get_str()
 


matthewkuner:
What should the new logic be for setting ISMEAR/SIGMA/KPOINTS? @esoteric-ephemera 

Maybe something like this?
```
if self.bandgap == None:
    self._config_dict["INCAR"].update({"ISMEAR": 0, "SIGMA": 0.05})
    self._config_dict["KPOINTS"]["reciprocal_density"] = 200 #kpoint density for metals (higher)
elif self.bandgap < 1e-4:
    self._config_dict["INCAR"].update({"ISMEAR": 1, "SIGMA": 0.2})
    self._config_dict["KPOINTS"]["reciprocal_density"] = 200 #kpoint density for metals (higher)
else:
    self._config_dict["INCAR"].update({"ISMEAR": 0, "SIGMA": 0.05})
    self._config_dict["KPOINTS"]["reciprocal_density"] = 64 #kpoint density for nonmetals (lower)
```

janosh:
Do we want to change the input sets in place or create new versions? I think new versions are better.

shyuep:
Just change the existing sets.

matthewkuner:
I agree w/ changing existing sets. Otherwise we would have >10 different versions of MPRelaxSet

janosh:
We also have 377 versions of `pymatgen`.

PR #3420, opened 2023-10-24
Drop `black` for `ruff format`

janosh:
`ruff format` [was just released](https://astral.sh/blog/the-ruff-formatter) as part of [v0.1.2](https://github.com/astral-sh/ruff/releases/tag/v0.1.2). It's in preview but considered production-ready. About 30x faster than `black` and 99% compatible (counted by lines of code with conflicting format in empirical testing). Appears to be much more than 99% compatible in `pymatgen`.


janosh:
You know most of these are auto-changes? Just ran `time black pymatgen --check`. Takes 2 sec on my machine and several times that to install in CI.

janosh:
`ruff` has ~10x the development speed of `black`, is committed to supporting `mojo` (which imo is the most promising avenue for making `pymatgen` more performant in the future). `black` will likely never support `mojo` since it's written in Python and so relies on Python to parse Python. Dropping it means one less tool for contributors to install and know about. I also noticed `ruff-format` is faster and appears to be less buggy than `black` in [auto-format on save in VS Code](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff&ssr=false#overview).

shyuep:
I don't have a problem with this change being made. So go ahead. I am merely pointing out that there are bigger things to deal with.

PR #3422, opened 2023-10-25
Add functionality to read ASE *.traj file in Trajectory class method from_file()

exenGT:
Summary:
Add functionality to read ASE *.traj file in Trajectory class method from_file(), located in core/trajectory.py.

Purpose:
As ASE *.traj file is a standard format for generating and storing relaxation / MD trajectories, it would be nice to have a method which can read a Pymatgen Trajectory instance from an ASE *.traj file.


pymatgen/core/trajectory.py
```diff
@@ -542,15 +544,30 @@ def from_file(cls, filename: str | Path, constant_lattice: bool = True, **kwargs
...
         else:
             supported = ("XDATCAR", "vasprun.xml")
             raise ValueError(f"Expect file to be one of {supported}; got {filename}.")
 
-        return cls.from_structures(
-            structures,
-            constant_lattice=constant_lattice,
-            **kwargs,
-        )
+        if "structures" in locals():
```
janosh, line 559: Using `locals()` here is unnecessary. Cheaper to define `is_struct_traj = any(ase_traj[0].pbc)` and return based on that.

janosh (changes requested):
We need a test for this. Could you upload an ASE traj file and read it in `tests/core/test_trajectory.py`?

pymatgen/core/trajectory.py
exenGT: Here I think the reason I use `locals()` is because `from_file()` can also read other file formats (`XDATCAR` and `vasprun.xml`), and `any(ase_traj[0].pbc)` does not apply to those cases. Is there a better way to handle this?

janosh: Using `locals` seems too brittle. How about you call it `structures` either way and define  `is_mol = False` at the top of the method and then modify it later as needed?

exenGT: Thanks! Just commited a new update accordingly. (The name should be "Add flag `is_mol`", somehow git doesn't recognize the quotation marks)

pymatgen/core/trajectory.py
```diff
@@ -12,10 +12,12 @@
 from typing import Any, Union
 
 import numpy as np
+from ase.io.trajectory import Trajectory as ASE_Trajectory
```
Andrew-S-Rosen, line 15: ASE is not a core requirement of pymatgen, so you can't do this.

pymatgen/core/trajectory.py
```diff
@@ -537,20 +539,36 @@ def from_file(cls, filename: str | Path, constant_lattice: bool = True, **kwargs
...
-            supported = ("XDATCAR", "vasprun.xml")
+            supported = ("XDATCAR", "vasprun.xml", "*.traj")
             raise ValueError(f"Expect file to be one of {supported}; got {filename}.")
 
-        return cls.from_structures(
-            structures,
-            constant_lattice=constant_lattice,
-            **kwargs,
-        )
+        if not is_mol:
```
Andrew-S-Rosen, line 561: You should do `if is_mol` (return from_molecules) and then `else` (return from_structures). Makes more sense to start with a positive logic statement rather than a negative.

Andrew-S-Rosen (commented):
Some small comments.

pymatgen/core/trajectory.py
exenGT: Oops... Is there an alternative option for this?

Andrew-S-Rosen: Yes, move the import statement to this line: https://github.com/materialsproject/pymatgen/pull/3422/files#diff-4330101375652bb97d6774e06676d30a6c51537d6df49f01eba66863561839d9R549, within the function and immediately abovve where it's being called. You may also wish to wrap it within a `try`/`except ImportError` block to raise a helpful error message if the import statement does not work.

pymatgen/core/trajectory.py
```diff
@@ -537,20 +539,36 @@ def from_file(cls, filename: str | Path, constant_lattice: bool = True, **kwargs
...
+            return cls.from_structures(
+                structures,
+                constant_lattice=constant_lattice,
+                **kwargs,
+            )
+        else:
+            return cls.from_molecules(
+                molecules,
+                **kwargs,
+            )
```
Andrew-S-Rosen, line 571: As noted in the CI linter, you do not need to have an `else` statement with a return. You can just do the return.

tests/core/test_trajectory.py
```diff
@@ -471,3 +471,7 @@ def test_xdatcar_write(self):
         written_traj = Trajectory.from_file("traj_test_XDATCAR")
         self._check_traj_equality(self.traj, written_traj)
         os.remove("traj_test_XDATCAR")
+
+    def test_read_file(self):
```
janosh, line 477: best match the method name in the test name: i.e. `test_from_file`. Also, maybe check the traj length and composition of the 1st frame for extra safety.

janosh (changes requested):
Thanks @exenGT, almost ready to go! 👍

janosh (changes requested):
The test is failing due to file not found. Also, could you compress `tests/files/LiMnO2_chgnet_relax.traj`?

PR #3424, opened 2023-10-25
Allow BaderAnalysis to load from existing results of bader analysis

JiQi535:
## Summary
This PR reformats the __init__ function of BaderAnalysis class and adds new functionality without changing its default behaviors. 

## Major changes:
1. Allow users to load from existing bader charge analysis results from folder. This means that when users already have bader analysis results, pymatgen does not need to call bader executable, which was previously not available. This new functionality can make the analysis and result storage more convenient.

2. Reformat the __init__ function by separating different analysis code chunks to functions with readable names, i.e., _load_charge_data and _load_bader_data.  

## Thoughts
I think the ideal case should be like the EnumlibAdaptor, another command_line class in pymatgen. It does not run the external executable with __init__ function. But I do not want to change any default behaviors of BaderAnalysis, as I expect workflows out there are using this current version. So this PR aims for a pure reformating and addition of functionality without changing any default behavior.

@janosh @shyuep Please kindly review. Thanks.

### DIFF
--- a/pymatgen/analysis/pourbaix_diagram.py
+++ b/pymatgen/analysis/pourbaix_diagram.py
@@ -287,7 +287,7 @@ class MultiEntry(PourbaixEntry):
         # Attributes that are weighted averages of entry attributes
         if attr in ["energy", "npH", "nH2O", "nPhi", "conc_term", "composition", "uncorrected_energy", "elements"]:
             # TODO: Composition could be changed for compat with sum
-            start = Composition({}) if attr == "composition" else 0
+            start = Composition() if attr == "composition" else 0
             weighted_values = (getattr(entry, attr) * weight for entry, weight in zip(self.entry_list, self.weights))
             return sum(weighted_values, start)
 
--- a/pymatgen/analysis/reaction_calculator.py
+++ b/pymatgen/analysis/reaction_calculator.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import re
 from itertools import chain, combinations
+from typing import TYPE_CHECKING, no_type_check
 
 import numpy as np
 from monty.fractions import gcd_float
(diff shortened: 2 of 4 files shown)


shyuep:
This breaks legacy code. We don't do that. I would rather you create a separate BaderData class that stores the analysis and you can write the init whatever way you want. 

Basic principle of writing a code base used by others - don't break stuff unless you have no choice.

JiQi535:
@shyuep Thanks for the quick feedback. Would you let me know how these edits would break legacy codes and workflows? I deliberately avoid any possible changes to default behaviors.

shyuep:
I understand that. But this is not a good API design.

janosh:
I think we need to give a bit more specific feedback on what needs to change to get this merged. From my side, if we can now load an existing Bader analysis without needing to run the `bader` exectuable we definitely need a test for that functionality.

JiQi535:
Thanks for the feedback. @janosh  I would be happy to add a test for this functionality. 
And I mostly agree with @shyuep that as a command_line caller (an API), the BaderAnalysis class might not be the most appropriate place to get bader analysis results (ACF.dat, etc.) loaded and processed. Maybe I can define the "separate BaderData class" also in this bader_caller.py?

shyuep:
I would define it simply as ChargeData object in core. For now, this will be updated from a BaderAnalysis. But as a more general principle, since the output is mostly the charge on each site, isn't it better to decorate a structure with site property "bader_charges", which can also be serialized? I think as a general principle, that should be the way this ways?

JaGeo:
> I would define it simply as ChargeData object in core. For now, this will be updated from a BaderAnalysis. But as a more general principle, since the output is mostly the charge on each site, isn't it better to decorate a structure with site property "bader_charges", which can also be serialized? I think as a general principle, that should be the way this works for all kinds of structure-based anlysis? That's why we expanded site_properties. @mkhorton thoughts?

We also have charges from Lobster and provide a decorated structure. Maybe interesting to unify this in pymatgen.

shyuep:
@JaGeo Agreed. I think this is actually the best way forward. Everything that is site-wise goes into Structure (or more generally, SiteCollection), which already supports all forms of serialization and deserialization. Unification would also make it a lot easier to develop featurization approaches based on Structure/Molecule.

PR #3428, opened 2023-10-27
Add lattice velocities to Poscar

gpetretto:
## Summary

NpT MD simulation in VASP provide additional information about lattice dynamics in the CONTCAR that are used when restarting.

In order to allow restarting of generic MD simulations in workflows (https://github.com/materialsproject/atomate2/pull/489) I added the parsing and writing of this section to the `Poscar` object.

PR #3433, opened 2023-10-30
Fix lattice velocities formatting

gpetretto:
## Summary

This PR fixes a problem in #3428 that I encountered while performing some more tests with NpT MD simulations: in the case of lattice velocities VASP seems to be strict in the format when reading the POSCAR. After some tests it seems that this should be the best format to allow the correct parsing of this quantity in VASP.

Sorry for the error in the previous PR.


pymatgen/io/vasp/inputs.py
```diff
@@ -545,7 +547,8 @@ def get_str(self, direct: bool = True, vasp4_compatible: bool = False, significa
                 lines.append("Lattice velocities and vectors")
                 lines.append("  1")
                 for v in self.lattice_velocities:
-                    lines.append(" ".join(format_str.format(i) for i in v))
+                    # vasp is strict about the format when reading this quantity
+                    lines.append(" ".join(f" {i: .7E}" for i in v))
```
janosh, line 551: If VASP is strict about this, could you add a test that checks for the right float precision?

janosh (changes requested):

gpetretto:
Hi @janosh, I added a check on the format of the produced POSCAR. It should be noted that in this case it just checks that the format is equivalent to the one written from the `Poscar` object. This seems to be the only option that is correctly parsed by VASP, even though it is not exactly equilant to what is produced by fortran, since in the scientific notation python has a number >0 as the first digit, while the floats written from fortran start with a 0. 
This will prevent regressions, but does not strictly check the correctness of the format. I have found one library that reproduces the i/o operations as done in fortran: https://github.com/brendanarnold/py-fortranformat. This would provide a more stringent test, but I suppose it might not be worth to add this as a dependency (even though only for dev) just for this.

janosh (approved):
Thanks @gpetretto! 👍

Issue #3437, opened 2023-10-31
It seems that there is a bug to write the lobsterin file


RedStar-Iron:
https://github.com/materialsproject/pymatgen/blob/15dc77da508c6824a2e7cc41e983f797c64815c4/pymatgen/io/lobster/inputs.py#L271-L272

`key` in `self.get(key)` should be lower, then it can be found in Lobsterin.FLOAT_KEYWORDS.

And the cohpsteps might need to be a positive integer.

So I just revised it to:
```
if key.lower() == "cohpsteps":
    f.write(f"{key} {int(self.get(key.lower()))}\n")
else:
    f.write(f"{key} {self.get(key.lower())}\n")
```

JaGeo:
Thanks for looking at the Lobster part of the code. @naik-aakash and I will check your suggestion carefully.

JaGeo:
I think the bug is coming from the fact that the overwriting of the dict. We will look into it more carefully to not break other parts of the code!

naik-aakash:
Hi @RedStar-Iron, Thanks for the suggestion. Can you also please post the steps to reproduce the issue? It would help me address it properly

JaGeo:
Yup, I have also already reproduced the issue and have started with a fix. Will start the pull request in a WIP minute.

JaGeo:
@RedStar-Iron we have worked on a fix for both of your issues. However, we investigated that it does not matter if cohpsteps is a float or int (see here https://github.com/materialsproject/pymatgen/pull/3439#issuecomment-1791469518)

Issue #3438, opened 2023-10-31
This part might not be robust.


RedStar-Iron:
https://github.com/materialsproject/pymatgen/blob/eee88ab63b66a0b92757524975c7290838b4bf01/pymatgen/io/lobster/inputs.py#L584-L589

It will fail if there is a '\t' in my lobsterin file. 

I recommand to use the `re`.

```
raw_datum = re.findall(r'(?i)\b[a-z]+(?:\d+)?\b|\b\d+[a-z]+\b|[-]?\d*\.?\d+', datum)
```

Above code might meet the requirement.

Andrew-S-Rosen:
Tagging @JaGeo.

JaGeo:
We are on it as a part of https://github.com/materialsproject/pymatgen/pull/3439. Thank you, @Andrew-S-Rosen !

JaGeo:
@janosh can also be closed. It was adressed as a part of #3439

PR #3439, opened 2023-11-01
Fix lobsterin dict inheritance and treat \t in lobsterins correctly

JaGeo:
## Summary

Closes https://github.com/materialsproject/pymatgen/issues/3437 and https://github.com/materialsproject/pymatgen/issues/3438.
Inheritance of the dict was not carried out correctly. I have started here with inheriting from UserDict. We will add more tests to avoid similar issues in the future.


JaGeo:
@naik-aakash , I have started with the fix. I will test methods like update, delete as well. 

We could also directly implement a fix for reading lobsterins with \t

naik-aakash:
Thanks! Thats great. So you will also fix the `reading lobsterins with \t` issue in this PR itself?

JaGeo:
@naik-aakash You are invited to contribute to this PR. I am in the train and I don't really expect to be able to work much 😅

naik-aakash:
> @naik-aakash You are invited to contribute to this PR. I am in the train today and I don't really expect to be able to work much on the remote connection 😅

Ah okay. Just got the invitation. Thank you 😄

JaGeo:
I think I have now correctly implemented the inheritance from a UserDict. All standard functionalities work.

Things left to do:
- clarify how lobster reads cohpsteps (int, float): i.e., should we include a new category of lobster arguments?
-  treat \t in lobsterins incluing whether Lobster can process them and test every different functionality

naik-aakash:
I think treating `\t` is solved as well, have updated the existing test file to check the implementation and it seems to work fine

JaGeo:
Update: lobster can read floats and integers. It will automatically round to the next larger integer independent on whether it is a float or an integer (in version 4.1.0 and 5.0.0). I think it doesn't really matter that we print a float... Also, I think this behavior of Lobster is a bit unexpected.

JaGeo:
@janosh , I think this would be ready for review in case you have time.

janosh (approved):

PR #3447, opened 2023-11-01
Refresh OPTIMADE aliases and update docstrings

ml-evs:
## Summary

*Apropos* of nothing in particular... here is PR that makes some updates to the OPTIMADE client (and maybe starts a discussion about its future).

Major changes:

- Refreshes the baked-in OPTIMADE providers list with the latest new APIs
- Expands upon some of the docstrings

Happy to talk about how we could integrate the models and client at optimade-python-tools if that would be preferable. It would allow for async pulling data from all databases, pre-flight validating filters and validating responses (dangerous stuff for some databases...) and the code here could focus on just making sure the returned objects work with pymatgen. Should be releasing o-p-t v1.0 (https://github.com/Materials-Consortia/optimade-python-tools/issues/1831) before the end of the year which will support pydantic v2 and hopefully have better isolation, so wouldn't require many additional deps (httpx for async requests, Lark for parsing filters: see [here](https://github.com/Materials-Consortia/optimade-python-tools/blob/master/requirements.txt))

## Checklist

### DIFF
--- a/pymatgen/ext/optimade.py
+++ b/pymatgen/ext/optimade.py
@@ -16,9 +16,6 @@ from pymatgen.core.structure import Structure
 from pymatgen.util.due import Doi, due
 from pymatgen.util.provenance import StructureNL
 
-# from retrying import retry
-
-
 # TODO: importing optimade-python-tool's data structures will make more sense
 Provider = namedtuple("Provider", ["name", "base_url", "description", "homepage", "prefix"])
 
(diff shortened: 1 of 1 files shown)


janosh:
Thanks @ml-evs! 👍 

I expect new deps, esp. `pydantic` might be a blocker. We might consider a soft dependency that prompts the user to `pip install pydantic` when actually using functionality that requires it.

ml-evs:
> Thanks @ml-evs! 👍
> 
> I expect new deps, esp. `pydantic` might be a blocker. We might consider a soft dependency that prompts the user to `pip install pydantic` when actually using functionality that requires it.

Haha yes, after my strop about pymatgen indirectly needing pydantic a few months ago, I don't want to be blamed for reintroducing it :P Thankfully the client can be used separately from pydantic so with a bit more separation of concerns at our end we could make it a standalone installable. I'll have a think about it -- not necessary for this PR of course.

PR #3452, opened 2023-11-03
Fix Vasprun not interpreting float overflow as nan

tawe141:
VASP will sometimes print "***********" when there's a float overflow. There's already a function in here to interpret these as `np.nan`, but it looks like it wasn't applied everywhere. I found one place where it wasn't used (ie just using Python's `float` function) that helped me process a large batch of `vasprun.xml`s correctly, but maybe the same function can be used elsewhere.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -1302,7 +1302,7 @@ class Vasprun(MSONable):
 
     def _parse_calculation(self, elem):
         try:
-            istep = {i.attrib["name"]: float(i.text) for i in elem.find("energy").findall("i")}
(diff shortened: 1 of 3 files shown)


Andrew-S-Rosen:
Good to see you around these parts, @tawe141!!!

janosh (changes requested):
Thanks @tawe141! This should go along with a new test that applies `_parse_calculation` to a file containing a VASP overflow.

tawe141:
Added test and confirmed it passes on my end!

janosh (approved):
Thanks a lot for fixing @tawe141! 👍

PR #3456, opened 2023-11-04
Add a save parameter for chargemol calculation to select whether save the output files, use "link" rather than "copy" for the input file of chargemol calc, add mpi function to run chargemol.

RedStar-Iron:
## Revised Content

1. Add `save` parameter to control whether save the output files from `chargemol` calculation
2. Use `link` rather than `copy` for `CHGCAR, POTCAR, ARCCAR0 and ARCCAR2` during the calculation
3. Support `mpirun` for `chargemol` calculation

### DIFF
--- a/pymatgen/command_line/chargemol_caller.py
+++ b/pymatgen/command_line/chargemol_caller.py
@@ -196,14 +196,14 @@ class ChargemolAnalysis:
         """
         if mpi:
             if ncores:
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
@RedStar-Iron: Apologies, but I unfortunately don't have time to look at this PR. @chiang-yuan and/or @janosh might be interested, however.

shyuep:

PR #3458, opened 2023-11-05
Issue strong warning if `bader` is run without the `AECCAR`s

### LINKED ISSUE #3448: [Feature Request]: Clear warning if Bader is run without the AECCARs as reference
### Problem

In pymatgen, it's possible to run Bader both with and without the AECCARs as reference. However, in the latter case, the results of a Bader analysis can be severely inaccurate. I don't have statistics off-hand, but U had discussed this with _someone_ in Kristin's group (can't remember who... @janosh?) where we had a few systems with completely bizarre charges that were entirely resolved when using the AECCARs as reference. Anyway, it might be worth making this clearer because many people seem to not be aware of this and run Bader with just the CHGCAR even when they have valuable AECCARs lying around...

In `bader_analysis_from_path`, a warning is already raised to address this:

https://github.com/materialsproject/pymatgen/blob/f15f4b231668970c827964601b77def64b392af0/pymatgen/command_line/bader_caller.py#L460-L464

However, if someone is using the `BaderAnalysis` class, there's no similar information in the docstring or source code itself.

### Proposed Solution

Either log a warning or (if that's too intrusive), make it much clearer in the docstring.

It should be pretty easy to address this, and I'm happy to open the PR for it, but I wanted to use the issue report to ask what people suggest the ideal approach should be.

CC @chiang-yuan.

### Alternatives

_No response_

janosh:
Closes #3448.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.3
+    rev: v0.1.4
     hooks:
       - id: ruff
         args: [--fix]
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -371,8 +371,7 @@ class BaderAnalysis:
         to perform Bader analysis.
 
         Args:
-            path (str): Name of directory where VASP output files are
-                stored.
+            path (str): Name of directory where VASP output files are stored.
             suffix (str): specific suffix to look for (e.g. '.relax1' for 'CHGCAR.relax1.gz').
 
         Returns:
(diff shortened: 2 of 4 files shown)


JiQi535 (commented):
I'm quite sure that this PR is causing the "bader_analysis_from_path" method to fail for any of its usage. I'm very confused how could this PR pass the tests without failing. 

The problematic line is line 455 in this PR and now line 451 in the latest pymatgen. The method is now looking for CHGCAR in the "Could not find CHGCAR!" folder path...

I think the tests and this code should be corrected.
https://github.com/materialsproject/pymatgen/blob/5514ff5418b23da809b23056ea876c0a4c244c96/pymatgen/command_line/bader_caller.py#L438-L452

janosh:
@JiQi535 you're right, thanks for reporting. the reason this wasn't caught is that the only test for `bader_analysis_from_path` is skipped

https://github.com/materialsproject/pymatgen/blob/5514ff5418b23da809b23056ea876c0a4c244c96/tests/command_line/test_bader_caller.py#L84-L99

PR #3464, opened 2023-11-09
Module-scoped imports

janosh:
Follow up to #3462.


shyuep:
There are good reasons to not use module scope in many instances. E.g., when a function is rarely used.

janosh:
There are 2 more conditions for it to make sense to have function-scoped imports. The import must be expensive and not already in the import tree. I'm looking out for that in these refactors but those 3 conditions almost never coincide.

PR #3465, opened 2023-11-09
`ChargemolAnalysis` auto-download atomic densities if not found

janosh:
Related to https://github.com/materialsproject/atomate2/pull/587 and https://github.com/materialsproject/emmet/pull/869.

Pinging @chiang-yuan in case you want to take this over. The tests are still WIP and I didn't do any real-life robustness testing yet. E.g. do we gracefully bail if running on HPC with no network connection?

PR #3474, opened 2023-11-15
Chargemol autodownloads

chiang-yuan:
Try to close #3465 

## Checklist

- [ ] ~~If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))~~

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -44,6 +44,9 @@ from pymatgen.core.periodic_table import DummySpecies, Element, Species, get_el_
 from pymatgen.core.sites import PeriodicSite, Site
 from pymatgen.core.units import Length, Mass
 from pymatgen.electronic_structure.core import Magmom
+from pymatgen.io.cif import CifWriter
+from pymatgen.io.gaussian import GaussianInput
+from pymatgen.io.xyz import XYZ
 from pymatgen.symmetry.maggroups import MagneticSpaceGroup
 from pymatgen.util.coord import all_distances, get_angle, lattice_points_in_supercell
 
@@ -2663,12 +2666,8 @@ class IStructure(SiteCollection, MSONable):
         fmt = fmt.lower()
 
         if fmt == "cif" or fnmatch(filename.lower(), "*.cif*"):
-            from pymatgen.io.cif import CifWriter
-
             writer = CifWriter(self, **kwargs)
         elif fmt == "mcif" or fnmatch(filename.lower(), "*.mcif*"):
-            from pymatgen.io.cif import CifWriter
-
             writer = CifWriter(self, write_magmoms=True, **kwargs)
         elif fmt == "poscar" or fnmatch(filename, "*POSCAR*"):
             from pymatgen.io.vasp import Poscar
@@ -3550,9 +3549,6 @@ class IMolecule(SiteCollection, MSONable):
         Returns:
             IMolecule or Molecule.
         """
-        from pymatgen.io.gaussian import GaussianInput
-        from pymatgen.io.xyz import XYZ
-
         fmt = fmt.lower()  # type: ignore[assignment]
         if fmt == "xyz":
             mol = XYZ.from_str(input_string).molecule
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -44,9 +44,6 @@ from pymatgen.core.periodic_table import DummySpecies, Element, Species, get_el_
 from pymatgen.core.sites import PeriodicSite, Site
 from pymatgen.core.units import Length, Mass
 from pymatgen.electronic_structure.core import Magmom
-from pymatgen.io.cif import CifWriter
-from pymatgen.io.gaussian import GaussianInput
-from pymatgen.io.xyz import XYZ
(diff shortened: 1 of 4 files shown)


tests/command_line/test_chargemol_caller.py
```diff
@@ -86,31 +80,61 @@ def mock_xyz(tmp_path: Path):
...
     ca._atomic_densities_path = os.path.expanduser("~/.cache/pymatgen/ddec/fake_file")
 
     assert ca.ddec_charges == [0.8432, -0.8432]
 
 
-def test_auto_download(monkeypatch):
-    # mock getenv to simulate that DDEC6_ATOMIC_DENSITIES_DIR is not set
-    monkeypatch.setattr(os, "getenv", lambda *args, **kwargs: None)
+def fake_download(self, version: str = "latest", verbose: bool = True) -> None:
+    extraction_path = "~/.cache/pymatgen/ddec"
```
Andrew-S-Rosen, line 95: Be careful with path handling. You don't want to impose a given OS.

I recommend using the `pathlib` library instead.

tests/command_line/test_chargemol_caller.py
```diff
@@ -86,31 +80,61 @@ def mock_xyz(tmp_path: Path):
...
+    os.makedirs(os.path.expanduser(extraction_path), exist_ok=True)
 
+    # Create a fake zipfile with a folder named "atomic_densities"
+    fake_zip_path = os.path.expanduser(f"{extraction_path}/fake_file.zip")
+    fake_folder_path = os.path.expanduser(f"{extraction_path}/atomic_densities")
```
Andrew-S-Rosen, line 100: Same comments here regarding path handling. Do not hard-code slashes.

tests/command_line/test_chargemol_caller.py
janosh: Actually, forward slashes work under windows. I've only encountered a few arcane cases where they cause problems. See the `pmg` test suite. Runs on Windows in GHA and it's all forward slashes.

Andrew-S-Rosen: The `~` is not valid on Windows, although I guess you get out of that pickle by doing `os.path.expanduser`.

Andrew-S-Rosen: Doing `os.path.expanduser(extraction_path)` expands to `C:\\Users\\asros/.cache/pymatgen/ddec'` which will work with `os.makedir` since `os` is smart. So, tests will pass, but you're playing with fire a bit... 😅

janosh: Yeah, i'm running a live experiment. So far it's only caused issues 2 or 3 times and they were niche. Feel like so far the readability gain from not using `os.path.join` was well worth it. Need to reconsider if sth non-niche comes up.

Andrew-S-Rosen: Feel free to ignore these comments then @chiang-yuan. Good to know, @janosh.

chiang-yuan: Thanks @Andrew-S-Rosen . I actually prefer `pathlib` since it is a newer standard over `os.path`. I was just following what we already have in the code 😉 Will change it whenever appropriate :)

Andrew-S-Rosen: Seems it doesn't really matter :)

janosh: while `pathlib` has some nice features, i find it a bit too OOP with resulting perf overhead compared to `os.path` (see e.g. https://youtu.be/tFrh9hKMS6Y). although admittedly the speed diff doesn't matter in practice.

pymatgen/core/structure.py
janosh, line 1: can you revert the trailing comma insertions in this file and `tests/command_line/test_chargemol_caller.py`? makes it hard to spot meaningful changes

janosh (changes requested):
@chiang-yuan thanks! i think there are a lot of unintended white space/indentation changes from trailing commas in this PR?

pymatgen/core/structure.py
chiang-yuan: hmmm sorry let me check... perhaps my ruff is not working correctly

shyuep:
@chiang-yuan I would like to merge this, but no tests are being run for some reason. Only pre-commit was run. Pls make sure the test suite is run properly. Thanks.

chiang-yuan:
@shyuep sorry this is a bit old and I am not sure why we need to test fake download as that does not really test if we download the file but mocks it offline.  I could take care of autodownloads  with #3778 together and the test there will really download chargmol during CI workflow

janosh:
i'll close this for the same reason as #3465 (i.e. stalled). feel free to re-open if you want to resolve merge conflicts @chiang-yuan.

PR #3476, opened 2023-11-15
Convert all FHI-aims stresses to be 3x3 instead of Voigt notation

tpurcell90:
Previously used voigt

## Summary

Major changes:

- fix 1: all stresses are 3x3 now

## Checklist

### DIFF
--- a/pymatgen/io/aims/parsers.py
+++ b/pymatgen/io/aims/parsers.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any
 import numpy as np
 
 from pymatgen.core import Lattice, Molecule, Structure
+from pymatgen.core.tensors import Tensor
 
 if TYPE_CHECKING:
     from collections.abc import Generator, Sequence
@@ -56,12 +57,6 @@ SCALAR_PROPERTY_TO_LINE_KEY = {
     "electronic_temp": ["Occupation type:"],
     "fermi_energy": ["| Chemical potential (Fermi level)"],
 }
-
-
-def voigt_to_full_stress_conv(voigt_stress: Sequence[float]) -> Matrix3D:
-    """Transform voigt vector to a full 3x3 matrix."""
-    xx, yy, zz, yz, xz, xy = np.array(voigt_stress).flatten()
-    return np.array([[xx, xy, xz], [xy, yy, yz], [xz, yz, zz]])
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


pymatgen/io/aims/parsers.py
```diff
@@ -633,7 +639,7 @@ def forces(self) -> list[Vector3D] | None:
         )
 
     @property
-    def stresses(self) -> list[Matrix3D] | None:
+    def stresses(self) -> Sequence[Matrix3D] | None:
```
janosh, line 642: Better not use sequence here. best to be as generic as possible in arguments and as specific as possible in return values.

janosh (changes requested):

pymatgen/io/aims/parsers.py
```diff
@@ -58,6 +58,12 @@ def __init__(self, message: str) -> None:
 }
 
 
+def voigt_to_full_stress_conv(voigt_stress: Sequence[float]) -> Matrix3D:
+    """Transform voigt vector to a full 3x3 matrix."""
+    xx, yy, zz, yz, xz, xy = np.array(voigt_stress).flatten()
+    return np.array([[xx, xy, xz], [xy, yy, yz], [xz, yz, zz]])
```
janosh, line 64: Can't we use `ase.stress.voigt_6_to_full_3x3_stress` instead of defining our own?

tpurcell90: We can I was not sure the recommendation was for the use of ASE functions in non-ASE parts since it is an optional dependency

pymatgen/io/aims/parsers.py
tpurcell90: This is fixed now, but I am not sure what caused the CI errors for VASP

pymatgen/io/aims/parsers.py
janosh: You were already importing from `ase.stress` but I just realized that was in a test file. You're right, in a non-test file, we have to wrap in `try: import ase except ImportError: ...`.

Better still though, we already have the [`Tensor` class for this](https://github.com/tpurcell90/pymatgen/blob/db52f3bf37224cfff2c8df363ad151059196f338/pymatgen/core/tensors.py#L40). It has a `from_voigt` method.

tpurcell90: I converted this to use the Tensor object instead (sorry for the delay I got my new personal computer so can now actually work again)

janosh: Thanks! 👍

janosh (approved):

PR #3483, opened 2023-11-26
Add gradient, Hessian, and orbital coeffs scratch file parsers to `pymatgen.io.qchem.outputs`

Andrew-S-Rosen:
## Summary

This PR adds function to `pymatgen.io.qchem.outputs` to parse (scratch) gradient, Hessian, and previous orbital coefficient files made by Q-Chem.

**Note**: 99% of the credit for this goes to @samblau. I am simply moving it into pymatgen where it belongs.

## Todos

I need to add tests. There are also some things that are definitely better done with numpy vectorization, but it's not a huge issue.

## Checklist

- [] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

```sh
pip install -U pre-commit
pre-commit install
```


Andrew-S-Rosen:
@janosh: This should be ready for review. As I mentioned above, this is code mostly written my @samblau --- I added some simple tests and cleaned up a few things. Longer term, it would probably make sense to create a new `QCScratch(MSONable)` class that parses scratch files in a directory. However, I have not implemented that here (unless you feel it necessary).

tests/files/qchem/53.0
janosh, line 1: Given the file size probably worth compressing these test files, esp. this one?

janosh (changes requested):
Thanks @Andrew-S-Rosen, looks great! Just a minor suggestion re test file compression.

tests/io/qchem/test_outputs.py
```diff
@@ -492,6 +501,37 @@ def test_nbo_3_c(self):
         assert perturb_ene[0]["perturbation energy"][3209] == 3.94
 
 
+def test_gradient(tmpdir):
```
janosh, line 504: I spoke to a `pytest` developer once who said `tmpdir` is slated for eventual deprecation so better to use `tmp_path` here.

Andrew-S-Rosen: @janosh: Thanks! Didn't know about that. I just changed them out. Resolving this.

Andrew-S-Rosen: Ew, that's gross. I forgot since it's `Path` now, it doesn't have that. Changed. Hopefully this works.

Andrew-S-Rosen:
@janosh: All good now!

janosh (approved):
Thanks @Andrew-S-Rosen! 👍

PR #3489, opened 2023-12-01
Updates for Vasprun with MD simulations

gpetretto:
## Summary
I propose a few minor changes in the `Vasprun` object when parsing MD simulations and especially those produced with the ML option.

* the `converged_ionic` property was False for an MD simulation. But the check did not seems reasonable for an MD
* the number of steps for a simulation that involves ML MD is not given by the length of `ionic_steps`
* in case of NPT simulation the parsing of stresses for ML MD simulations was missing


tests/files/vasprun.xml.md
janosh, line 1: Can you compress?

janosh (changes requested):
The changes sound good to me! Thanks a lot

tests/files/vasprun.xml.md
janosh: Actually, there's already `tests/files/vasprun.xml.ml_md`. Is that file incompatible with this test?

janosh: If the new file covers a superset of test cases, you can remove the old one and plug in the path to the new one instead (only 1 test needs updating).

gpetretto: no, the new one will not cover the ML MD, that produces the md_data attribute. The data are partially overlapping, so will not cover the same set of lines. If I need to remove one, I think that removing the `vasprun.xml.md` would be better.

janosh: no need. where you load the file

```py
vasp_run = Vasprun(f"{TEST_FILES_DIR}/vasprun.xml.md.gz")
```

could you add a brief comment what differentiates this file from `vasprun.xml.ml_md` to leave a paper trail of our discussion in the code.

janosh (approved):

PR #3490, opened 2023-12-04
Add multipole parsing for Q-Chem IO

espottesmith:
Minor PR. Just adding electric multipole (up through hexadecapole) parsing for Q-Chem.

Tests for relevant classes pass locally, black and ruff seem happy enough for the files I changes. Didn't check mypy, so here's praying.


pymatgen/io/qchem/outputs.py
```diff
@@ -976,6 +977,137 @@ def _read_charges_and_dipoles(self):
...
+                            "XXXZ": float(hpole[5]),
+                            "XXYZ": float(hpole[6]),
+                            "XYYZ": float(hpole[7]),
+                            "YYYZ": float(hpole[8]),
+                            "XXZZ": float(hpole[9]),
+                            "XYZZ": float(hpole[10]),
+                            "YYZZ": float(hpole[11]),
+                            "XZZZ": float(hpole[12]),
+                            "YZZZ": float(hpole[13]),
+                            "ZZZZ": float(hpole[14]),
```
janosh, line 1107: This could be made more concise with a list comprehension

```py
keys = ["XXXX", "XXXY", "XXYY", "XYYY", "YYYY", "XXXZ", "XXYZ", "XYYZ", "YYYZ", "XXZZ", "XYZZ", "YYZZ", "XZZZ", "YZZZ", "ZZZZ"]
hexadecapole = [{key: float(hpole[idx]) for idx, key in enumerate(keys)}]
```

Likewise above.

janosh (changes requested):
Thanks @espottesmith! 👍 Looking good. Just a few small refactors needed.

espottesmith:
How does this look, @janosh?

janosh (approved):
Looks great, thanks! 👍

espottesmith:
I think your most recent proposed changes (particularly the use of str.split() to form the key lists) save a small number of characters at the expense of increasing cognitive load.

As long as this gets merged, I don't really care, but at some point I'd love to chat with you about coding style in PMG. Ultimately code needs to be read by human beings; as such, I strongly believe that efficiency (in terms of LOC or characters or whatever) should not be the ultimate goal.

(For clarity, this critique isn't aimed at your previous requested changes. Using the key lists and enumerate() probably also somewhat increase cognitive load, but in that case I'd say the cleanup was well worth it.)

janosh:
@espottesmith Fair enough reverted in 93078ff. I know @shyuep will definitely agree with you there. I like `str.split` if it avoids multiple line breaks. Personally don't find it hard to read/reason about and the more code fits on one screen height, usually the easier to debug in my experience...

PR #3493, opened 2023-12-05
Fix pdplotter.show with matplotlib backend

lbluque:
## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


janosh (changes requested):
Thanks @lbluque! 👍 

Can we change this method to return the `ax` (`fig` for plotly)? Don't see a reason not to return. And add a test for the return type?

lbluque:
> Thanks @lbluque! 👍 
> 
> Can we change this method to return the `ax` (`fig` for plotly)? Don't see a reason not to return. And add a test for the return type?

That looks like the `get_plot` method already does. This method is to display the plot.

janosh:
> That looks like the `get_plot` method already does.

I know but does it hurt to return anyway? I guess it could actually. I think in Jupyter if some code calls `plt.show()` and you don't assign the returned figure to a variable the cell will render the returned figure and so it would be shown twice? If that's the case then you're right and we shouldn't return.

janosh (approved):
I added a bunch more tests including ones that mock `PDPlotter.show()` in CI. This is good to go. 🎉

lbluque:
Awesome thanks for adding the tests! lgtm

PR #3494, opened 2023-12-05
Sherif

sheriftawfikabbas:
## Summary

Major changes:

- feature 1: add_interstitial() method

## Todos

Optimisation of the loops for larger structures.

## Checklist

- [x ] Google format doc strings added. Check with `ruff`.
- [x ] Type annotations included. Check with `mypy`.

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -625,6 +625,35 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
                 new_species[Species(sym, oxidation_state=oxi_state, spin=spin)] = occu
             site.species = Composition(new_species)
 
+    def add_interstitial(self, atom) -> None:
+        """
+        Creates an interstitial defect in the structure by adding the atom to the available voids.
+        The structure is embedded in a 3D grid. For simplicity, the grid is defined based on fractional coordinates. Each lattice side is divided into 100 points.
+        Scans voids by adding densities (gaussians) and choosing the density minima.
+        """
+        def density(r,R):
+            return np.exp(-np.dot(r - R,r - R))
+        supercell = self.copy()
+        supercell.make_supercell([3,3,3])
+        divisions = np.array([10,10,10])*3
+        n = np.zeros(divisions)
+        for i in range(divisions[0]):
+            for j in range(divisions[1]):
+                for k in range(divisions[2]):
+                    f = [i/divisions[0],j/divisions[1],k/divisions[2]]
+                    r = np.dot(f, supercell.lattice.matrix)
+
+                    for R in supercell.cart_coords:
+                        if self.distance(R,r) < 4:
+                            n[i,j,k] += density(r,R)
+        n = n[10:20,10:20,10:20]
+        p = np.where(n == np.min(n))
+        p = [p[0][0]/(divisions[0]/3),p[1][0]/(divisions[1]/3),p[2][0]/(divisions[2]/3)]
+        print(p)
+        self.append(atom, p)
+            
+
+
     def remove_spin(self) -> None:
         """Remove spin states from structure."""
         for site in self:
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -625,34 +625,34 @@ class SiteCollection(collections.abc.Sequence, metaclass=ABCMeta):
                 new_species[Species(sym, oxidation_state=oxi_state, spin=spin)] = occu
             site.species = Composition(new_species)
 
-    def add_interstitial(self, atom:str) -> None:
+    def add_interstitial(self, atom: str) -> None:
         """Creates an interstitial defect in the structure in-place by adding the atom to the available voids. The structure is embedded in a 3D grid. For simplicity, the grid is defined based on fractional coordinates. Each lattice side is divided into 100 points. Scans voids by adding densities (gaussians) and choosing the density minima.
 
         Args:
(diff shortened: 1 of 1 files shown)


pymatgen/core/structure.py
```diff
@@ -625,6 +625,35 @@ def add_spin_by_site(self, spins: Sequence[float]) -> None:
                 new_species[Species(sym, oxidation_state=oxi_state, spin=spin)] = occu
             site.species = Composition(new_species)
 
+    def add_interstitial(self, atom: str) -> None:
```
Andrew-S-Rosen, line 628: It would probably be better to use an [`Element`](https://pymatgen.org/pymatgen.core.html#pymatgen.core.periodic_table.Element) here instead of a string.

pymatgen/core/structure.py
```diff
@@ -625,6 +625,35 @@ def add_spin_by_site(self, spins: Sequence[float]) -> None:
...
+        Args:
+            atom (str): e.g. "H", "Li" etc.
+        """
+
+        def density(r, R):
+            return np.exp(-np.dot(r - R, r - R))
+
+        supercell = self.copy()
+        supercell.make_supercell([3, 3, 3])
+        divisions = np.array([10, 10, 10]) * 3
```
Andrew-S-Rosen, line 640: It might be worth thinking about if we can generalize this in any way. I assume the 3x3x3 was to address PBCs? If so, it could make more sense to test for the minimum image distance rather than make a supercell, particularly if the starting structure is large to begin with.

pymatgen/core/structure.py
```diff
@@ -625,6 +625,35 @@ def add_spin_by_site(self, spins: Sequence[float]) -> None:
...
+        n = np.zeros(divisions)
+        for i in range(divisions[0]):
+            for j in range(divisions[1]):
+                for k in range(divisions[2]):
+                    f = [i / divisions[0], j / divisions[1], k / divisions[2]]
+                    r = np.dot(f, supercell.lattice.matrix)
+
+                    for R in supercell.cart_coords:
+                        if self.distance(R, r) < 4:
```
Andrew-S-Rosen, line 649: The 4 A cutoff here might be better suited as an optional keyword argument that the user can specify, which would default to 4.

Andrew-S-Rosen (commented):
Hi @sheriftawfikabbas! Thank you for your contribution!

While I'm not a maintainer, I left a few brief comments here which you may wish to consider. @janosh will be a better reviewer than me. :)

I'm tagging @jmmshn here because he has thought a lot about intercalation. I wonder if he already added some functionality somewhere in pymatgen or elsewhere?

utf:
Just a general point, there are several approaches to adding interstitial atoms in the [pymatgen-analysis-defects](https://github.com/materialsproject/pymatgen-analysis-defects/) package, including Voronoi and charge based methods:

https://github.com/materialsproject/pymatgen-analysis-defects/blob/aaa68e29ffa84599f306984b0f95eaa212081e8f/pymatgen/analysis/defects/generators.py#L307

https://github.com/materialsproject/pymatgen-analysis-defects/blob/aaa68e29ffa84599f306984b0f95eaa212081e8f/pymatgen/analysis/defects/generators.py#L397

@jmmshn can comment more, but perhaps pymatgen-analysis-defects is a better place for this code?

jmmshn:
@sheriftawfikabbas there are a couple of helper functions here 

https://github.com/materialsproject/pymatgen-analysis-defects/blob/main/pymatgen/analysis/defects/utils.py

That might do similar things.  Also, I noticed that there is a choice of a 3x3x3 supercell implicit in the code.  I think that is something we should avoid doing.
And example of how that is handled is here:
https://github.com/materialsproject/pymatgen-analysis-defects/blob/main/pymatgen/analysis/defects/supercells.py

janosh:
I'll go ahead and close this for now.

sheriftawfikabbas:
Thanks everyone for the great comments, will apply and resubmit the PR.

Sherif

janosh:
@sheriftawfikabbas If the helper functions @jmmshn mentioned don't do the trick, consider submitting the PR to https://github.com/materialsproject/pymatgen-analysis-defects instead.

jmmshn:
So in my experience, Voronoi doesn't work great but charge density requires additional data.  So if this works and is roughly symmetry-aware like the charge density is I think it could be really useful.

Since this is a density field you can probably just subclass the ChargeInterstitialGenerator.  I am curious about whether the minima you find this way will be meaningful but I guess we will wait and see.

Also, I agree that this should go into the defects package since the infrastructure for dealing with interstitials is already there.

sheriftawfikabbas:
Awesome, will move the code to the defects package and try your suggestion.

Sherif

sheriftawfikabbas:
Interesting, just found that the [3x3x3] supercell trick was also used in
the QModel class to overcome the PBC issue

Quoting
https://github.com/materialsproject/pymatgen-analysis-defects/blob/main/pymatgen/analysis/defects/utils.py
lines 286,287:
    # Make 3x3x3 supercell
    # This is a trick to resolve the periodical boundary issue.

Will try to find an alternative strategy for that.

Sherif

jmmshn:
Oh yeah, I didn't read the initial PR carefully enough.  I thought the 3x3x3 was enforced on an output or something.  I didn't realized it was to help with finding point at the boundary.  Either way it looks like subclassing `ChargeInterstitialGenerator` is what you want.

PR #3495, opened 2023-12-05
fix TypeError when attr force_field not exists

xjf729:
## Summary

Major changes:

- feature 1: ...
- fix 1: Fix bug "TypeError: argument of type 'NoneType' is not iterable". Not all LammpsData in CombinedData should have attribute force_field, replace this `NoneType` attribute with `[]` to skip unexisting force_field.

## Todos

- feature 2: ...
- fix 2:

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


janosh (changes requested):
Can you please add a test for this?

PR #3496, opened 2023-12-05
Avoid redirects in `MPRester` requests

tschaume:
add trailing slash to all MPRester requests to avoid unnecessary redirects


janosh (approved):
Thanks @tschaume, I'll merge this as is. Do you think it's worth adding tests for this? We'd need to intercept HTTP requests and check the endpoint URL.

PR #3502, opened 2023-12-06
 Enforce `zval` to be an integer to avoid improper syntax in `.cri` file

### LINKED ISSUE #3501: Enforce zval to be an integer to avoid improper syntax in .cri file
https://github.com/materialsproject/pymatgen/blob/ebd776900edfc45bd3b9f045e1e04db1e2d2752b/pymatgen/command_line/critic2_caller.py#L198C1-L198C17

Main issue:
**formatting for zval is incorrect, causes failure when running .cri file**

Default behavior of the pmg Potcar object that is initialized by the Critic2Caller.from_path() method returns zval as a float rather than an int:

Source from line 1640 in pymatgen.io.vasp
```python
    parse_functions = dict(
        LULTRA=_parse_bool,
        LUNSCR=_parse_bool,
        LCOR=_parse_bool,
        LPAW=_parse_bool,
        EATOM=_parse_float,
        RPACOR=_parse_float,
        POMASS=_parse_float,
        ZVAL=_parse_float,
```

Which causes an error because the float is not converted to int

```python
        if not zpsp:
            potcar_path = get_filepath(
                "POTCAR",
                "Could not find POTCAR, will not be able to calculate charge transfer.",
                path,
                suffix,
            )

            if potcar_path:
                potcar = Potcar.from_file(potcar_path)
                zpsp = {p.element: p.zval for p in potcar}
```

For anyone with the same issue, remove POTCAR from the directory and pass zpsp as an argument. 

Otherwise you will get a similar error in the output:

```
%% load int.CHGCAR id chg_int zpsp H 1.0 Au 11.0 C 4.0 S 6.0

ERROR : load int.CHGCAR id chg_int zpsp H 1.0 Au 11.0 C 4.0 S 6.0
ERROR (load): wrong syntax in ZPSP
```

wladerer:
Closes #3501.

Formatting for `zval` in Critic2Caller.from_path() method wrote `zval` as a `float` rather than an `int`, causing failure when running `.cri` file.


janosh:
Thanks a lot! We need a test for this fix.

PR #3508, opened 2023-12-10
Reduce the lower-bound pin on numpy

Andrew-S-Rosen:
This is a test to see if the >=1.25.0 is actually needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
         "matplotlib>=1.5",
         "monty>=3.0.2",
         "networkx>=2.2",
-        "numpy>=1.25.0",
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
@janosh: I was looking to see if I could lower the numpy bound (i.e. see if the suite still passes)

janosh:
The errors here are not from your `numpy` version change but would have also happened on `master`.

I just fixed them in https://github.com/materialsproject/pymatgen/pull/3507.

Andrew-S-Rosen:
@janosh: Yup, just noticed that. Either way, my test was a failure anyway because I was still installing the new numpy version.

PR #3517, opened 2023-12-15
`Composition` raise `ValueError` if `formula` string is only numbers and spaces

janosh:
Controlled by new `strict: bool = True` parameter to `_parse_formula`.

### DIFF
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -522,10 +522,12 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
             return any(category[0] in el.block for el in self.elements)
         return any(getattr(el, f"is_{category}") for el in self.elements)
 
-    def _parse_formula(self, formula: str) -> dict[str, float]:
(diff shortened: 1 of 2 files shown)


CompRhys:
Doesn't catch the "NaN" -> float("NaN") edge case
```
>>> Composition(float("NaN"))
  File "/Users/chemix-rhys/Chemix/pymatgen/pymatgen/core/composition.py", line 127, in __init__
    elem_map = dict(*args, **kwargs)  # type: ignore
TypeError: 'float' object is not iterable
>>> float("NaN")
nan
```

I was going to add another check:

```python
from math import isnan
...
        if len(args) == 1 and isinstance(args[0], Composition):
            elem_map = args[0]
        elif len(args) == 1 and isinstance(args[0], str):
            elem_map = self._parse_formula(args[0])  # type: ignore
        elif len(args) == 1 and isnan(args[0]):
            raise ValueError("float('NaN') is not a valid Composition, did you mean str('NaN')?")
        else:
            elem_map = dict(*args, **kwargs)  # type: ignore
```

PR #3520, opened 2023-12-16
Add `PhononDos.mae()` and `PhononBandStructure.has_imaginary_gamma_freq()` methods

janosh:
@JaGeo Would be great to get your review on this, esp. the new `PhononDos.mae()` and `PhononBandStructure.has_imaginary_gamma_freq()` methods.

### DIFF
--- a/pymatgen/electronic_structure/bandstructure.py
+++ b/pymatgen/electronic_structure/bandstructure.py
@@ -779,7 +779,7 @@ class BandStructureSymmLine(BandStructure, MSONable):
         TODO: now it uses the label we might want to use coordinates instead
         (in case there was a mislabel)
         """
-        # if the kpoint has no label it can"t have a repetition along the band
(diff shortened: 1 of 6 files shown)


JaGeo:
Have you checked the tolerance in detail? I think it might be very tight. Similar for the overall tolerance

janosh:
I just copied the `tol=1e-5` from the existing `has_imaginary_freq` method but use `tol=1e-3` in my own analysis. Should we increase to `tol=1e-3` by default?

JaGeo:
> I just copied the `tol=1e-5` from the existing `has_imaginary_freq` method but use `tol=1e-3` in my own analysis. Should we increase to `tol=1e-3` by default?

Yes, this would be great! 

I haven't spotted anything else when reading the pull-request. However, I haven't tested it in detail.

PR #3534, opened 2023-12-27
Fix vasp run type handling for GGA+U scheme

### LINKED ISSUE #3506: Vasprun return invalid run type PBE
Hi, I would like to apply the legacy corrections with MaterialsProjectCompatibility. However, an invalid run-type was returned.
```
from pymatgen.io.vasp import Vasprun
from pymatgen.entries.compatibility import MaterialsProjectCompatibility, AnionCorrection
vasprun1 = Vasprun("data/mp_1595_vasprun.xml")
entry1 = vasprun1.get_computed_entry(inc_structure=True)
compatibility = MaterialsProjectCompatibility()
entry_ = compatibility.get_adjustments(entry1)
```
Error
```
CompatibilityError: Entry vasprun-2023-12-08 10:44:39.983016 has invalid run type PBE. Discarding.
```
[mp_1595_vasprun.xml.zip](https://github.com/materialsproject/pymatgen/files/13608078/mp_1595_vasprun.xml.zip)

DanielYang59:
## Summary

- Fixed inconsistent non-GGA run type handling, to fix #3506.

## Need suggestions from maintainer

As discussed in #3506, the run type of "PBE" is not treated as "GGA" which seems to be inconsistent with the defination in the following part: https://github.com/materialsproject/pymatgen/blob/1dd74a8479ee700552accea770c3ceda85ef89c9/pymatgen/io/vasp/outputs.py#L612-L630
Is this intended, or is it a compatibility issue? I maybe able to help fix it but would need input from maintainers on what is expected.


janosh (changes requested):
This is a breaking change but I'm not sure if it has any downstream implications to no longer raise `CompatibilityError` on non-GGA run types. I expect it's fine.

@DanielYang59 Could you update the test for the new warning?

DanielYang59:
@janosh Yes sure. I would try to get my hands on that ASAP. 

I adjusted the imports to alphabet order while I was fixing the run type issue, but that commit seems to have been reversed by workflow or something?

janosh:
Imports are auto-sorted by `ruff`. No need to worry about that.

DanielYang59:
Thanks for clarifying @janosh . However I just realize insteading of diving in and removing `raise error` altogether, I would need to figure out **why `PBE` is not considered `GGA` in the first place**. Do you have any idea? Could this be some consistency issue (for example the original code is intended for generate GGA, and then more specific GGA functional names for example PBE are introduced.)

As from my understanding, PBE is a specific GGA exchange-correlation functional as defined in:https://github.com/materialsproject/pymatgen/blob/1dd74a8479ee700552accea770c3ceda85ef89c9/pymatgen/io/vasp/outputs.py#L612-L630

Meanwhile, the Hubbard+U method should work for PBE functional [here](https://www.vasp.at/wiki/index.php/NiO_GGA%2BU) for example, and thus there is no good reason to rule out PBE from GGA+U scheme?

janosh:
> I would need to figure out why PBE is not considered GGA in the first place. Do you have any idea?

@rkingsbury Might know what the thought process behind this was? Might have something to do with the MP correction scheme only being designed to be applied to MP compatible VASP runs using e.g. `MPRelaxSet`. If we get clarity on this, would be good to leave a code comment to address future questions.

pymatgen/entries/compatibility.py
```diff
@@ -463,12 +463,13 @@ def get_correction(self, entry) -> ufloat:
...
         """
-        if entry.parameters.get("run_type") not in ("GGA", "GGA+U"):
-            raise CompatibilityError(
-                f"Entry {entry.entry_id} has invalid run type {entry.parameters.get('run_type')}. Discarding."
-            )
+        # Only correct GGA or GGA+U entries
+        run_type = entry.parameters.get("run_type")
+        if run_type not in ("GGA", "GGA+U"):
+            warnings.warn(f"Entry {entry.entry_id} has invalid {run_type=}. Applying no corrections.")
+            return ufloat(0.0, 0.0)
```
rkingsbury, line 472: This will need testing, but I'm fairly certain this entire block should be deleted. The `get_corrections` methods should not raise `CompatibilityError` anymore; rather they should just return `ufloat(0,0)` for any entries to which the correction does not apply.

`CompatibilityError` is raised by `get_adjustments`

pymatgen/entries/compatibility.py
```diff
@@ -477,10 +478,6 @@ def get_correction(self, entry) -> ufloat:
         most_electroneg = elements[-1].symbol
         correction = ufloat(0.0, 0.0)
 
-        # only correct GGA or GGA+U entries
-        if entry.parameters.get("run_type") not in ("GGA", "GGA+U"):
-            return ufloat(0.0, 0.0)
-
```
rkingsbury, line 483: This block should then stay in place, if the above block is deleted

rkingsbury:
Hi @DanielYang59 , thanks for raising this; it should be addressed. However it is important not to change the error into a warning. It's intended that a user calls `process_entries` to apply corrections rather than `get_adjustments`. If called that way, there is a kwarg that defines the behavior when `get_adjustments` raises an error (by default the error is ignored). 

I believe this default causes the entry to be passed through to the returned list, albeit without any energy corrections being applied. See docstring:

```
    def process_entries(
        self,
        entries: AnyComputedEntry | list[AnyComputedEntry],
        clean: bool = True,
        verbose: bool = False,
        inplace: bool = True,
        on_error: Literal["ignore", "warn", "raise"] = "ignore",
    ) -> list[AnyComputedEntry]:
        """Process a sequence of entries with the chosen Compatibility scheme.

        Warning: This method changes entries in place! All changes can be undone and original entries
        restored by setting entry.energy_adjustments = [].

        Args:
            entries (AnyComputedEntry | list[AnyComputedEntry]): A sequence of
                Computed(Structure)Entry objects.
            clean (bool): Whether to remove any previously-applied energy adjustments.
                If True, all EnergyAdjustment are removed prior to processing the Entry.
                Defaults to True.
            verbose (bool): Whether to display progress bar for processing multiple entries.
                Defaults to False.
            inplace (bool): Whether to adjust input entries in place. Defaults to True.
            on_error ('ignore' | 'warn' | 'raise'): What to do when get_adjustments(entry)
                raises CompatibilityError. Defaults to 'ignore'.

        Returns:
            list[AnyComputedEntry]: Adjusted entries. Entries in the original list incompatible with
                chosen correction scheme are excluded from the returned list.
        """
```

As to the core of your question - you are correct that PBE is one type of GGA functional, but within the MP Compatibility classes we often refer to PBE as "GGA" and "PBE+U" as "GGA+U". I suspect this is because we have to return something if there is no explicitly set functional in the INCAR, and VASP defaults to PBE.

See my code review comments for my suggested fix (which will need to be confirmed by passing unit tests). I think this is the cleanest way forward.

rkingsbury (changes requested):
There are 6 instances of this pattern in `pymatgen.entries.compatibility`:

```
# only correct GGA or GGA+U entries
        if entry.parameters.get("run_type") not in ("GGA", "GGA+U"):
            return ufloat(0.0, 0.0)
```

I think the cleanest fix is to expand the set of tested `run_type` in all 6 cases to be `("GGA", "PBE", "GGA+U", "PBE+U")` to cover cases where `PE` is explicitly set as the functional in the INCAR.

pymatgen/entries/compatibility.py
janosh: Thanks for thorough review @rkingsbury! 👍 

Tip: add `py` to your code fences for syntax highlighting.

````diff
- ```
+ ```py
````

**Comment by DanielYang59** [comment]
Hi @rkingsbury thanks so much for the valuable inputs and suggestions, I will have a look and try to think of a proper improvement for this (as I'm pretty much a beginner).

File: tests/entries/test_compatibility.py [hunk 3]
```diff
@@ -80,9 +80,7 @@ def setUp(self):
         self.compat = MaterialsProjectCompatibility(check_potcar_hash=False)
 
     def test_correction_specificity(self):
-        processed = self.compat.process_entries([self.entry1, self.entry2, self.entry3])
-
-        assert len(processed) == 2
```
DanielYang59, line 85: Please help me check carefully here @rkingsbury as I'm not 100% confident about these two changes in unit tests. Prior to this change: https://github.com/materialsproject/pymatgen/blob/8aa1cf89a93e12c2ac16ddde517105e307a04b91/pymatgen/entries/compatibility.py#L468-L482
if run_type is not in {"GGA", "GGA+U"}, an error would be raise thus no correction would be applied. After this change, a correction of 0.0 would be returned thus the number of entries would increase from 2 to 3. Correct me if I'm wrong. Thanks.

rkingsbury: Great catch @DanielYang59 . We should not make a change that causes these tests to fail (for backward compatibility reasons).

I've studied the code again, and actually you are right that the `raise` statement should remain.

The logic is the following:
- when `process_entries` is called, it calls `get_adjustments` for every entry provided
- Any entry which raises a `CompatibilityError` is excluded from the returned list.
- In the specific case of `MaterialsProjectCompatibility`, `get_adjustments` calls `get_corrections_dict`, which loop through the `get_correction` method of all 4 of the underlying `Correction` classes (we eliminated these in the newer scheme for simplicity)
- It just so happens that the `raise` statement that occurs in `UCorrection` (which processes any entry sent through `get_Adjustments`) serves the purpose of excluding incompatible entries. 
- Strictly speaking, this `raise` statement could/should occur in `get_adjustments` itself, because that is really the only place that the run_type check needs to happen (an entry which is incompatible with one of the corrections is incompatible with all of them, for this purpose)
- The `return ufloat(0,0)` statements in the `Correction` classes do not serve an important purpose as far as I can tell.

I think the cleanest fix would be to move the run type check and `raise` statement to `get_adjustments` itself and remove the other statements, as illustrated [here](https://github.com/materialsproject/pymatgen/pull/3540/files)

pymatgen/entries/compatibility.py
```diff
@@ -206,7 +206,7 @@ def get_correction(self, entry) -> ufloat:
         # set error to 0 because old MPCompatibility doesn't have errors
 
         # only correct GGA or GGA+U entries
-        if entry.parameters.get("run_type") not in ("GGA", "GGA+U"):
+        if entry.parameters.get("run_type") not in {"GGA", "PBE", "GGA+U", "PBE+U"}:
             return ufloat(0.0, 0.0)
```
rkingsbury, line 210: All statements of this form in `get_correction` methods can be removed.

pymatgen/entries/compatibility.py
```diff
@@ -931,7 +923,7 @@ def get_adjustments(self, entry: AnyComputedEntry) -> list[EnergyAdjustment]:
         Raises:
             CompatibilityError if the entry is not compatible
         """
-        if entry.parameters.get("run_type") not in ("GGA", "GGA+U"):
+        if entry.parameters.get("run_type") not in {"GGA", "PBE", "GGA+U", "PBE+U"}:
```
rkingsbury, line 926: This complete block (plus a comment about only correcting GGA/GGA+U entries) can be  moved to `get_adjustments`

tests/entries/test_compatibility.py
```diff
@@ -472,7 +470,7 @@ def test_get_corrections_dict(self):
 
     def test_process_entries(self):
         entries = self.compat.process_entries([self.entry1, self.entry2, self.entry3, self.entry4])
-        assert len(entries) == 2
+        assert len(entries) == 3
```
rkingsbury, line 474: revert change. Tests needs to pass.

tests/entries/test_compatibility.py
rkingsbury: Tests pass for me after making that change

rkingsbury (changes requested):

DanielYang59:
Thanks @rkingsbury for the explanation! Your fix looks much cleaner than mine (also no need to change the tests). I might close this PR here. Good work!

rkingsbury:
> Thanks @rkingsbury for the explanation! Your fix looks much cleaner than mine (also no need to change the tests). I might close this PR here. Good work!

Thanks @DanielYang59 , I didn't mean to pre-empt you (that's why I didn't open a PR against the upstream repo, just my own fork). I just thought that was the easiest way to explain what I was talking about. If you're up for it, please copy what I have there into your solution, because you were the one that identified this issue, and I like the changes you made to the docstrings!

We also need to make sure the tests pass on the upstream CI, which I have not done (I only tested locally).

DanielYang59:
Hi @rkingsbury thanks a lot for that. That's totally alright, you fixed it nice and clean, so the honor should go to you (Janosh revised the type hint for the docstring). There is no reason for me to copy your paste as mine. Also I have already committed the minor changes to the other PR #3539 (would also appreciate it if you could put some comments there if possible).

The most important thing for me now is that I learned a lot from you and everyone here. Please go ahead to merge the fix. Thanks!

PR #3536, opened 2023-12-31
Add codespace container for reproducing issues

DanielYang59:
Tasks finished to resolve #2753:
- Added a codespace container configuration file.
- Added a brief guide to use codespace container for issue reproduction.
- Added link in issue template to encourage users to try this new workspace.

However I still need advice on handling user feedbacks, as I failed to identify any way to create a link or share a snapshot to an existing codespace, as @janosh suggested.

> We'd link this codespace in our issue template and invite people to launch it, copy-paste their code into it and then paste the link to the codespace into their issue.

The closest I could find currently is the [Live Share](https://docs.github.com/en/codespaces/developing-in-a-codespace/working-collaboratively-in-a-codespace), which allows multiple users to work on a single codespace in real-time. But I don't think this is what we're looking for though. This would require the user to publicly share a link to a codespace and might have security risks.

So far the best solution to me still remains to run the code in the codespace, and **paste the output back to the issue** manually (until a feature such as sharing a snapshot is out). In doing so, we could help pre-configure the reproduction environment and information collection mechanisms, also avoiding all sorts of non-pymatgen issues (python not installed properly, outdated pymatgen or dependency, and so on).

PR #3537, opened 2024-01-01
Add codespace container for reproducing issues

### LINKED ISSUE #2753: Repro template
Many big projects offer repro templates that spin up an environment with everything needed make a minimal reproducible example simply by clicking a link.

Having something like this for `pymatgen` would make it much easier for users (esp. newcomers) to point out problems and reduce workload on maintainers if they get high-quality reproductions.

We could try out [GitHub's new Codespaces](https://docs.github.com/en/codespaces) for this.

Here's what that might look like:

Screenshot 2022-11-21 at 14 10 54

We'd link this codespace in our issue template and invite people to launch it, copy-paste their code into it and then paste the link to the codespace into their issue. (This would also ensure the error occurs on `master` and isn't due to outdated versions.

All we'd need to do is create a `.devcontainer/devcontainer.json` file that tells GitHub how to spin up our codespace and create a file like `test_files/repro-template.ipynb` with some common imports and maybe a dummy structure or simple `MPRester` query to get people up and running.

```json5
// .devcontainer/devcontainer.json
{
  "image": "mcr.microsoft.com/devcontainers/universal:2",
  "hostRequirements": {
    "cpus": 4
  },
  "waitFor": "onCreateCommand",
  "updateContentCommand": "python -m pip install -r requirements.txt && python -m pip install -e .",
  "customizations": {
    "codespaces": {
      "openFiles": ["test_files/repro_tempalte.ipynb"]
    },
    "vscode": {
      "extensions": [
        "ms-toolsai.jupyter",
        "ms-python.python"
      ]
    }
  }
}
```

DanielYang59:
Tasks finished to resolve #2753 (#3536 is messed up):
- Added a codespace container configuration file.
- Added a brief guide to use codespace container for issue reproduction.
- Added link in issue template to encourage users to try this new workspace.

However I still need advice on handling user feedbacks, as I failed to identify any way to create a link or share a snapshot to an existing codespace, as @janosh suggested.

> We'd link this codespace in our issue template and invite people to launch it, copy-paste their code into it and then paste the link to the codespace into their issue.

The closest I could find currently is the [Live Share](https://docs.github.com/en/codespaces/developing-in-a-codespace/working-collaboratively-in-a-codespace), which allows multiple users to work on a single codespace in real-time. But I don't think this is what we're looking for though. This would require the user to publicly share a link to a codespace and might have security risks.

So far the best solution to me still remains to run the code in the codespace, and **paste the output back to the issue** manually (until a feature such as sharing a codespace snapshot is out). In doing so, we could help pre-configure the reproduction environment and information collection mechanisms, also avoiding all sorts of non-pymatgen issues (python not installed properly, outdated pymatgen or dependency, and so on).


DanielYang59:
Also to answer your @janosh question in #3536 that:

> Shouldn't this markdown be part of the issue template?

My idea is to keep the issue template clean (and it is not a tutorial at the end of the day), and keep codespace optional (instead of "pushing" users to use codespace). However I might move the guidance**figure** into the issue template, and put the introductory staff into the repro template (as if the use follows the figure and open up the codespace, the repro template would be presented)? What do you think?

Again I'm really sorry for messing around with the commit history. But after one hour's experiment I'm now much more confident with git workflow now. Thanks.

DanielYang59:
OMG, I'm frustrated. I combined the commits by rebasing in my laptop but my desktop just pushed uncombined commits again and flooded the commit tree. I would try to fix it and I'm really sorry for this.

janosh:
> In doing this we could avoid an extra readme file. What do you think?

yes exactly, i think avoiding an extra readme file which many people might miss would be nice. could you make that change?

DanielYang59:
Done @janosh. I merged the separate guide README into the issume template (folded by default to minimize disturbance to the template overall). Let me know if you would like to see more changes.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -2,6 +2,19 @@ name: Report a Bug
...
+      value: |
+        ### Try the new Codespaces, setup env and collect info made easy.
+        <details>
+          <summary>🚀 Click to Show the Codespaces Guide</summary>
+          <img width="300" alt="Create codespace" src="https://github.com/materialsproject/pymatgen/assets/80093591/482181f8-d7e1-49b4-8382-2d948ed4e550">
+
+          1. Open up the [pymatgen repo](https://github.com/materialsproject/pymatgen) in your web browser.
+          2. Click the green `Code` button at the top-right corner of the repo, and switch to the `Codespaces` panel.
+          3. Click `Create code space on master` to create a new codespace, or to return to an existing codespace.
+        </details>
```
janosh, line 16: Rather than this multistep process, can we not just show an SVG badge with a deeplink to open a new codespace? see [Open in GitHub Codespaces docs](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/setting-up-your-repository/facilitating-quick-creation-and-resumption-of-codespaces#creating-an-open-in-github-codespaces-badge)

DanielYang59: @janosh Wonderful suggestion! Thanks and implemented.

janosh (approved):
Thanks @DanielYang59! In `devcontainer.json`, I decreased `hostRequirements.cpus: 4->1`  since Python only uses 1 CPU anyway. Unless I'm missing something?

DanielYang59:
@janosh Look good to me. Good work @janosh. 1 core should suffice.

DanielYang59:
@janosh It seems you might have left the outputs inside the repro template ipynb?

> date: 2024-01-08
> Python version: 3.11.6
> pymatgen version: 2023.11.12
> pymatgen version: 2023.11.12

janosh:
Thanks @DanielYang59, good catch! 👍

PR #3539, opened 2024-01-02
Handle invalid selective dynamics info in POSCAR

### LINKED ISSUE #3008: POSCAR is ignoring selective dynamics
I am using the pymatgen.io.vasp.inputs Module. According to the documentation (https://pymatgen.org/pymatgen.io.vasp.inputs.html) the from_file(filename) method of the Poscar class should handle selective dynamics. However it does ignore selective dynamics or read it in the wrong way.

Here is some test file called POSCAR_test:

```
Test
   1.000000000
    20.0000000000000000    0.0000000000000000    0.0000000000000000
     0.0000000000000000   23.0000000000000000    0.0000000000000000
     0.0000000000000000    0.0000000000000000   15.0000000000000000
   C     H  
   2     1
Selective dynamics
Cartesian
     2.201545747940     2.214990562232     1.000000000000  T  T  T  
     3.625859708182     2.193941473662     1.000000000000  T  T  T  
     4.326973622066     3.379594386626     1.000000000000  T  T  T  
```

Here is the code:

```
from pymatgen.io.vasp.inputs import Poscar
poscar = Poscar.from_file('POSCAR_test', False, False)
print(poscar)
```

The output:

```
Test
1.0
  20.0000000000000000    0.0000000000000000    0.0000000000000000
   0.0000000000000000   23.0000000000000000    0.0000000000000000
   0.0000000000000000    0.0000000000000000   15.0000000000000000
C H
2 1
direct
   0.1100772873970000    0.0963039374883478    0.0666666666666667 C
   0.1812929854091000    0.0953887597244348    0.0666666666666667 C
   0.2163486811033000    0.1469388863750435    0.0666666666666667 H
```

As you can see, selective dynamics are getting ignored, which should not be the case.

If you change the POSCAR_test to this:

```
Test
   1.000000000
    20.0000000000000000    0.0000000000000000    0.0000000000000000
     0.0000000000000000   23.0000000000000000    0.0000000000000000
     0.0000000000000000    0.0000000000000000   15.0000000000000000
   C     H  
   2     1
Selective dynamics
Cartesian
     2.201545747940     2.214990562232     1.000000000000 C T  T  T  
     3.625859708182     2.193941473662     1.000000000000 C T  T  T  
     4.326973622066     3.379594386626     1.000000000000 H T  T  T  
```

The output will be this:

```
Test
1.0
  20.0000000000000000    0.0000000000000000    0.0000000000000000
   0.0000000000000000   23.0000000000000000    0.0000000000000000
   0.0000000000000000    0.0000000000000000   15.0000000000000000
C H
2 1
Selective dynamics
direct
   0.1100772873970000    0.0963039374883478    0.0666666666666667 F T T C
   0.1812929854091000    0.0953887597244348    0.0666666666666667 F T T C
   0.2163486811033000    0.1469388863750435    0.0666666666666667 F T T H
```

Now the module seems to get that I want to use selective dynamics, but it sets all the x-coordinates to F and not T.

I also tried this POSCAR_test:

```
Test
   1.000000000
    20.0000000000000000    0.0000000000000000    0.0000000000000000
     0.0000000000000000   23.0000000000000000    0.0000000000000000
     0.0000000000000000    0.0000000000000000   15.0000000000000000
   C     H  
   2     1
Selective dynamics
Cartesian
     2.201545747940     2.214990562232     1.000000000000 T  T  T  C 
     3.625859708182     2.193941473662     1.000000000000 T  T  T  C
     4.326973622066     3.379594386626     1.000000000000 T  T  T  H
```

which outputs the same as the first POSCAR_test, totally ignoring selective dynamics.

DanielYang59:
## Summary

Major changes:

- Issue warning in `from_str` method when invalid selective dynamics info would found in POSCAR, to resolve #3008.
- Clean up INCAR tag checkings in `check_params` method.
- Fix missing type checkings in`test_check_params` of `tests/io/test_inputs.py`.

Minor changes:

- Issue warning if elements cannot be determined when reading POSCAR `from_file`
- Minor format improvements 

## Analysis of "Invalid selective dynamics info in POSCAR"

In the second POSCAR provided in #3008:
```
Test POSCAR
   1.000000000
    20.0000000000000000    0.0000000000000000    0.0000000000000000
     0.0000000000000000   23.0000000000000000    0.0000000000000000
     0.0000000000000000    0.0000000000000000   15.0000000000000000
   C     H  
   2     1
Selective dynamics
Cartesian
     2.201545747940     2.214990562232     1.000000000000 C T  T  T  
     3.625859708182     2.193941473662     1.000000000000 C T  T  T  
     4.326973622066     3.379594386626     1.000000000000 H T  T  T  
```

Which interprets to:
```
Test POSCAR
1.0
  20.0000000000000000    0.0000000000000000    0.0000000000000000
   0.0000000000000000   23.0000000000000000    0.0000000000000000
   0.0000000000000000    0.0000000000000000   15.0000000000000000
C H
2 1
Selective dynamics
direct
   0.1100772873970000    0.0963039374883478    0.0666666666666667 F T T C
   0.1812929854091000    0.0953887597244348    0.0666666666666667 F T T C
   0.2163486811033000    0.1469388863750435    0.0666666666666667 F T T H
```

Seems intended by `pymatgen` (anything other than "T" would be treated "F"):
https://github.com/materialsproject/pymatgen/blob/8aa1cf89a93e12c2ac16ddde517105e307a04b91/pymatgen/io/vasp/inputs.py#L430-L431

I looked into the `VASP` source code `poscar.F (version 5.4.4.pl2)` line `385` to `392` (I cannot paste the code here for copyright reasons), which defines the parsing mechanism of POSCAR by `VASP`. And it seems `VASP` reads the 3-5 (0-indexed) elements of the position line as selective dynamics, which deems such POSCAR invalid. 

## `Selective dynamics` tag dropped when all DOFs relaxed

Regarding the 1st POSCAR in the original issue. There is: https://github.com/materialsproject/pymatgen/blob/8aa1cf89a93e12c2ac16ddde517105e307a04b91/pymatgen/io/vasp/inputs.py#L121-L124

So selective dynamics array would only be kept when there exists at least one degree of freedom (DOF) frozen. And the user used an example POSCAR where all DOFs are relaxed. 

Should we issue a warning to avoid confusion? I think it's unnecessary (but why would people relax all DOFs and toggle selective dynamics at the same time in the first place).

## TODO list

- [x] Issue warning when invalid selective dynamics info detected in POSCAR
- [x] (**Need suggestions on this**) Confirm such POSCARs are truly invalid
- [x] Update unit tests for such invalid POSCARs


DanielYang59:
Sorry I realized the I should not assign a dict with only "selective_dynamics" which breaks the whole workflow. I'm working on fixing it.

DanielYang59:
Meanwhile for the second POSCAR:
```
Test
   1.000000000
    20.0000000000000000    0.0000000000000000    0.0000000000000000
     0.0000000000000000   23.0000000000000000    0.0000000000000000
     0.0000000000000000    0.0000000000000000   15.0000000000000000
   C     H  
   2     1
Selective dynamics
Cartesian
     2.201545747940     2.214990562232     1.000000000000 C T  T  T  
     3.625859708182     2.193941473662     1.000000000000 C T  T  T  
     4.326973622066     3.379594386626     1.000000000000 H T  T  T  
```

Which interprets to:
```
Test
1.0
  20.0000000000000000    0.0000000000000000    0.0000000000000000
   0.0000000000000000   23.0000000000000000    0.0000000000000000
   0.0000000000000000    0.0000000000000000   15.0000000000000000
C H
2 1
Selective dynamics
direct
   0.1100772873970000    0.0963039374883478    0.0666666666666667 F T T C
   0.1812929854091000    0.0953887597244348    0.0666666666666667 F T T C
   0.2163486811033000    0.1469388863750435    0.0666666666666667 F T T H
```

Seems still intended (anything other than "T" would be treated "F"):
https://github.com/materialsproject/pymatgen/blob/8aa1cf89a93e12c2ac16ddde517105e307a04b91/pymatgen/io/vasp/inputs.py#L430-L431

But is this a legal POSCAR in the first place? I looked into [vasp wiki](https://www.vasp.at/wiki/index.php/POSCAR) and didn't find much info about comments in position lines. As I seldom like to write comments like this.

DanielYang59:
I just looked into the VASP source code `poscar.F version 5.4.4.pl2` line 385 to 392 (I cannot paste the code here, or I'm pretty sure I would get sued for copyright issues), which defines the parsing mechanism of POSCAR. And it seems VASP reads the 3-5 (0-indexed) elements as selective dynamics, which deems this POSCAR illegal. If so, I might proceed and issue a warning from our end (or would you suggest an error) @janosh ?

pymatgen/io/vasp/inputs.py
```diff
@@ -244,9 +244,9 @@ def from_file(cls, filename, check_for_potcar=True, read_velocities=True, **kwar
 
         If all else fails, the code will just assign the first n elements in
         increasing atomic number, where n is the number of species, to the
-        Poscar. For example, H, He, Li, .... This will ensure at least a
-        unique element is assigned to each site and any analysis that does not
-        require specific elemental properties should work fine.
+        Poscar, where a warning would be issued. For example, H, He, Li, ....
+        This will ensure at least a unique element is assigned to each site and
+        any analysis that does not require specific elemental properties should work.
```
DanielYang59, line 249: I inserted a comment about warning here, so adjust the lines try not to make lines too long.

pymatgen/io/vasp/inputs.py
```diff
@@ -531,20 +532,18 @@ def get_str(self, direct: bool = True, vasp4_compatible: bool = False, significa
             lines.append("Selective dynamics")
         lines.append("direct" if direct else "cartesian")
 
-        selective_dynamics = self.selective_dynamics
```
DanielYang59, line 534: I didn't see much point to re-assign an attribute here, so I removed it.

pymatgen/io/vasp/inputs.py
```diff
@@ -978,7 +976,7 @@ def check_params(self):
                 if type(incar_params[key]).__name__ == "str":
                     # Now we check if this is an appropriate parameter type
                     if incar_params[key] == "float":
-                        if not type(val) not in ["float", "int"]:
+                        if type(val) in ["float", "int"]:
                             warnings.warn(f"{key}: {val} is not real", BadIncarWarning, stacklevel=2)
```
DanielYang59, line 980: This checking mechanism doesn't quite make sense to me... Why a warning is issued when the real type and expected type match?

pymatgen/io/vasp/inputs.py
```diff
@@ -531,20 +532,18 @@ def get_str(self, direct: bool = True, vasp4_compatible: bool = False, significa
             lines.append("Selective dynamics")
         lines.append("direct" if direct else "cartesian")
 
-        selective_dynamics = self.selective_dynamics
         for i, site in enumerate(self.structure):
             coords = site.frac_coords if direct else site.coords
             line = " ".join(format_str.format(c) for c in coords)
-            if selective_dynamics is not None:
-                sd = ["T" if j else "F" for j in selective_dynamics[i]]
```
DanielYang59, line 539: I didn't see much point to re-assign an attribute here, so I removed it.

pymatgen/io/vasp/inputs.py
```diff
@@ -271,6 +271,7 @@ def from_file(cls, filename, check_for_potcar=True, read_velocities=True, **kwar
                     potcar = Potcar.from_file(sorted(potcars)[0])
                     names = [sym.split("_")[0] for sym in potcar.symbols]
                     [get_el_sp(n) for n in names]  # ensure valid names
```
DanielYang59, line 273: I guess this floating list comprehension should be fixed.

pymatgen/io/vasp/inputs.py
DanielYang59, line 539: I didn't see much point to re-assign an attribute here, so I removed it.

DanielYang59:
I was feeling boring so went ahead and tried to fix some minor issues and formats. Hope this is acceptable (let me know if I should be opening a new PR, but I think that might be too overkill).

DanielYang59:
Meanwhile I noticed this warning here: https://github.com/materialsproject/pymatgen/blob/8aa1cf89a93e12c2ac16ddde517105e307a04b91/pymatgen/io/lobster/inputs.py#L637-L644

I had some experience with this issue, and was lucky enough to get direct input from Dr. Ryky Nelson @rnels12 (first author of the [LOBSTER paper](https://onlinelibrary.wiley.com/doi/abs/10.1002/jcc.26353)). When this imcompatibility issue happens, the charge spilling could jump from under 5% to as much as around 40%. Meanwhile the fix is pretty straightforward, by simply removing the lines in POTCAR with `SHA256` and `COPYR`, which is suggested by Dr. Nelson himself. I have tested this the fix and could confirm this could significant reduce charge spilling to normal range. 

Should I proceed and implement this fix directly in pymatgen (make a backup copy of POTCAR and modify the current POTCAR)? Meanwhile I think I might need to send Dr. Nelson another email to confirm the fix, because unfortunately my previous communication with Dr. Nelson is lost (my previous university set up an email purging policy for some reason).

JaGeo:
> Meanwhile I noticed this warning here:
> 
> https://github.com/materialsproject/pymatgen/blob/8aa1cf89a93e12c2ac16ddde517105e307a04b91/pymatgen/io/lobster/inputs.py#L637-L644
> 
> I had some experience with this issue, and was lucky enough to get direct input from Dr. Ryky Nelson @rnels12 (first author of the [LOBSTER paper](https://onlinelibrary.wiley.com/doi/abs/10.1002/jcc.26353)). When this imcompatibility issue happens, the charge spilling could jump from under 5% to as much as around 40%. Meanwhile the fix is pretty straightforward, by simply removing the lines in POTCAR with `SHA256` and `COPYR`, which is suggested by Dr. Nelson himself. I have tested this the fix and could confirm this could significant reduce charge spilling to normal range.
> 
> Should I proceed and implement this fix directly in pymatgen (make a backup copy of POTCAR and modify the current POTCAR)? Meanwhile I think I might need to send Dr. Nelson another email to confirm the fix, because unfortunately my previous communication with Dr. Nelson is lost (my previous university set up an email purging policy for some reason).

We implemented this warning after talking to @rnels12 and @QuantumChemist . If this happens, the POTCARs are not read correctly and the results you get are wrong. Of course, you can implement a fix but I am not sure what you are intending to do: generating POTCARs without the line?

JaGeo:
Btw, many of the pymatgen implementations have been used for the above LOBSTER paper. 🙃

JaGeo:
Side note: we should test if the latest LOBSTER version fixes this. If so, I would not spend time on this fix.

JaGeo:
@naik-aakash might also be able to give further input and could probably check if the latest LOBSTER fixes the issue.

> BTW, I happend to find your LobsterPy package several weeks ago and it's awesome. I would see if I could make some contributions to that soon.

Btw, We work on a JOSS at the moment.

JaGeo:
And, yes, the fix mentioned above is essentially the fix that we implemented manually.

naik-aakash:
Hi @DanielYang59, I can test if the latest LOBSTER version fixes this issue and update you here asap

DanielYang59:
@naik-aakash That would be much appreciated! My bad, I seem to be flooding this PR with irrelevant topic...

DanielYang59:
> And, yes, the fix mentioned above is essentially the fix that we implemented manually.

@JaGeo Thanks for the input. If you have time you might open up some issues to resolve and I would see if I could solve any (as I'm not super familiar with your package) so may not actually know where to get started. All the best with your paper then!

JaGeo:
> > And, yes, the fix mentioned above is essentially the fix that we implemented manually.
> 
> @JaGeo Thanks for the input. If you have time you might open up some issues and I would see if I could solve any (as I'm not quite familiar with your package at this moment) so may not actually know where to get started. All the best with your paper then!

I don't have time to implement a fix. Sorry. Let's just see if the latest LOBSTER resolves it. Thanks in any case.

DanielYang59:
Sorry @JaGeo I didn't express myself clear enough. I mean I might divide into your Lobsterpy package and see if I could help contribute something. If you happen to have some work to do for Lobsterpy, I might try to fix some. Cheers!

pymatgen/entries/compatibility.py
```diff
@@ -803,16 +803,14 @@ def __init__(
...
-"""
-Note from Ryan Kingsbury (2022-10-14): MaterialsProject2020Compatibility inherits from Compatibility
-instead of CorrectionsList which came before it because CorrectionsList had technical limitations.
-When we did the new scheme (MP2020) we decided to refactor the base Compatibility class to not
-require CorrectionsList.
-
-This was particularly helpful for the AqueousCorrection class. The new system gives complete
-flexibility to process entries however needed inside the get_adjustments() method, rather than
-having to create a list of separate correction classes.
-"""
```
DanielYang59, line 815: I migrated this minor format fix from closed PR #3534 as I don't think it's worth opening up a PR for just this

pymatgen/io/vasp/inputs.py
```diff
@@ -964,29 +960,22 @@ def __add__(self, other):
...
+            elif isinstance(incar_params[tag], str) and type(val).__name__ != incar_params[tag]:
+                warnings.warn(f"{tag}: {val} is not a {incar_params[tag]}", BadIncarWarning, stacklevel=2)
 
-                # if we have a list of possible parameters, check
-                # if the user given parameter is in this list
-                elif type(incar_params[key]).__name__ == "list" and val not in incar_params[key]:
-                    warnings.warn(f"{key}: Cannot find {val} in the list of parameters", BadIncarWarning, stacklevel=2)
+            # Check if the given value is in the list
+            elif isinstance(incar_params[tag], list) and val not in incar_params[tag]:
+                warnings.warn(f"{tag}: Cannot find {val} in the list of values", BadIncarWarning, stacklevel=2)
```
DanielYang59, line 978: I tried to clean up the INCAR check method here. Please help me check @janosh  as I'm not quite confident if changing the unit test would break anything. Several things to clarify here:

1. I tried to use "tag" to refer to the name of keys in INCAR files, trying to be consistent with [VASP wiki](https://www.vasp.at/wiki/index.php/Category:INCAR_tag). Prior to this, a mixture of "key", "parameter", "flag" and "tag" are used and was quite confusing.

2. The checking mechanism for real numbers is buggy and unnecessary (and the checking is missing in unit test), (plus number ) so I merged it to a type check.

tests/io/vasp/test_inputs.py
```diff
@@ -721,51 +721,31 @@ def test_proc_types(self):
...
     def test_check_params(self):
-        # Triggers warnings when running into nonsensical parameters
+        # Triggers warnings when running into invalid parameters
         with pytest.warns(BadIncarWarning) as record:
             incar = Incar(
                 {
                     "ADDGRID": True,
                     "ALGO": "Normal",
                     "AMIN": 0.01,
-                    "AMIX": 0.2,
```
DanielYang59, line 731: Simplify the test INCAR a bit, keep only one entry of each data type.

tests/io/vasp/test_inputs.py
```diff
@@ -721,51 +721,31 @@ def test_proc_types(self):
...
-                    "PREC": "Accurate",
-                    "SIGMA": 0.2,
-                    "NBAND": 250,  # spelling mistake
-                    "PHON_TLIST": "is_a_str",  # this parameter should be a list
-                    "LATTICE_CONSTRAINTS": [
-                        True,
-                        False,
-                        "f",
-                    ],  # Should be a list of bools
-                    "M_CONSTR": [True, 1, "string"],  # Should be a list of real numbers
```
DanielYang59, line 759: Meanwhile I noticed the type checking for the values of "LATTICE_CONSTRAINTS" and "M_CONSTR" was not implemented in the `check_params` method. As there isn't explicit definition for the type of list elements in `incar_parameters.json` (such as List[bool]).

tests/io/vasp/test_inputs.py
```diff
@@ -721,51 +721,31 @@ def test_proc_types(self):
...
             incar.check_params()
 
-        assert "ISIF: Cannot find 9 in the list of parameters" in record[0].message.args
-        assert "LASPH: 5 is not a bool" in record[1].message.args
-        assert "METAGGA: Cannot find SCAM in the list of parameters" in record[2].message.args
-        assert "Cannot find NBAND in the list of INCAR flags" in record[3].message.args
-        assert "PHON_TLIST: is_a_str is not a list" in record[4].message.args
+        assert record[0].message.args[0] == "Cannot find NBAND in the list of INCAR tags"
+        assert record[1].message.args[0] == "METAGGA: Cannot find SCAM in the list of values"
+        assert record[2].message.args[0] == "EDIFF: (5+1j) is not a float"
```
DanielYang59, line 745: warning for `"EDIFF": 5 + 1j, # value should be a float` was missing

pymatgen/io/vasp/inputs.py
```diff
@@ -2262,20 +2244,18 @@ def parse_fortran_style_str(input_str: str) -> Any:
...
-                        The (E10.0) means write an exponential-format number with 10
-                            characters before the decimal, and 0 characters after
-                    """
+                    # NB: fortran style floats always include a decimal point.
+                    #     While you can set, e.g., x = 1E4, you cannot print/write x without
+                    #     a decimal point:
+                    #         `write(6,*) x`          -->   `10000.0000` in stdout
+                    #         `write(6,'(E10.0)') x`  -->   segfault
+                    #     The (E10.0) means write an exponential-format number with 10
+                    #         characters before the decimal, and 0 characters after
```
DanielYang59, line 2258: converted string to real comment

pymatgen/io/vasp/inputs.py
```diff
@@ -2234,14 +2218,12 @@ def is_valid(self) -> bool:
...
-        LDA pseudopotential.
+        # Some POTCARs have an LEXCH (functional used to generate the POTCAR)
+        # with the expected functional, e.g., the C_d POTCAR for PBE is actually an
+        # LDA pseudopotential.
 
-        Thus we have to look for matches in all POTCAR dirs, not just the ones with
-        consistent values of LEXCH
-        """
+        # Thus we have to look for matches in all POTCAR dirs, not just the ones with
+        # consistent values of LEXCH
```
DanielYang59, line 2226: converted string to real comment

pymatgen/io/vasp/inputs.py
```diff
@@ -1614,13 +1602,11 @@ class PotcarSingle:
...
-        the original (univie) releases include no SHA256 hashes nor COPYR fields
-        in the PSCTR/header field.
-    We indicate the older release in `functional_dir` as PBE_52, PBE_54, LDA_52, LDA_54.
-    The newer release is indicated as PBE_52_W_HASH, etc.
-    """
+    # NB: there are multiple releases of the {LDA,PBE} {52,54} POTCARs
+    #     the original (univie) releases include no SHA256 hashes nor COPYR fields
+    #     in the PSCTR/header field.
+    # We indicate the older release in `functional_dir` as PBE_52, PBE_54, LDA_52, LDA_54.
+    # The newer release is indicated as PBE_52_W_HASH, etc.
```
DanielYang59, line 1609: converted string to real comment

pymatgen/io/vasp/inputs.py
```diff
@@ -345,23 +346,21 @@ def from_str(cls, data, default_names=None, read_velocities=True):
...
+            #    1.00000000000000
+            #      8.5415010000000002   -0.0077670000000000   -0.0007960000000000
+            #     -0.0077730000000000    8.5224019999999996    0.0105580000000000
+            #     -0.0007970000000000    0.0105720000000000    8.5356889999999996
+            #    Fe   Cr   Fe   Cr   Fe   Cr   Fe   Cr   Fe   Cr   Fe   Cr   Fe   Cr   Fe   Ni   Fe   Cr   Fe   Cr
+            #    Fe   Ni   Fe   Cr   Fe
+            #      1   1   2   4   2   1   1   1     2     1     1     1     4     1     1     1     5     3     6     1
+            #      2   1   3   2   5
+            # Direct
+            #   ...
```
DanielYang59, line 363: converted string to real comment

pymatgen/entries/compatibility.py
```diff
@@ -803,16 +803,14 @@ def __init__(
...
+# Note from Ryan Kingsbury (2022-10-14): MaterialsProject2020Compatibility inherits from Compatibility
+# instead of CorrectionsList which came before it because CorrectionsList had technical limitations.
+# When we did the new scheme (MP2020) we decided to refactor the base Compatibility class to not
+# require CorrectionsList.
+
+# This was particularly helpful for the AqueousCorrection class. The new system gives complete
+# flexibility to process entries however needed inside the get_adjustments() method, rather than
+# having to create a list of separate correction classes.
```
DanielYang59, line 813: converted string to real comment

pymatgen/io/vasp/inputs.py
```diff
@@ -428,7 +427,10 @@ def from_str(cls, data, default_names=None, read_velocities=True):
             crd_scale = scale if cart else 1
             coords.append([float(j) * crd_scale for j in tokens[:3]])
             if has_selective_dynamics:
-                selective_dynamics.append([tok.upper()[0] == "T" for tok in tokens[3:6]])
+                if all(value in {'T', 'F'} for value in tokens[3:6]):
+                    selective_dynamics.append([value == 'T' for value in tokens[3:6]])
```
DanielYang59: Please help me confirm that VASP indeed only takes the `token[3:6]` as selective dynamics values.

pymatgen/io/vasp/inputs.py
```diff
@@ -663,14 +663,10 @@ def set_temperature(self, temperature: float):
         self.structure.add_site_property("velocities", velocities.tolist())
 
 
-with open(f"{module_dir}/incar_parameters.json") as json_file:
+with open(f"{module_dir}/incar_parameters.json", encoding="utf-8") as json_file:
     incar_params = json.loads(json_file.read())
 
 
-class BadIncarWarning(UserWarning):
-    """Warning class for bad Incar parameters."""
```
DanielYang59, line 671: I didn't see much point in defining a UserWarning when it's only called three times within the same method (`check_params`).

naik-aakash:
> @naik-aakash That would be much appreciated! My bad, I seem to be flooding this PR with irrelevant topic...

The issue with high spilling when POTCARs have `SHA256` and `COPYR` seems to be solved in the latest [LOBSTER version 5.0.0](http://www.cohp.de/)

JaGeo:
Great. Then, we just leave the warning for now (and potentially update the LOBSTER version in the text). I wouldn't attempt a fix as people should always use the latest LOBSTER version. It's also not hard to switch as it has an academic license and does not need much setup in contrast to codes like VASP.

DanielYang59:
Great! Thanks a lot for the confirmation @naik-aakash. Yes I agree with you, pymatgen should not patch every downstream issue, especially when it could be fixed with negligible cost by the user (it seems to be the `lobsterin` for 4.1.0 should also work for 5.0.0, but I could be wrong.). 

Meanwhile it might be good to implement a lobster version check before issuing a warning? What do you think? @JaGeo I currently cannot think of a better method than running command `lobster` or check the first line of `lobsterout` file?

JaGeo:
> Great! Thanks a lot for the confirmation @naik-aakash. Yes I agree with you, pymatgen should not patch every downstream issue, especially when it could be fixed with negligible cost by the user (it seems to be the `lobsterin` for 4.1.0 should also work for 5.0.0, but I could be wrong.).
> 
> Meanwhile it might be good to implement a lobster version check before issuing a warning? What do you think? @JaGeoBut I currently cannot think of a better method than running command `lobster` or check the first line of `lobsterout` file?

We could do this. However, not sure how consistent versioning is ... Especially for test versions.

DanielYang59:
@JaGeo That would be wonderful. Meanwhile it appears to me lobster doesn't seem to have a version checking command? Maybe I'm wrong? I tried `lobster --version`/`lobster -v`/`lobster -V` and none seems to be. Also I didn't find any from the Manual.

pymatgen/entries/compatibility.py
rkingsbury: I'm not sure what the 'official' best practice on multi-line comments is. This seems fine to me though.

DanielYang59: I'm not 100% confident about the "official" multi-line comments either, and I would like to hear more comments. But here is some info I have so far:
From the [PEP-8 doc](https://peps.python.org/pep-0008/#comments) under the `comments` section, there is:

> [Block Comments](https://peps.python.org/pep-0008/#block-comments)
Block comments generally apply to some (or all) code that follows them, and are indented to the same level as that code. **Each line of a block comment starts with a # and a single space** (unless it is indented text inside the comment).
    Paragraphs inside a block comment are separated by a line containing a single #.

So I assume it might be slightly better (though it should work almost the same) to use a stack of "#" signs, instead of a "floating" string (which should be specific to docstrings from my understanding), meanwhile `pylint` would give me `String statement has no effectPylint [W0105:pointless-string-statement]` warning.

DanielYang59:
Can you please review this? @janosh . I need confirmation on the handling of selective dynamics info by `VASP`. Thanks!

janosh:
> I need confirmation on the handling of selective dynamics info by VASP

I'm not the best person for that. @JaGeo @esoteric-ephemera @Andrew-S-Rosen and others are much more qualified here

DanielYang59:
Sorry I'm not quite familiar with git rebase and it seems I have messed up the commit history again. I'm working on rebase and squashing the above commits. Sorry for the inconvenience.

janosh:
> I'm working on rebase and squashing the above commits. Sorry for the inconvenience.

No need. this PR is mergeable as is. once this is fully reviewed i'll take of any diverging git history

DanielYang59:
Thanks @janosh, it's so nice of you. I'm feeling so bad for messing with the commit history😭. I would go the learn all these again.

JaGeo:
Sorry, I am still sick (third week) and therefore behind on many things. 🙃

DanielYang59:
So sorry to hear that @JaGeo. No rush at all. Hope you could get well soon.

esoteric-ephemera:
Hey @DanielYang59 and @janosh - I just posted a reply to issue #3008. That issue can probably be closed with the explanation given

The short answer is yes, the author of 3008 had incorrectly-formatted POSCARs, and was unaware that at least one degree of freedom had to be disabled to use selective dynamics

The warning in this PR won't be raised if, e.g., a POSCAR contains fluorine and reverses the order of the site labels and selective dynamics tags, since the first letter is then F.  The only way around that is to check if the labels at the end of the line match the expected composition, which they don't have to (might be wrong there)

The point of having specific warnings like `BadIncarWarning` is to help the user in debugging. Even if it's used quite sparingly within PMG, the warning points you to INCAR being incorrectly formatted. I don't see the rationale in removing this

DanielYang59:
Good point @esoteric-ephemera. I didn't realize `fluorine` would be a special case and it's quite tricky to handle. I cannot think it a good solution for this and might wait to see if others have a more elegant way around.

Sure I totally agree with your point to keep the `BadIncarWarning`. The reason I wish to remove it previously is that I browsed through the `inputs.py` module and notice custom warning is issued only for bad `INCAR`, and within the same method, not for any other input files (not `BadPoscarWarning` or such). I would reverse this deletion for now and maybe we can add a specific warning for each input file.

DanielYang59:
It appears to me the above checking should be able to capture most invalid POSCARs, except for some rare cases including Fluorine. But I think our safety margin is still generous.

As far as I can think of, only the following invalid POSCARs cannot be captured by the above check.

### Some invalid POSCARs could bypass the check

```POSCAR
Invalid POSCAR with one Fluorine only
1
 0.50000000 0.50000000 0.00000000
 0.00000000 0.50000000 0.50000000
 0.50000000 0.00000000 0.50000000
  F
  1
Selective dynamics
Cartesian
0.00000000 0.00000000 0.00000000 F T T F
```
There is no way to tell whether the intended selective dynamics info is `F T T` or `T T F`. But why would people attempt to relax a single atom strutucre in the first place?

```POSCAR
Invalid POSCAR with multiple Fluorine only
1
 0.50000000 0.50000000 0.00000000
 0.00000000 0.50000000 0.50000000
 0.50000000 0.00000000 0.50000000
  F
  3
Selective dynamics
Cartesian
0.00000000 0.00000000 0.00000000 F T T F
0.00000000 0.00000000 0.15000000 F T T F
0.00000000 0.00000000 0.30000000 F T T F
```
This time the user has a good reason to relax the structure, but I'm not aware of any chemical structure consisting Fluorine only myself.

```POSCAR
Invalid POSCAR with Fluorine, but only Fluorine position has comment
1
 0.50000000 0.50000000 0.00000000
 0.00000000 0.50000000 0.50000000
 0.50000000 0.00000000 0.50000000
  Na F
  1   1
Selective dynamics
Cartesian
0.00000000 0.00000000 0.00000000 T T T      # Does not has chemical symbol tag
0.00000000 0.00000000 0.15000000 F T T F    # Has a confusing chemical symbol tag
```
In this case, if the user intend to apply a `T T F` selective dynamics constrain on Flurione, the check would be bypassed 😓. But...I'm out of it... There is no way to tell...

### Some invalid POSCARs could not bypass the check

```POSCAR
Invalid POSCAR with Fluorine
1
 0.50000000 0.50000000 0.00000000
 0.00000000 0.50000000 0.50000000
 0.50000000 0.00000000 0.50000000
  Na F
  1   1
Selective dynamics
Cartesian
0.00000000 0.00000000 0.00000000 Na T T T
0.00000000 0.00000000 0.15000000 F  T T F    
```
In this slightly rational case, `F T T F` would bypass the chec, but `Na T T T` would still be captured, and therefore a warning would be issued.

So this check should be fairly robust, and only fail in some extreme cases. What do you think? @esoteric-ephemera

esoteric-ephemera:
We have about 12,100 structures on the Materials Project that contain fluorine, so this isn't really an edge case (for example, we have fewer structures that contain nitrogen).

Can we also issue a warning when there are four items on that line, and the structure is known to contain fluorine? Just to tell the user to double check the format, and give the correct format

Also, would be good to issue a warning when the user requests selective dynamics, but sets all degrees of freedom to relax, as in #3008. Just to let the user know that selective dynamics can only be used when at least one DOF should not be relaxed

DanielYang59:
Hi @esoteric-ephemera, thanks for the input and great points as always! Sure I would implement these as you suggested today.

Yes I totally agree to raise a warning when selective dynamics are toggle with all DOFs relaxed to avoid confusion.

pymatgen/io/vasp/inputs.py
```diff
@@ -419,16 +420,36 @@ def from_str(cls, data, default_names=None, read_velocities=True):
...
             crd_scale = scale if cart else 1
             coords.append([float(j) * crd_scale for j in tokens[:3]])
             if has_selective_dynamics:
-                selective_dynamics.append([tok.upper()[0] == "T" for tok in tokens[3:6]])
+                # Warn when values contain suspicious entries
+                if any(value not in {"T", "F"} for value in tokens[3:6]):
+                    warnings.warn("Selective dynamics values must be either 'T' or 'F'.", BadPoscarWarning)
+
+                # Warn when elements contains Fluorine (F) (#3539)
+                if atomic_symbols[i] == "F" and len(tokens[3:]) >= 4 and "F" in tokens[3:7]:
```
DanielYang59, line 441: The warning is intended to be species, and is only issued when all three conditions are met:
1. The element for this line is Fluorine;
2. This line does have extra information (comment or such);
3. There is "F" within the suspicious range.

QuantumChemist:
> > > And, yes, the fix mentioned above is essentially the fix that we implemented manually.
> > 
> > 
> > @JaGeo Thanks for the input. If you have time you might open up some issues and I would see if I could solve any (as I'm not quite familiar with your package at this moment) so may not actually know where to get started. All the best with your paper then!
> 
> I don't have time to implement a fix. Sorry. Let's just see if the latest LOBSTER resolves it. Thanks in any case.

Hi @DanielYang59 ,
just wanted to let you know that one of the current LOBSTER maintainers and lead developer is [David Schnieders](https://www.iac.rwth-aachen.de/cms/IAC/Das-Institut/Personen/~plqh/Mitarbeiter-CAMPUS-/?gguid=0x54966D61721C6A488FEAC97A65197DD9&lidx=1&allou=1)  who might also help you and there's also the possibility to contact current LOBSTER developers via [email](mailto:mail@cohp.de) for queries and technical questions.

pymatgen/io/vasp/inputs.py
esoteric-ephemera: Thanks @DanielYang59, sorry I missed this last week. This looks good!

pymatgen/io/vasp/inputs.py
```diff
@@ -419,16 +420,36 @@ def from_str(cls, data, default_names=None, read_velocities=True):
...
+                    warnings.warn(
+                        "Selective dynamics with Fluorine. Make sure the 4th-6th entry is selective dynamics.",
+                        BadPoscarWarning,
+                    )
+
+                selective_dynamics.append([value == "T" for value in tokens[3:6]])
+
+        # Warn when ALL degrees of freedom relaxed (#3539)
+        if has_selective_dynamics and all(all(i is True for i in in_list) for in_list in selective_dynamics):
+            warnings.warn("Selective dynamics toggled with ALL degrees of freedom relaxed.", BadPoscarWarning)
```
esoteric-ephemera, line 451: Maybe make this warning more descriptive: "Selective dynamics can only be used when at least one degree of freedom is not permitted to relax. All degrees of freedom are currently allowed to relax."

pymatgen/io/vasp/inputs.py
DanielYang59: @esoteric-ephemera No rush at all. Thanks for reviewing. I was down for a whole week because of flu.

DanielYang59:
@QuantumChemist Thanks a lot for the information and really glad to know that. I would certain reach out to them whenever possible. It's so nice of you and your team members!

pymatgen/io/vasp/inputs.py
DanielYang59: Hi @esoteric-ephemera thanks for reviewing and the suggestion. Would that be a little too verbose? I think we just need to make users aware that they might be doing something improper? Also to avoid confusion from the user such as in the original issue #3008?


Meanwhile I think `VASP` would be perfectly happy to toggle selective dynamics with all DOFs relaxed, so I don't quite agree with the following message:

> Selective dynamics can only be used when at least one degree of freedom is not permitted to relax.

Also from [VASP wiki](https://www.vasp.at/wiki/index.php/POSCAR):

> If the selective dynamics feature is enabled on each coordinate triplet is followed by three additional logical flags, i.e. each is either T or F for true and false, respectively. This determines whether to allow changes of the coordinates or not. If the line selective dynamics is removed from the POSCAR file this flag will be ignored (and internally set to T).

So I think what we should do is to remain the user some change (in this case the selective dynamics flag would be dropped) is expected to avoid confusion. What do you think?

esoteric-ephemera: To avoid confusion, I think it's better to have a descriptive warning message that explains why the warning is thrown. "Ignoring `selective dynamics` tag, as no ionic degrees of freedom were disabled." is maybe clearer and more to the point.

Just a suggestion, hope you're feeling better!

pymatgen/io/vasp/inputs.py
```diff
@@ -419,16 +420,43 @@ def from_str(cls, data, default_names=None, read_velocities=True):
...
+                        "Selective dynamics toggled with Fluorine element detected. Make sure the 4th-6th entry each position line is selective dynamics info.",
+                        BadPoscarWarning,
+                    )
+
+                selective_dynamics.append([value == "T" for value in tokens[3:6]])
+
+        # Warn when ALL degrees of freedom relaxed (#3539)
+        if has_selective_dynamics and all(all(i is True for i in in_list) for in_list in selective_dynamics):
+            warnings.warn(
+                "Ignoring selective dynamics tag, as no ionic degrees of freedom were fixed.",
```
DanielYang59, line 456: Is it looking better? @esoteric-ephemera Thanks!

esoteric-ephemera: Perfect - both warnings should be clear to the user

tests/io/vasp/test_inputs.py
esoteric-ephemera: I don't know that it's possible to do more careful type checking the way with the current framework (e.g., `LATTICE_CONSTRAINTS : Sequence[bool, bool, bool]`) but that's good to note.

esoteric-ephemera (approved):

pymatgen/io/vasp/inputs.py
DanielYang59: Thanks for reviewing @esoteric-ephemera

DanielYang59:
> I don't know that it's possible to do more careful type checking the way with the current framework (e.g., LATTICE_CONSTRAINTS : Sequence[bool, bool, bool]) but that's good to note.

It should be possible and I would have a closer look soon. 

It would require updating the `incar_parameters.json` file first. I'm not sure if updating the `incar_parameters.json` file would break anything else, but I just did a search over the entire `pymatgen` repo and didn't it being called elsewhere. So it might be safe to update this json file I assume?

DanielYang59:
Please don't merge yet. @janosh I'm working on updating the INCAR tag checking mechanism as suggested by @esoteric-ephemera and would let you know once it's ready. Thanks!

DanielYang59:
I think we need to update the `incar_parameters.json` file a little bit. @esoteric-ephemera Currently the json file has a mixture of type hints and possible values, making it hard to tell one from another. Do you have any suggestions?

```json
"ADDGRID": "bool",
 "AEXX": "float",
 "AGGAC": "float",
 "AGGAX": "float",
 "ALDAC": "float",
 "ALGO": [
  "Normal",
  "VeryFast",
  "Fast",
  "Conjugate",
  "All",
  "Damped",
  "Subrot",
  "Eigenval",
  "Exact",
  "None",
  "Nothing",
  "CHI",
  "G0W0",
  "GW0",
  "GW",
  "scGW0",
  "scGW",
  "G0W0R",
  "GW0R",
  "GWR",
  "scGW0R",
  "scGWR",
  "ACFDT",
  "RPA",
  "ACFDTR",
  "RPAR",
  "BSE",
  "TDHF",
  "Timeev",
  "EVGW0",
  "EVGW",
  "QPGW0",
  "QPGW",
  "QPGW0R",
  "QPGWR"
 ],
 "AMIN": "float",
 "AMIX": "float",
 "AMIX_MAG": "float",
 "ANDERSEN_PROB": "float",
 "ANTIRES": [
  0,
  1,
  2
 ],
```

pymatgen/io/vasp/incar_parameters.json
```diff
@@ -1,653 +1,1447 @@
...
+ },
+ "LREAL": {
+   "values": [
+      false,
+      true,
+      "O",
+      "A",
+      "Auto",
+      "On"
+     ]
```
DanielYang59, line 660: I'm still looking into possible way to record more complex type hints like `bool | str`

pymatgen/io/vasp/inputs.py
```diff
@@ -964,29 +986,44 @@ def __add__(self, other):
...
+            # Check value and its type
+            param_type = incar_params[tag].get("type")
+            allowed_values = incar_params[tag].get("values")
 
-                # if we have a list of possible parameters, check
-                # if the user given parameter is in this list
-                elif type(incar_params[key]).__name__ == "list" and val not in incar_params[key]:
-                    warnings.warn(f"{key}: Cannot find {val} in the list of parameters", BadIncarWarning, stacklevel=2)
+            if param_type is not None \
+                    and not isinstance(val, eval(param_type)):
```
DanielYang59, line 1012: `eval()` might potentially bring security vulnerabilities but as we're not reading input from users here, it should be safe. For some reason `literal_eval` doesn't work and I would see if I could fix it.

DanielYang59:
Can you help give me some suggestions? @esoteric-ephemera I'm totally running out of ideas to revise the checking mechanism of `INCAR` parameters.

Current situation being:
- I still stick to `eval()` to cast a type string to type (string `"int"` to type `int`)), as the type is taken from `incar_parameters.json` and I didn't expect significant security issues.
- The safer alternative to `eval` `literal_eval` function from `ast` doesn't seem to be able to convert string `"bool"` to type `bool`.
- Current check of more complex types might potential break the program if the complex type is not imported in `inputs.py`, for example if there is a new type in `incar_parameters.json` (say `Union`) and is not imported in `inputs.py`, a `NameError: name 'Union' is not defined` would be thrown. This is quite risky as:
    - The import of types is quite implicit in `inputs.py`, and could be easily removed by any format checking tools as the import appears unused.
    - Anyone who update the `incar_parameters.json` could easily forgot to update the type in `inputs.py` as well.

So I think I cannot think of an elegant way to handle more complex type checkings... Any suggestion would be appreciated!😆

esoteric-ephemera:
Hey @DanielYang59  - I don't have good suggestions re security issues / type checkign. For the sake of getting the previous changes into PMG, it would make more sense to open a separate PR for type checking INCAR params

DanielYang59:
Great point! I would remove the type checking part and move them to a separate PR! Thanks

DanielYang59:
Can you please review? @janosh @esoteric-ephemera I reverted the changes in "improve INCAR tag checking" and would push them in a separate PR. Thanks!

pymatgen/io/vasp/inputs.py
```diff
@@ -1828,7 +1848,7 @@ def from_file(cls, filename: str) -> PotcarSingle:
         Returns:
             PotcarSingle
         """
-        match = re.search(r"(?<=POTCAR\.)(.*)(?=.gz)", str(filename))
+        match = re.search(r"(?<=POTCAR\.)(.*)(?=.gz)", filename)
```
janosh, line 1851: i assume the `str(filename)` was there to handle `pathlib.Path` so let's keep it

janosh (changes requested):
sorry about the slow review and resulting merge conflicts! just a minor comment from me but after conflict resolution this lgtm

pymatgen/io/vasp/inputs.py
DanielYang59: Thanks for reviewing @janosh. Sure let's keep the `str()` for now. I would resolve these conflicts and @esoteric-ephemera should have already reviewed this as the INCAR tag checking changes have been reverted, and would be pushed in a separate PR.

janosh:
@esoteric-ephemera if you have time, could you take a quick look at why `PotcarSingle` `test_is_valid` is now failing?

```py
    def test_is_valid(self):
>       assert self.psingle_Fe.is_valid
E       AssertionError: assert False
E        +  where False = PotcarSingle(symbol='Fe', functional='PBE', TITEL='PAW_PBE Fe 06Sep2000', VRHFIN='Fe:  d7 s1', n_valence_elec=8).is_valid
```

DanielYang59:
I noticed another test is failing too @esoteric-ephemera :
```bash
___________________________________ TestPotcarSingle.test_multi_potcar_with_and_without_sha256 ____________________________________ 

self = <tests.io.vasp.test_inputs.TestPotcarSingle testMethod=test_multi_potcar_with_and_without_sha256>

    def test_multi_potcar_with_and_without_sha256(self):
        filename = f"{FAKE_POTCAR_DIR}/POT_GGA_PAW_PBE_54/POTCAR.Fe_O.gz"
        potcars = Potcar.from_file(filename)
        # Still need to test the if POTCAR can be read.
        # No longer testing for hashes
        for psingle in potcars:
            if psingle.hash_sha256_from_file:
                assert psingle.sha256_computed_file_hash == psingle.hash_sha256_from_file
            else:
>               assert psingle.is_valid
E               AssertionError: assert False
E                +  where False = PotcarSingle(symbol='Fe_pv', functional='PBE', TITEL='PAW_PBE Fe_pv 02Aug2007', VRHFIN='Fe:  3pd7s1', n_valence_elec=14).is_valid
```

pymatgen/io/vasp/inputs.py
```diff
@@ -321,7 +321,7 @@ def from_str(cls, data, default_names=None, read_velocities=True) -> Poscar:
             raise ValueError("Empty POSCAR")
 
         # Parse positions
-        lines = tuple(clean_lines(chunks[0].split("\n"), remove_empty_lines=False))
+        lines = list(clean_lines(chunks[0].split("\n"), remove_empty_lines=False))
```
DanielYang59, line 324: Fixed the `mypy` type annotation error: `pymatgen\io\vasp\inputs.py:481: error: Incompatible types in assignment (expression has type "list[str]", variable has type "tuple[Any, ...]")  [assignment]`

pymatgen/io/vasp/inputs.py
```diff
@@ -312,7 +312,7 @@ def from_str(cls, data, default_names=None, read_velocities=True) -> Poscar:
             Poscar object.
         """
         # "^\s*$" doesn't match lines with no whitespace
-        chunks = re.split(r"\n\s*\n", data.rstrip(), flags=re.MULTILINE)
+        chunks: list[str] = re.split(r"\n\s*\n", data.rstrip(), flags=re.MULTILINE)
```
DanielYang59, line 315: Fixed another `mypy` error: `pymatgen\io\vasp\inputs.py:481: error: Incompatible types in assignment (expression has type "list[str] | Any", variable has type "tuple[Any, ...]")  [assignment]`

pymatgen/io/vasp/inputs.py
```diff
@@ -443,7 +443,7 @@ def from_str(cls, data, default_names=None, read_velocities=True) -> Poscar:
                 selective_dynamics.append([value == "T" for value in tokens[3:6]])
 
         # Warn when ALL degrees of freedom relaxed (#3539)
-        if has_selective_dynamics and all(all(i is True for i in in_list) for in_list in selective_dynamics):
+        if selective_dynamics is not None and all(all(i is True for i in in_list) for in_list in selective_dynamics):
```
DanielYang59, line 446: Use `selective_dynamics` instead of `has_selective_dynamics` to avoid: `pymatgen\io\vasp\inputs.py:443: error: Item "None" of "list[Any] | None" has no attribute "append"  [union-attr]`

DanielYang59:
There is one last `mypy` type annotation error (unrelated to this PR but still be good if we could fix it), from the last line of this code block: https://github.com/materialsproject/pymatgen/blob/12ab947261cda104ac87548429c69ce5014bbf02/pymatgen/io/vasp/inputs.py#L434-L472

I got:
```bash
pymatgen\io\vasp\inputs.py:468: error: Item "None" of "list[Any] | None" has no attribute "append"  [union-attr]
pymatgen\io\vasp\inputs.py:474: error: Item "None" of "list[Any] | None" has no attribute "append"  [union-attr]
pymatgen\io\vasp\inputs.py:496: error: Item "None" of "list[Any] | None" has no attribute "append"  [union-attr]
```

A simple example to reproduce this type annotation error:
```python
met_condition = True

if met_condition:
    foo = []

    for i in range(3):
        foo.append(i)

else:
    foo = None
```
I got `error: Incompatible types in assignment (expression has type "None", variable has type "list[int]")  [assignment]`.

If I change it to:
```python
from typing import Union


met_condition = True

if met_condition:
    foo: Union[list, None] = []

    for i in range(3):
        foo.append(i)

else:
    foo = None
```
then the error message would become: `error: Item "None" of "list[Any] | None" has no attribute "append"  [union-attr]`

Do you have any suggestions? @janosh Thanks!

janosh:
> Do you have any suggestions? @janosh Thanks!

instead of setting to `None`, you could initialize those variables to empty lists. that way, there's no type union. you'd only have to change the `Poscar.__init__` to accommodate:

```diff
- if velocities is not None:
+ if velocities:
```

and likewise for the other variables

DanielYang59:
Thanks a lot @janosh. Updated and confirmed to fix the `mypy` error.

pymatgen/io/vasp/inputs.py
```diff
@@ -2377,24 +2387,24 @@ def _gen_potcar_summary_stats(
...
-    for func in PotcarSingle.functional_dir:
-        cpsp_dir = f"{vasp_psp_dir}/{PotcarSingle.functional_dir[func]}"
+    for func, func_dir in PotcarSingle.functional_dir.items():
+        cpsp_dir = f"{vasp_psp_dir}/{func_dir}"
         if os.path.isdir(cpsp_dir):
-            func_dir_exist[func] = PotcarSingle.functional_dir[func]
+            func_dir_exist[func] = func_dir
         else:
-            warnings.warn(f"missing {PotcarSingle.functional_dir[func]} POTCAR directory")
+            warnings.warn(f"missing {func_dir} POTCAR directory")
```
DanielYang59, line 2395: > @DanielYang59, see my comment above, these tests pass outside this PR. They're failing because of changes made by this PR that are easy to fix - opened a PR on your fork to fix it.

@esoteric-ephemera Thanks a lot for pointing out! I didn't even realize its my changes breaking the unit tests. (Didn't see the comments for some reason). But I fixed it here (removed previous re-assignment of `casp_dir` to `func_dir`).

@janosh Should be ready to merge. Thanks a lot.

pymatgen/entries/compatibility.py
janosh: i think both versions are fine here

janosh (approved):
thanks a lot @DanielYang59! this is much appreciated. 👍

DanielYang59:
Thanks a lot for reviewing @janosh and everyone for giving me suggestions @esoteric-ephemera @JaGeo @rkingsbury

PR #3541, opened 2024-01-03
Fix bug in SQSTransformation

esoteric-ephemera:
The `SQSTransformation` class in `pymatgen.transformations.advanced_transformations` has a `return_ranked_list` kwarg in `apply_transformation`. 

Because this can be either boolean (return all best SQS structures from parallel runs) or an int (return only the `return_ranked_list` best SQS structures) and `bool` subclasses `int`in python, a previous check using `isinstance(return_ranked_list,int)` lead to returning only the first element in the ranked structure list (`int(True)=1`) rather than all structures in the list.

Fixed this bug and added to the existing TestSQSTransformation class to ensure this is checked.


mkhorton:
Thanks Aaron!

mkhorton:
Also noting another (Pythonic) SQS generator in `icet` although I have not tried it myself.

janosh (approved):
Thanks a lot @esoteric-ephemera! 👍

pymatgen/transformations/advanced_transformations.py
```diff
@@ -1874,31 +1874,33 @@ def _proj(b, a):
...
-                       scaling=4 would lead to a 32 atom supercell
-                    b. A sequence of three scaling factors, e.g., [2, 1, 1], which
-                       specifies that the supercell should have dimensions 2a x b x c
+                a. (preferred) Scales number of atoms, e.g., for a structure with 8 atoms,
+                    scaling=4 would lead to a 32 atom supercell
+                b. A sequence of three scaling factors, e.g., [2, 1, 1], which
+                    specifies that the supercell should have dimensions 2a x b x c
             cluster_size_and_shell (Optional[Dict[int, int]]): Dictionary of cluster interactions with entries in
                 the form number of atoms: nearest neighbor shell
         Keyword Args:
```
janosh, line 1906: @esoteric-ephemera Could you check if this doc string is incorrect?

> Currently supported options are "niggli" or "LLL". "False" does not reduce structure.

Does `"False"` actually changes the behavior? Don't have ATAT installed myself.

esoteric-ephemera: The structure reduction is actually done through PMG's `IStructure.get_reduced_structure` function, which only supports "niggli" or "LLL". Passing "False" would return a ValueError, but passing `False` (bool) would skip structure reduction. That docstr should probably be amended

janosh: will do

Issue #3503, opened 2023-12-06
Error with Locpot.from_file/Species.from_str


VicTrqt:
<b> Problem </b>
When I try to instantiate a ```pymatgen.io.vasp.outputs.Locpot``` object from the file, I get an ```IndexError``` from the ```Species.from_str``` call. I am using pymatgen v2023.11.12.
<b> Example </b>
```
Python 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0] on linux
>>> from pymatgen.io.vasp import Locpot
>>> Locpot.from_file("LOCPOT")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vtrinquet/anaconda3/envs/test_pmg/lib/python3.11/site-packages/pymatgen/io/vasp/outputs.py", line 3438, in from_file
    (poscar, data, data_aug) = VolumetricData.parse_file(filename)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vtrinquet/anaconda3/envs/test_pmg/lib/python3.11/site-packages/pymatgen/io/vasp/outputs.py", line 3281, in parse_file
    poscar = Poscar.from_str("\n".join(poscar_string))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vtrinquet/anaconda3/envs/test_pmg/lib/python3.11/site-packages/pymatgen/io/vasp/inputs.py", line 432, in from_str
    struct = Structure(
             ^^^^^^^^^^
  File "/home/vtrinquet/anaconda3/envs/test_pmg/lib/python3.11/site-packages/pymatgen/core/structure.py", line 3670, in __init__
    super().__init__(
  File "/home/vtrinquet/anaconda3/envs/test_pmg/lib/python3.11/site-packages/pymatgen/core/structure.py", line 929, in __init__
    site = PeriodicSite(
           ^^^^^^^^^^^^^
  File "/home/vtrinquet/anaconda3/envs/test_pmg/lib/python3.11/site-packages/pymatgen/core/sites.py", line 334, in __init__
    species = Composition({get_el_sp(species): 1})  # type: ignore
                           ^^^^^^^^^^^^^^^^^^
[...]

Issue #2151, opened 2021-05-21
how to handle properties of  tritium and deuterium? 


kjappelbaum:
**Is your feature request related to a problem? Please describe.**
In `cif`s (e.g., from the Cambridge Crystallographic Database) one sometimes finds symbols for common isotopes such as Tritium (`T`) or Deuterium (`D`). `pymatgen` parses them without issues, but it complains when one likes to access some properties like the atomic mass. 

**Describe alternatives you've considered**
- A "quick and dirty" solution I tried is to just add `D` to `periodic_table.json` with the `Atomic mass` and some other physical properties. It is somehow ugly because adding the `Atomic no` will lead to some issues when one wants to create an `Element.from_z` because then there is ambiguity. For this reason, one cannot simply add the `Atomic no` which might lead to unexpected issues, when users want to use this property. 
-  ASE simply replaces `Atom` with the symbol `D`. with an atom with symbol `H` and mass 2. But this behavior can sometimes be quite unexpected I feel. 
- Explicitly handle isotopes, https://github.com/pkienzle/periodictable does this. This seems like a major change. 
- One could also simply raise an error if there is a symbol `pymatgen` does not know. But this is not really pythonic and there are many functionalities for which one does not need to know any physical properties. 

**Additional context**
 I would be happy to contribute a PR to fix this but I'd like to get some feedback on the design you would use before I go for it.

shyuep:
I would actually go for option 1. I don't think it is that difficult to make it such that isotopes can be handled. The main problem, though, is that I don't have a source of compilation for all the other properties that most elements have (e.g., melting point, etc.) for isotopes.

If it is just D and T, I think it is a simple matter to just add those as additional element symbolsm, like what ASE does.

Alternatively, we can just translate D/T to H unless there is a very good reason why we want D or T to be present. 

I don't think most of the analysis we do differentiates between D or T or H (except maybe diffusion where the mass differences can be large). For most other isotopes, the properties are similar enough.

mkhorton:
I think this is a great question.

I agree that symbols `D` and `T` should be handled automatically, which would involve coding a special case somewhere.

If we did want to add isotope support (one might imagine this useful for other elements too, e.g. in the diffusion case mentioned, I've seen MD runs where specific isotopes are used for example), then one option is to add it as a supported property on `Species`. Currently, `Species` only formally supports `spin` and `oxidation_state`, but we could add `mass_number`. The string representation of this might be `Fe,A=56` for example.

[...]

Issue #662, opened 2017-05-09
DeformedStructureSet returns dict_keys object in python 3


Atraii:
## System

* Pymatgen version: **master 4.7.5**
* Python version: **3.6.1**
* OS version: **Fedora 25**

## Summary

* analysis/elasticity/strain.py: DeformedStructureSet() returns a [dict_keys](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects) object in python 3 (rather than a list in python 2)
* this is harder to understand as the default argument `symmetry=False` returns a list rather than a dictionary view object (even in python 3).
* returning different objects makes it difficult to determine how to use the object

## Example code

```python
poscar = Poscar.from_file('POSCAR_orig')
structure = poscar.structure
dss = elasticity.strain.DeformedStructureSet(structure, symmetry=True)
for i, st_deformed in enumerate(dss.def_structs):
    runvasp()
    vasprun = Vasprun('vasprun.xml')
    deformation_matrix = dss.deformations[i].tolist()
    sm = elasticity.strain.IndependentStrain(deformation_matrix)
    st = elasticity.stress.Stress(vasprun.ionic_steps[-1]['stress'])
    ss_dict[sm] = st
```

## Error message

```
Traceback (most recent call last):
  File "holy-potatoes-a-bug.py", line 28, in <module>
    deformation_matrix = dss.deformations[i].tolist()
TypeError: 'dict_keys' object does not support indexing
```

## Suggested solution (if any)

* From DeformedStructureSet(), could we get the same functionality from 
```python
 self.deformations = list(self.sym_dict)
```
as opposed to 
```python
 self.deformations = self.sym_dict.keys()
```

* Alternatively, I'd be okay with returning a dict_keys object if it also returned the same type **with symmetry=False**

montoyjh:
Let's do the first.  Thanks for finding the cause.

As an aside, I wrote DeformedStructureSet mostly to preserve some legacy functionality of an older codebase that I inherited, but I've never used it for anything practical (I mostly use atomate/MPWorks for managing calculations of ECs).  If there's any additional functionality that you think would be useful there, feel free to implement and submit a PR.

Issue #3373, opened 2023-10-04
Bug with structure read from cif file


epatyukova:
Structure with collection code 133424 from ICSD (Pr Sm Pb Si2 S8) is read incorrectly by Structure.from_file(). In particular there should be 18 Pb2+ sites, but there are 30 after the read. There is clearly a bug.

The cif file  [ICSD_CollCode133424.txt](https://github.com/materialsproject/pymatgen/files/12803589/ICSD_CollCode133424.txt)[extension changed to txt in order to add it here)

janosh:
@epatyukova You're right, there is a problem here. Fwiw, the problem is more widespread than `pymatgen`.

@esoteric-ephemera Did some investigation and it turns out `Vesta` and `ase` also parse this CIF file incorrectly. As @computron has said in the past, CIF is a troublesome format.

`ase` completely drops the `Sm`(Samarium) from the CIF file. Vesta also has the wrong stoichiometry as well though doesn't drop any elements outright.

```py
# %%
from ase.io import read as ase_read
from pymatgen.io.ase import AseAtomsAdaptor
from pymatgen.io.cif import CifParser

# %%
cif_filename = "./ICSD_CollCode133424.cif"
cif_ase = ase_read(cif_filename)
NPb_sites_ase = len([site for site in cif_ase.get_chemical_symbols() if site == "Pb"])

# %%
pmg_from_ase = AseAtomsAdaptor.get_structure(cif_ase)
pmg_prim_from_ase = pmg_from_ase.get_primitive_structure()
print(
    f"ASE atoms --> PMG: composition = {dict(pmg_from_ase.composition.as_dict())}, {pmg_from_ase.get_space_group_info()}"
)
print(
    f" --> primitive cell: composition = {dict(pmg_prim_from_ase.composition.as_dict())}, {pmg_prim_from_ase.get_space_group_info()}"
)

# %%
for i, site_tol in enumerate([1.0e-2, 1.0e-3, 1.0e-4, 1.0e-5]):
    cif_pmg = CifParser(cif_filename, site_tolerance=site_tol).get_structures()[0]
    print(f"{site_tol=}")
    rounded_comp = {
        element: round(cif_pmg.composition.as_dict()[element], 3) for element in cif_pmg.composition.as_dict()
    }
    print(f"  Composition={rounded_comp}")
    try:
        print(f"  Space group = {cif_pmg.get_space_group_info()}")
    except Exception:
        pass
[...]

PR #3542, opened 2024-01-05
Improve CIF checking, support for isotopes, and correct handling of new VASP 6.4.2 POSCAR format incl. slashes in header

esoteric-ephemera:
This PR does a few things:
- Adds a `CifAssesor` class to `pymatgen.io.cif` that checks whether the composition reported in a CIF matches that of the resultant pymatgen `Structure`. This would warn the user when elements are missing (as Hydrogens often are in experimental CIFs), or when the parsing has gone awry and the stoichiometries differ
- Adds support for isotopes, particularly Deuterium and Tritium. 
  - `pymatgen.core.periodic_table.Element` now has a few more attributes, such as the atomic mass number (number of protons and neutrons) as the `.A` attr
  - A structure containing an isotope will appear to only have the base element in it (e.g., D2O will look like H2O), and when files are written, the isotopes are converted to the element. However, the isotope names, mass, etc. are preserved in the element attr
  - This closes issue [#2151](https://github.com/materialsproject/pymatgen/issues/2151)
- As of VASP version 6.4.2, when using POTCARs with a SHA, all files like CONTCAR, CHG*, LOCPOT, AECCAR* will have the element names followed by a slash
   - This closes issue [#3503](https://github.com/materialsproject/pymatgen/issues/3503) and atomate2 issue [#662](https://github.com/materialsproject/atomate2/issues/662) 
   - Allow for correct parsing of these via changes to `pymatgen.io.vasp.inputs.Poscar`
- Tests for these additions


mkhorton:
Thanks for these nice improvements @esoteric-ephemera 

Since you’re looking at this part of the code, do you have a view of how H2O or OH sites should be handled? It’s one remaining weakness of the CIF parser I think.

mkhorton:
Also adding a note that the `mp` prefix has been registered with the IUCr, if we wanted to add support for writing out custom pymatgen fields (like site_properties). There’s definitely a way in which pymatgen could losslessly convert `Structure` to/from CIF. It’s been something on the back burner for me for a while.

esoteric-ephemera:
> Since you’re looking at this part of the code, do you have a view of how H2O or OH sites should be handled? It’s one remaining weakness of the CIF parser I think.

Are you thinking of experimental structures where the H sites are undetermined/unlisted? For organics, @Andrew-S-Rosen  had a good suggestion of trying to mimic how Olex (used by crystallographers) [adds hydrogen atoms](https://www.olexsys.org/olex2/docs/tutorials/sucrose/refining/#add-hydrogen-atoms) to structures where their positions are unknown

@kristinpersson had a suggestion that was based on a geometric construction - that would also be a cheap solution

I have a few other ideas that would use a DFT charge density, but these are more intensive options

> Also adding a note that the mp prefix has been registered with the IUCr, if we wanted to add support for writing out custom pymatgen fields (like site_properties). There’s definitely a way in which pymatgen could losslessly convert Structure to/from CIF. It’s been something on the back burner for me for a while.

That's great to know! The precision of atomic sites in CIFs seems to be a bottleneck to lossless conversion, since this can result in composition mismatches. Since a lot of CIFs are written by hand (especially ICSD ones), this happens frequently, see e.g., issue [#3373](https://github.com/materialsproject/pymatgen/issues/3373)

shyuep:
NIce work! One minor thing: does the cif assessor really need to be a class? I would think a CifParser.assess or CifParser.validate method would suffice.

mkhorton:
> Are you thinking of experimental structures where the H sites are undetermined/unlisted?

Not quite, there are two problems: (1) “implicit hydrogens” whereby the hydrogens are not listed in the atoms at all (so we might not realise hydrogens are missing unless we do a composition check as suggested), (2) the CIF files which list a molecule on a single atomic site, eg “OH” or “H2O” etc. Here, the CIF file at least tells us that a H should be present, but we still do not parse appropriately (I believe for OH, we parse as just O, and provide a warning).

For (2), it’s a difficult situation, because all options are unideal in some way. Omitting the H completely is wrong. Putting the H in an arbitrary location (eg correct bond length, perhaps do a bond valence sum or similar to find a plausible position) is tricky. Defining the site as containing a molecule is probably most correct, but does not fit with the current pymatgen object model.

esoteric-ephemera:
> NIce work! One minor thing: does the cif assessor really need to be a class? I would think a CifParser.assess or CifParser.validate method would suffice. I am heavy proponent for OOP but in this case, the CifAssessor doesn't seem to conceptually an object per se.

Thanks! Yes I'll try to refactor the assessor

mkhorton:
Sorry, sent that message too soon. I’m not familiar with the Olex method. My point anyway is just that there are cases where we have to figure out the hydrogen is missing based on composition, but also cases where we’re explicitly told a hydrogen should be there.

In terms of lossless conversion, correct me if I’m wrong but there’s no limit on the precision with which we write out atomic positions? We can write out as many significant figures as we do with JSON. I know the issue with the handwritten CIFs which is why we put in the rounding logic inside the parser (eg converting 0.333 -> 1/3 etc), but that seems like a separate issue.

shyuep:
@mkhorton @esoteric-ephemera I would just say that there are now universal potentials. For an OH, you can take the coordinates as the position of the O and then initial a H at 0.5A away. Then use a UP to relax the H positions. :-) For this very constrained use case, I think it would lead to very reasonable results (much more than pure geometry / empirical constructions).

esoteric-ephemera:
> (2) the CIF files which list a molecule on a single atomic site, eg “OH” or “H2O” etc. Here, the CIF file at least tells us that a H should be present, but we still do not parse appropriately (I believe for OH, we parse as just O, and provide a warning).

OK I see what you mean. The best solution there in keeping with pymatgen's model is probably putting O on the site position and adding in H

> My point anyway is just that there are cases where we have to figure out the hydrogen is missing based on composition, but also cases where we’re explicitly told a hydrogen should be there.
This PR warns the user in at least the latter case: if H is reported in the CIF but missing in the PMG structure, the user is warned. It **usually** seems to be the case that H's are reported in the CIF composition even when no H site positions are given. But if they're not, we might want to include a charge balance check

> I know the issue with the handwritten CIFs which is why we put in the rounding logic inside the parser (eg converting 0.333 -> 1/3 etc), but that seems like a separate issue.
We can specify arbitrary position, but a lot of the ICSD CIFs seem to have lower precision than the default rounding tolerance in `CifParser`. ICSD-191196 is a good example where the composition is only reported to two digits of precision

> I would just say that there are now universal potentials.
Might be good in this case to freeze the known positions and only relax the H atoms. My only concern here is that H is often underrepresented in ML-UIP training data. Maybe @janosh can comment more on that

shyuep:
Yes H is often underrepresented in UP training. But I think for a limited case like OH, it should be ok. Regardless, it is better than any of the rule of thumb alternatives. The only better option is DFT (with certain functionals).

esoteric-ephemera:
Just adding the WIP tag for now, since I realized that oxidation state handling for isotopes isn't working as intended (also pending refactor of CifAssessor)

janosh:
The hydrogen convex hull distance error for MLIPs is definitely elevated based on Matbench Discovery testing.
See "Per-Element Model Error Heatmaps" at https://matbench-discovery.materialsproject.org/models with M3GNet and MACE at 260 eV/atom, CHGNet at 240.

I think this not so much stems from underrepresentation (MPtrj contains 142k relaxations frames with hydrogen) but hints that the models are confused due to inconsistent training data, perhaps due to issues like the one addressed in this PR.

But I agree with @shyuep, doing H-only relaxation with any of the above mentioned models could still get you closer to true H positions.

esoteric-ephemera:
This is almost ready for review, I think. One thing I'm not satisfied with: the user can manually specify an oxidation state for an isotope, however calling convenience functions like `Composition(...).add_charges_from_oxi_state_guesses()` or `Structure(...).add_oxidation_state_by_guess()` will convert any isotope to its parent bare `Element`

For example, The composition `Composition({"D+": 2, "O2-": 1})` includes D+ in the composition. However calling `add_charges_from_oxi_state_guesses()` on  `Composition({"D": 2, "O": 1})` will return the equivalent of `Composition({"H+": 2 "O2-": 1})`.

Maybe this isn't such an issue since this represents an extreme edge case. Issuing a warning to the user when calling any of the `add_*` oxi state methods would be sufficient?

tests/io/test_cif.py
```diff
@@ -998,3 +998,68 @@ def test_bibtex(self):
 }
 """
         assert self.mcif_ncl.get_bibtex_string() == ref_bibtex_string
+
+
+class TestCifParserAssess(PymatgenTest):
```
janosh, line 1003: I think we can merge this into the `TestCifParser` class given `CifParserAssess` is no longer a standalone class.

dev_scripts/periodic_table.yaml
janosh, line 1: Not that it matters that much but could we keep the long lines in this file to minimize the diff in this PR?

janosh (changes requested):

tests/io/test_cif.py
janosh: done

dev_scripts/periodic_table.yaml
esoteric-ephemera: Yep sorry about that! I had basically used `monty.serialization.dumpfn` to overwrite the yaml file when I added entries to it, and didn't think about the formatting change

janosh: Thanks!

janosh (approved):
Thanks a lot @esoteric-ephemera! 👍 Super valuable guard rails for everyone parsing CIFs with `pymatgen`.

PR #3550, opened 2024-01-13
Allow for writing of `Structure.site_properties` as `_atom_site_` flags in `CifWriter`

### LINKED ISSUE #3548: Allow for custom `_atom_site_` fields in `CifWriter`
### Feature Requested

Right now, there is not too much flexibility with the `CifWriter` when it comes to what fields can be written out to the resulting CIF. It does have a keyword argument for writing out magnetic moments, but that's really it. It would be ideal to support a more generic option where the user can write out custom `_atom_site_<NAME>` fields based on the `Structure` object's site properties.

### Proposed Solution

The solution is I will do it myself imminently and open a PR.

Andrew-S-Rosen:
## Summary

Closes #3548 by adding a `write_stite_properties: bool = False` keyword argument to `CifWriter`.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/io/cif.py
+++ b/pymatgen/io/cif.py
@@ -121,7 +121,11 @@ class CifBlock:
         # add quotes if necessary
         if v == "":
             return '""'
-        if (" " in v or v[0] == "_") and not (v[0] == "'" and v[-1] == "'") and not (v[0] == '"' and v[-1] == '"'):
+        if (
+            (" " in v or v[0] == "_")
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
@janosh there is something _super_ bizarre happening with this test. When I run it locally, it passes ~2/3rds of the time and then randomly does not pass. Wild.

janosh (approved):
Thanks @Andrew-S-Rosen, nice addition! 👍

PR #3555, opened 2024-01-15
Support `magmoms` in `get_phonopy_structure()`

tomdemeyere:
## Summary

Phonopy needs magnetic moments to calculate the symmetry etc, currently pymatgen does not send it.

## Checklist

python -m pytest tests/io/test_phonopy.py fails with "ImportError: cannot import name 'coord_cython' from 'pymatgen.util'", any ideas?

### DIFF
--- a/pymatgen/core/structure.py
+++ b/pymatgen/core/structure.py
@@ -927,7 +927,7 @@ class IStructure(SiteCollection, MSONable):
         for idx, specie in enumerate(species):
             prop = None
             if site_properties:
-                prop = {key: val[idx] for key, val in site_properties.items()}
+                prop = {key: val[idx] for key, val in site_properties.items() if val is not None}
 
             label = labels[idx] if labels else None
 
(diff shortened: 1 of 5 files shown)


JaGeo:
Tagging @JonathanSchmidt1 as I have seen a similar fix in one of his atomate2 branches.

janosh (approved):
Thanks @tomdemeyere! 👍

JonathanSchmidt1:
@JaGeo  My solution basically was the same as here.
 Maybe sth relevant for you is that when I added magnetic moments it broke sth  in atomate2 because some of the functionalities of phonopy did not work with magnetic moments at the time.
I assume this change in the materials project will lead to the workflow accepting magnetic structures  automatically.
```   
 get_phonopy_structure(structure)
 if use_symmetrized_structure == "primitive" and kpath_scheme != "seekpath":
        primitive_matrix: list[list[float]] | str = [
            [1.0, 0.0, 0.0],
            [0.0, 1.0, 0.0],
            [0.0, 0.0, 1.0],
        ]
    else:
        primitive_matrix = "auto"
```
For me auto was not working with magmoms so I had to remove it. If this is still the case, it might be an option to give a warning in atomate2 and specifically disable auto for magnetic structures.

PR #3561, opened 2024-01-17
Remove deprecated `(to|from|as|get)_string` methods

janosh:
Deprecation period started with #3158 on Jul 15 2023 (6+ months ago).

Migration to the new API in all cases is to replace `(to|from|as|get)_string` with `(to|from|as|get)_str`.


mkhorton:
Hi @janosh, thanks for keeping on top of deprecation notices. I think we may have been a bit premature here:

* There was no notice on the [compatibility](https://pymatgen.org/compatibility.html) page.
* An updated atomate2 has not been released yet. I see your PR  [https://github.com/materialsproject/atomate2/pull/690] got merged but since there was no subsequent release, the latest atomate2 is not compatible with latest pymatgen.

I realize from a pymatgen maintainer point of view that it's not our responsibility to keep downstream codes working, but I do want to make sure we're mindful of community usage, especially for breaking changes which are not critical.

janosh:
apologies, i had forgotten about the existence of the compatibility page. both the [initial deprecation](https://github.com/materialsproject/pymatgen/releases/tag/v2023.8.10) and [the recent removal](https://github.com/materialsproject/pymatgen/releases/tag/v2024.1.26) were in the release notes of course but perhaps a bit buried.

the deprecation period of 6 months was agreed upon with @shyuep who also green-lighted making the to-string API consistent (i.e. replace mixture of `get_str` and `get_string` with `get_str` everywhere).
i suspect in `atomate2`, the deprecation warnings did not prompt any code updates until it actually broke because the warnings only show up in the logs when executing a workflow. the logs are quite verbose and so usually only inspected when a workflow fails. other community codes may have transitioned earlier and faster on account of the warnings being less buried. i suspect waiting longer would have much difference ito community transition (though I could be wrong)

i'll add an entry to the compatibility page right now

mkhorton:
Thanks @janosh! No worries, these things happen. But we should be careful, any breaking change has the potential to create a lot of work for a lot of people downstream, so we need to make sure such changes are strongly motivated (in this example, removing the deprecated functions does improve code cleanliness but given that the functions were already marked as deprecated, and their inclusion does not cause any performance issue, they could have been left around longer).

The point on logs is well taken. Ultimately, the lack of full, real integration tests has always been an atomate issue, but one that is difficult to address!

janosh:
> The point on logs is well taken. Ultimately, the lack of full, real integration tests has always been an atomate issue, but one that is difficult to address!

i was chatting with @Andrew-S-Rosen about that at some point who managed to setup very sophisticated CI which we tested in https://github.com/Quantum-Accelerators/quacc/pull/1285. it triggers integration tests on Princeton HPC that actually run VASP (even on 3rd party PRs). IIRC he suggested to run a small subset of atomate2 tests as well just to keep a tap on it. but long-term would be better to setup sth similar in `atomate2` itself

PR #3566, opened 2024-01-20
Alias `VaspInputSet` to `VaspInputGenerator` 

### LINKED ISSUE #3565: Add alias `VaspInputGenerator` for `VaspInputSet`
My decision is that I have no problems with having an alias called VaspInputGenerator. But VASPInputSet will not be deprecated. Let's be pragmatic - there are many things we probably want to rename. But deprecation of classes/methods creates headaches and unless there is an actual practical benefit, i.e., a large change in functionality or the API, these should not be done.

_Originally posted by @shyuep in https://github.com/materialsproject/pymatgen/issues/3484#issuecomment-1838877706_

janosh:
To accomodate `atomate2` terminology.

Closes #3565.

@utf @davidwaroquiers Let me know if this serves your purpose and if you think more changes along similar lines would make sense.


davidwaroquiers:
Thanks @janosh !
I would add the aliases to the subclasses too. For the base VaspInputSet, it's almost never used directly so it would not clarify things for the users. I can do that and make a PR to your vasp-input-gen-alias. I think it would be useful to add some explanations or note on this in the documentation. I will see to add this somewhere, at least in the API docstrings somewhere, maybe also somewhere in the main documentation, but I don't think there is anything (yet) about the io sets there.

janosh:
> I can do that and make a PR to your vasp-input-gen-alias. I think it would be useful to add some explanations or note on this in the documentation

please, doc improvements are always welcome!

> I would add the aliases to the subclasses too.

there are 23 subclasses that inherit from `VaspInputSet` via `DictSet`. not sure we need aliases for all of them. but only adding it for some seems inconsistent. not sure what's best

janosh:
@davidwaroquiers this PR is conflicting now so I'm leaning towards resolving and merging now. you can then make a subsequent PR straight to `master`

PR #3574, opened 2024-01-23
Fix floating point imprecision error in ordering property of CollinearMagneticStructureAnalyzer

kaueltzen:
## Summary

The ordering property of the CollinearMagneticStructureAnalyzer class is set depending on the `total_magnetization = abs(sum(magmoms))`. Due to floating point imprecision, the sum of magmoms can adapt values close to zero instead of zero. This can, for example, lead to collinear AFM structures being misclassified as FiM.

This was resolved by replacing `if total_magnetization > 0` with `if not isclose(total_magnetization, 0, abs_tol=1e-8)`.
A test to determine the magnetic ordering of MAGNDATA entry 1.62 (CuO) which would have failed with the previous code was added.

Source mcif: 
https://www.cryst.ehu.es/magndata/index.php?index=1.62
doi: 10.1088/0022-3719/21/15/023

## Open questions
@janosh 

1. Is an absolute tolerance of `abs_tol=1e-8` a reasonable choice for "being close to zero" here?
2. I included a test file (tests/files/magnetic.example.CuO.mcif) from MAGNDATA because this was also done for magnetic.example.NiO.mcif - hope this is okay?


janosh (approved):
thanks @kaueltzen, esp. for adding a test right away! 👍  this looks great.

pymatgen/analysis/magnetism/analyzer.py
```diff
@@ -481,10 +481,15 @@ def number_of_unique_magnetic_sites(self, symprec: float = 1e-3, angle_tolerance
     @property
     def ordering(self) -> Ordering:
         """Applies heuristics to return a magnetic ordering for a collinear
-        magnetic structure. Result is not guaranteed for correctness.
+        magnetic structure. Result is not guaranteed to be correct, just a best
+        guess. Tolerance for minimum total magnetization to be considered
+        ferro/ferrimagnetic is 1e-8.
```
JaGeo, line 486: I am wondering if we should make this adaptable when defining the instance the class if we are not sure about the tolerance yet.

janosh: ah, wanted to add that but forgot. PRs welcome!

JaGeo: @kaueltzen will hopefully open another PR as it was one of the questions in the PR!

PR #3575, opened 2024-01-23
`tests/io/aims` use `numpy.testing.assert_allclose` and `pytest.MonkeyPatch`

janosh:
for better error msgs and auto-reverted side-effects resp.


janosh:
failing tests pass locally. checking to see if they fail again on `master`

PR #3577, opened 2024-01-24
Added threshold_ordering parameter to CollinearMagneticStructureAnalyzer in addition to PR #3574

kaueltzen:
## Summary

A parameter threshold_ordering was added to CollinearMagneticStructureAnalyzer below which the total magnetization is treated as zero when defining the magnetic ordering.

@JaGeo @janosh `self.total_magmoms = sum(magmoms)` and `self.magnetization = sum(magmoms) / structure.volume` are defined before processing the input magmoms. They could also be affected by the error described in PR #3574 - do you think it's a good idea to add a threshold here as well?


janosh:
> They could also be affected by the error described in PR https://github.com/materialsproject/pymatgen/pull/3574 - do you think it's a good idea to add a threshold here as well?

what would change based on the threshold? setting `self.total_magmoms` and `self.magnetization` to exactly 0 when below the threshold?

could you add a test case that passes a custom `threshold_ordering` and verifies different behavior?

kaueltzen:
> > They could also be affected by the error described in PR #3574 - do you think it's a good idea to add a threshold here as well?
> 
> what would change based on the threshold? setting `self.total_magmoms` and `self.magnetization` to exactly 0 when below the threshold?

Yes - if, for example, a database is scanned for structures with zero magnetization based on `self.magnetization` the imprecision issue could be relevant. but i don't know if this important enough to be included?

JaGeo:
> > > They could also be affected by the error described in PR #3574 - do you think it's a good idea to add a threshold here as well?
> > 
> > 
> > what would change based on the threshold? setting `self.total_magmoms` and `self.magnetization` to exactly 0 when below the threshold?
> 
> Yes - if, for example, a database is scanned for structures with zero magnetization based on `self.magnetization` the imprecision issue could be relevant. but i don't know if this important enough to be included?

I would not change the data that is stored on the object. I think this would be unexpected behavior. If someone needs this, they will implement it themself

janosh (approved):
thanks @kaueltzen! 👍

Issue #3586, opened 2024-01-26
`Vasprun` parser raises a `FileNotFoundError` when searching for POTCARs with pymatgen 2024.1.26


Andrew-S-Rosen:
### Python version

3.11

### Pymatgen version

2024.1.26

### Operating system version

Ubuntu

### Current behavior

Do the following in a completed folder containing a VASP run.

```Python
from pymatgen.io.vasp.outputs import Vasprun

Vasprun('vasprun.xml')
```

```python
File /scratch/network/asrosen/software/miniconda/envs/quacc/lib/python3.10/site-packages/pymatgen/i
o/vasp/outputs.py:296, in Vasprun.__init__(self, filename, ionic_step_skip, ionic_step_offset, pars
e_dos, parse_eigen, parse_projected_eigen, parse_potcar_file, occu_tol, separate_spins, exception_o
n_bad_xml)
    293         self.nionic_steps = len(self.ionic_steps)
    295     if parse_potcar_file:
--> 296         self.update_potcar_spec(parse_potcar_file)
    297         self.update_charge_from_potcar(parse_potcar_file)
    299 if self.incar.get("ALGO") not in ["CHI", "BSE"] and not self.converged and self.parameters.
get("IBRION") != 0:
File /scratch/network/asrosen/software/miniconda/envs/quacc/lib/python3.10/site-packages/pymatgen/i
o/vasp/outputs.py:1054, in Vasprun.update_potcar_spec(self, path)
   1046 def update_potcar_spec(self, path):
   1047     """
   1048     Args:
   1049         path: Path to search for POTCARs
   (...)
   1052         Potcar spec from path.
   1053     """
-> 1054     if potcar := self.get_potcars(path):
   1055         self.potcar_spec = [
   1056             {"titel": sym, "hash": ps.md5_header_hash, "summary_stats": ps._summary_stats}
   1057             for sym in self.potcar_symbols
   1058             for ps in potcar
   1059             if ps.symbol == sym.split()[1]
   1060         ]
File /scratch/network/asrosen/software/miniconda/envs/quacc/lib/python3.10/site-packages/pymatgen/i
o/vasp/outputs.py:1010, in Vasprun.get_potcars(self, path)
   1007 else:
   1008     search_path = os.path.split(self.filename)[0] if path is True else str(path)
   1009     potcar_paths = [
-> 1010         f"{search_path}/{fn}" for fn in os.listdir(search_path) if fn.startswith("POTCAR") 
[...]

PR #3587, opened 2024-01-26
Fix `Vasprun.get_potcars` search method; tweak fake POTCARs

esoteric-ephemera:
The main change here is to correct the `update_potcar_spec` method of `Vasprun` found by @Andrew-S-Rosen in issue #3586. Specifying `Vasprun("vasprun.xml")` would not find POTCARs, but specifying the leading `Vasprun("./vasprun.xml")` would. That has been corrected.

Other fixes:
- Add test for this behavior, `io/test_outputs.TestVasprun.test_parse_potcar_cwd_relative`
- Update potcar scrambling to ensure that scrambled POTCARs have same TITELs as originals. A "FAKE" comment still exists in the scrambled POTCARs to show that they're gobbled, but this is no longer parsed into `PotcarSingle.TITEL`


pymatgen/io/vasp/inputs.py
```diff
@@ -272,6 +272,7 @@ def from_file(cls, filename, check_for_potcar=True, read_velocities=True, **kwar
                     [get_el_sp(n) for n in names]  # ensure valid names
                 except Exception:
                     names = None
+        print("shitfucker", names, dirname, filename)
```
janosh, line 275: my favorite line in this PR! 😄

janosh (approved):
thanks @esoteric-ephemera for the quick fix!

pymatgen/io/vasp/inputs.py
Andrew-S-Rosen: LOL I am losing it over here.

esoteric-ephemera: Ahhhhhh I'm so sorry! I was maybe working a bit too fast lol

PR #3591, opened 2024-01-28
Use `name` instead of `value` in `Spin.__str__`

### LINKED ISSUE #3590: String representation of enum
Should the string representations for Spin.up/Spin.down be "up" and "down" respectively?

https://github.com/materialsproject/pymatgen/blob/ac8a7e9cd93ba9f7afeaced45dc28af6d4e151d4/pymatgen/electronic_structure/core.py#L29-L30

janosh:
Closes #3590.

### DIFF
--- a/pymatgen/electronic_structure/core.py
+++ b/pymatgen/electronic_structure/core.py
@@ -29,6 +29,14 @@ class Spin(Enum):
     def __str__(self) -> str:
         return str(self.name)
 
+    @classmethod
+    def from_str(cls, val: str) -> Spin:
+        """Returns Spin from dict representation."""
+        value = {"spin.up": 1, "up": 1, "spin.down": -1, "down": -1}.get(val.lower(), val)
+        if value not in (1, -1, "1", "-1"):
+            raise ValueError(f"Invalid Spin {value=}")
+        return Spin(int(value))
+
 
 @unique
 class OrbitalType(Enum):
--- a/pymatgen/electronic_structure/core.py
+++ b/pymatgen/electronic_structure/core.py
@@ -27,7 +27,7 @@ class Spin(Enum):
         return float(self.value)
 
     def __str__(self) -> str:
-        return str(self.value)
+        return str(self.name)
 
 
 @unique
--- a/pymatgen/electronic_structure/bandstructure.py
+++ b/pymatgen/electronic_structure/bandstructure.py
@@ -598,9 +598,9 @@ class BandStructure:
         projections = {}
         structure = None
         if isinstance(next(iter(dct["bands"].values())), dict):
-            eigen_vals = {Spin(int(k)): np.array(dct["bands"][k]["data"]) for k in dct["bands"]}
+            eigen_vals = {Spin.from_str(key): np.array(dct["bands"][key]["data"]) for key in dct["bands"]}
         else:
-            eigen_vals = {Spin(int(k)): dct["bands"][k] for k in dct["bands"]}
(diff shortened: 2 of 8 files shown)


R1j1t:
Thank @janosh! I was thinking of contributing myself and was waiting for comments from maintainers if this is a change that is required. No worries, I will try to fix the failed test here and open a PR to `fix-gh-3590` :)

janosh:
thanks @R1j1t. i think this change makes sense which is why i wanted to make a PR to see how much it would break. the `Spin.from_str` method can be made backward compatible to handle both the old and new serialization ("up"/"down" vs "1"/"-1"). but of course, it's still a breaking change in that JSON documents (e.g. for DOS and band structure objects) in existing databases will become inhomogeneous. old ones will have "1"/"-1" as keys for spin projections, new ones "up"/"down". this is not a change to commit to lightly and will require community discussion.

janosh:
@shyuep what's your stance here? keep serializing `Spin.up`/`Spin.down` as 1/-1 or change to `"up"`/`"down"`? we could also choose `"Spin.up"`/`"Spin.down"` to be even more explicit

pymatgen/electronic_structure/core.py
```diff
@@ -27,7 +27,15 @@ def __float__(self) -> float:
...
     def __str__(self) -> str:
-        return str(self.value)
+        return str(self.name)
+
+    @classmethod
+    def from_str(cls, val: str) -> Spin:
+        """Returns Spin from dict representation."""
+        value = {"spin.up": 1, "up": 1, "spin.down": -1, "down": -1}.get(val.lower(), val)
+        if value not in (1, -1, "1", "-1"):
+            raise ValueError(f"Invalid Spin {value=}")
```
R1j1t, line 37: Can we change this to `__init__` and something like below:

```py
    def __init__(self, val) -> Spin:
         """Returns Spin from possible values."""
         value = {"spin.up": 1, "up": 1, "spin.down": -1, "down": -1}
         if val in {-1, 1}:
                  raise DeprecationWarning('Int initializing is deprecated, please use "up"/"down" instead')
                  return Spin(int(value))
         if val in value:
                  return Spin(int(value[val]))

         raise ValueError(f"Invalid Spin {value=}")

```

R1j1t: This might prevent it from becoming breaking change. Thus people you have `Spin(str(1))` in their codes(workflows/scripts) will correctly initialize the  `Spin` and also leave a warning for the future.

mkhorton:
> @shyuep what's your stance here? keep serializing Spin.up/Spin.down as 1/-1 or change to "up"/"down"? we could also choose "Spin.up"/"Spin.down" to be even more explicit

Responding to this: we should definitely keep the existing 1/-1. We should absolutely not make changes to how things are serialized unless absolutely necessary.

Ultimately, good arguments could be made for a different value here, but at this point it's an aesthetic choice. Changing it would not bring a clear benefit, but would introduce a lot of potential data incompatibilities or chores for downstream users, as noted:

> but of course, it's still a breaking change in that JSON documents (e.g. for DOS and band structure objects) in existing databases will become inhomogeneous

shyuep:
I would definitely not do this. It is purely aesthetic.

R1j1t:
I can't entirely agree with the argument that it is not required because it's an aesthetic change. Choices like these make the outputs more readable. However, I completely understand that it is a significant change. Hence, it will have a much lower priority.

shyuep:
It is not about required or not. It is a balance between the value of the change and the headache it will create for the community.

PR #3593, opened 2024-01-30
Add interface to icet SQS tools through SQSTransformation

esoteric-ephemera:
In a recent PR #3541, @mkhorton pointed out the existence of another SQS generation tool already natively in python, icet. This has nice features that complement the ATAT MCSQS interface already in PMG, namely generation of SQS by enumeration of all possible supercells. 

This is possible for smaller supercells, and guarantees that you get the "optimal" SQS for a given supercell size (when feasible).

This PR adds: 
- An interface to icet through `io.icet`
- Access to icet's enumeration and Monte Carlo features through `transformations.advanced_transformations.SQSTransformation`
- Tests for the icet features in `SQSTransformation`

Looking to get feedback on if people think these are sensible additions, or if they belong somewhere else (like atomate2), or if there are features that should be added.


esoteric-ephemera:
Thanks @mkhorton! Maybe nothing more to add on the pymatgen side, you might want to wait to merge until [this icet issue](https://gitlab.com/materials-modeling/icet/-/issues/614) is resolved

I've had to install from git and lightly modify icet to get it running, but it works quite well besides that

pymatgen/io/icet.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from _icet import _ClusterSpace
+    from ase import Atoms
+
+
+@requires(ClusterSpace is not None, "IcetSQS requires the icet package. Use `pip install icet`")
```
janosh, line 31: can we add a test for this error? would require `monkeypatching` the `ClusterSpace` variable in this module to `None`.

also would be good to change the error class from `Runtime` to `ImportError` once https://github.com/materialsvirtuallab/monty/pull/625 is merged

janosh (changes requested):
nice work @esoteric-ephemera! 👍

janosh:
also, we can use GH's draft PR feature instead of prefixing PR titles with `[WIP]` as it's commonly forgotten to remove once PR is ready

pymatgen/io/icet.py
esoteric-ephemera: I've added a test for the import but removed the `@requires` decorator, which I've found to be troublesome:
- It makes top-level functions no-longer top-level / non-pickle-able, which removes multiprocessing capability
- It is unaffected by monkeypatching. The following code, which should work with the `@requires` decorator, does not:
```
from pymatgen.io import icet as icet_interface
with pytest.MonkeyPatch.context() as monkeypatch:
    monkeypatch.setattr(icet_interface,"ClusterSpace",None)
    with pytest.raises(RuntimeError):
        icet_interface.IcetSQS(
        structure=self.stored_run["disordered_structure"],
        scaling=self.scaling,
        instances=None,
        cluster_cutoffs={2: 5.0}
    )
```
I consistently get `Failed: DID NOT RAISE <class 'RuntimeError'>`

janosh: thanks @esoteric-ephemera! 

> which removes multiprocessing capability

interesting. @shyuep is this a known limitation? otherwise i'll open a monty issue to fix or document this.

either way thanks for fixing tests @esoteric-ephemera.


> you might want to wait to merge until [this icet issue](https://gitlab.com/materials-modeling/icet/-/issues/614) is resolved

is the draft status on this PR to wait for the above? otherwise lgtm

janosh (approved):

PR #3598, opened 2024-02-01
Rename `master` branch to `main`

janosh:


shyuep:
I have said no to this before and I will not say it again. Do not merge this.

mkhorton:
Agree with @janosh here, regardless of views on other factors, consistency has its benefits. 2 to 1?

PR #3601, opened 2024-02-04
Minor update to avoid deprecation warning

kavanase:
This is just a _tiny_ update to avoid the newly-added deprecation warnings for `get_vasp_input()`.
`get_vasp_input()` is still used in `DictSet.write_input()`, but should be changed to `get_input_set()` in line with the deprection warning (as implemented here).

PS: The recent updates to the VASP sets code is very nice @utf!


janosh (approved):
thanks @kavanase!

PR #3602, opened 2024-02-05
VASP IO `copy()` methods

janosh:
as pointed out by @esoteric-ephemera, calling `.copy()` on VASP IO classes like `Incar` and `VaspInput` returns a dict rather than a new instance of the class as most users would expect. this PR adds explicit `copy()` methods to minimize user surprise.

@esoteric-ephemera that mixin snippet i sent doesn't work as the `__init__` signatures of IO classes are too different. for the record, we considered defining

```py
class CopyableMixin:
    def copy(self):
        return type(self)(self)
```

and making all Vasp IO classes inherit from it


esoteric-ephemera:
Thanks @janosh! Question for @shyuep: @janosh and I were discussing if it makes more sense to define the `self.copy` method for `MSONable` - do you think it makes more sense to define the `copy` method for specific classes that inherit from `MSONable`, or for `MSONable` itself?

shyuep:
Depends on how general the copy method is. If it is generic and uses the MSONAble capabilities to do the copy, then it makes sense.

shyuep:
Just an addendum that when I say "makes sense", it means it makes sense for any MSONable class, not just VASP io related stuff.

PR #3613, opened 2024-02-09
Fix URL joining in OptimadeRester

rdamaral:
## Summary

Major changes:

- fix 1: 
  - Fix URL joining in OptimadeRester. The original os.path.join function is not the correct function to use for joining URLs. It's designed for joining file paths, and its behavior depends on the operating system. On Windows, it uses a backslash (\) as the separator, while on Unix-based systems, it uses a forward slash (/). 
  - This is a small fix, so no tests were added to 'test_optimade.py'
  - This fix should resolve the issue reported at: https://matsci.org/t/could-not-retrieve-information-from-optimade/39469

## Todos

None

## Checklist

- [N/A] Google format doc strings added. Check with `ruff`.
- [N/A] Type annotations included. Check with `mypy`.
- [N/A] Tests added for new features/fixes.
- [N/A] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/pymatgen/ext/optimade.py
+++ b/pymatgen/ext/optimade.py
@@ -5,9 +5,7 @@ from __future__ import annotations
 import logging
 import sys
 from collections import namedtuple
-from urllib.parse import urljoin
-from os.path import join
-from urllib.parse import urlparse
+from urllib.parse import urljoin, urlparse
 
 import requests
 from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


janosh (approved):
thanks a lot @rdamaral for submitting a fix!

PR #3615, opened 2024-02-09
Add missing `MPSCANRelaxSet.yaml` parameters and alphabetize

Andrew-S-Rosen:
Tagging @esoteric-ephemera for review, @utf for context, and @shyuep & @mkhorton as code owners of the MP sets.

This PR ensures that the `MPSCANRelaxSet.yaml` in Pymatgen matches the [`BaseMPR2SCANRelaxSet.yaml`](https://github.com/materialsproject/atomate2/blob/main/src/atomate2/vasp/sets/BaseMPR2SCANRelaxSet.yaml) used in Atomate2 (and thereby, in MP production calculations). Namely, the `MPSCANRelaxSet.yaml` in pymatgen was missing the "sensible defaults" added by @esoteric-ephemera, which includes: `ISMEAR: 0`, `KSPACING: 0.22`, `SIGMA: 0.05`. Everything else is just a reordering for alphabetization purposes.


Andrew-S-Rosen:
@esoteric-ephemera --- I see the thumbs up. Does that mean this is good to go, or are you still planning to review this?

mkhorton:
Looks good to me. I think `ISMEAR` was probably omitted since this was intended to be set dynamically based on gap. I have no objection to include a default value.

I believe there was also some talk about switching `LREAL` to False, but this might require a separate, updated input set, since it would technically break compatibility with the published input set.

esoteric-ephemera:
@Andrew-S-Rosen thanks for putting this together! @mkhorton, yeah for the new input sets, putting them in a separate PR would be good. I'll reach out to the foundation soonish with that

Andrew-S-Rosen:
Great, thank you both! I also agree that some minor modifications should be made but that will be elsewhere.

@shyuep, I think we have the necessary blessings for this one.

Andrew-S-Rosen:
@janosh should I update the hash?

janosh:
yes please

janosh (approved):

PR #3619, opened 2024-02-13
Adds support for an `MSONAtoms` class that's an `MSONable` form of an ASE `Atoms` object

Andrew-S-Rosen:
## Summary

Here, I introduce the `MSONAtoms` class, which is simply an `MSONable` version of the ASE `Atoms` object, introducing an `.as_dict()` and `from_dict()` method that uses ASE's JSON encode/decoder to do the serialization. By default, the object returned from `ASEAtomsAdaptor.get_atoms()` is now an `MSONableAtoms` object. The original `Atoms` class can be returned via the `msonable=False` keyword argument, although this should rarely be needed in practice since the functionality of `MSONAtoms` is otherwise identical to `Atoms`. This is a follow-up to https://github.com/materialsvirtuallab/monty/pull/619.

## Checklist


pymatgen/io/ase.py
```diff
@@ -38,18 +40,41 @@
 __date__ = "Mar 8, 2012"
 
 
+class PMGAtoms(Atoms, MSONable):
```
janosh, line 43: one thing to consider is how this affects any future plans we might hatch to develop a 1st class `Atoms` object in `pymatgen`. there were times where i thought a single class to handle periodic and non-periodic systems would be nice. just to keep that option open, maybe it would be better to reserve the name `PmgAtoms` (even if we later just call it `Atoms` or similar) to avoid future confusion and name the new class in this PR `AseAtoms` instead?

Andrew-S-Rosen: Personally, I think `AseAtoms` is somewhat misleading, as it is not the ASE `Atoms` object. I'm open to other suggestions though.

Sidenote: We might want to consider `MSONableAtoms` as another option to make the purpose clearer, although that doesn't directly address your concern. Also, `PmgAtoms` or `PMGAtoms`?

janosh: i don't know if i'd consider it misleading given it's in every way identical except for some added serialization behavior. after all, according to Python 

```py
isinstance(pymatgen.io.ase.AseAtoms(), ase.atoms.Atoms)
>>> True
```

that said, i like `MSONableAtoms` even better as, like you said, it communicates intent.

i find pascal case `PmgAtoms` more readable than `PMGAtoms` but i'm sure that's debatable 😄

Andrew-S-Rosen:
> MSONAbleAtoms is a mouthful.How about just call it AtomsM?

My personal opinion is that descriptive is better than concise. It's not immediately clear what an `AtomsM` means at a glance. Having it be unambiguous is particularly helpful with type hinting, among other things.

That said, I agree it's somewhat long. I'm open to other alternatives.

JaGeo:
I would also vote for MSONAbleAtoms or something descriptive.

shyuep:
Then MSONAtoms?

Andrew-S-Rosen:
Works for me. I'll make the change when I'm at my laptop.

mkhorton (approved):
Looks good to me.

I do have a question, perhaps @Andrew-S-Rosen knows, is lossless conversion to/from `Structure` and `Atoms` possible? Or are some attributes potentially lost?

Andrew-S-Rosen:
@mkhorton Good question. I have spent a fair bit of time on this module to ensure that it is as lossless as possible, and there are back-and-forth tests for this as well. If you spot an edge case, please let me know and I will patch it!

janosh:
> Or are some attributes potentially lost?

@mkhorton there's no remaining known loss of information during `Atoms`<->`Structure` round trips. many edge cases were reported in the past which have all been addressed, most recently in

https://github.com/materialsproject/pymatgen/pull/3359
https://github.com/materialsproject/pymatgen/pull/3338
https://github.com/materialsproject/pymatgen/pull/3270
https://github.com/materialsproject/pymatgen/pull/3151

unknown remaining edge cases are possible but likely rare

shyuep:
Thanks @Andrew-S-Rosen . This is great!

shyuep:
I just pushed a change that ensures MSONable is returned by from_dict. This is supposed to be the case. Otherwise it loses functionality with deserializaation.

pymatgen/io/ase.py
```diff
@@ -38,18 +40,41 @@
 __date__ = "Mar 8, 2012"
 
 
+class MSONAtoms(Atoms, MSONable):
```
ml-evs, line 43: Something here breaks imports in cases where the user doesn't have ASE installed. This seems to be breaking the latest `mp_api` as a result:

```python
from mp_api.client.mprester import MPRester
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/mp_api/client/__init__.py", line 8, in <module>
    from .mprester import MPRester
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/mp_api/client/mprester.py", line 10, in <module>
    from emmet.core.electronic_structure import BSPathType
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/emmet/core/electronic_structure.py", line 11, in <module>
    from pymatgen.analysis.magnetism.analyzer import (
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/pymatgen/analysis/magnetism/__init__.py", line 5, in <module>
    from pymatgen.analysis.magnetism.analyzer import (
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/pymatgen/analysis/magnetism/analyzer.py", line 24, in <module>
    from pymatgen.transformations.advanced_transformations import MagOrderingTransformation, MagOrderParameterConstraint
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/pymatgen/transformations/advanced_transformations.py", line 33, in <module>
  File "/home/mevans/repos/re2fractive/re2fractive/experiments/2_featurizing_refractive_index_data/.venv-mp-api/lib/python3.11/site-packages/pymatgen/io/ase.py", line 44, in <module>
    class MSONAtoms(Atoms, MSONable):
                    ^^^^^
NameError: name 'Atoms' is not defined
```

Will raise an issue

ml-evs: see #3644

PR #3621, opened 2024-02-13
Improve INCAR tag check

DanielYang59:
## Summary

Some legacy from #3539.

- Clean up INCAR parameters recording file `incar_parameters.json` to separate types and values.
- Clean up INCAR checking mechanism accordingly.

## TODOs

Enable checking of more complex type annotations like `LATTICE_CONSTRAINTS : Sequence[bool, bool, bool])` suggested by @esoteric-ephemera in https://github.com/materialsproject/pymatgen/pull/3539#discussion_r1471615138.


pymatgen/io/vasp/inputs.py
```diff
@@ -50,12 +50,6 @@
 logger = logging.getLogger(__name__)
 module_dir = os.path.dirname(os.path.abspath(__file__))
 
-# hashes computed from the full POTCAR file contents by pymatgen (not 1st-party VASP hashes)
-PYMATGEN_POTCAR_HASHES = loadfn(f"{module_dir}/vasp_potcar_pymatgen_hashes.json")
-# written to some newer POTCARs by VASP
-VASP_POTCAR_HASHES = loadfn(f"{module_dir}/vasp_potcar_file_hashes.json")
-POTCAR_STATS_PATH = os.path.join(module_dir, "potcar-summary-stats.json.bz2")
```
DanielYang59, line 57: I didn't really see a good reason to define these at such a "global" position when they are exclusively used by `PotcarSingle` and in the definition of `_gen_potcar_summary_stats` function. Meanwhile the `_gen_potcar_summary_stats` seems to be called only inside the unit test. Maybe `_gen_potcar_summary_stats` should be moved to the unit test as well? Is there any underlying thoughts I didn't know about @esoteric-ephemera? Thanks!

esoteric-ephemera: I'm pretty sure that loading them here rather than every time a `PotcarSingle` is instantiated reduces load time. These were formerly loaded in at global spec, but that no longer is the case (not sure when that happened).

The `Potcar` class basically concatenates multiple `PotcarSingle` instances, so the way that these are loaded now might increase load time substantially. Probably smarter to load it in `Potcar` as well and feed that to `PotcarSingle`. Does that sound reasonable @Janosh?

`_gen_potcar_summary_stats` is/was used to generate the file potcar-summary-stats.json.bz2 and `PotcarSingle._potcar_summary_stats` attr. It could go into `dev_tools`, but we decided to keep it where it is for discoverability (#3370)

DanielYang59: Thanks a lot for the input @esoteric-ephemera 👍 . I'm not sure about the impact of above path definitions on the instantiation overhead of `Potcar` and `PotcarSingle`, but I would give it a test and discuss with you on the necessity of repositioning them then (in a separate PR because I think it might be better to make PRs more narrow on specific topic).

DanielYang59:
I did some research and found [typeguard](https://pypi.org/project/typeguard/) seems to be what we're looking for. 
For example:
```python
from typeguard import check_type, CollectionCheckStrategy


check_type(
    value=[1, "hi"], expected_type=list[int],
    collection_check_strategy=CollectionCheckStrategy.ALL_ITEMS
)

check_type(
    value=[1, "hi"], expected_type=list[int, int],
    collection_check_strategy=CollectionCheckStrategy.ALL_ITEMS
)

check_type(
    value=[1, "hi"], expected_type=eval("list[int, int]"),
    collection_check_strategy=CollectionCheckStrategy.ALL_ITEMS
)
```

However there are still some concerns to address:
1. Is it essentially necessary to add an extra dependency `typeguard` to `pymatgen`
2. I currently don't know any better approach to cast string to object (`"list"` to `list`) than `eval`, but in our case the source is internal (from file `incar_parameters.json`), so it should not lead to any securify issues
3. type check each item might incur a performance penalty (https://github.com/agronholm/typeguard/issues/418#issuecomment-1817515059), thus we would need to discuss the necessity of this and have a switch for strict mode (check every item or just check the first item, where check first item of a collection is the default for `typeguard`).

DanielYang59:
Can you give me some advice on this @janosh? Do you think it's necessary to have an item-wise type check for INCAR tags (and might need to include an extra dependency `typeguard`)? Or should we stick to a more general type check (`list` instead of `list[int, int, int] for instance`). Thanks for any input in advance.

janosh:
> Do you think it's necessary to have an item-wise type check for INCAR tags (and might need to include an extra dependency `typeguard`)?

i think adding `typeguard` as a dependency just for this is a tall ask. i'm guessing @shyuep will be strongly against.
unless there's a strong selling point i would also lean against

DanielYang59:
> i think adding `typeguard` as a dependency just for this is a tall ask. i'm guessing @shyuep will be strongly against. unless there's a strong selling point i would also lean against

Yes I was thinking the same, adding a dependency to be used only once would be too much overhead. Actually this should be questions:
- Do you think it would be necessary to perform item-wise check for INCAR tags (or just keep current more general type checking)? @janosh I'm pretty unsure and worried about the potential performance penalty here.
- If so, including an extra dependency should certainly be discouraged. I would see if I could find another way around.

janosh:
> Do you think it would be necessary to perform item-wise check for INCAR tags

i don't think so. i don't remember a lot of GitHub issues about this.

pymatgen/io/vasp/inputs.py
```diff
@@ -2570,6 +2571,10 @@ def set_symbols(
             self.extend(PotcarSingle.from_symbol_and_functional(el, functional) for el in symbols)
 
 
+class UnknownPotcarWarning(UserWarning):
+    """Warning raised when POTCAR hashes do not pass validation."""
```
DanielYang59, line 2575: Relocated Warning class after its corresponding class for consistency.

DanielYang59:
> i don't think so. i don't remember a lot of GitHub issues about this.

Let's keep the type checking unchanged then, and implement more careful checking when necessary. This PR should be good for review now @janosh . Thanks a lot!

DanielYang59:
Guess this PR get forgotten... Can you please review this @janosh ? Thanks.

janosh (approved):
thanks @DanielYang59! lgtm :shipit:

DanielYang59:
Thanks for reviewing @janosh

PR #3625, opened 2024-02-14
Fixed Incar object to allow for ML_MODE vasp tag

davidwaroquiers:
Fixed Incar object to allow for ML_MODE vasp tag which does not want to have capitalized values (train, run, ...). By default, pymatgen capitalizes the values of the INCAR tags (e.g. "ALGO = Fast", even if you set incar["ALGO"] = "fast"). This is not working for the ML_MODE tag.
Also fixed (i.e. ignored specific mypy code) random mypy errors in files that were not touched ...

## Summary

Major changes:

- Allow the use of ML_MODE tag in INCAR (not capitalizing, lowering instead)


janosh (approved):
i reverted the `mypy` ignores to see if they actually show up in our CI cause I couldn't repro locally. maybe check your mypy version and/or reinstall commit hooks

PR #3627, opened 2024-02-14
Lobster io improvements

naik-aakash:
# Todo

Make MSONable

- [x] Charge
- [x] Sitepotential
- [x] Grosspop
- [x] Bandoverlap 
- [x] MadelungEngeries 
- [x] Icohplist
- [x] Update tests


naik-aakash:
Hi @JaGeo , I made this changes as per our prep to update our schema in atomate2.

JaGeo:
Thank you!

@naik-aakash , could you add tests for from_dict as well? And, doesn't monty already implement as_dict. Why do you need to implement this yourself?

JaGeo:
Just to add @naik-aakash and I will discuss this and then update the PR.

davidwaroquiers:
Hello,


If I may, here are some possibly useful comments.


I would avoid having the filename in the init, but only have the data itself and have a from_file method that reads in.
Example for Charge object:

```
class Charge(MSONable):
    def __init__(self, atomlist, types, mulliken, loewdin):
        self.atomlist = atomlist
        self.types= types
        self.mulliken= mulliken
        self.loewdin= loewdin

    @classmethod
    def from_file(cls, filename: str = "CHARGE.lobster"):
        with zopen(filename, mode="rt") as file:
            data = file.read().split("\n")[3:-3]
        if len(data) == 0:
            raise OSError("CHARGES file contains no data.")

        atomlist: list[str] = []
        types: list[str] = []
        mulliken: list[float] = []
        loewdin: list[float] = []
        for line in data:
            sp = line.split()
            atomlist.append(sp[1] + sp[0])
            types.append(sp[1])
            mulliken.append(float(sp[2]))
            loewdin.append(float(sp[3]))
        return cls(atomlist=atomlist, types=types, mulliken=mulliken, loewdin=loewdin)

    @property
    def num_atoms(self)
        return len(self.atomlist)
```

Such an organization of the object is usually cleaner than init with filenames (which are actually not exactly compatible with MSONable, which is probably why @naik-aakash needed to reimplement as_dict). For standard MSONable to work, the init method should only set the attributes and do nothing else no "post-processing" of the arguments passed to init ideally.

Then indeed you don't need to implement as_dict and from_dict in principle.

Also I would avoid having capitalized attributes (herabove mulliken instead of Mulliken), otherwise you may think it's a class.

Same applies to the other objects of course. I don't know if it's possible or easy to change for you on the packages/scripts using these but I would say it would improve the lobster io if it is possible.

Best

JaGeo:
Thanks, @davidwaroquiers !

JaGeo:
Some first thoughts: 

It would be good to keep the breaking changes minimal as these classes are used by other people. Thus, I am not sure if implementing a from_file method at this point is a good idea.

Vasp output io implements the classes similarily. And I think I took them as an example when I initially started with this development, e.g., https://github.com/materialsproject/pymatgen/blob/098b2a9663d889f766458a4c7283c3d776e3913a/pymatgen/io/vasp/outputs.py#L143

In any case, if we implement an as_dict method, the from_dict one needs to work.

JaGeo:
My answer sounds a bit too direct. Sorry for this. I
n any case, the suggestions are very helpful but we need to keep a balance between correcting all poor design choices from some years ago, consistency with other code parts and few breaking changes.

davidwaroquiers:
No offense taken @JaGeo ;)
I fully agree there is a trade-of to take into account when making backward incompatible changes.
Nevertheless, I also tend to compare with other (much more used) packages, such as numpy, matplotlib, pandas, pydantic and many others, which did (recently) huge changes with big backward incompatible changes. While I agree it was (and still is for some of these changes) annoying to adapt the codes, scripts and the like, I believe these changes made sense as they were all adding value: cleaner api, more standardized usage, removal of weird (or wrong) methods/features, easier development of new features, more modularity, ...
Anyway, I'm currently not a user of lobster (nor a user or developer of lobster_io) and you should decide what's best for the lobster community.
As for the vasp io, I believe it is the same trade-off, with a larger user base. The choices made 10-15 years ago for these seemed good at the time and now we are living with these choices, which will probably not change unless a big backward incompatible refactoring happens all at once (not in any current plan of pymatgen I believe).

Best,

naik-aakash:
Thank you @davidwaroquiers , for your suggestions. They were certainly insightful and would be nice to keep in mind when implementing new stuff here on.

naik-aakash:
Hi @JaGeo , I have adapted Charge class now, without introducing breaking change am able to get as_dict and from_dict working out of the box without having to define it myself. If you think this is fine, will add a similar approach to other classes as well.

pymatgen/io/lobster/outputs.py
```diff
@@ -709,26 +733,11 @@ def get_structure_with_charges(self, structure_filename):
             Structure Object with Mulliken and Loewdin charges as site properties.
         """
         struct = Structure.from_file(structure_filename)
-        Mulliken = self.Mulliken
-        Loewdin = self.Loewdin
-        site_properties = {"Mulliken Charges": Mulliken, "Loewdin Charges": Loewdin}
+        mulliken = self.mulliken
```
JaGeo, line 736: Can you please keep the old names for a while and deprecate them with a warning? see https://dev.to/hckjck/python-deprecation-2mof

naik-aakash: Done.

tests/io/lobster/test_inputs.py
```diff
@@ -770,8 +771,8 @@ def test_attributes(self):
         atomlist = ["O1", "Mn2"]
         types = ["O", "Mn"]
         num_atoms = 2
-        assert charge_Mulliken == self.charge2.Mulliken
-        assert charge_Loewdin == self.charge2.Loewdin
+        assert charge_Mulliken == self.charge2.mulliken
```
JaGeo, line 774: Could you keep the old tests for now as well?

tests/io/lobster/test_inputs.py
```diff
@@ -815,6 +816,14 @@ def test_get_structure_with_charges(self):
         s2 = Structure.from_dict(structure_dict2)
         assert s2 == self.charge2.get_structure_with_charges(f"{TEST_FILES_DIR}/POSCAR.MnO")
 
+    def test_as_dict(self):
```
JaGeo, line 819: Please add a test for from_dict

tests/io/lobster/test_inputs.py
naik-aakash: Sure

tests/io/lobster/test_inputs.py
naik-aakash: Sure

tests/io/lobster/test_inputs.py
naik-aakash: Done

tests/io/lobster/test_inputs.py
naik-aakash: Done

janosh (commented):
@JaGeo i'm happy for you to make merge decisions! no one more qualified to review this code than you! 😄 

don't want to speak for @shyuep and @mkhorton but i imagine they too much appreciate your help!

janosh:
fwiw, i had the same thought as @davidwaroquiers and was planning to make similar comments but they've already been discussed so feel free to merge

JaGeo:
@janosh , great thanks. 

Is there any io class that you would recommend to model other io classes after? Might be helpful for future implementations. I think we have very different strategies depending on the io module at the moment.

janosh:
good question! maybe the recently added FHI-aims IO modules? e.g.

https://github.com/materialsproject/pymatgen/blob/0f474f67bd4f5a2cfc6abcc5e44c921308669c72/pymatgen/io/aims/outputs.py#L30-L50

the `io.res` module also has pretty good code quality. e.g. `ResProvider` and `AirssProvider`

https://github.com/materialsproject/pymatgen/blob/0f474f67bd4f5a2cfc6abcc5e44c921308669c72/pymatgen/io/res.py#L323-L353

`io.vasp` uses `InputFile` and similar classes feel a bit "boilerplatey" so i wouldn't always model after it even though it's widely used

https://github.com/materialsproject/pymatgen/blob/0f474f67bd4f5a2cfc6abcc5e44c921308669c72/pymatgen/io/core.py#L48

Issue #3626, opened 2024-02-14
CifParser check() method does not account for different site multiplicities


kaueltzen:
### Python version

3.10.0

### Pymatgen version

2024.2.8 or master

### Current behavior

The check() method of the CifParser class performs checks on the compositions reported in both cif and Structure object parsed from cif. Besides other things, it is checked if the relative stochiometry from cif and structure are the same.  

If the cif file is missing both the keys `_chemical_formula_sum` and `_chemical_formula_structural`, `cif_formula` is constructed from the atom types of all crystallographically unique sites in the structure:

https://github.com/materialsproject/pymatgen/blob/098b2a9663d889f766458a4c7283c3d776e3913a/pymatgen/io/cif.py#L1345

However, if the site multiplicities differ within the structure, the formula of the Structure object will have a different relative stochiometry. This leads to a `UserWarning: Incorrect stoichiometry` even if the cif file is not faulty.

### Expected Behavior

No `UserWarning` to be raised for the below mentioned example.

In my opinion, checking the relative stochiometry should be skipped if both `_chemical_formula_sum` and `_chemical_formula_structural` are not present in the cif. To obtain the true stochiometry (including multiplicities) from the cif without these keys, its symmetry operations would need to be applied as it is also done for constructing the Structure object - therefore, this would no longer be a sanity check (?).

### Minimal example

```Python
from pymatgen.io.cif import CifParser
import warnings

warnings.filterwarnings("error")

struct = CifParser("path-to-example-cif").parse_structures(primitive=True)[0]
```

### Relevant files to reproduce this bug

The bug can be for example reproduced with the mcif file of MAGNDATA entry 0.296:
https://www.cryst.ehu.es/magndata/index.php?this_label=0.296

This structure contains three crystallographically unique Cu sites, out of which two each have a site multiplicity of 2 and one has a multiplicity of 4. The site multiplicity of all unique anion sites is also 4. This results in a warning:
`UserWarning: Incorrect stoichiometry:
  CIF={'Cu': 3.0, 'Cl': 1.0, 'O': 3.0}
  PMG={'Cu': 8.0, 'Cl': 4.0, 'O': 12.0}
  ratios={'Cl': 4.0, 'O': 4.0, 'Cu': 2.6666666666666665}`

kaueltzen:
Mentioned this in PR #3628

PR #3628, opened 2024-02-16
Modified CifParser.check() as one possible solution for issue #3626

kaueltzen:
## Summary

As mentioned in issue #3626 , `CifParser.check()` raises a `UserWarning: Incorrect stoichiometry` for a cif, if
- the keys `_chemical_formula_sum` and `_chemical_formula_structural` are missing and
- crystallographic sites have different multiplicities

For the relative stoichiometry to be checked correctly, the site multiplicities would have to be included by applying the cif's symmetry operations to the crystallographically unique sites (like for construction for the structure).

As this would be no longer a sanity check, i propose skipping the relative stoichiometry check in this case (but still including all the other checks for valid elements etc.).

For the case of skipping the relative stoichiometry check, i appended a message to `self.warnings`, but I'm not sure if there are better ways to handle this case.


kaueltzen:
If you are okay with this approach i could adapt the 2 cif tests that are failing now (and for example add a formula key so the checks are not skipped).

pymatgen/io/cif.py
```diff
@@ -1370,17 +1374,20 @@ def check(self, structure: Structure) -> str | None:
...
+                    for elt_a in orig_comp_elts
+                    for elt_b in orig_comp_elts
+                )
 
-            if not same_stoich:
-                failure_reason = f"Incorrect stoichiometry:\n  CIF={orig_comp}\n  PMG={struct_comp}\n  {ratios=}"
+                if not same_stoich:
+                    failure_reason = f"Incorrect stoichiometry:\n  CIF={orig_comp}\n  PMG={struct_comp}\n  {ratios=}"
+            else:
+                self.warnings.append("Skipping relative stoichiometry check because CIF does not contain formula keys.")
```
janosh, line 1390: can we add a test for this warning?

esoteric-ephemera:
Thanks @kaueltzen  for pointing this out! I like your solution of warning that the stoichiometry check couldn't be accomplished - the fallback I added originally clearly isn't robust enough in general

Adding a test for this (using the magnetic cif you sent with the issue) would be great to move this forward

kaueltzen:
Thank you for your quick replies! I added a test both for skipping the relative stoichiometry check and appending a warning in the case of missing formula keys. There was already a test file in pymatgen, Li10GeP2S12.cif, that matched the criteria (it has no formula keys and sites with different multiplicities and occupancies).

janosh (approved):
very nice work! thanks a lot @kaueltzen 👍

PR #3630, opened 2024-02-17
Make the POTCAR setup instructions clearer

### LINKED ISSUE #3629: POTCAR setup instructions should be made clearer
### Python version

3.10+

### Pymatgen version

2024.2.8

### Operating system version

_No response_

### Current behavior

Currently, when Pymatgen searches for a pseudopotential, it has a default set of paths it looks in. For instance, if we look for the PBE_54 pseudos and symbol Al, Pymatgen will search as follows:

```
OSError: You do not have the right POTCAR with functional='PBE_54' and symbol='Al' in your 
PMG_VASP_PSP_DIR='/home/rosen/software/vasp/pseudos'. Paths tried: 
['/home/rosen/software/vasp/pseudos/POT_GGA_PAW_PBE_54/POTCAR.Al', 
'/home/rosen/software/vasp/pseudos/POT_GGA_PAW_PBE_54/Al/POTCAR']
```

However, for quite a while now, VASP has been using `potpaw_LDA.54` and `potpaw_PBE.54` as the folder names. We should really add these as defaults so users don't have to deal with modifying this. The same folder structure is used in the v64 pseudos as well.

To be clear, this is the expected path:

```
VASP_PSP_DIR/potpaw_PBE.54/Al/POTCAR
```

### Expected Behavior

Pymatgen should register pseudos out-of-the-box.

### Minimal example

```python
from pymatgen.io.vasp.inputs import Potcar
Potcar(["Al"],functional="PBE_54")
```

_No response_

Andrew-S-Rosen:
Closes #3629. I have always found the POTCAR setup instructions to be incredibly confusing, in part because some parts were never defined (e.g. `<MY_PSP>`) and in other parts because things are quite out-of-sync with the POTCAR structure offered by VASP as of the last several years. This is an attempt to make it clearer.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -191,7 +191,7 @@ In practice, this entire process might look something like the following:
 ```bash
 pmg config -p /path/to/pseudos/potcar_PBE.54/ /path/to/pseudos/pmg_potcars/
 pmg config -p /path/to/pseudos/potcar_LDA.54/ /path/to/pseudos/pmg_potcars/
-pmg config --add PMG_VASP_PSP_DIR pmg_potcars
+pmg config --add PMG_VASP_PSP_DIR /path/to/pseudos/pmg_potcars/pmg_potcars
 ```
 
 If desired, you may specify a default version and type of pseudopotentials as follows:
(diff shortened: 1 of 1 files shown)


janosh (approved):
nice! thanks for improving these docs 👍

PR #3639, opened 2024-02-21
[BugFix] Subclass Construction Locpot<:VolumetricData 

jmmshn:
## Subclass Construction Locpot<:VolumetricData

The constructor signature for `Locpot` is different from the parent.
This makes it hard to write general code to handle different kinds of `VolumetricData` objects.

Just adding a `**kwarg` there should work.


janosh (changes requested):
at the risk of sounding pedantic, can we add a test for this?

jmmshn:
No problem!  The test I added passes with the new code but not the old code.

janosh (approved):
thanks @jmmshn!

PR #3642, opened 2024-02-22
Remove properties from abivars dict as this breaks the interface with…

gmatteo:
Hi @janosh , is it possible to merge this fix and make a new release of Pymatgen 
as this bug breaks the interface with Abinit? Thanks


janosh (approved):
sorry for the breakage @gmatteo and thanks for fixing. 🙏 

not sure i fully understand this comment and why this prevents structure properties but since I'm not an abinit user and maybe shouldn't have messed with this code in the first place i'll take your word for it 😄 

> Properties are not supported if fmt="abivars" as its not a serialization protocol

janosh:
we can make a new release once this and #3645 is merged.

PR #3645, opened 2024-02-22
Guard `MSONAtoms` definition behind ASE package availability

ml-evs:
## Summary

Hopefully closes #3644. Not sure if this is the approach you want to take.

## Todo

It would be nice to add a test or linter rule that would have caught the initial issue.

PR #3646, opened 2024-02-22
Add a CI run that only tests non-optional deps (optimistically)

ml-evs:
Hopefully all tests have been written in such a way that they will be skipped if the underlying required package is not found... this would hopefully catch issues like #3645 ahead of time (although better linting would do the same too and avoid the potential need to update tests to skip if packages are missing -- only the implementation code would need to include package guard).

This should be updated after #3645 is merged to see what's left: not sure if adding a whole additional test run is "worth it" for the maintainers here but it definitely shows up the issues presented and would also be robust to any weirdness with optional deps changing the values of expected tests. There might be some kind of pytest plugin that lets you re-run only the skipped tests in the previous step when considering all deps.


janosh:
wow, that's way more possibly unbound variables than expected... 😅

ml-evs:
@coderabbitai ignore

janosh:
my impression is coderabbitai is too verbose. if it has useful suggestions, they are buried in a lot of boilerplate language that you don't want to read

DanielYang59:
Here I come! Giving you a hand on all those `pyright` fixes. I wish you don't mind :)

ml-evs:
Just to be clear I was never planning on making these fixes, I just wanted to show the issue by adding it to the CI. So feel free to close/take over/supersede this PR

DanielYang59:
Thanks for the input. I could help fix them and I could either push my fixes to your branch or maybe open a separate PR  :) ? Which do you prefer?

ml-evs:
Probably best to close this one and start again, the CI changes I suggested here are already out of date it seems. pyright was just being used to look for symptoms of the problem, the real change that needs to be made is running the test suite with optional deps only and making sure nothing fails.

DanielYang59:
Thanks for clarifying. I would redirect the PR directly to upstream if you don't mind (wasn't planning to pre-empt you)?

ml-evs:
I won't be doing anything further on this so please go ahead!

DanielYang59:
Thanks!

PR #3653, opened 2024-02-24
Clean up test files: VASP outputs

DanielYang59:
## Clean up test files: VASP outputs

This is the first step of cleaning up test files: VASP outputs, relocate VASP output files to a dedicated dir (`vasp/outputs`) under `tests` dir to tidy up the `files` dir. 

With the aim to clean up test files so that future developers could easily find and reuse existing test files whenever possible.

## TODOs

- [x] Relocate `vasprun.xml` c68d5c3d15791d148b4e6e6f2e09b6900c5249b7 and 2c8eba79da95f040ec2baa6c915a4ad7987e7f07
- [x] Relocate `OUTCAR` ee6df808d69a0bcd9694c45535c268eb0bed5511
- [x] Relocate `CHGCAR` 37868cedcfddc8491ff117c669722d3decbbe126
- [x] Relocate `DOSCAR` 649152a9cd2f24fe834dc321a4f5825a69e68f4b
- [x] Relocate `EIGENVAL` 1bba1bf3cfe5abdfbff7f9370b3de2b0a90630db
- [x] Relocate `PROCAR` 099c2058257c4d42746a7a73f3ae92f24414a07d
- [x] Relocate `WAVECAR` 428c326bc515cbbab4d8279e016048f1dd2d72a3
- [x] Relocate `XDATCAR` 0044e4984a71e0b3f1408d83ee2fef803a277327 and bd07afd6a4b26f6918bdd0c515864a45a0369e0f
- [x] Relocate `LOCPOT` 897fa39ee37ae596fc865e374275ef97527b335d
- [x] Relocate `CONTCAR` cb9fd86985140b7dd1039d9fe1c97f881ebe181e
- [x] Relocate other VASP output files (`DYNMAT` 8c691924455d042f0684d7496226ef2814b32123, `ELFCAR` 71964ad5eb4f436b51dd53135a9fcf6b9e1cd728, `IBZKPT` eb33d4082305fd76c0d56202d2b0ebdcef22826d, `WAVEDER` 002225f9ab4f584e77dc6b51591a7d5e780aa26b, `OSZICAR` b2d9040289f3eaf070d13c7c646e1b9d615b7766). 
- [x] Remove duplicates a7301ba0df2cb3b364de0cdd5050eb0dfff83666, 82c37296eb8ce00b429634b4cc62f5e8131d28ac
- [x] Unify test file namings a1c25f7e35abc23aa5dd5a0b0b7c13a84a5c6dc0, c4d7e6af84385b2fb5bdddc6caa751daa9aeb731, 0fa962d58324dcd847877d1d83a5f42d41236e2e, f2e0526e169a97029a714f9606d511b6bb16262e

## Future follow up PRs

- Clean up VASP input files
- Group and clean up dedicated VASP dirs, for example `tests/files/relaxation`, `tests/files/kpoints_opt` and more.


DanielYang59:
Guess I would need to improve the in-place decompression behaviour of `bader_caller` in this PR, otherwise tests cannot pass (as test file `CHGCAR.Fe3O4` would be decompressed and following tests dependending on it would fail with `FileNotFoundError`):
https://github.com/materialsproject/pymatgen/pull/3656#issuecomment-1963538547

It could be bypassed though by altering the test order but doesn't resolve this problem. So better just fix it.

janosh:
> Do you have any comments so far before I proceed? @janosh

sorry, i must have missed this!

> In this case, maybe either vasprun_something.xml.gz or vasprun.something.xml.gz? Which one would you suggest? 

i prefer `vasprun_something.xml.gz`

> Meanwhile my relocations seem to break tests for borg, as the outputs and inputs for VASP are now not inside the same dir:

i think the way to go about this is copy all input files to `self.tmp_path` first and then pass that tmp dir to `borg`. let me know if that doesn't work

DanielYang59:
> sorry, i must have missed this!

No worries at all and no rush.

> > In this case, maybe either vasprun_something.xml.gz or vasprun.something.xml.gz? Which one would you suggest?
> 
> i prefer `vasprun_something.xml.gz`

Good news! As I'm leaning towards the same and have already made the changes in c4d7e6af84385b2fb5bdddc6caa751daa9aeb731, 0fa962d58324dcd847877d1d83a5f42d41236e2e, f2e0526e169a97029a714f9606d511b6bb16262e


> i think the way to go about this is copy all input files to `self.tmp_path` first and then pass that tmp dir to `borg`. let me know if that doesn't work

That makes sense (with some file IO overhead).

janosh:
> That makes sense (with some file IO overhead).

might be enough actually to just symlink the files instead of making actual copies if the decompression leaves the original file in place

janosh:
i noticed that as well. not sure why but also haven't investigated this at all yet. my suspicion was the windows runner just use slower hardware but maybe that's wrong and it's something in our test setup.

either way it makes sense to occasionally the `.pytest-durations.json` file that determines how the workload is optimally split across concurrent test runners. and as the code changes, that file needs to be updated once in a while

DanielYang59:
> my suspicion was the windows runner just use slower hardware but maybe that's wrong and it's something in our test setup.

Not sure (don't have much knowledge about Github Actions...). But I guess it has something to do with the runners themselves as I notice there're others reporting [this](https://github.com/actions/runner-images/issues/7320) as well. 

> either way it makes sense to occasionally update the `.pytest-durations.json` file that determines how the workload is optimally split across concurrent test runners so they all take roughly equally long. 

Well I don't think this would help much though... As currently Windows runners seem consistently slower than Ubuntu, so even we balance the runtime of both sides, the final run time would still be determined (and slowed) by Windows 😿

DanielYang59:
Leave a quick note for myself:

I'm working on fixing the tests for `borg` but find the situation is quite tricky... 

I never use `borg` before nor read the docs. But there seems to be some hard-coded parts for example:
https://github.com/materialsproject/pymatgen/blob/442d740eb7794d360cb35f187e271b3debb7f047/pymatgen/apps/borg/hive.py#L104-L125

Which seems to be explained in https://github.com/materialsproject/pymatgen/blob/442d740eb7794d360cb35f187e271b3debb7f047/pymatgen/apps/borg/hive.py#L70-L75

I guess I would need to symlink a dedicated test dir for it, because I noticed changing the file names may very likely to change which dir it would `assimilate` and yield the test quite fragile.

Also in the test itself the target dir for `assimilate` is a bit implicit:https://github.com/materialsproject/pymatgen/blob/442d740eb7794d360cb35f187e271b3debb7f047/tests/apps/borg/test_hive.py#L27-L38

But I would find out the target later.

DanielYang59:
Please review @janosh.

I have to keep the test file name for `borg` as `vasprun.xml.xe.gz` instead of the expected `vasprun.xe.xml.gz` because it is hardcoded to collect info from vasprun files of pattern `vasprun.xml*`:
https://github.com/materialsproject/pymatgen/blob/b96d9e147e21489b377166f8839de8fab47e3e69/pymatgen/apps/borg/hive.py#L104-L133

I wouldn't touch the source code because I don't have proper knowledge of `borg`.

Noticed some tests are failing because `pydantic` and `matgl` are missing. Not sure about the reason though.

janosh:
> Well I don't think this would help much though... As currently Windows runners seem consistently slower than Ubuntu, so even we balance the runtime of both sides, the final run time would still be determined (and slowed) by Windows 😿

that's certainly true

> I have to keep the test file name for `borg` as `vasprun.xml.xe.gz` instead of the expected `vasprun.xe.xml.gz` because it is hardcoded to collect info from vasprun files of pattern `vasprun.xml*`:

sure, makes sense

> Noticed some tests are failing because `pydantic` and `matgl` are missing. Not sure about the reason though.

not sure either `pydantic` should not be needed at all and `matgl` should be installed in CI. i can try to debug tomorrow

https://github.com/materialsproject/pymatgen/blob/b96d9e147e21489b377166f8839de8fab47e3e69/setup.py#L81

DanielYang59:
> I have to keep the test file name for `borg` as `vasprun.xml.xe.gz` instead of the expected `vasprun.xe.xml.gz` because it is hardcoded to collect info from vasprun files of pattern `vasprun.xml*`:

Well I came up with a solution: create a symlink with `ScratchDir`, now all files follow the same pattern 😃 .

janosh (approved):
perfect! thank you so much @DanielYang59! 👍 this is a huge PR and hugely appreciated! you've been a big help cleaning up the pymatgen code base!

i can take a look at the unrelated CI errors in a separate PR

DanielYang59:
Thanks for reviewing @janosh! (I was about to redefine a `VASP_OUT_DIR` after cleaning up all vasp files, but glad you helped me out, thanks a ton 😃). I would proceed to other clean up once I have time and glad I could help.

PR #3655, opened 2024-02-24
Lobsterout update doc-string to match renamed class variable

naik-aakash:


naik-aakash:
Hi @JaGeo,@janosh  this change in doc-string got leftout with small changes done before merging. So would be great if this could be merged as well 😄

JaGeo:
@janosh I merged it as it was a tiny doc change.

Issue #3652, opened 2024-02-24
[Dev] Tests for `Bader` not running properly


DanielYang59:
### Summary
Noticed two `Bader` related issues:
- GitHub test workflow doesn't download Bader properly
- Tests for Bader fails

### Test workflow not running properly
While trying to clean up test files for VASP, I noticed tests for `Bader` in (tests/command_line/test_bader_caller.py) (apparently depends on VASP output files) in the GitHub workflow doesn't seem to run properly (though the tests seem to pass, until one checks the details):

```bash
Run wget http://theory.cm.utexas.edu/henkelman/code/bader/download/bader_lnx_64.tar.gz
--2024-02-23 16:46:06--  http://theory.cm.utexas.edu/henkelman/code/bader/download/bader_lnx_64.tar.gz
Resolving theory.cm.utexas.edu (theory.cm.utexas.edu)... 146.6.145.114
Connecting to theory.cm.utexas.edu (theory.cm.utexas.edu)|146.6.145.114|:80... connected.
HTTP request sent, awaiting response... 403 Forbidden
2024-02-23 16:46:06 ERROR 403: Forbidden.

Error: Process completed with exit code 8.
```

Seen the screenshot:
![test_fail](https://github.com/materialsproject/pymatgen/assets/80093591/bac6ba31-1e98-4ef1-9be7-081b37332cb9)

### Test for `Bader` fails
If running on a local machine with Bader properly installed, I then get:
```python
__________________ TestBaderAnalysis.test_atom_parsing ___________________

self = <tests.command_line.test_bader_caller.TestBaderAnalysis testMethod=test_atom_parsing>

    def test_atom_parsing(self):
        # test with reference file
        analysis = BaderAnalysis(
            chgcar_filename=f"{TEST_FILES_DIR}/CHGCAR.Fe3O4",
            potcar_filename=f"{TEST_FILES_DIR}/POTCAR.Fe3O4",
            chgref_filename=f"{TEST_FILES_DIR}/CHGCAR.Fe3O4_ref",
            parse_atomic_densities=True,
        )
    
        assert len(analysis.atomic_densities) == len(analysis.chgcar.structure)
    
>       assert np.sum(analysis.chgcar.data["total"]) == approx(
            np.sum([dct["data"] for dct in analysis.atomic_densities])
        )

/home/yang/Developer/pymatgen/tests/command_line/test_bader_caller.py:129: 
[...]

PR #3656, opened 2024-02-26
Deprecate `_parse_atomic_densities` in `BaderAnalysis` and fix `Bader` test setup

DanielYang59:
### Summary

- Clean up `__init__` of `BaderAnalysis`
- Mark `parse_atomic_densities` in `BaderAnalysis` as deprecated, to be removed on 26 Feb 2025, as discussed in #3652.

### DIFF
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -194,9 +194,11 @@ class BaderAnalysis:
                 self._parse_atomic_densities()
 
     @deprecated(
-        message="parse_atomic_densities was deprecated on 2024-02-26 "
-        "and will be removed on 2025-02-26. See"
-        " https://github.com/materialsproject/pymatgen/issues/3652 for reason."
+        message=(
+            "parse_atomic_densities was deprecated on 2024-02-26 "
+            "and will be removed on 2025-02-26. See https://"
+            "github.com/materialsproject/pymatgen/issues/3652 for details."
+        )
     )
     def _parse_atomic_densities(
         self,
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -201,9 +201,7 @@ class BaderAnalysis:
             "github.com/materialsproject/pymatgen/issues/3652 for details."
         )
     )
-    def _parse_atomic_densities(
-        self,
-    ):
+    def _parse_atomic_densities(self):
         # Deprecation tracker
         if datetime(2025, 2, 26) < datetime.now() and "CI" in os.environ:
             raise RuntimeError("This method should have been removed, see #3656.")
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -164,7 +164,7 @@ class BaderAnalysis:
                 self.version = -1  # Unknown
             if self.version < 1.0:
                 warnings.warn(
-                    "Your installed version of Bader is outdated. Calculation of vacuum charge may be incorrect.",
+                    "Your installed version of Bader is outdated, calculation of vacuum charge may be incorrect.",
                     UserWarning,
                 )
 
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import os
 import subprocess
 import warnings
+from datetime import datetime
 from glob import glob
 from shutil import which
 from tempfile import TemporaryDirectory
@@ -203,6 +204,10 @@ class BaderAnalysis:
     def _parse_atomic_densities(
         self,
     ):
+        # Deprecation tracker
+        if datetime(2025, 2, 26) < datetime.now() and "CI" in os.environ:
+            raise RuntimeError("This method should have been removed, see #3656.")
+
         def slice_from_center(data, x_width, y_width, z_width):
             """Slices a central window from the data array."""
             x, y, z = data.shape
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -193,7 +193,7 @@ class BaderAnalysis:
             if self.parse_atomic_densities:
                 self._parse_atomic_densities()
 
-    @deprecated("parse_atomic_densities deprecated and would be removed on 26Feb2024.")
+    @deprecated(message="parse_atomic_densities deprecated and would be removed on 26Feb2024.")
     def _parse_atomic_densities(
         self,
     ):
--- a/pymatgen/command_line/bader_caller.py
+++ b/pymatgen/command_line/bader_caller.py
@@ -193,7 +193,11 @@ class BaderAnalysis:
             if self.parse_atomic_densities:
                 self._parse_atomic_densities()
 
-    @deprecated(message="parse_atomic_densities deprecated and would be removed on 26Feb2024.")
+    @deprecated(
+        message="parse_atomic_densities was deprecated on 2024-02-26 "
+        "and will be removed on 2025-02-26. See"
+        " https://github.com/materialsproject/pymatgen/issues/3652 for reason."
+    )
     def _parse_atomic_densities(
         self,
     ):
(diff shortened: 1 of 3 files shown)


pymatgen/command_line/bader_caller.py
```diff
@@ -179,49 +191,63 @@ def __init__(
...
+        def slice_from_center(data, x_width, y_width, z_width):
+            """Slices a central window from the data array."""
+            x, y, z = data.shape
+            start_x = x // 2 - (x_width // 2)
+            start_y = y // 2 - (y_width // 2)
+            start_z = z // 2 - (z_width // 2)
+            return data[
+                start_x : start_x + x_width,
+                start_y : start_y + y_width,
+                start_z : start_z + z_width,
```
DanielYang59, line 209: For some reason unknown to me, `ruff-format` seems against removal of `whitespace before ':'`? I'm a little confused.

DanielYang59:
Meanwhile this PR only fixes second half of #3652.

First half of the GitHub workflow run still needs to be fixed (`wget` failed with `403 Forbidden` error) @janosh .

DanielYang59:
Test setup for `Bader` seems to have been fixed using `https` instead of `http`. Then every test workflow would get the following error:
```python
_____________________ TestBaderAnalysis.test_atom_parsing ______________________

self = <tests.command_line.test_bader_caller.TestBaderAnalysis testMethod=test_atom_parsing>

    def test_atom_parsing(self):
        # test with reference file
        analysis = BaderAnalysis(
            chgcar_filename=f"{TEST_FILES_DIR}/CHGCAR.Fe3O4",
            potcar_filename=f"{TEST_FILES_DIR}/POTCAR.Fe3O4",
            chgref_filename=f"{TEST_FILES_DIR}/CHGCAR.Fe3O4_ref",
            parse_atomic_densities=True,
        )
    
        assert len(analysis.atomic_densities) == len(analysis.chgcar.structure)
    
>       assert np.sum(analysis.chgcar.data["total"]) == approx(
            np.sum([dct["data"] for dct in analysis.atomic_densities])
        )
```

Maybe I still should zero-pad them in the test (hopefully fix the issue)? @janosh  (Or just remove/supress this test if no one is using it anyway?)

janosh:
> Maybe I still should zero-pad them in the test

yes, we should fix the test instead of removing it while the functionality is still available

pymatgen/command_line/bader_caller.py
```diff
@@ -179,49 +191,63 @@ def __init__(
...
-                    dct = {
-                        "data": find_encompassing_vol(shifted_data),
-                        "shift": shift,
-                        "dim": self.chgcar.dim,
-                    }
-                    atomic_densities.append(dct)
-                self.atomic_densities = atomic_densities
+                self._parse_atomic_densities()
+
+    @deprecated(message="parse_atomic_densities deprecated and would be removed on 26Feb2024.")
```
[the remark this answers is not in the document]
DanielYang59: Ah... Apologies...~~Time machine out of tune~~

tests/command_line/test_bader_caller.py
```diff
@@ -127,7 +127,7 @@ def test_atom_parsing(self):
         assert len(analysis.atomic_densities) == len(analysis.chgcar.structure)
 
         assert np.sum(analysis.chgcar.data["total"]) == approx(
-            np.sum([dct["data"] for dct in analysis.atomic_densities])
+            np.sum([np.sum(dct["data"]) for dct in analysis.atomic_densities])
```
DanielYang59, line 130: I misunderstood the implementation a little. The implementation seems to calculate the total num instead of element-wise sum:
https://github.com/materialsproject/pymatgen/blob/dc60d35a21fd7b55d82c65bb078afc4a2c0e9bbf/pymatgen/command_line/bader_caller.py#L209-L216

The `np.sum` must have been removed by accident in ff45527caf5de3bc76ed859bab0d978cab38972d, adding it back is what we need to fix this.

DanielYang59:
Should be good for now, please review @janosh .

Though everything _seems to be working_, there is still a minor issue here: As `Bader` itself cannot work directly on compressed files like `.gz`, `BaderAnalysis` would need to decompress corresponding files during runtime:
https://github.com/materialsproject/pymatgen/blob/dc60d35a21fd7b55d82c65bb078afc4a2c0e9bbf/pymatgen/command_line/bader_caller.py#L99-L131

This leads to some issues:
- `file.gz` would be converted to `file` after `BaderAnalysis` initialization.
- Run unit test with `test_bader_caller` would decompress test files too (As we're working towards compressing test files).

Meanwhile I think it might be better not to alter files in place, both during runtime and test time. As such I think it might make sense to alter the behaviour of `BaderAnalysis` a bit to create a temp dir during run time, make a copy (or decompress file) there, perform the analysis and remove the temp dir. This might lead to some performance overhead owing to file operations (But hopefully won't impact users as we don't expect them to have data compressed). But I would pin myself here to do some optimizations anyway.

janosh:
>  decompress the test file in to a duplicate (and remove it afterwards). I would pin myself here to do this after cleaning up test files.

that sounds good! by "after cleaning up test files" do you mean in or after #3653?

DanielYang59:
> by "after cleaning up test files" do you mean in or after #3653?

I think after #3653, still to make each PR on a dedicated topic 😃

pymatgen/command_line/bader_caller.py
```diff
@@ -179,49 +191,69 @@ def __init__(
...
+            "parse_atomic_densities was deprecated on 2024-02-26 "
+            "and will be removed on 2025-02-26. See https://"
+            "github.com/materialsproject/pymatgen/issues/3652 for details."
+        )
+    )
+    def _parse_atomic_densities(
+        self,
+    ):
+        def slice_from_center(data, x_width, y_width, z_width):
+            """Slices a central window from the data array."""
```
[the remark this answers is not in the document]
DanielYang59: Sure. We should do this. (But I didn't see any `reminder` stuff in GitHub for such scheduled tasks?)

What about I go ahead and add a `deadline` arg to the `deprecated` method to `monty`?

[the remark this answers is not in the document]
janosh: sounds good. i would make it a bit more explicit and call the monty keyword:

```py
raise_in_ci_after: tuple[datetime, str]  # removal date and error message
```

e.g.

```py
raise_in_ci_after=(datetime(2025, 2, 26), "remove this method and its tests")
```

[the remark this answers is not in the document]
DanielYang59: Thanks for the suggestion. I would make a PR there soon.

janosh (approved):
great work @DanielYang59! love all your contributions 👍

pymatgen/command_line/bader_caller.py
DanielYang59: What about this? @janosh It seems `flake8` and `ruff-format` go against each other...

DanielYang59:
Thanks for reviewing and those helpful suggestions @janosh

pymatgen/command_line/bader_caller.py
janosh: we don't actually use `flake8` so `ruff-format` determines formatting

DanielYang59: Okay thanks... Sorry didn't notice that.

pymatgen/command_line/bader_caller.py
```diff
@@ -179,49 +192,71 @@ def __init__(
...
+                start_y : start_y + y_width,
+                start_z : start_z + z_width,
+            ]
+
+        def find_encompassing_vol(data: np.ndarray):
+            """Find the central encompassing volume which
+            holds all the data within a precision.
+            """
+            total = np.sum(data)
+            for idx in range(np.max(data.shape)):
```
DanielYang59, line 226: Another note for my side: 

- This method `bader_caller` is overall slow, or the bottleneck could be called `bader` itself, and might need some profiling to identify the bottleneck,

- Try replacing linear search with bisection method to see if performance could be improved. However it's also important to note this should search for the **MINIMUM** encompassing volume, pure bisection might not be suitable.

PR #3660, opened 2024-02-28
Avoid `bader_caller` from altering compressed file in place

DanielYang59:
### Summary: Avoid bader_caller from altering compressed file in place

- Prevent `bader_caller` from altering compressed test in place, see https://github.com/materialsproject/pymatgen/pull/3656#issuecomment-1963538547
- Removed `bader_exe_path` to encourage users to add `Bader` executable to PATH for simplicity
- Other format clean ups

### Details

Added a utility function `temp_decompress` to copy compressed file and decompress the copied file in `ScratchDir`. If the file is not compressed, would just pass to avoid unnecessary file OPs.


tests/command_line/test_bader_caller.py
```diff
@@ -74,7 +74,7 @@ def test_from_path(self):
         chgref_path = f"{direct_dir}/_CHGCAR_sum.gz"
 
         analysis = BaderAnalysis(chgcar_filename=chgcar_path, chgref_filename=chgref_path)
-        analysis_from_path = BaderAnalysis.from_path(from_path_dir)
+        analysis_from_path = analysis.from_path(from_path_dir)
```
DanielYang59, line 77: Wondering how this was passing before... I'm pretty confused 😕 .

pymatgen/command_line/bader_caller.py
```diff
@@ -80,33 +80,31 @@ def __init__(
...
-                f"You cannot parse a cube and a CHGCAR at the same time. \n{cube_filename=}\n{chgcar_filename=}"
-            )
+            raise ValueError("Cannot parse cube and CHGCAR at the same time.")
+
+        # Parse atomic partitioned charge density
         self.parse_atomic_densities = parse_atomic_densities
 
         with ScratchDir("."):
             if chgcar_filename:
-                self.is_vasp = True
```
DanielYang59, line 108: Didn't see this being used.

pymatgen/command_line/bader_caller.py
```diff
@@ -485,7 +482,8 @@ def _get_filepath(filename: str, msg: str = "") -> str | None:
         return paths[0]
 
     chgcar_path = _get_filepath("CHGCAR", "Could not find CHGCAR!")
-    chgcar = Chgcar.from_file(chgcar_path)
+    if chgcar_path is not None:
+        chgcar = Chgcar.from_file(chgcar_path)
```
DanielYang59, line 486: Fixed `mypy` error

pymatgen/command_line/bader_caller.py
```diff
@@ -57,21 +54,19 @@ class BaderAnalysis:
...
     """
 
     def __init__(
         self,
         chgcar_filename: str = "",
         potcar_filename: str = "",
         chgref_filename: str = "",
-        parse_atomic_densities: bool = False,
         cube_filename: str = "",
-        bader_exe_path: str | None = BADER_EXE,
```
DanielYang59, line 74: I hope I'm not breaking anything here by removing `bader_exe_path`. As I just noticed @janosh you just added this in #3191.

While I agree having an extra `bader_exe_path` does give user the flexiblity to specify the path to `Bader` but I'm "guessing" this is not the standard approach (maybe)? I think having executables in the `PATH` is the more general approach.

I must have been a minimalist so would rather keep number of var minimal unless necessary. Let me know if you think otherwise and I'm happy to revert this deletion.

tests/command_line/test_bader_caller.py
janosh: `from_path` is a `classmethod`. i.e. it's not using any instance attributes so your change shouldn't make any difference. i think the code was clearer before though that the arguments of `BaderAnalysis(chgcar_filename=chgcar_path, chgref_filename=chgref_path)` were not used so i would revert

https://github.com/materialsproject/pymatgen/blob/d871fdfacbb4b08f200bdb64e76b0aaefccaf690/pymatgen/command_line/bader_caller.py#L398-L400

DanielYang59: Make sense. Sorry I didn't look into the `from_path` careful enough. I guess I changed this because I got an error somehow and cannot recreate it now. Would revert this. Thanks.

DanielYang59:
Please review @janosh. Thanks.

Maybe relocate the utility function to a more general accessible location (as I notice there are other tests seem to alter compressed file in place as well, cannot remember which one at the moment, would tag them in the future once I find any)?

Tests for `test_cod` seems to fail but this should not be related to this PR, `ERROR 2003 (HY000): Can't connect to MySQL server on 'www.crystallography.net:3306' (10060)`

pymatgen/command_line/bader_caller.py
```diff
@@ -42,11 +44,8 @@
 __date__ = "4/5/13"
 
 
-BADER_EXE = which("bader") or which("bader.exe")
```
DanielYang59, line 45: `monty.dev.requires` is cleaner but would break all `mypy` tests, as decorator seems too implicit to be understood properly by `mypy` at this moment

pymatgen/command_line/bader_caller.py
janosh: i added `bader_exe_path` keyword in preparation for a PR to `atomate2` which would have allowed setting the path to bader in the [`atomate2` config file](https://github.com/materialsproject/atomate2/blob/2f5ec957723ca23a13dc2bed52056e06a93a83c1/src/atomate2/settings.py#L83) though that 2nd bit stalled and i think i never finished it. still hoping to get to it at some point so would prefer to keep this `bader_exe_path`

DanielYang59: Sure thanks for the explanation. I would revert this removal.

pymatgen/command_line/bader_caller.py
```diff
@@ -80,44 +78,66 @@ def __init__(
...
             cube_filename (str, optional): The filename of the cube file.
-            bader_exe_path (str, optional): The path to the bader executable.
+            bader_path (str, optional): The path to the bader executable.
+            parse_atomic_densities (bool, optional): Enable atomic partition of the
+                charge density. Charge densities are atom centered. Defaults to False.
         """
-        bader_exe = which(bader_exe_path or "")
-        if bader_exe is None:
+
+        def temp_decompress(file: str | Path, target_dir: str = ".") -> str:
```
janosh, line 87: is this guaranteed to make `temp` files? where are the decompressed files deleted?

DanielYang59: Hi thanks for asking and reviewing. This `temp_decompress` works more as an internal utility to work with `Scratch` (from monty.tempfile.Scratch) as described in its docstring

https://github.com/DanielYang59/pymatgen/blob/b3a989a37fd7e4e727410536a1a1f96077f1d940/pymatgen/command_line/bader_caller.py#L131-L161

As such the decompressed files should be created inside the `Scratch` dir and would be deleted once the context manager exists.

But I have a feeling this could be improved so let me know if you have any suggestion. Thanks! 😃

pymatgen/command_line/bader_caller.py
```diff
@@ -100,14 +98,14 @@ def temp_decompress(file: str | Path, target_dir: str = ".") -> str:
             """
             file = Path(file)
 
-            if file.suffix.lower() in [".bz2", ".gz", ".z"]:
+            if file.suffix.lower() in (".bz2", ".gz", ".z"):
```
DanielYang59, line 101: My oversight, should use set instead `{".bz2", ".gz", ".z"}`

pymatgen/command_line/bader_caller.py
```diff
@@ -531,8 +527,9 @@ def _get_filepath(filename: str, msg: str = "") -> str | None:
         return paths[0]
 
     chgcar_path = _get_filepath("CHGCAR", "Could not find CHGCAR!")
-    if chgcar_path is not None:
-        chgcar = Chgcar.from_file(chgcar_path)
+    if chgcar_path is None:
+        raise FileNotFoundError("Could not find CHGCAR!")
```
DanielYang59, line 531: The `_get_filepath` method already issued a warning in line 529, maybe this is not necessary?

pymatgen/command_line/bader_caller.py
```diff
@@ -150,20 +148,19 @@ def temp_decompress(file: str | Path, target_dir: str = ".") -> str:
...
             else:
-                fpath = cube_fpath = temp_decompress(cube_filename)
+                filepath = cube_fpath = temp_decompress(cube_filename)
 
                 self.cube = VolumetricData.from_cube(cube_fpath)
                 self.structure = self.cube.structure
                 self.nelects = []
 
-            # Compile CHGCAR reference file
+            # prepare CHGCAR file
```
DanielYang59, line 159: Maybe `Prepare CHGCAR reference file`? As there is both `CHGCAR` with `chgcar` and reference CHGCAR with `chgref`

pymatgen/command_line/bader_caller.py
```diff
@@ -57,21 +56,20 @@ class BaderAnalysis:
...
     """
 
     def __init__(
         self,
         chgcar_filename: str = "",
         potcar_filename: str = "",
         chgref_filename: str = "",
-        parse_atomic_densities: bool = False,
         cube_filename: str = "",
-        bader_exe_path: str | None = BADER_EXE,
```
DanielYang59, line 74: Also I switched to a single `bader_path` throughout this caller to avoid confusion. As there were three similar variables: `bader_exe_path`, `bader_exe` and `BADER_EXE`

pymatgen/command_line/bader_caller.py
janosh: i guess `Chgcar.from_file(chgcar_path)` will already raise if the file is missing?

pymatgen/command_line/bader_caller.py
janosh: sounds good

pymatgen/command_line/bader_caller.py
janosh: me too. this code feels quite messy so i'm glad you're having a go at cleaning it up 👍

pymatgen/command_line/bader_caller.py
DanielYang59: Maybe no... By `Chgcar.from_file`
https://github.com/materialsproject/pymatgen/blob/d07164fd27b955010ad411033d262ef3d935d09c/pymatgen/io/vasp/outputs.py#L3661-L3672

And then `VolumetricData.parse_file`:
https://github.com/materialsproject/pymatgen/blob/d07164fd27b955010ad411033d262ef3d935d09c/pymatgen/io/vasp/outputs.py#L3431-L3535

Then`zopen`: https://github.com/materialsvirtuallab/monty/blob/4d60fd4745f840354e7b3f48346eaf3cd68f2b35/monty/io.py#L19-L45

But certainly `open` would complain (maybe not so descriptive). But `_get_filepath` would issue a warning anyway, so maybe don't bother issure another here 😄

DanielYang59: By the way, I added the `if chgcar_path not None` check to avoid `mypy` error, because `_get_filepath` could return `None`

pymatgen/command_line/bader_caller.py
DanielYang59: > this code feels quite messy so i'm glad you're having a go at cleaning it up 👍

I actually wanted to say my `temp_decompress` could be improved 😄... But well... `BaderAnalysis` does have quite some code crammed inside `__init__` so cleaning it up a bit makes it easier to read. Would see if I could further make some improvement soon. Glad I could help.

janosh (approved):
thanks @DanielYang59! 👍

DanielYang59:
Thanks for viewing @janosh . Glad I could help.

pymatgen/command_line/bader_caller.py
DanielYang59: Add a quick note for myself, `Sourcery` would recommend using `set` over `list` or `tuple` for membership check, see [here](https://docs.sourcery.ai/Coding-Assistant/Reference/Rules-and-In-Line-Suggestions/Python/Default-Rules/collection-into-set/). 

The advantage of `set/tuple` over `list` seems quite obvious, for example discussed [here](https://stackoverflow.com/questions/25368337/tuple-or-list-when-using-in-in-an-if-clause)

However I still don't quite know the difference between `tuple` and `set` in this scenario, [some](https://stackoverflow.com/questions/45017774/python-set-vs-tuple-lookup-is-lookup-in-tuple-o1) claims it's related to performance.

PR #3662, opened 2024-02-29
Remove BadPoscarWarning when POSCAR elements set by POTCAR

esoteric-ephemera:
A `BadPoscarWarning` was recently added to `pymatgen.io.vasp.inputs.Poscar` when the elements in the POSCAR are identified by the order of elements in the POTCAR file. Reading in the element names from POTCAR is valid VASP input, and it's [even permitted to omit the species names in POSCAR](https://www.vasp.at/wiki/index.php/POSCAR) because VASP reads them from POTCAR:
```This line lists the species of the present ions. The given order should match the order of species appearing in the POTCAR file. This line is optional, if omitted the species names are taken from the POTCAR file.```

This PR just removes the warning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.2.1
+    rev: v0.3.0
     hooks:
       - id: ruff
         args: [--fix, --unsafe-fixes]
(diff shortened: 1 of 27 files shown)


janosh (approved):
thanks @esoteric-ephemera, nicely ruffled! 😄

DanielYang59:
Just came across this PR randomly... I guess I added this warning in #3539. I just had another look and I think this PR is right and I misunderstood the implementation. This https://github.com/materialsproject/pymatgen/blob/d07164fd27b955010ad411033d262ef3d935d09c/pymatgen/io/vasp/inputs.py#L259-L267
is nested under `check_for_potcar`. (I thought `from_file` is falling to element check from `POTCAR` because `POSCAR` is badly formatted, which seems not true). I'm sorry for this oversight and thanks for fixing this @esoteric-ephemera .

janosh:
> I'm sorry for this oversight

no worries. we live and learn :)

PR #3675, opened 2024-03-07
Migrate CI dependency installation from `pip` to `uv`

janosh:
hoping the migration to `uv`, besides being faster than `pip`, can also fix the CI import errors for `matgl` and `pydantic` that started 2 days ago. not optimistic about `pydantic` which shouldn't be imported by pymatgen or any of its deps anyway but maybe `matgl`.

related PR: https://github.com/CederGroupHub/chgnet/pull/133

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,8 +10,7 @@ on:
     branches: [master]
     paths-ignore: ["**/*.md", docs/**]
   workflow_dispatch:
-  # make this workflow reusable by release.yml
-  workflow_call:
+  workflow_call: # make this workflow reusable by release.yml
 
 permissions:
   contents: read
@@ -42,8 +41,6 @@ jobs:
 
     env:
       PMG_MAPI_KEY: ${{ secrets.PMG_MAPI_KEY }}
-      MPLBACKEND: Agg # https://github.com/orgs/community/discussions/26434
-      PMG_TEST_FILES_DIR: ${{ github.workspace }}/tests/files
       GULP_LIB: ${{ github.workspace }}/cmd_line/gulp/Libraries
       PMG_VASP_PSP_DIR: ${{ github.workspace }}/tests/files
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,17 +48,24 @@ jobs:
       PMG_VASP_PSP_DIR: ${{ github.workspace }}/tests/files
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Check out repo
+        uses: actions/checkout@v4
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
           cache-dependency-path: setup.py
+
+      - name: Install uv
+        run: pip install uv
+
       - name: Copy GULP to bin
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo cp cmd_line/gulp/Linux_64bit/* /usr/local/bin/
+
       - name: Install Bader
         if: matrix.os == 'ubuntu-latest'
         run: |
@@ -66,6 +73,7 @@ jobs:
           tar xvzf bader_lnx_64.tar.gz
           sudo mv bader /usr/local/bin/
         continue-on-error: true # This is not critical to succeed.
+
       - name: Install Enumlib
         if: matrix.os == 'ubuntu-latest'
         run: |
@@ -79,6 +87,7 @@ jobs:
           cd ..
           sudo cp aux_src/makeStr.py /usr/local/bin/
         continue-on-error: true # This is not critical to succeed.
+
       - name: Install Packmol
         if: matrix.os == 'ubuntu-latest'
         run: |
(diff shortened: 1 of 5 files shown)


janosh:
good news: looks like linux CI dependency install time went from 2.5 to 1 min (2.5x speed up) and windows install time went from 9-14 min to 3-5 min (~3x speed up)!

also, `matgl` import error was indeed automatically fixed by `uv` and remaining failing tests due to `pydantic` import error are skipped until https://github.com/materialsvirtuallab/matgl/issues/238 is resolved.

pinging a few people in case they want to migrate their repos to `uv` as well for (hopefully) similar speedups.
- @utf for `atomate2`
- @Andrew-S-Rosen for `quacc`
- @shyuep / @esoteric-ephemera for `custodian`
- @DanielYang59  for `pymatviz`
- @mkhorton for CTK

Andrew-S-Rosen:
Amazing! I have a few upstream issues in uv I need fixed before I can switch. Already tried. But I hopeful it will be soon!

janosh:
> I'm quite curious why the original import issue in the first place because these two should function the same?

`uv` is meant as a drop-in replacement for `pip`, yes, but their module resolution works differently. i assume `pip` failed to install `matgl` due conflicting version reqs where `uv` succeeded. but just guessing, didn't look into it

DanielYang59:
> i assume `pip` failed to install `matgl` due conflicting version reqs where `uv` succeeded. but just guessing, didn't look into it

Well...At lease from this [run](https://github.com/materialsproject/pymatgen/actions/runs/8184520269/job/22379106771), `matgl` is successfully installed but still the weird `importError`. Not idea why.
```
Successfully installed ..... matgl-1.0.0 ....
```

DanielYang59:
A quick update @janosh , it seems `uv` timed out for some reason [here](https://github.com/materialsproject/pymatgen/actions/runs/8188398603/job/22390876432?pr=3674).
```bash
Run uv pip install numpy cython --system
Resolved 2 packages in 98ms
Downloaded 2 packages in 419ms
Installed 2 packages in 17ms
 + cython==3.0.9
 + numpy==1.26.4
Built 1 editable in 28.18s
Resolved 194 packages in 2.55s
error: Failed to download distributions
  Caused by: Failed to fetch wheel: torch==2.2.1
  Caused by: Failed to extract source distribution
  Caused by: Failed to download distribution due to network timeout. Try increasing UV_HTTP_TIMEOUT (current value: 300s).
```

.github/workflows/test.yml
```diff
@@ -91,14 +97,15 @@ jobs:
...
         continue-on-error: true # This is not critical to succeed.
+
       - name: Install dependencies
         run: |
-          python -m pip install numpy cython
+          uv pip install numpy cython --system
 
-          # TODO remove next line installing ase from main branch until FrechetCellFilter is released
-          pip install git+https://gitlab.com/ase/ase
+          uv pip install -e '.[dev,optional]' --system
```
DanielYang59, line 105: @janosh, sorry for digging out this. But why do we need to install "editable" for CI?

janosh: i don't remember exactly. i'm fairly sure i tried non-editable first and it failed for some reason. but that was with an earlier `uv` and might have changed

DanielYang59: I could confirm [install "non-editable" indeed breaks the tests](https://github.com/DanielYang59/pymatgen/actions/runs/8628590384/job/23651200610). The `pip install` process seems to run without any issue, but `pytest` seems unable to access test files. I don't quite understand this behavior.

An off-topic question: do we really need: https://github.com/materialsproject/pymatgen/blob/7064c439cbd57d90a39af9659fb206d9eaf22154/.github/workflows/test.yml#L18-L21
By default workflows would not run on forks until enabled by fork owner. Having this would prevent fork from running workflows, which seems unnecessary (some people might prefer to test on their own fork before pushing to upstream)? Or in my case I have to remove this line to run tests on my own fork.

Screenshot 2024-04-10 at 16 57 58

janosh: editable install basically just symlinks the repo into `site-packages`, meaning the whole repo contents including test files are symlinked. i think with non-editable install `TEST_FILES_DIR` points at a non-existent directory.

> why do we need:
```yml
     if: github.repository == 'materialsproject/pymatgen' 
```

not sure if this changed but people used to get every CI failure email twice from GitHub, once on their fork and once on open PRs which was very annoying

janosh: i'm guessing if anything, `--editable` offers a performance gain given how fast it is to symlink

> Otherwise people who want to run workflow on their fork for any reason would need to remove this line?

not sure that's an issue? people usually just fork to make PRs in which CI runs on this repo. if they want to maintain a fork long-term, removing this line is not much work?

DanielYang59: > i'm guessing if anything, `--editable` offers a performance gain given how fast it is to symlink

Not sure, but I would test later and let you know.

> people usually just fork to make PRs in which CI runs on this repo. 

Yes. But my point is... By default workflow would NOT run on forks (I assume people would enable workflow on their fork only when they have a good reason to), such that people could easily toggle workflow on and off easily on their fork (and have more control over their fork).

But if we disable workflow for them, they can not easily enable it. (I guess click a button would be easier than changing a line and make a commit? Also if they want to to push their change to upstream, they would have to revert it.)

PR #3681, opened 2024-03-09
Clean up test files: dedicated VASP directories, `xyz`, `mcif`, `cssr`, `exciting`, `wannier90`

DanielYang59:
## Clean up test files: dedicated VASP directories

The third (hopefully the last) episode of cleaning up VASP test files: relocate dedicated VASP directories to `vasp/fixtures`, following #3653 and #3674.

## Tasks

- [x] Relocate dedicated VASP directories
- [x] Compress some hidden VASP outputs
- [x] Group some files: xyz, mcif, cssr, exciting, wannier90


DanielYang59:
`borg` is a mystery for me because I never use it before. It seems like a parallelized VASP results collection tool. But in the unit tests, what files are being collected is not explicitly declared, nor commented. And the test data length is hard-coded. I would need to add some comments after I find out what files are being collected.

DanielYang59:
There is a test skipped because `free_sph.cif` is absent:
https://github.com/materialsproject/pymatgen/blob/daa0bb01477321828aa324c9ce9d07383db778e1/tests/io/test_zeopp.py#L173-L192

Maybe we should somehow recover the test file or remove this test altogether?

janosh:
> Maybe we should somehow recover the test file or remove this test altogether?

looks like the file never existed which is odd. however, even if we had the file, the test would fail with

> RuntimeError: get_voronoi_nodes requires Zeo++ cython extension to be installed. Please contact developers of Zeo++ to obtain it.

i would keep the test for now even if it's not running.

DanielYang59:
I might need your advice on fixing tests for `borg` @janosh, thanks (sorry I pushed the comment in the wrong repo...).

Current the `drone` is being deployed to `TEST_FILES_DIR`. And as the test files are dynamic under `TEST_FILES_DIR` while the number of `valid_paths` is hard-coded to be `16`: 
https://github.com/materialsproject/pymatgen/blob/910e5719770647e6ed7fa41cb2358d5f8f42161f/tests/apps/borg/test_queen.py#L16-L21
Any time a vasp test directory is added or removed, the test is very likely to fail.

So I guess we should either:
- Use a more "general" condition like `assert len(data)` over `assert len(data) == 16`
- Or 
- Create a dedicated test directories for `borg` with static test files (symlinks maybe).

I would prefer the latter to resolve this once and for all, what do you think?

tests/apps/borg/test_queen.py
```diff
@@ -1,5 +1,3 @@
-"""Created on Mar 18, 2012."""
```
DanielYang59, line 1: Removed as there is already a `__date__` below.

tests/apps/borg/test_hive.py
```diff
@@ -26,25 +25,21 @@ def test_get_valid_paths(self):
                 assert len(self.drone.get_valid_paths(path)) > 0
 
     def test_assimilate(self):
-        """Test assimilate data from "vasprun.xe.xml.gz" file."""
-
-        with ScratchDir("."):
```
DanielYang59, line 31: `ScratchDir` is not necessary as we switched to a dedicated test dir for `borg`

janosh:
> UPDATE: already created a dedicated dir for borg.

great! i would also have gone with this approach 👍

pymatgen/apps/borg/hive.py
```diff
@@ -69,7 +69,7 @@ class VaspToComputedEntryDrone(AbstractDrone):
 
     There are some restrictions on the valid directory structures:
 
-    1. There can be only one vasp run in each directory.
+    1. There can be only one vasprun in each directory.
```
DanielYang59, line 72: I would need to revert this. This doesn't look like a type. I assume this means a VASP job

tests/files/app_borg/test_dir/vasprun.xml.xe.gz
janosh, line 1: do we actually need this level of nesting? what about removing `test_dir`? i.e. move to

```
tests/files/app_borg/vasprun.xml.xe.gz
```

DanielYang59: I assume we do need a nested structure as `queen` is deploying `drones` to search for `valid_paths`: https://github.com/materialsproject/pymatgen/blob/910e5719770647e6ed7fa41cb2358d5f8f42161f/pymatgen/apps/borg/queen.py#L71-L83

And now we are deploying `queen` to `f"{TEST_FILES_DIR}/app_borg"`, which would return `f"{TEST_FILES_DIR}/app_borg/test_dir"` as a `valid_path`? 😄

janosh: makes sense, let's keep it the way it is

janosh (approved):
really nice to bring the repo size down by compressing more test files! thanks @DanielYang59 👍

DanielYang59:
> really nice to bring the repo size down by compressing more test files!

Thanks for reviewing. Didn't compress much files 😃 but just grouped different types to tidy up the test files.

janosh:
quarter million lines gone. i call that a win 😄

DanielYang59:
> quarter million lines gone. 

Yes I indeed removed some unused test directories. Tried my best to search globally for their names and at least confirmed they're not explicitly called. Also committed all removals individually in case any need to be recovered.

With this said, I'm not sure what is the best practice, but I think it's better to explicitly declare file name in unit tests wherever possible? Or if this is not possible, at least put a comment about the target directory/file.

For example this gave me quite a painful time:
https://github.com/materialsproject/pymatgen/blob/d07164fd27b955010ad411033d262ef3d935d09c/tests/apps/borg/test_hive.py#L27-L33

Because:
- No comment nor explicit declaration of test file name.
- A ton of possible target `vasprun.xml` files exist in directory and don't know which is the target file.
- Result hard coded.

Can you please encourage developers to add a comment or declare explicitly in the future if anyone is adding test files? Thanks a ton 😃 .

janosh:
> Can you please encourage developers to add a comment or declare explicitly in the future if anyone is adding test files? Thanks a ton 😃 .

of course! this test predates my becoming a `pymatgen` co-maintainer but i'm glad you refactored to clarify the test ↔ test-file association here

tests/apps/borg/test_queen.py
```diff
@@ -12,16 +10,20 @@
...
     def test_get_data(self):
+        """Test get data from vasprun.xml.xe.gz file."""
         drone = VaspToComputedEntryDrone()
-        self.queen = BorgQueen(drone, TEST_FILES_DIR, 1)
-        data = self.queen.get_data()
-        assert len(data) == 16
+        queen = BorgQueen(drone, TEST_DIR, 1)
+        data = queen.get_data()
+        assert len(data) == 1
+        assert data[0].energy == 0.5559329
```
DanielYang59, line 23: I should have used `approx` here and I would make this change in a future PR somewhere.

PR #3688, opened 2024-03-13
Generalize fatband plots from Lobster

### LINKED ISSUE #3686: Fatband plot with lobster: not only vasp should be supported
### Feature Requested

Not only fatband plots with lobster and VASP should be covered. 

### Proposed Solution

Use POSCAR.lobster instead of vasprun.xml, POSCAR.lobster will be used. Efermi can be read from vasprun.xml

### Relevant Information

_No response_

JaGeo:
Closes #3686 

Todo: 

- [x] More tests


pymatgen/io/lobster/outputs.py
```diff
@@ -1116,62 +1116,84 @@ class Fatband:
...
-        kpoints_object = Kpoints.from_file(Kpointsfile)
+        if structure_file is None and structure is None:
+            raise ValueError("structure_file or structure have to be provided")
+        if vasprun_file is None and efermi is None:
+            raise ValueError("vasprun_file or efermi have to be provided")
+        if structure_file is not None:
+            self.structure = Structure.from_file(structure_file)  # type: ignore
+        else:
+            self.structure = structure  # type: ignore
+        self.lattice = self.structure.lattice.reciprocal_lattice  # type: ignore
```
DanielYang59, line 1154: I wish you don't mind me pop in...But if `self.structure` is `None`, would this `self.lattice = self.structure.lattice.reciprocal_lattice` fail?

DanielYang59, line 1154: I wish you don't mind me pop in, try this:
```python
        warnings.warn("Make sure all relevant FATBAND files were generated and read in!")
        warnings.warn("Use Lobster 3.2.0 or newer for fatband calculations!")

        if vasprun_file is None and efermi is None:
            raise ValueError("vasprun_file or efermi have to be provided")

        if structure_file is not None:
            self.structure = Structure.from_file(structure_file)
        elif structure is not None:
            self.structure = structure
        else:
            raise ValueError("structure_file or structure have to be provided")
```

pymatgen/io/lobster/outputs.py
```diff
@@ -1116,62 +1116,84 @@ class Fatband:
...
-        structure (Structure): Structure read in from vasprun.xml.
+        structure (Structure): Structure read in from structure_file or Structure object.
     """
 
-    def __init__(self, filenames=".", vasprun="vasprun.xml", Kpointsfile="KPOINTS"):
+    def __init__(
+        self,
+        filenames: str | list = ".",
+        kpoints_file: str = "KPOINTS",
+        structure_file: str | None = "POSCAR.lobster",
```
janosh, line 1126: following Python's "there should be one obvious way to do something" i'd advocate for removing the `structure_file` keyword. afaict, it's only saving the user 1 line

```py
Structure.from_file(structure_file)
```

tests/io/lobster/test_inputs.py
```diff
@@ -1274,27 +1274,78 @@ class TestFatband(PymatgenTest):
...
+                ionic_step_skip=None,
+                ionic_step_offset=0,
+                parse_dos=True,
+                parse_eigen=False,
+                parse_projected_eigen=False,
+                parse_potcar_file=False,
+                occu_tol=1e-8,
+                exception_on_bad_xml=True,
+            ).final_structure,
+        )
```
janosh, line 1310: maybe extract repeated kwargs into a separate dict to reuse in each `Vasprun` call:

```py
kwargs = dict(
        ionic_step_skip=None,
        ionic_step_offset=0,
        parse_dos=True,
        parse_eigen=False,
        parse_projected_eigen=False,
        parse_potcar_file=False,
        occu_tol=1e-8,
        exception_on_bad_xml=True,
        # ...
)

Vasprun(filenames="...", **kwargs)
```

janosh (changes requested):

tests/io/lobster/test_inputs.py
JaGeo: I decided to rather define a structure in the setup and then reuse that. But I also got rid of lot of repetitions. Thanks for the suggestion!

JaGeo: Not sure why I did not do this in the first place. I guess I was too distracted with getting it to work in general.

DanielYang59:
Looks like the same download timeout error as #3675 from `uv`, see discussion [here](https://github.com/astral-sh/uv/issues/1921), maybe we should increase the `UV_HTTP_TIMEOUT`?

JaGeo:
@janosh weird failures during the dependency installations

janosh:
no point in increasing `UV_HTTP_TIMEOUT` since it's already very high (300s). hopefully, `uv` comes up with a fix soon-ish else we might have to pre-install `pytorch` with `pip` before calling `uv`

DanielYang59:
> we might have to pre-install `pytorch` with `pip` before calling `uv`

This sounds like a good temporary fix 😃

JaGeo:
@janosh are the changes fine? Tests have run through after I triggered them again

janosh (approved):
thanks @JaGeo, lgtm. 👍

janosh:
> Todo:
> More tests

is the todo resolved?

JaGeo:
> > Todo:
> > More tests
> 
> is the todo resolved?

Yes, i added several new tests.

PR #3689, opened 2024-03-14
Improve Bandoverlaps parser

naik-aakash:
# Issue
Current implementation of  `Bandoverlaps` parser adds keys with '.' in `band_overlaps_dict` attribute leading to failure of pydantic checks in atomate2 lobster workflow https://github.com/materialsproject/atomate2/issues/769

# Changes

1. Rewrite output format of `band_overlaps_dict` attribute of `Bandoverlaps` class
2. Update and add associated tests


pymatgen/io/lobster/outputs.py
```diff
@@ -1363,24 +1365,31 @@ def _read(self, contents: list, spin_numbers: list):
...
 
             else:
-                overlaps = []
+                rows = []
                 for el in line.split(" "):
                     if el != "":
-                        overlaps.append(float(el))
-                self.band_overlaps_dict[spin][" ".join(kpoint_array)]["matrix"].append(overlaps)
+                        rows += [float(el)]
+                overlaps += [rows]
```
janosh, line 1390: `overlaps` is possibly unbound now from renaming `overlaps = []` to `rows` above.

we don't check for this in CI yet due to way too many legacy violations but really should. related discussion in https://github.com/materialsproject/pymatgen/pull/3646

janosh (changes requested):

pymatgen/io/lobster/outputs.py
naik-aakash: Hi @janosh, I did not get what you meant here. Can you please elaborate a bit more? I need the overlaps variable initialized in the previous if clause to properly store matrices for each k-point.

Is there anything I need to do ? I could not think of any other way to do this.

janosh: the problem is, if the code enters the `else` case `overlaps` won't be declared, so Python will throw a `NameError`

naik-aakash: I think, such case won't happen as per the current file format, it will always pass through the [elif clause ](https://github.com/naik-aakash/pymatgen/blob/b93fc2331a2a5f8e69483f318762599208642ab4/pymatgen/io/lobster/outputs.py#L1370).

But I Will check for a few more examples that I have and see If I encounter any errors

naik-aakash: Hi @janosh , I checked for about 20 different example files which I have, does not result in `NameError `due to file format , logic seems to work without breaking it 😃

But if you have any better idea to get same outcome, I am happy to implement it. Just at this point I seem to have blanked out 😅 and cannot think of any alternatives

janosh: you could `pip install pyright` and run it on this file. it'll show you the error i mean. you just need to ensure `overlaps` is declared by all code paths

naik-aakash: Hi @janosh , I have now address the errors which I could using pyright. If you think these are okay it could be merged.

janosh (approved):
thanks @naik-aakash! 👍

Issue #3682, opened 2024-03-09
Create monolayer MoS2 from the bulk.


hongyi-zhao:
I try to generate the monolayer MoS2 from the bulk as follows:

```python
from pymatgen.core.structure import Structure
from pymatgen.core.surface import SlabGenerator
from pymatgen.io.vasp import Poscar

# Load the MoS2 structure (assuming you have the CIF file)
MoS2 = Structure.from_file("MoS2_mp-1434.cif")

# Add oxidation states
MoS2.add_oxidation_state_by_element({"Mo": 4, "S": -2})

# Initialize SlabGenerator for a single layer
slabgen = SlabGenerator(MoS2,
                        miller_index=(0, 0, 1),
                        min_slab_size=MoS2.lattice.c/3,  # Approximate single layer thickness in Å
                        in_unit_planes=False, 
                        min_vacuum_size=10)

# Generate slabs
slabs = slabgen.get_slabs()

# Assuming the first slab is the desired single layer, this may need verification
single_layer_slab = slabs[0]

# Write the single layer structure to a POSCAR file
poscar = Poscar(single_layer_slab)
poscar.write_file("MoS2_single_layer_POSCAR")
```

But the result is as follows:

![image](https://github.com/materialsproject/pymatgen/assets/11155854/cbb704b1-798a-40aa-811d-ea3f568c6b86)

As you can see, obviously it includes multiple layers alongside the (001) plane. Attached are all the files mentioned above:

[MoS2_ML.zip](https://github.com/materialsproject/pymatgen/files/14546257/MoS2_ML.zip)

Any tips for achieving this goal will be helpful.

Regards,
Zhao

Issue #3412, opened 2023-10-17
Inconsistency in oriented_unit_cell when generating (001)/(010)/(100) surface


hezhengda:
### Bug description
When I generated the slab of TiNb(010) surface, I have found that the thickness of the slab is smaller than `min_slab_size`, although `primitive=False` has been applied. 

### Reproduce the bug
```python
from pymatgen.core.surface import generate_all_slabs
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
from pymatgen.core import Structure

bulk_structure = Structure.from_file('TiNb_conv.cif')
sga = SpacegroupAnalyzer(bulk_structure)
conv_structure = sga.get_conventional_standard_structure()

slabs = generate_all_slabs(structure=conv_structure, 
                           max_index=2, 
                           min_slab_size=8,
                           min_vacuum_size=15, 
                           center_slab=True,
                           primitive=False)
```

The cif file for `TiNb_conv.cif` is:
```
# generated using pymatgen
data_TiNb
_symmetry_space_group_name_H-M   'P 1'
_cell_length_a   3.28200886
_cell_length_b   4.63609090
_cell_length_c   4.64594597
_cell_angle_alpha   90.00000000
_cell_angle_beta   90.00000000
_cell_angle_gamma   90.00000000
_symmetry_Int_Tables_number   1
_chemical_formula_structural   TiNb
_chemical_formula_sum   'Ti2 Nb2'
_cell_volume   70.69128019
_cell_formula_units_Z   2
loop_
 _symmetry_equiv_pos_site_id
 _symmetry_equiv_pos_as_xyz
  1  'x, y, z'
loop_
 _atom_site_type_symbol
 _atom_site_label
 _atom_site_symmetry_multiplicity
 _atom_site_fract_x
 _atom_site_fract_y
 _atom_site_fract_z
 _atom_site_occupancy
  Ti  Ti0  1  0.50000000  0.00000000  0.50000000  1.0
  Ti  Ti1  1  0.00000000  0.50000000  0.50000000  1.0
  Nb  Nb2  1  0.00000000  0.00000000  0.00000000  1.0
  Nb  Nb3  1  0.50000000  0.50000000  0.00000000  1.0
```
---
### Try to pin point the bug
I have done some digging to understand the source of the bug.

The lattice constant of `oriented_unit_cell` for (010) surface is:
```
The lattice constants of oriented_unit_cell is:
-3.28,0.00,-0.00
[...]

Issue #3411, opened 2023-10-17
Thickness of generated slab is smaller than min_slab_size


hezhengda:
### Bug Description
When using `generate_all_slabs` to generate surfaces by `primitive=True`, I have found some surfaces will have thickness smaller than the `min_slab_size` value.

### Reproduce of the bug
```python
from pymatgen.core.surface import generate_all_slabs
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
from pymatgen.core import Structure

bulk_structure = Structure.from_file('TiNb_conv.cif') # I attach this as `TiNb_conv.cif`
sga = SpacegroupAnalyzer(bulk_structure)
conv_structure = sga.get_conventional_standard_structure()

slabs = generate_all_slabs(structure=conv_structure, 
                           max_index=2, 
                           min_slab_size=8,
                           min_vacuum_size=15, 
                           center_slab=True,
                           primitive=True)
```
The slab thickness of TiNb(111) surface is 7.06Å, which is less than `min_slab_size`, which is 8Å.

Slab thickness is calculated by the maximal distance of z-component of cartesian coordinates:
```python
coords = structure.cart_coords
sorted_coords = sorted(coords, key=lambda x: x[2])
thickness = sorted_coords[-1][2] - sorted_coords[0][2]
```

Since cif file cannot be uploaded, I paste the cif file (`TiNb_conv.cif`, conventional cell of TiNb) in the following:
```
# generated using pymatgen
data_TiNb
_symmetry_space_group_name_H-M   'P 1'
_cell_length_a   3.28200886
_cell_length_b   4.63609090
_cell_length_c   4.64594597
_cell_angle_alpha   90.00000000
_cell_angle_beta   90.00000000
_cell_angle_gamma   90.00000000
_symmetry_Int_Tables_number   1
_chemical_formula_structural   TiNb
_chemical_formula_sum   'Ti2 Nb2'
_cell_volume   70.69128019
_cell_formula_units_Z   2
loop_
 _symmetry_equiv_pos_site_id
 _symmetry_equiv_pos_as_xyz
  1  'x, y, z'
loop_
 _atom_site_type_symbol
 _atom_site_label
 _atom_site_symmetry_multiplicity
 _atom_site_fract_x
 _atom_site_fract_y
 _atom_site_fract_z
 _atom_site_occupancy
  Ti  Ti0  1  0.50000000  0.00000000  0.50000000  1.0
[...]

PR #3691, opened 2024-03-17
Clean up `core.surface` comments and docstrings

### LINKED ISSUE #2591: Atom position overlap after repairing the slab
Hey, I'm trying to generate the (010) facet structures of θ-Al2O3 with pymatgen 2022.5.26
`
from pymatgen.core.surface import *

struct = Structure.from_file('Al2O3CONTCAR')
struct = SpacegroupAnalyzer(struct).get_conventional_standard_structure()
slab = SlabGenerator(struct, miller_index=[0, 1, 0], min_slab_size=10,min_vacuum_size=15) 

for n, s in enumerate(slab.get_slabs(bonds = {('Al','O'):2.1},repair=True)):
    s=s.get_sorted_structure() 
    s.to(filename='POSCAR-'+str(n),fmt='poscar')
`
When I opened one of the generated POSCAR with VESTA, I found the label of the 30th Al atom is invisible because of the overlap,
vesta

which caused the failure of the subsequent calculation by VASP
vasp

It is strange that such a problem will not happen when there is no bond constraints and repair requirement.

Here is the input CONTCAR and the generated POSCAR
[Al2O3CONTCAR.txt](https://github.com/materialsproject/pymatgen/files/9189365/Al2O3CONTCAR.txt)
[POSCAR-0.txt](https://github.com/materialsproject/pymatgen/files/9189368/POSCAR-0.txt)

DanielYang59:
## Summary

Major clean up of `core.surface`, I decide to separate this task into two for readability: this PR to revise comments/docstrings with some relocation of code. Another follow up PR for functional changes/fixes.

### Non-breaking changes

- Major clean up of all comments/docstrings (made more concise and add missing)
- Sort ordering of methods in `Slab` and `SlabGenerator`

### TODOs:
- [x] Add type annotations and format clean up in `core.surface` and `core.structure`
- [x] Clarify wording of `site` and `point` in `surface.Slab`, currently a mixture of both is used and might be confusing
- [x] Legacy fix in `queen` unit test from #3681

### Breaking changes (need discussion and confirmation for each)

- [ ] Use `isclose` with tolerance instead of full equality (`==`) to check if two positions overlap, [details](https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1554563901).
- [ ] Revert default `shift` position (currently a positive shift value would move the atoms "downwards"), [details](https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1553711017). The unit of `shift` needs to be clarified as well (Cartesian/fractional).
- [ ] `Shift` values generated from `SlabGenerator.get_slabs.gen_possible_shifts` is weird (default `shift` is 0.5, also the center of cluster is returned instead of the shift itself), need a closer look, [details](https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1554518586).

### Fix some `Slab` related issues
Don't think I could resolve them in one shot, but good to collect them:
- #2591
- #3411
- #3412
- #3682

## Summary by CodeRabbit

- **Refactor**
	- Improved efficiency and clarity in methods handling site properties, oxidation states, and structure interpolation.
	- Enhanced type specificity and error handling in various structure management methods.

- **Documentation**
	- Updated comments and docstrings for better clarity and understanding of the code.

- **Tests**
	- Enhanced accuracy in tests by using better assertions for floating-point comparisons.

- **Bug Fixes**
	- Adjusted fractional coordinates calculation to honor the correct conditions.


DanielYang59:
Notice multiple funcs with unused `include_index` and `include_image` args across `core.structure`, for example: https://github.com/materialsproject/pymatgen/blob/884946d0d3c022ae197e4188364db2b7ad9cde9a/pymatgen/core/structure.py#L1825-L1911

These args were marked "Deprecated" **five years ago** in #1610 so I assume I should be safe to remove them?

Or perhaps even the function altogether? https://github.com/materialsproject/pymatgen/blob/884946d0d3c022ae197e4188364db2b7ad9cde9a/pymatgen/core/structure.py#L1990-L2007

Can you please advice on this? @janosh

pymatgen/core/surface.py
```diff
@@ -449,9 +446,6 @@ def add_adsorbate_atom(self, indices, specie, distance) -> Slab:
         return self
 
     def __str__(self) -> str:
-        def to_str(x) -> str:
```
DanielYang59, line 452: Unused func

pymatgen/core/surface.py
```diff
@@ -86,7 +86,6 @@ def __init__(
         coords_are_cartesian=False,
         site_properties=None,
         energy=None,
-        properties=None,
```
DanielYang59, line 89: Unused arg

pymatgen/core/surface.py
```diff
@@ -1633,10 +1626,10 @@ def hkl_transformation(transf, miller_index):
     """
     # Get a matrix of whole numbers (ints)
 
-    def lcm(a, b):
```
DanielYang59, line 1636: Avoid redefine `from monty.fractions import lcm`

DanielYang59:
I must have opened the Pandora's box of `mypy` errors... Still quite some errors to fix 😭

pymatgen/core/surface.py
```diff
@@ -339,26 +341,17 @@ def get_sorted_structure(self, key=None, reverse=False):
...
         site properties.
 
         Args:
-            site_properties (dict): Properties to add or override. The
+            site_properties (dict): Properties to update. The
                 properties are specified in the same way as the constructor,
-                i.e., as a dict of the form {property: [values]}. The
-                properties should be in the order of the *original* structure
-                if you are performing sanitization.
-            sanitize (bool): If True, this method will return a sanitized
```
DanielYang59, line 352: Unused arg `sanitize`

pymatgen/core/surface.py
```diff
@@ -66,37 +69,35 @@ class Slab(Structure):
 
     Attributes:
         miller_index (tuple): Miller index of plane parallel to surface.
-        scale_factor (float): Final computed scale factor that brings the parent cell to the surface cell.
```
DanielYang59, line 69: This is super confusing. Need a closer look. `scale_factor` seems to be a `np.ndarray` not a `float`

DanielYang59:
Can you pls assign me to the above mentions issues so other would know someone is working on those to avoid overlap? @janosh, thanks.

janosh:
i can only assign members of MP to issues or people who've commented on issues. so if you leave a comment i could assign you. but given the 1st issue has been open since Oct 23, i don't see a high risk of other people working on it

DanielYang59:
> but given the 1st issue has been open since Oct 23, i don't see a high risk of other people working on it

I think it should be fine. People should see this PR being linked to those issues. Thanks for the help. 

But I'm afraid this would take some time as it seems I would need to clean up the entire `surface` module

janosh:
> But I'm afraid this would take some time as it seems I would need to clean up the entire surface module

that would be great. i can definitely help out here and there :)

DanielYang59:
> Notice multiple funcs with unused `include_index` and `include_image` args across `core.structure`, for example:
> 
> https://github.com/materialsproject/pymatgen/blob/884946d0d3c022ae197e4188364db2b7ad9cde9a/pymatgen/core/structure.py#L1825-L1911
> 
> These args were marked "Deprecated" **five years ago** in #1610 so I assume it should be safe to remove them?
> 
> Or perhaps even the function altogether?
> 
> https://github.com/materialsproject/pymatgen/blob/884946d0d3c022ae197e4188364db2b7ad9cde9a/pymatgen/core/structure.py#L1990-L2007

In case you missed this @janosh. Thanks!

janosh:
sorry, i did miss this. 

`get_all_neighbors_old` needs to stay. it's only there for testing purposes (slow but easy to implement and therefore less error prone).

https://github.com/materialsproject/pymatgen/blob/48af0386a80b5a6e59d97b42f62529358c43cbb3/tests/core/test_structure.py#L815

the unused keywords `include_(index|image)` are to unsure compatible function signatures though i never checked if anything would actually break by removing them. you're welcome to try :)

DanielYang59:
> sorry, i did miss this.
> 
> `get_all_neighbors_old` needs to stay. it's only there for testing purposes (slow but easy to implement and therefore less error prone).
> 
> https://github.com/materialsproject/pymatgen/blob/48af0386a80b5a6e59d97b42f62529358c43cbb3/tests/core/test_structure.py#L815

Sure let's keep it for testing

> the unused keywords `include_(index|image)` are to unsure compatible function signatures though i never checked if anything would actually break by removing them. you're welcome to try :)

Thanks. I might go ahead and give it a try. But I'm not sure if any downstream packages call these two args, if they do, this might break them.

For example if any downstream code goes like:
 ```python
get_all_neighbors( 
include_index = False, 
include_image = False
)
```

These two args would be just ignored but if we remove it the code might break (`include_index  is an invalid keyword argument for get_all_neighbors`).

 I guess it's better to add some deprecation tag for some time just in case?

**Comment by janosh** [comment]
given those keywords have no effect, downstream code is unlikely to use them but of course a deprecation period doesn't hurt

File: pymatgen/core/structure.py [hunk 6]
```diff
@@ -49,6 +49,7 @@
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Sequence
     from pathlib import Path
+    from typing import Self
```
DanielYang59, line 52: A new trick for me. But I get:
```python
pymatgen/core/structure.py:52: error: Module "typing" has no attribute "Self"  [attr-defined]
pymatgen/core/structure.py:52: note: Use `from typing_extensions import Self` instead
pymatgen/core/structure.py:52: note: See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module
```

I might replace these. @janosh

DanielYang59:
Dependency installation with `uv` is [failing again](https://github.com/materialsproject/pymatgen/actions/runs/8320310950/job/22764834050?pr=3691). @janosh

pymatgen/core/structure.py
janosh: feel free to remove. i forgot to remove myself. it's only supported on 3.11+ but i was curious to check if it would solve the `mypy` error (it didn't)

janosh:
> Maybe we should consider reverting torch installation with pip for the time being? @janosh

the thing we need to find out first is how to make `uv` ignore packages that are already installed. i think by default it will try to overwrite existing `pytorch`

DanielYang59:
> the thing we need to find out first is how to make `uv` ignore packages that are already installed. i think by default it will try to overwrite existing `pytorch`

It seems to me `uv` would ignore already installed package? Maybe I'm wrong
```bash
(venv) yang@Yang-MacStudio pymatgen % pip show torch
Name: torch
Version: 2.2.1
Summary: Tensors and Dynamic neural networks in Python with strong GPU acceleration
Home-page: https://pytorch.org/
Author: PyTorch Team
Author-email: packages@pytorch.org
License: BSD-3
Location: /Users/yang/developer/pymatgen/venv/lib/python3.9/site-packages
Requires: filelock, fsspec, jinja2, networkx, sympy, typing-extensions
Required-by: matgl, pytorch-lightning, torchdata, torchmetrics

(venv) yang@Yang-MacStudio pymatgen % uv pip install torch
Audited 1 package in 2ms
```

janosh:
you need to test `uv pip install .`. i think in that case it will overwrite but maybe they changed that in a recent release

DanielYang59:
This is what I got running locally on my MacOS machine:
```bash
(venv) yang@Yang-MacStudio pymatgen % uv pip install -e .
   Built file:///Users/yang/developer/pymatgen                                  Built 1 editable in 31.62s
Resolved 40 packages in 2.03s
Installed 1 package in 1ms
 - pymatgen==2024.3.1 (from file:///Users/yang/developer/pymatgen)
 + pymatgen==2024.3.1 (from file:///Users/yang/developer/pymatgen)
```

Hope this would be helpful 😄

pymatgen/core/structure.py
DanielYang59: Look like you are using it in the absolutely correct form, just replacing it with `from typing_extensions import Self` is sufficient to fix this issue below Python 3.11.

pymatgen/core/surface.py
```diff
@@ -165,131 +165,24 @@ def __init__(
...
-                slabs.append(slab)
-        s = StructureMatcher()
-        return [ss[0] for ss in s.group_structures(slabs)]
+        for idx, site in enumerate(self):
+            outs.append(f"{idx + 1} {site.species_string} {' '.join(f'{j:0.6f}'.rjust(12) for j in site.frac_coords)}")
+        return "\n".join(outs)
 
-    def is_symmetric(self, symprec: float = 0.1):
+    @property
+    def is_symmetric(self, symprec: float = 0.1) -> bool:
```
DanielYang59, line 185: Default value of `symprec` of `0.1` is different from the default value in `SpacegroupAnalyzer` (0.01)

pymatgen/core/surface.py
```diff
@@ -165,204 +160,108 @@ def __init__(
...
+        """
+        centroid = np.sum(self.cart_coords, axis=0) / len(self)
 
-            for selection in itertools.product(*combinations):
-                species = [site.species for site in fixed]
-                frac_coords = [site.frac_coords for site in fixed]
+        dipole = np.zeros(3)
+        for site in self:
+            charge = sum(getattr(sp, "oxi_state", 0) * amt for sp, amt in site.species.items())
+            dipole += charge * np.dot(site.coords - centroid, self.normal) * self.normal
```
DanielYang59, line 191: Need confirmation: The formula for calculating dipole moment seems not correct. Should be `charge * np.dot(site.coords - centroid, self.normal) / self.normal`?

pymatgen/core/surface.py
```diff
@@ -515,8 +323,8 @@ def get_surface_sites(self, tag=False):
         for analysis involving broken bonds and for finding adsorption sites.
```
DanielYang59, line 324: What is a "elemental system" here?

pymatgen/core/surface.py
```diff
@@ -515,8 +323,8 @@ def get_surface_sites(self, tag=False):
...
```
DanielYang59, line 323: What is "elemental system" in the docstring
```
        We use VoronoiNN to determine the coordination number of sites.
        Due to the pathological error resulting from some surface sites in the
        VoronoiNN, we assume any site that has this error is a surface
        site as well. This will work for elemental systems only for now.
        Useful for analysis involving broken bonds and for finding adsorption sites.
```

janosh: systems consisting only of one type of element

janosh:
> Maybe we should consider reverting `torch` installation with `pip` for the time being? @janosh

that was added in #3697 and is now on `master`

pymatgen/core/surface.py
janosh: feel free to replace with "single-element system" for clarity

DanielYang59: Thanks will do. Would appreciate it if you could give me a hand with #3694. Tried my best to fix most `mypy` errors but still a few left. Thanks! 😃

pymatgen/core/surface.py
```diff
@@ -339,311 +590,56 @@ def get_sorted_structure(self, key=None, reverse=False):
...
 
         Returns:
             Slab: self with adsorbed atom.
         """
-        # Let's work in Cartesian coords
+        # Calculate target site as the center of sites
         center = np.sum([self[idx].coords for idx in indices], axis=0) / len(indices)
 
-        coords = center + self.normal * distance / np.linalg.norm(self.normal)
+        coords = center + self.normal * distance
```
DanielYang59, line 614: The surface normal vector should have been normalize to unit length already.

pymatgen/core/surface.py
```diff
@@ -339,311 +590,56 @@ def get_sorted_structure(self, key=None, reverse=False):
...
-
-        return site2
-
-    def symmetrically_add_atom(self, specie, point, coords_are_cartesian=False) -> None:
-        """Class method for adding a site at a specified point in a slab.
-            Will add the corresponding site on the other side of the
-            slab to maintain equivalent surfaces.
+    def symmetrically_add_atom(
+        self, specie: str | Element | Species, point: ArrayLike, coords_are_cartesian: bool = False
+    ) -> None:
```
DanielYang59, line 622: Seems like a type, should be `species`?

pymatgen/core/surface.py
```diff
@@ -339,311 +590,56 @@ def get_sorted_structure(self, key=None, reverse=False):
...
+            specie (str | Element | Species): The specie to add
+            point (ArrayLike): The coordinate of the site in the slab to add.
             coords_are_cartesian (bool): Is the point in Cartesian coordinates
-
-        Returns:
-            Slab: The modified slab
         """
-        # For now just use the species of the
-        # surface atom as the element to add
+        # For now just use the species of the surface atom as the element to add
```
DanielYang59, line 631: Well...I don't quite understand what this comment means...

pymatgen/core/surface.py
```diff
@@ -869,25 +917,23 @@ def __init__(
...
         _a, _b, c = self.oriented_unit_cell.lattice.matrix
         self._proj_height = abs(np.dot(normal, c))
-        self.reorient_lattice = reorient_lattice
 
-    def get_slab(self, shift=0, tol: float = 0.1, energy=None):
-        """This method takes in shift value for the c lattice direction and
-        generates a slab based on the given shift. You should rarely use this
-        method. Instead, it is used by other generation algorithms to obtain
-        all slabs.
+    def _get_slab(self, shift: float=0, tol: float = 0.1, energy: float | None=None) -> Slab:
```
DanielYang59: Need discussion on this rename @janosh.

Though seemingly breaking, I guess this still should be renamed as a private method for two reasons:
- The original author of this code block explicitly suggested against using this method directly: https://github.com/materialsproject/pymatgen/blob/81dc677462d164182115903e28318cb5cf2ea804/pymatgen/core/surface.py#L877-L881
- There is another `get_slabs` method side by side, which is confusing: https://github.com/materialsproject/pymatgen/blob/81dc677462d164182115903e28318cb5cf2ea804/pymatgen/core/surface.py#L1038-L1052

janosh: i suspect it's quite possible the `get_slab` method has seen external use despite it's doc string warning against it. so i don't think we can easily rename it at this point. i would keep it as is

JaGeo: Is it really a warning against it or rather a suggestion about how the method should be used?

janosh: yeah, warning was too strong a word. like you said, it's a suggestion

DanielYang59: > i suspect it's quite possible the `get_slab` method has seen external use despite it's doc string warning against it. so i don't think we can easily rename it at this point. i would keep it as is

Yes I totally agree with you on this point, as I noticed multiple external calls to this method in d3da8e29e89dba34568fccafdcd82e1a282bfe55.

> Is it really a warning against it or rather a suggestion about how the method should be used?

Yes I assume it's a suggestion too because `You should rarely use this method. Instead, it is used by other generation algorithms to obtain all slabs`.

But at the end of day, it's confusing to have two similar (in both name and function) methods side by side (`get_slab` and `get_slabs`), should we do something about this?

janosh: i think it's not that confusing, especially with the `get_slab` doc string recommending to look at other methods

DanielYang59: Okay let's know it as is and I would revert this renaming.

pymatgen/core/surface.py
```diff
@@ -869,63 +917,71 @@ def __init__(
         self.min_slab_size = min_slab_size
         self.in_unit_planes = in_unit_planes
         self.primitive = primitive
-        self._normal = normal
+        self._normal = normal  # TODO (@DanielYang59): used only in unit test
```
DanielYang59, line 920: This protected attribute `self._normal` seems only used for unit test (to confirm surface normal is indeed perpendicular to vector a/b): https://github.com/materialsproject/pymatgen/blob/a2ab944bf8e1b90e51ca2b65f853972392177286/tests/core/test_surface.py#L400-L404 

This is certainly not wrong but I'm just wondering if there is a cleaner way to achieve this? As having some code here just for test purpose "smells suspicious", but I don't know how to do better frankly.

janosh: normal vector seems like a useful thing to know about a surface. surprised that it's only used in testing. we could turn it into a `@property` of `Slab` and give it it's own unit test

DanielYang59: Thanks for sharing your thoughts. It's actually a property for `Slab` already, see: https://github.com/materialsproject/pymatgen/blob/a2ab944bf8e1b90e51ca2b65f853972392177286/pymatgen/core/surface.py#L410-L415

janosh: i see and https://github.com/materialsproject/pymatgen/blob/a2ab944bf8e1b90e51ca2b65f853972392177286/pymatgen/core/surface.py#L872 is on `SlabGenerator`, not `Slab` itself. but shouldn't it also be a `@property` on `SlabGenerator`?

DanielYang59: I'm still reading and consuming the `core.surface` module, so I may not be having a good enough understanding. And I'm thinking of cleaning up this module "superficially" in the first go here and then dive deeper later, right?

The `SlabGenerator` seems like a generator (quite obviously) to generator `Slab`s from a given condition, as such I wouldn't expect the generator to carry the normal vector of the `Slab` (which sounds like a property of the `Slab` itself). What do you think?

janosh: > I wouldn't expect the generator to carry the normal vector of the Slab

but it already does? unless you want to remove it in which case i'm open to arguments. it would be good to know if all generated slabs are guaranteed to have the same normal vector. in that case, i think it makes sense to have it on the generator. if not, i don't understand why it's there in the 1st place

DanielYang59: > but it already does? 

Yes it does but as a protected attribute instead of a `property`.

> it would be good to know if all generated slabs are guaranteed to have the same normal vector.

I think this should be an internal checking mechanism for `SlabGenerator`. Currently the checking mechanism for surface normal seems rely on the external unit test, which feels weird (and thus I tagged it here), maybe we should migrate this check into `SlabGenerator`?

> it would be good to know if all generated slabs are guaranteed to have the same normal vector.

I'm not certain about this, but at least from the original author and also the [original paper](https://www.sciencedirect.com/science/article/pii/S003960281300160X): `Normality is not guaranteed, but the oriented cell will have the c vector as normal as possible to the surface.` here: https://github.com/materialsproject/pymatgen/blob/a2ab944bf8e1b90e51ca2b65f853972392177286/pymatgen/core/surface.py#L759-L770

Maybe we could put it aside for now and revisit this after I finish consuming the entire `core.surface` module (By which time I would have a much better understanding of all these)? Sorry for drawing your attention too early.

janosh: sounds good

pymatgen/core/surface.py
```diff
@@ -869,63 +917,79 @@ def __init__(
...
+        new_lattice = [a, b, n_layers * c]
+
         species = self.oriented_unit_cell.species_and_occu
-        props = self.oriented_unit_cell.site_properties
-        props = {k: v * n_layers_slab for k, v in props.items()}  # type: ignore[operator, misc]
+
+        # Shift all atoms
+        # DEBUG(@DanielYang59): shift value in Angstrom inconsistent with frac_coordis
         frac_coords = self.oriented_unit_cell.frac_coords
         frac_coords = np.array(frac_coords) + np.array([0, 0, -shift])[None, :]
```
DanielYang59, line 963: This shifting method looks suspicious: 
- As per the docstring, the `shift` should be in Angstrom, but is applied directly to fractional coordinates.
- I would expect to a positive `shift` to move the atom towards the positive direction of z-axis, but this does the contrary.

pymatgen/core/surface.py
```diff
@@ -961,124 +1037,118 @@ def get_slab(self, shift=0, tol: float = 0.1, energy=None):
...
-        c_ranges = [] if bonds is None else self._get_c_ranges(bonds)
+
+        def calculate_possible_shifts(tol: float = 0.1) -> list[float]:
+            frac_coords = self.oriented_unit_cell.frac_coords
+            n_atoms = len(frac_coords)
+
+            # Clustering does not work when there is only one atom
+            if n_atoms == 1:
+                # TODO (@DanielYang59): why this magic number?
+                shift = frac_coords[0][2] + 0.5
```
DanielYang59, line 1086: Here is a magic number `0.5`, when the parent structure has only one atom, the atom would be shifted by `0.5` (fractional coordinate), which is super weird?

I would expect not to shift at all, or to shift it to the center.

pymatgen/core/surface.py
```diff
@@ -961,124 +1037,123 @@ def get_slab(self, shift=0, tol: float = 0.1, energy=None):
...
+            frac_coords = self.oriented_unit_cell.frac_coords
+            n_atoms = len(frac_coords)
+
+            # Clustering does not work when there is only one atom
+            if n_atoms == 1:
+                # TODO (@DanielYang59): why this magic number 0.5?
+                shift = frac_coords[0][2] + 0.5
+                return [shift - math.floor(shift)]
+
+            # Compute a Cartesian z-coordinate distance matrix
```
DanielYang59, line 1095: The implement differs from the comment: https://github.com/materialsproject/pymatgen/blob/a2ab944bf8e1b90e51ca2b65f853972392177286/pymatgen/core/surface.py#L978-L990

where I'm not seeing the periodic boundary condition (PBC) being accounted for. The comment was corrected for now but it's certainly better to actually revise the implementation to account for PBC.

pymatgen/core/surface.py
```diff
@@ -961,126 +1037,140 @@ def get_slab(self, shift=0, tol: float = 0.1, energy=None):
...
+                                # beyond the upper boundary
+                                if z_range[1] > 1:
+                                    z_ranges.extend([(z_range[0], 1), (0, z_range[1] - 1)])
+
+                                # When z coordinate is below the lower boundary
+                                elif z_range[0] < 0:
+                                    z_ranges.extend([(0, z_range[1]), (z_range[0] + 1, 1)])
+
+                                # Neglect overlapping positions
+                                elif z_range[0] != z_range[1]:
```
DanielYang59, line 1161: Using `==` for detecting equality of z coordinates seems improper, but replacing it with `isclose(z_range[0], z_range[1], abs_tol=tol)` breaks some unit tests. Need a closer look.

pymatgen/core/surface.py
```diff
@@ -1217,105 +1217,122 @@ def get_z_ranges(
...
-                        # and then move those NNs along with the central
+                        # Move those NNs along with the center
                         # atom back to the other side of the slab again
                         slab = self.move_to_other_side(slab, to_move)
 
         return slab
 
-    def move_to_other_side(self, init_slab: Slab, index_of_sites: list[int]) -> Slab:
-        """Move a set of sites to the opposite surface of the slab.
+    def move_to_other_side(
```
DanielYang59, line 1289: There are multiple issues with this `move_to_other_side` method, ranked by severity from most to least: 

1. Wrong sign for the moving vector: 
`slab.translate_sites(top_site_index, vector=[0, 0, slab_ratio])`
where the `slab_ratio` is the slab/total_cell ratio (positive). This means the sites on the top is moved further upwards, instead of the intended downwards (the opposite side).

2. Confusing and imprecise calculation of moving vector `slab_ratio` as:
```python
  nlayers_slab: int = math.ceil(self.min_slab_size / height)
  nlayers_vac: int = math.ceil(self.min_vac_size / height)
  nlayers: int = nlayers_slab + nlayers_vac
  
  slab_ratio: float = nlayers_slab / nlayers
```
As of (1), currently the sites are moved by a vector `slab_ratio`, but using the slab/total_cell **layer** ratio is very confusing and imprecise(two roundings are included) in this use case. I would prefer just calculate the vector by the slab/cell **height ratio** instead of the **number of layers ratio**.

3. The use case of this method should be limited to **surface sites** instead of **any site** inside the Slab (need clarify in docstring). Reason being this method actually move the sites by a length equal to the thickness of the Slab, if people use it to move some site closer to the center, the site would then end up deep in the vacuum layer.

pymatgen/core/surface.py
```diff
@@ -801,8 +801,8 @@ def __init__(
 
         def reduce_vector(vector: tuple[int, int, int]) -> tuple[int, int, int]:
             """Helper function to reduce vectors."""
-            divisor = abs(reduce(gcd, vector))
-            return tuple(int(idx / divisor) for idx in vector)
+            divisor = abs(reduce(gcd, vector))  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
janosh: `# type: ignore[arg-type]` is fine here. not sure there's a good solution

pymatgen/core/surface.py
janosh: thanks for taking a deep dive on this module. such doc string clarifications are very valuable!

pymatgen/core/surface.py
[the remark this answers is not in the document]
DanielYang59: Thanks for the input. I would keep this in mind and go back to it if we could come up with a "solution". :)

pymatgen/core/surface.py
janosh: re (1), could you comment on why the tests did not catch this? could you add a new test first that demonstrates the bug?

https://github.com/materialsproject/pymatgen/blob/2a3608f31d44b7ed8c23f64b9c9bbe07f518a5ae/tests/core/test_surface.py#L561-L579

re (2): 

> I would prefer just calculate the vector by the slab/cell height ratio instead of the number of layers ratio.

would we lose backwards compat? if not, feel free to simplify/clarify the implementation

DanielYang59: > re (1), could you comment on why the tests did not catch this? could you add a new test first that demonstrates the bug?

That's a good point. I wasn't looking into the test (but I should). I would have a look into the test and update this comment. But I just tried to reverse the moving vector and the unit test still passed, so I assume the test didn't capture the issue.

> re (2):
> 
> > I would prefer just calculate the vector by the slab/cell height ratio instead of the number of layers ratio.
> 
> would we lose backwards compat? if not, feel free to simplify/clarify the implementation

I believe it would not. It's an internal "moving distance" calculation method, and I believe replacing it with another method would not bring up any compat issue (as it only affect how much the site is move when `move_to_other_side`, and current implement use the ratio of slab_layers versus total_cell_layers to determine the proportion of Slab inside the whole Cell).

DanielYang59: I decide maybe it's wise to separate such function tweaks into a separate PR. Current PR has too many changes already and it would be hard to notice any change/issue.

DanielYang59:
Can you please review this? @janosh. Sorry I'm making this PR overly large, but I don't expect functional changes within this PR (mostly cleaning up of comment/docstring/namings). 

And I would also expect (multiple) follow up PRs, as I did find some issues (all tagged, you could search `@DanielYang59` to find tags). And I think it would be easier to view the difference if I separate them with this PR.

tests/core/test_surface.py
```diff
@@ -321,7 +321,7 @@ def test_as_dict(self):
             self.zno55.miller_index,
             self.zno55.oriented_unit_cell,
             0,
-            self.zno55.scale_factor.tolist(),
```
DanielYang59, line 324: I migrated this cast inside `as_dict`:
```python
def as_dict(self, **kwargs) -> dict:  # type: ignore[override]
  """MSONable dict."""
  dct = super().as_dict(**kwargs)
  dct["@module"] = type(self).__module__
  dct["@class"] = type(self).__name__
  dct["oriented_unit_cell"] = self.oriented_unit_cell.as_dict()
  dct["miller_index"] = self.miller_index
  dct["shift"] = self.shift
  dct["scale_factor"] = self.scale_factor.tolist()  # np.ndarray is not JSON serializable
  dct["reconstruction"] = self.reconstruction
  dct["energy"] = self.energy
  return dct
```

janosh (approved):
> And I would also expect (multiple) follow up PRs, as I did find some issues (all tagged, you could search `@DanielYang59` to find tags). And I think it would be easier to view the difference if I separate them with this PR.

thanks a lot for polishing this module! doc string improvements are great for helping usability

DanielYang59:
> > And I would also expect (multiple) follow up PRs, as I did find some issues (all tagged, you could search `@DanielYang59` to find tags). And I think it would be easier to view the difference if I separate them with this PR.
> 
> thanks a lot for polishing this module! doc string improvements are great for helping usability

Thanks a lot for reviewing. This really took me a lot of energy but I'm now much much familiar with the module, which should hugely facilitate future functional tweaks.

ftherrien:
@janosh I think this PR may have removed the definition of `_calculate_possible_shifts` or was it renamed?

DanielYang59:
Hi @ftherrien thanks for asking. It's now renamed to:
https://github.com/materialsproject/pymatgen/blob/b28c9371d684b8a605830cd13060bef9accf68cf/src/pymatgen/core/surface.py#L1226-L1232
as a private function of `get_slabs` for two reasons:
- It was a private method and depended on by `get_slabs` alone.
- The original name is a bit misleading, because it doesn't really generate the shift but the termination (negative of shift).

ftherrien:
Ah! Thanks so much that makes a lot of sense

DanielYang59:
All good! Thanks for asking and happy coding!

PR #3694, opened 2024-03-18
Reformat docstrings to Google style and add type annotations

DanielYang59:
## Reformat docstrings to Google style and add type annotations

There should be no functional changes. Any change made to the code is to resolve `mypy` type errors.

Tasks:
- Corrected some already Google styled docstrings. Replaced keywords `Arg/Return/Raise` with its plural form `Args/Returns/Raises` even when there is only one `Arg/Return/Raise`. [Reference](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings).
- Replaced `reStructuredText` format by searching for keywords: `:param`, `:type`, `:returns`, `:rtype`, `::`.
- Did not find `Scipy/Numpy/Epytext` styled docstrings.


DanielYang59:
Opened another Pandora's box of `mypy` errors.

tasks.py
```diff
@@ -70,19 +76,20 @@ def make_doc(ctx):
...
+    Args:
+        ctx (Context): The context.
     """
     ctx.run("rm dist/*.*", warn=True)
     ctx.run("python setup.py sdist bdist_wheel")
     ctx.run("twine upload dist/*")
 
 
 @task
-def set_ver(ctx, version):
```
DanielYang59, line 85: Can you please help me check this to make sure removal of unused arg doesn't break any workflow or anything? @janosh Thanks.

janosh: `ctx` must be present in all `@task` decorated functions. see `invoke` docs. not sure we still use any of these functions but maybe @shyuep does and i know he likes to keep them around

pymatgen/analysis/bond_dissociation.py
```diff
@@ -110,11 +110,11 @@ def __init__(
             for bond_pair in self.bond_pairs:
                 self.fragment_and_process(bond_pair)
 
-    def fragment_and_process(self, bonds):
+    def fragment_and_process(self, bonds: list) -> None:
         """Fragment and process bonds.
 
         Args:
-            bonds (list): bonds to process.
+            bonds (list): list of bonds to process.
```
janosh, line 117: `list of ` is a bit redundant here as the type already says `list`. not a strongly held opinion but i think we can keep it shorter by not repeating `list of`

tasks.py
DanielYang59: > `ctx` must be present in all `@task` decorated functions. see `invoke` docs. 

Thanks for the input.

> not sure we still use any of these functions but maybe @shyuep does and i know he likes to keep them around

Well I don't have any intention to remove them 😃

pymatgen/analysis/bond_dissociation.py
DanielYang59: Good point. I would change these. 

Also the types of `bond` don't look well annotated. Any help would be appreciated. For example I don't quite know what is the actual type of `bond`, is doesn't look like a `CovalentBond` but a `tuple[int, int]`... 

I would need a closer look into the types.

pymatgen/analysis/graphs.py
```diff
@@ -1597,7 +1597,7 @@ def with_empty_graph(cls, molecule, name="bonds", edge_weight_name=None, edge_we
         return cls(molecule, graph_data=graph_data)
 
     @staticmethod
-    def with_edges(molecule: Molecule, edges: dict[tuple[int, int], dict]):
+    def with_edges(molecule: Molecule, edges: dict[tuple[int, int], None | dict]) -> MoleculeGraph:
```
DanielYang59, line 1600: type annotation of `edges` was likely wrong. Apparently the values could be `None`:
```python
for edge, props in edges.items():
            try:
                from_index = edge[0]
                to_index = edge[1]
            except TypeError:
                raise ValueError("Edges must be given as (from_index, to_index) tuples")

            if props is None:
                weight = None

            else:
                weight = props.pop("weight", None)
                if len(props.items()) == 0:
                    props = None
```

But still get `pymatgen/analysis/fragmenter.py:86: error: Argument 2 to "with_edges" of "MoleculeGraph" has incompatible type "dict[tuple[Any, Any], None]"; expected "dict[tuple[int, int], Optional[dict[Any, Any]]]"  [arg-type]`

pymatgen/analysis/graphs.py
```diff
@@ -1002,30 +1023,30 @@ def types_of_coordination_environments(self, anonymous=False):
...
-                labels = [(label[0], mapping[label[1]]) for label in labels]
+                sp_counts = [(label[0], mapping[label[1]]) for label in sp_counts]
 
-            labels = [f"{label[1]}({label[0]})" for label in labels]
+            labels = [f"{label[1]}({label[0]})" for label in sp_counts]
             motif = f"{centre_sp}-{','.join(labels)}"
             motifs.add(motif)
 
-        return sorted(motifs)
+        return list(set(motifs))
```
DanielYang59, line 1047: Added a `sp_counts` for "species counts" because type changes during runtime and triggered `mypy` error.

However this looks funny, maybe return `set` instead? Would using `set` break anything?

pymatgen/analysis/fragmenter.py
```diff
@@ -78,8 +82,8 @@ def __init__(
         if edges is None:
             self.mol_graph = MoleculeGraph.with_local_env_strategy(molecule, OpenBabelNN())
         else:
-            edges = {(e[0], e[1]): None for e in edges}
-            self.mol_graph = MoleculeGraph.with_edges(molecule, edges)
+            _edges = {(edge[0], edge[1]): None for edge in edges}
```
DanielYang59, line 85: Can you please give me a hand on this `mypy` error:

```python
pymatgen/analysis/fragmenter.py:86: error: Argument 2 to "with_edges" of "MoleculeGraph" has incompatible type "dict[tuple[int, int], None]"; expected "dict[tuple[int, int], Optional[dict[Any, Any]]]"  [arg-type]
pymatgen/analysis/fragmenter.py:86: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
pymatgen/analysis/fragmenter.py:86: note: Consider using "Mapping" instead, which is covariant in the value type
```

I'm not sure what is the reason. Thanks a lot.

DanielYang59: @janosh

pymatgen/analysis/graphs.py
```diff
@@ -369,6 +382,7 @@ def add_edge(
             return
 
         # sanitize types
+        # TODO(@DanielYang59) fix mypy error
```
DanielYang59, line 385: Not sure why I'm getting this following error:
```
pymatgen/analysis/graphs.py:386: error: Incompatible types in assignment (expression has type "tuple[int, ...]", variable has type "tuple[int, int, int]")  [assignment]
pymatgen/analysis/graphs.py:386: error: Incompatible types in assignment (expression has type "tuple[int, ...]", variable has type "Optional[tuple[int, int, int]]")  [assignment]
```

A simple code to recreate this:
```python
t: tuple[int, int, int] = (1, 0, 0)

t = tuple(map(int, t))
```
would trigger the same error.

DanielYang59:
I guess it's not a good idea to reformat docstring and add type annotations (for modules I'm not quite familiar with especially) at the same time. I should focus on reformatting docstring for this PR and stop adding type annotations.

I tried my best to fix most `mypy` error but still get quite some left (some even seems like programming bugs).

Could you please give me a hand when you have time? @janosh. Thanks a ton in advance. 😄

pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
```diff
@@ -498,10 +528,10 @@ class SimplestChemenvStrategy(AbstractChemenvStrategy):
...
 
     STRATEGY_DESCRIPTION = (
-        "    Simplest ChemenvStrategy using fixed angle and distance parameters \n"
-        "    for the definition of neighbors in the Voronoi approach. \n"
-        "    The coordination environment is then given as the one with the \n"
-        "    lowest continuous symmetry measure."
+        "Simplest ChemenvStrategy using fixed angle and distance parameters \n"
+        "for the definition of neighbors in the Voronoi approach. \n"
+        "The coordination environment is then given as the one with the \n"
+        "lowest continuous symmetry measure."
```
DanielYang59, line 534: Do want have a good reason to indent this description? The unit test seems to be expecting indentation, if not I would remove the indentation from unit test as well.

pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
```diff
@@ -855,10 +898,10 @@ class SimpleAbundanceChemenvStrategy(AbstractChemenvStrategy):
...
     )
     STRATEGY_DESCRIPTION = (
-        '    Simple Abundance ChemenvStrategy using the most "abundant" neighbors map \n'
-        "    for the definition of neighbors in the Voronoi approach. \n"
-        "    The coordination environment is then given as the one with the \n"
-        "    lowest continuous symmetry measure."
+        'Simple Abundance ChemenvStrategy using the most "abundant" neighbors map \n'
+        "for the definition of neighbors in the Voronoi approach. \n"
+        "The coordination environment is then given as the one with the \n"
+        "lowest continuous symmetry measure."
```
DanielYang59, line 904: Same here, is indentation necessary?

pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
JaGeo: @davidwaroquiers ?

davidwaroquiers: Fine with me to remove indentation.

pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
davidwaroquiers: Same, fine to remove indentation.

pymatgen/analysis/graphs.py
```diff
@@ -739,6 +760,8 @@ def get_connected_sites(self, n, jimage=(0, 0, 0)):
             site = PeriodicSite.from_dict(site_d)
 
             # from_site if jimage arg != (0, 0, 0)
+            # TODO (@DanielYang59): distance should only take one arg
+            # DEBUG
```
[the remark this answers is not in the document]
janosh: this is `mypy` not understanding `site` is a `PeriodicSite` which takes `jimage` in `distance`

janosh:
thanks so much @DanielYang59! this looks like a lot of work! `mypy` is happy now 26a2c8c :)

pymatgen/analysis/fragmenter.py
DanielYang59: No worries. Any help would be appreciated!

pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
```diff
@@ -30,6 +30,9 @@
 from pymatgen.core.sites import PeriodicSite
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
 
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
```
janosh, line 35: sneaky :)

`typing_extensions` is actually a [PyPI package](https://pypi.org/project/typing-extensions/) but given we never import it at runtime, looks like we can actually get away with not listing it as dependency and installing it in CI.
still, i think that might be bad practice as contributors who want to run type checks and don't have it installed will get errors (i assume). so maybe better to add to optional deps

DanielYang59: Good catch. I'm not even aware it's not a built-in package. We should be able to native `Self` support after Python3.11 but it looks we still need a legacy `typing_extensions` for now. Or I could still use the name of the Class itself instead of `Self`(but apparently `Self` is more proper), if you decide adding an extra optional dep is overkill for this.

After fixing it this PR should be good. Thanks!

janosh: i'm a fan of `Self` and i don't mind depending on `typing_extensions` until we reach oldest supported Python = 3.11.

but maybe it would be better to add it in a separate PR and then use `Self` on _all_ methods where it applies?

DanielYang59: I help do this but I guess reverting changes in this PR would pointless then? Maybe add an optional dep here and I would do the following later?

janosh: yes, that works

DanielYang59: Thanks. I would do that soon.

DanielYang59: (I'm hoping I'm not adding it to the wrong place)

janosh: that's correct. i'm still not sure what happens to people who use an IDE with type hint support that don't have `typing_extensions` installed. could be that their IDE doesn't understand `Self` but would have understood the `NameOfClass` hint. could be we're removing some type assistance in which case to fix that we would even need to make it a real dependency. but this is all just me guessing. and either way, it doesn't seem like a big loss

tasks.py
janosh, line 1: type hints in `invoke` functions used to be a problem but [looks like they fixed that in v2](https://github.com/pyinvoke/invoke/issues/357#issuecomment-1374042624)

janosh (approved):
thanks a lot @DanielYang59! 👍

pymatgen/analysis/fragmenter.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         if edges is None:
             self.mol_graph = MoleculeGraph.with_local_env_strategy(molecule, OpenBabelNN())
         else:
-            _edges = {(edge[0], edge[1]): None for edge in edges}
+            _edges: dict[tuple[int, int], dict | None] = {(edge[0], edge[1]): None for edge in edges}
```
DanielYang59, line 85: This looks like magic to me. Didn't expect declaring the a super set of its actual type would solve the problem.

DanielYang59:
> thanks a lot @DanielYang59! 👍

Thanks for reviewing! Glad I can help.

pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
DanielYang59: I'm working with VS Code and I tried to uninstall `typing_extensions` and everything seems to work just fine:
```bash
(venv) yang@Yang-PC:~/pymatgen$ pip show typing_extensions
WARNING: Package(s) not found: typing_extensions
(venv) yang@Yang-PC:~/pymatgen$ pre-commit run --all-files
check yaml...............................................................Passed
mypy.....................................................................Passed
cython-lint..............................................................Passed
double-quote Cython strings..............................................Passed
markdownlint.............................................................Passed
nbstripout...............................................................Passed
```

Meanwhile I'm pretty sure I didn't manually install `typing_extensions`, and it seems bundled with `venv` for some reason.

DanielYang59:
Ahh. I forgot to mention this but it seems we need to rebuild the `docs` after so many changes? @janosh

PR #3697, opened 2024-03-20
Prevent GitHub Actions from running document related CIs on forks

lan496:
`.github/workflows/jekyll-gh-pages.yml` run on forked repository, and caused CI fails if the forked repository is not set up  for GitHub Pages (e.g. https://github.com/lan496/pymatgen/actions/runs/8325702971).
As the CI for testing is already disabled on forked (https://github.com/materialsproject/pymatgen/commit/531561d8d1dd55b3104d923c585861340c2223f0), I think it will be reasonable to disable CI for Github Pages as well.

### DIFF
--- a/.github/workflows/jekyll-gh-pages.yml
+++ b/.github/workflows/jekyll-gh-pages.yml
@@ -24,6 +24,8 @@ concurrency:
 jobs:
   # Build job
   build:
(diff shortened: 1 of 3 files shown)


janosh (approved):
thanks @lan496, makes perfect sense

PR #3700, opened 2024-03-22
Add more flexibility to `PhononDOSPlotter` and `PhononBSPlotter`

ab5424:
## Summary

Major changes:

- feature 1: Add `invert_axes` to `PhononDOSPlotter` (similar to `DOSPlotter`)
- feature 2: Compare to more than one other `PhononBSPlotter`

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Added an `invert_axes` parameter to various plotting methods for axis inversion, enhancing plot customization.
	- Enhanced color handling and customization options for comparing multiple plots.
- **Refactor**
	- Renamed a parameter for clarity and adjusted plot dimensions based on new axis inversion feature.
- **Tests**
	- Updated plotting tests to reflect changes in plotting methods and configurations, including the use of `matplotlib.pyplot`.


pymatgen/phonon/plotter.py
```diff
@@ -590,7 +619,7 @@ def get_ticks(self) -> dict[str, list]:
 
     def plot_compare(
         self,
-        other_plotter: PhononBSPlotter,
+        other_plotter: PhononBSPlotter | Sequence[PhononBSPlotter],
```
janosh, line 622: how about we use `dict[str, PhononBSPlotter]` here so we can label other band structures? would be great to add a legend mapping colors to dict keys.

[see `pymatviz` for what that might look like](https://github.com/janosh/pymatviz/blob/49588d1cce3186322fae1a7c6d4af0211dbfedb2/pymatviz/phonons.py#L108)
phonon-bands-mp-2758

ab5424: There is also `labels` a few lines below, doesn't that do the same? I could refactor though.

janosh: yeah i think refactoring would make for a better API here. `labels` was only added 3 months in https://github.com/materialsproject/pymatgen/commit/9268942bb28076c53761882effd4cdb2ba65ec79 ago so i think even a breaking change would be acceptable

ab5424: I removed `labels`. I changed the logic so that no legend is shown if the Plotter is initialised without a label (i.e. if the "original" PhononBS has no label).

janosh: it's important to still support passing in single `PhononBSPlotters`. made some changes to that effect in 
26fb878. also i think if there are multiple structures, we should always show a legend

ab5424: I added a test for the single `PhononBSPlottter` case.

ab5424: @janosh Is there a best way to add custom colours for the comparison? With `PhononDOSPlotter`, it is done via `add_dos()`, but here it is not obvious to me. Is it better to add the colour to the init or passing multiple colours to `plot_compare`?

janosh: definitely `plot_compare`. usually it's best for arguments to be as close to where they're used as possible

ab5424: I added a `colors` variable to `plot_compare()`.

janosh (approved):
thanks a lot @ab5424, nice work!

janosh:
@ab5424 actually, maybe these two CodeRabbit comments are worth addressing. would you be able to add these checks to `tests/phonon/test_plotter.py`

> 46-49: Testing the invert_axes feature by asserting the swapped labels is a good approach. Additionally, consider verifying that the axes' limits are correctly inverted to ensure the feature works as expected.

> 89-101: Testing the plot_compare method with different scenarios and asserting the legend labels and colors is thorough. It ensures that the comparison feature works correctly for multiple cases. Consider adding a test to verify the handling of incompatible band structures if not already covered.

ab5424:
@janosh I changed the code according to the suggestions. Interestingly, I found that I was able to get the NaCl/STO comparison though the checking block because the distances lists had the same length (but not the same shape so there was an `IndexError` on line 691). I changed it so that the shape must be the same.

PR #3702, opened 2024-03-22
`Self` return type on `from_dict` methods

janosh:
follow up to @DanielYang59's start in #3694 of adding `Self` return types on `from_dict` methods. far from complete. should be set on `from_str` and other `from_...` methods as well


DanielYang59:
Great work. I can take care of the rest if you don't mind.

Meanwhile I noticed at least two modules are using naming conventions as `with_something` instead of `from_something`, do we have any plan to unify the namings as well?

image

janosh:
good catch. i think we can rename those to `from_...` with a year-long deprecation period.

DanielYang59:
Sure. I would do that.

janosh:
while we're at it, these classes should be renamed to `PascalCase` (e.g. `BrunnerNNReciprocal`) and also be deprecated. you can find them with this regex: `class [^(= ]+_[^(]+\(`

Screenshot 2024-03-23 at 06 42 45

DanielYang59:
Sure would do. 

Can you please rebuild the `docs` (should be quite outdated) as we changed quite a lot in #3694? 🌹

janosh:
`invoke make-doc` doesn't do a whole lot. not sure how doc updating is supposed to work atm. @shyuep?

Screenshot 2024-03-23 at 06 50 14

DanielYang59:
Did you install `sphinx` dependencies? Like `sphinx_markdown_builder` and `sphinx_rtd_theme`? I got quite a lot changes.

image

janosh:
i did have those installed. i just updated `sphinx_markdown_builder` and `sphinx_rtd_theme` to v2 and ran again. that seems to have helped. still not many meaningful changes though. a few changed line ranges mostly https://github.com/materialsproject/pymatgen/commit/af0425827650c05692882bacc0fff6646115f859

DanielYang59:
Great to know. Meanwhile I noticed this warning from 3 years ago #2149:
https://github.com/materialsproject/pymatgen/blob/af0425827650c05692882bacc0fff6646115f859/pymatgen/analysis/interface.py#L18-L23

I assume it might be old enough to be removed?

PR #3704, opened 2024-03-24
Support appending vectors to positions in XSF format

mturiansky:
## Summary

XSF allows you to append a vector after the positions (usually forces). This makes it easier to visualize phonons with VESTA.

## Checklist


janosh (approved):
thanks @mturiansky! 👍

PR #3705, opened 2024-03-24
Use `Self` type in Method Signatures

DanielYang59:
## Summary

Use `Self` for return type in Method Signatures, see reference [here](https://typing.readthedocs.io/en/latest/spec/generics.html#self).

### Major changes:

- Use `Self` type in Method Signatures
- Add related type annotations
- Some `sourcery` fixes

### Major changes:
- Docstring tweaks

## Summary by CodeRabbit

- **New Features**
  - Enhanced search functionality with more intuitive variable names and clearer method signatures.
  - Improved clarity and consistency in documentation across various modules.
- **Refactor**
  - Updated method return types to leverage `Self` for better type hinting and future maintainability.
  - Modified method signatures for clarity and consistency.
- **Chores**
  - Updated pre-commit hook configurations for improved code quality checks.
- **Style**
  - Clarified and condensed docstrings for better readability.
- **Bug Fixes**
  - Fixed potential issues with handling `None` values in data processing methods.


pymatgen/alchemy/transmuters.py
```diff
@@ -308,7 +310,7 @@ def from_filenames(poscar_filenames, transformations=None, extend_collection=Fal
         """
         trafo_structs = []
         for filename in poscar_filenames:
-            with open(filename) as file:
+            with open(filename, encoding="utf-8") as file:
                 trafo_structs.append(TransformedStructure.from_poscar_str(file.read(), []))
         return StandardTransmuter(trafo_structs, transformations, extend_collection=extend_collection)
```
DanielYang59, line 315: Need a closer look: this returns its super class, which differs from method docstring.

pymatgen/core/composition.py
```diff
@@ -25,7 +25,10 @@
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterator
 
-SpeciesLike = Union[str, Element, Species, DummySpecies]
```
DanielYang59, line 28: Make sure this `SpeciesLike` is the same as `from pymatgen.util.typing import SpeciesLike`

pymatgen/core/ion.py
```diff
@@ -19,15 +23,15 @@ class Ion(Composition, MSONable, Stringify):
     Mn[+2]. Note the order of the sign and magnitude in each representation.
     """
 
-    def __init__(self, composition, charge=0.0, _properties=None) -> None:
+    def __init__(self, composition: Composition, charge: float=0.0, _properties=None) -> None:
```
DanielYang59, line 26: `_properties` is not used. Any good reason to keep it here?

pymatgen/core/structure.py
```diff
@@ -177,12 +177,12 @@ def __getitem__(self, idx: int | slice):
         """Make neighbor Tuple-like to retain backwards compatibility."""
         return (self, self.nn_distance, self.index, self.image)[idx]
 
-    def as_dict(self) -> dict:  # type: ignore
+    def as_dict(self) -> dict:  # type: ignore[override]
```
DanielYang59, line 180: I guess use of "too general" `# type: ignore` should be avoided wherever possible.

janosh:
[i added `Self` return type to `__new__` and `__enter__` methods](https://github.com/materialsproject/pymatgen/pull/3705/commits/2adde295d16126b84e4f346ff3294420003c9069). created 66 `mypy` errors unfortunately so feel free to revert if we don't want to deal with this now

pymatgen/io/lammps/data.py
```diff
@@ -1373,13 +1375,14 @@ def disassemble(
             for mol in self.mols
         ]
 
+    # TODO(@DanielYang59): should the following two methods be removed?
```
DanielYang59, line 1378: Tag: The following two methods do nothing but raise an error. Should they be removed?

DanielYang59:
> [i added `Self` return type to `__new__` and `__enter__` methods](https://github.com/materialsproject/pymatgen/pull/3705/commits/2adde295d16126b84e4f346ff3294420003c9069). created 66 `mypy` errors unfortunately so feel free to revert if we don't want to deal with this now

Thanks for your help! And don't worry about the `mypy` errors. I would try to fix as many as I can later.

pymatgen/io/lammps/data.py
janosh: seems like those methods are needed to override parent class `LammpsData`

https://github.com/materialsproject/pymatgen/blob/ca8f608dec81743d506f365e91efd3bf76d468cf/pymatgen/io/lammps/data.py#L740-L748

pymatgen/io/adf.py
```diff
@@ -85,21 +85,15 @@ def __init__(self, name, options=None, subkeys=None):
         """
         Initialization method.
 
-        Parameters
-        ----------
-        name : str
-            The name of this key.
```
DanielYang59, line 91: Still noticed some non-Google styled docstrings. There might be some hidden format I'm not aware of. (Thought I searched keywords `Parameters` for some reason.)

janosh: yeah, i noticed them too. `Args:` would be better. likewise

https://github.com/materialsproject/pymatgen/blob/ca8f608dec81743d506f365e91efd3bf76d468cf/pymatgen/io/adf.py#L186-L190

```diff
  Returns:
- -------
```

DanielYang59: I tried to fix as much as I could find in e0c1366553c6a9642bb5f0e0addccbec85745705 and 48a8fd53858f6dec1033c4f80e8b2347e560789b 😄

pymatgen/core/composition.py
DanielYang59: Confirmed: https://github.com/materialsproject/pymatgen/blob/b1e50237cfd90a68440fb82806b58e59db2a4d5e/pymatgen/util/typing.py#L24

pymatgen/io/adf.py
janosh: looks good!

pymatgen/core/ion.py
janosh: not sure. maybe for `monty` `as/from_dict` compatibility? but since there's no comment explaining it, i'm open to removing it to see what happens

pymatgen/io/lammps/data.py
DanielYang59: Thanks for the input. I put a comment there (with your name, because you gave the info) in case some other people ask

pymatgen/io/adf.py
DanielYang59: Thanks. Let me know if I'm missing any.

pymatgen/core/ion.py
DanielYang59: Removed in 11a53c4b26a39bc686101dc888b0f680d9e36ab1. I would keep an eye this as well.

DanielYang59:
Naming of these two properties is a bit funny: https://github.com/materialsproject/pymatgen/blob/b1e50237cfd90a68440fb82806b58e59db2a4d5e/pymatgen/analysis/reaction_calculator.py#L455-L482

Deprecate them and use `reaction_energy(_uncertainty)` instead?

pymatgen/core/units.py
```diff
@@ -291,38 +294,7 @@ class FloatWithUnit(float):
 
     Error = UnitError
```
DanielYang59, line 295: Not sure if this exists for a good reason?

janosh:
> Deprecate them and use `reaction_energy(_uncertainty)` instead (in the future `deprecation` PR for `with_something`)?

i agree the naming is verbose but i would probably keep them as is to minimize breaking changes

DanielYang59:
> i would probably keep them as is to minimize breaking changes

Sure thing! Thanks for helping me catch more docstrings.

DanielYang59:
Naming aside, need a closer look at the `std_dev` in: https://github.com/materialsproject/pymatgen/blob/b1e50237cfd90a68440fb82806b58e59db2a4d5e/pymatgen/analysis/reaction_calculator.py#L469-L482

`self.calculate_energy(calc_energies)` should be a `float`, why `self.calculate_energy(calc_energies).std_dev` would even work...

Also the `mypy` error: 
```
pymatgen/analysis/reaction_calculator.py:484: error: "float" has no attribute "std_dev"  [attr-defined]
```

janosh:
> `self.calculate_energy(calc_energies)` should be a `float`, why `self.calculate_energy(calc_energies).std_dev` would even work ([0aba332](https://github.com/materialsproject/pymatgen/commit/0aba332b272cc7d122c54b822c534fcce0f7032e))...

it's not a regular float but a `ufloat` from the [`uncertainties` package](https://pythonhosted.org/uncertainties/user_guide.html)

DanielYang59:
> > `self.calculate_energy(calc_energies)` should be a `float`, why `self.calculate_energy(calc_energies).std_dev` would even work ([0aba332](https://github.com/materialsproject/pymatgen/commit/0aba332b272cc7d122c54b822c534fcce0f7032e))...
> 
> it's not a regular float but a `ufloat` from the [`uncertainties` package](https://pythonhosted.org/uncertainties/user_guide.html)

Ahh, thanks a lot for the quick response. Thought I would investigate tomorrow (22:21 in my time zone). Maybe I was adding the wrong type as `float` instead of `ufloat`. Would fix them tomorrow. Thanks!

pymatgen/analysis/adsorption.py
```diff
@@ -19,11 +19,12 @@
 from pymatgen.analysis.structure_matcher import StructureMatcher
 from pymatgen.core import Molecule, Structure
 from pymatgen.core.operations import SymmOp
-from pymatgen.core.surface import generate_all_slabs
+from pymatgen.core.surface import Slab, generate_all_slabs
```
DanielYang59, line 22: This looks like a typo? `Slab` should be imported only `if TYPE_CHECKING`?

janosh: right, feel free to change

pymatgen/analysis/ewald.py
```diff
@@ -619,7 +619,7 @@ def best_case(self, matrix, m_list, indices_left):
         interaction_correction = np.sum(step3)
 
         if self._algo == self.ALGO_TIME_LIMIT:
-            elapsed_time = datetime.datetime.now(datetime.timezone.utc) - self._start_time
+            elapsed_time = datetime.now(timezone.utc) - self._start_time
```
DanielYang59, line 622: This change need to be reverted unless the following issue can be resolved:
```python
import datetime

start_time_old = datetime.datetime.utcnow()  # 2024-03-25 08:46:23.748342 

start_time_new = datetime.datetime.now(datetime.UTC)  # 2024-03-25 08:46:23.748472+00:00

print(start_time_old, start_time_new)  


dt = datetime.datetime.strptime(
    str(start_time_old), "%Y-%m-%d %H:%M:%S.%f"
)
print(dt)


dt = datetime.datetime.strptime(
    str(start_time_new), "%Y-%m-%d %H:%M:%S.%f"
)
print(dt)
```

Which raises:
```python
  File "/Users/yang/developer/test/test.py", line 16, in <module>
    dt = datetime.datetime.strptime(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.2_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/_strptime.py", line 554, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "/opt/homebrew/Cellar/python@3.12/3.12.2_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/_strptime.py", line 336, in _strptime
    raise ValueError("unconverted data remains: %s" %
ValueError: unconverted data remains: +00:00
```

pymatgen/analysis/adsorption.py
```diff
@@ -599,11 +600,11 @@ def get_mi_vec(slab):
     return mvec / np.linalg.norm(mvec)
 
 
-def get_rot(slab):
+def get_rot(slab: Slab) -> SymmOp:
     """Gets the transformation to rotate the z axis into the miller index."""
     new_z = get_mi_vec(slab)
-    a, _b, _c = slab.lattice.matrix
-    new_x = a / np.linalg.norm(a)
```
DanielYang59, line 605: I would revert this too @janosh . As `a_len = slab.lattice.a` gives the length of lattice vector a, and `a, _b, _c = slab.lattice.matrix` extracted the vector a itself. This leads to multiple unit test errors.

pymatgen/analysis/ewald.py
janosh: if it's just there to measure elapsed time then it's not using the right API anyway and needlessly complicated as a result. have a look at [`time.perf_counter`](https://docs.python.org/3/library/time.html#time.perf_counter) instead

pymatgen/analysis/adsorption.py
janosh: oops, my mistake

pymatgen/analysis/ewald.py
DanielYang59: Thanks I would have a closer look, and perhaps report this to `Python` repo, which is quite similar to [this](https://github.com/python/cpython/issues/109964). 

It should have been a simple deprecation replacement (`datetime.datetime.utcnow()` -> `datetime.datetime.now(datetime.UTC)`). But for some reason these two doesn't return the exact same value (with an additional `+00:00` at the end from the new API), which leads to format error.

pymatgen/analysis/adsorption.py
DanielYang59: No worries at all. Feel free to push some related commit like #3706 to this PR, and I would (try my best to) take care of the `mypy` error later.

The thing I noticed is that the default value of `None` is triggering quite a lot of `mypy` errors, for example:
```python
pymatgen/io/exciting/inputs.py:86: error: Item "None" of "Optional[Element]" has no attribute "text"  [union-attr]
pymatgen/io/exciting/inputs.py:89: error: Item "None" of "Optional[str]" has no attribute "split"  [union-attr]
pymatgen/io/exciting/inputs.py:98: error: Item "None" of "Optional[str]" has no attribute "split"  [union-attr]
pymatgen/io/exciting/inputs.py:104: error: Item "None" of "Optional[str]" has no attribute "split"  [union-attr]
pymatgen/io/exciting/inputs.py:123: error: Item "None" of "Optional[Element]" has no attribute "get"  [union-attr]
pymatgen/io/exciting/inputs.py:126: error: Item "None" of "Optional[Element]" has no attribute "get"  [union-attr]
pymatgen/io/exciting/inputs.py:129: error: Item "None" of "Optional[Element]" has no attribute "iter"  [union-attr]
pymatgen/io/exciting/inputs.py:131: error: Item "None" of "Union[str, None, Any]" has no attribute "split"  [union-attr]
```

This is quite similar to this: https://github.com/materialsproject/pymatgen/blob/05d3f9f5643c9585e15b297a58697432c9de4dcf/pymatgen/io/vasp/optics.py#L175-L187

But I guess applying `None` check is better than just suppressing the error messages (which would break the code if the args are truly `None`).

janosh:
i think i would be good to open a `monty` PR that adds `Self` return type to `MSONable.from_dict` as well

pymatgen/analysis/ewald.py
DanielYang59: I decide to revert this deprecation change for now until I could figure out a way to avoid `datetime.datetime.now(datetime.UTC)` from output timezone info.

DanielYang59:
> i think i would be good to open a `monty` PR that adds `Self` return type to `MSONable.from_dict` as well

I just had a quick look and realized `monty` is missing quite some type annotations. I think it might be good to add type annotations for the entire code base (Given `monty` is way smaller in size than PMG.)? I can do this after I finished this PR, or do you want to do that yourself? 😄

janosh:
> I can do this after I finished this PR, or do you want to do that yourself? 😄

that would be nice. you could probably save some time by having an LLM like ChatGPT or Claude do the first pass and then fix the mistakes manually

DanielYang59:
>  you could probably save some time by having an LLM like ChatGPT or Claude do the first pass and then fix the mistakes manually

Thanks for the suggestion. I usually use LLM for docstring or descriptive stuff. Currently I'm still not confident about their capability to infer var types. But currently, you know, I need to train myself to consume code 😄 . (With this said, the NVIDIA GTC 2024 keynote is gorgeous)

pymatgen/io/core.py
```diff
@@ -255,11 +257,12 @@ class InputGenerator(MSONable):
     """
 
     @abc.abstractmethod
-    def get_input_set(self) -> InputSet:
+    def get_input_set(self, *args, **kwargs) -> Self:
```
DanielYang59, line 260: It seems there is still a unit test failure related to this @janosh

janosh: thanks for fixing!

DanielYang59: It seems our work is overlapping. My fix in 7b8227420f244cee4c468a05b1e9efbee397a699 does not need a `ignore` 😃

DanielYang59: Half done. Half to go.

DanielYang59: I guess you have misunderstood... I mean 8583e39b10cc8aaeeba98f1ff880817962f2e14f seems to have broken one of the unit tests (split 6), do you have time to have a look? thanks

janosh: yes, i'll fix it

DanielYang59: Appreciate that!

pymatgen/io/zeopp.py
```diff
@@ -117,7 +117,8 @@ def from_str(cls, string: str) -> Self:
         lengths.insert(0, a)
         alpha = angles.pop(-1)
         angles.insert(0, alpha)
-        lattice = Lattice.from_parameters(*lengths, *angles)
```
DanielYang59, line 114: Not sure if there is a better way out. It seems we have multiple `mypy` complaints related to unpacking args.

pymatgen/io/zeopp.py
```diff
@@ -117,7 +117,8 @@ def from_str(cls, string: str) -> Self:
...
+        lattice = Lattice.from_parameters(*lengths, *angles)  # type: ignore[arg-type]
```
DanielYang59, line 120: Not sure if there is a better way out. We're having multiple `mypy` complains related to unpacking args.

pymatgen/io/core.py
DanielYang59: Fixed in 2d5e0bb889ab7ccdfd0b7d3c5c65c8333145ed07. Seems like you got a typo in replacing the `LammpsInputSet` class name in:https://github.com/materialsproject/pymatgen/blob/0f3a938ed83fb4a831771904807cb55d39b39895/pymatgen/io/lammps/generators.py#L34-L82

pymatgen/io/vasp/inputs.py
```diff
@@ -2674,14 +2674,14 @@ def from_directory(cls, input_dir: str, optional_files: dict | None = None) -> S
...
-                sub_dct[fname.lower()] = ftype.from_file(full_zpath)
+                sub_dct[fname.lower()] = ftype.from_file(full_zpath)  # type: ignore[attr-defined]
             except FileNotFoundError:  # handle the case where there is no KPOINTS file
                 sub_dct[fname.lower()] = None
 
         sub_dct["optional_files"] = {}
         if optional_files is not None:
             for fname, ftype in optional_files.items():
-                sub_dct["optional_files"][fname] = ftype.from_file(os.path.join(input_dir, fname))
+                sub_dct["optional_files"][fname] = ftype.from_file(os.path.join(input_dir, fname))  # type: ignore[attr-defined]
```
DanielYang59, line 2684: Wondering if there is a better solution here

pymatgen/io/vasp/inputs.py
```diff
@@ -1384,18 +1384,25 @@ def from_str(cls, string: str) -> Self:
...
 
             return cls.gamma_automatic(kpts, kpts_shift) if style == "g" else cls.monkhorst_automatic(kpts, kpts_shift)
 
         # Automatic kpoints with basis
         if num_kpts <= 0:
             _style = cls.supported_modes.Cartesian if style in "ck" else cls.supported_modes.Reciprocal
-            kpts_shift = [float(i) for i in lines[6].split()]
+            _kpts_shift = tuple(float(i) for i in lines[6].split())
+            if len(_kpts_shift) == 3:
+                kpts_shift = _kpts_shift
```
DanielYang59, line 1404: This fix looks terrible. A cleaner fix would be much appreciated.

DanielYang59:
@janosh Could you please give me a hand? There is only eight (hardest) `mypy` errors left.

DanielYang59:
The unit test of `Kpoints` is inconsistent with type annotation (therefore the unit test failure), mostly `tuple` and `list` cannot be used interchangeably after adding type annotations. Need discussion on this as I'm not sure if any internal mechanism has any requirement.

pymatgen/analysis/reaction_calculator.py
```diff
@@ -240,7 +240,7 @@ def as_entry(self, energies) -> ComputedEntry:
         Returns a ComputedEntry representation of the reaction.
         """
         relevant_comp = [comp * abs(coeff) for coeff, comp in zip(self._coeffs, self._all_comp)]
-        comp = sum(relevant_comp, Composition())
+        comp: Composition = sum(relevant_comp, Composition())  # type: ignore[assignment]
```
DanielYang59, line 243: Not sure if `sum` could really return `Composition`, please help me double check.

janosh: yes it can. but `sum` has the wrong type hints in `mypy` restricting it to `int | float`. using `# type: ignore[assignment]` like you did is best

pymatgen/io/vasp/inputs.py
janosh: thanks! i'll take a look at the last error and this fix tomorrow

pymatgen/analysis/reaction_calculator.py
DanielYang59: Thanks for the input!

pymatgen/io/vasp/inputs.py
DanielYang59: That would be much appreciated. Previous error in 8583e39b10cc8aaeeba98f1ff880817962f2e14f has been fixed in 2d5e0bb889ab7ccdfd0b7d3c5c65c8333145ed07. The only error is related to this (`Kpoint`) as well, see [my comments](https://github.com/materialsproject/pymatgen/pull/3705#issuecomment-2024658023).

.pre-commit-config.yaml
```diff
@@ -25,6 +25,7 @@ repos:
     rev: v1.9.0
     hooks:
       - id: mypy
+        additional_dependencies: [types-python-dateutil]
```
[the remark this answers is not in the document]
janosh: i think it's better to refactor `ResParser._parse_date` to use standard lib `datetime` functionality instead. `dateutil` was never added as a `pymatgen` dependency so the only reason this worked is because some other `pmg` dependency already uses `dateutil`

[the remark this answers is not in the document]
DanielYang59: Thanks! Wonderful suggestion. Sorry I wasn't thinking about the code itself much (as there were too many `mypy` errors).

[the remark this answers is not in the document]
janosh: i definitely know the feeling. sometimes there are so many errors, you can't devote full attention to everyone. it's usually good to have a second look with fresh eyes in those cases 😄

DanielYang59:
Everything looks good! Only two minor comments [here](https://github.com/materialsproject/pymatgen/pull/3705#discussion_r1542264987) and [here](https://github.com/materialsproject/pymatgen/pull/3705#discussion_r1541061338).

Thanks a lot for helping me @janosh 

Sorry I'm making this PR too large. I would try to chop up future PRs into smaller chunks.

janosh (approved):
thank you so much @DanielYang59! really impressed how quickly you pulled off this major effort! big type safety improvement and hugely appreciated.

this brings us a step closer to using a type checker to error on possibly undefined variables as suggested by @ml-evs in #3646.

.pre-commit-config.yaml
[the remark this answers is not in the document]
DanielYang59: Yes. That's the reason I comment here and there so that me (and others) could have a closer look at those I'm unsure about. Thanks!

DanielYang59:
> thank you so much @DanielYang59! 

Glad I could help (as always)... Can never be done without your help. ❤️

pymatgen/io/vasp/inputs.py
janosh: my first though was to define a protocol

```py
from typing import Protocol

class FromFileable(Protocol):
    @classmethod
    def from_file(cls, filename: str) -> Self: ...
```

looks like this solves the problem for `pylance` (a different slightly better type checker from microsoft) but `mypy` doesn't seem to support this

janosh: i don't think there's a better solution

DanielYang59: Thanks. But that `Protocol` is certainly a good tool to have!

pymatgen/io/zeopp.py
DanielYang59: Last unresolved here @janosh.

Sorry I'm making this PR too large to navigate.

janosh: > Sorry I'm making this PR too large to navigate.

all good! there's a lot of legacy stuff to clean up in `pymatgen` mkaing large PRs necessary

[i think i found a cleaner solution to this problem](https://github.com/materialsproject/pymatgen/pull/3705/commits/6d06bef4eb10a9c1e3062c94fa8001757b686d39#diff-5215f3d451242a8bcb7a19dde68e5afa7583bafc7a598039da82f03f9fd91c77R322)

pymatgen/core/lattice.py
```diff
@@ -319,6 +319,7 @@ def from_parameters(
         alpha: float,
         beta: float,
         gamma: float,
+        *,  # help mypy separate positional and keyword-only arguments
```
DanielYang59, line 322: Thanks! New trick learned. (There should be another similar `mypy` suppressed by me somewhere)

DanielYang59: Thanks! I was about to remove that tag myself.

janosh:
> Looks like some internal error for the Windows runner.

very strange. did not notice this before in other PRs. if i see it a 2nd time I'll do some digging

pymatgen/io/zeopp.py
DanielYang59: Yes [I did notice that](https://github.com/materialsproject/pymatgen/pull/3705#discussion_r1544278631). But frankly I have never seen such usage before. Hopefully `mypy` could have a "smarter" way to detect this in the near future (otherwise we would need to do the same for all unpacking of args).

DanielYang59:
Cheers!

janosh:
with all the type fixes in this PR, we can probably consider unskipping `mypy` in CI. i.e. remove this line

https://github.com/materialsproject/pymatgen/blob/3a49dd918db2b64c48493ac6e1eb837d65930d86/.pre-commit-config.yaml#L5

DanielYang59:
> we can probably consider unskipping `mypy` in CI

Ahh... Frankly I don't know how would this potentially alter (impact) CI. I thought `mypy` has always been running in CI?

janosh:
> I thought `mypy` has always been running in CI?

it has but not in `pre-commit.ci`, only in GitHub actions. for some reason ([maybe different set of CLI args](https://github.com/pre-commit/mirrors-mypy/blob/9db9854e3041219b1eb619872a2dfaf58adfb20b/.pre-commit-hooks.yaml#L7)), the former reported more errors in the past which is why i skipped it. but i think we can try again now to see how many errors remain

DanielYang59:
Speaking of this. I might continue to add type annotations across the code base in the near future. But I guess it's better to start from the most used parts (`core` and `vasp` I guess?) instead of alphabetically, do you have any suggestions?

DanielYang59:
> it has but not in `pre-commit.ci`, only in GitHub actions. for some reason ([maybe different set of CLI args](https://github.com/pre-commit/mirrors-mypy/blob/9db9854e3041219b1eb619872a2dfaf58adfb20b/.pre-commit-hooks.yaml#L7)), the former reported more errors in the past which is why i skipped it. but i think we can try again now to see how many errors remain

Thanks for clarifying. This makes sense to me, but I guess it's better to run locally with the same args, instead of risk breaking everyone's CI workflow? I'm having a try now.

janosh:
instead of risk breaking everyone's CI workflow?

oh don't worry, `skip` applies only to `ci`. `mypy` was only skipped in `pre-commit.ci`

Screenshot 2024-03-29 at 13 02 22

> Speaking of this. I might continue to add type annotations across the code base in the near future. But I guess it's better to start from the most used parts (core and vasp I guess?) instead of alphabetically, do you have any suggestions?

that would be very welcome! `core` is definitely most important. `vasp`, `electronic_structure`, `phonons` are also important

DanielYang59:
I think I don't understand the difference between CI and `pre-commit` properly (currently doing some research on this). From my current understanding, `pre-commit` seems to check modified file for a PR only, right?

On the other hand, running `mypy pymatgen` gives `Found 732 errors in 96 files (checked 278 source files)`................................ OMG

janosh:
>  From my current understanding, pre-commit seems to check modified file for a PR only, right?

that's correct. you can still run it on all files with `pre-commit run --all-files`

> On the other hand, running `mypy pymatgen` gives:
> 
> ```
> Found 732 errors in 96 files (checked 278 source files)
> ```
> 
> ................................ OMG

😆 yep, lots of legacy issues to clean up. can't be done all at once. but some type errors could also go away with a smarter type checker in the future. hoping `ruff` becomes a type checker! 🤞

pymatgen/io/cp2k/inputs.py
```diff
@@ -2732,22 +2724,23 @@ class DataFile(MSONable):
...
-            data = cls.from_str(file.read())
-            for obj in data.objects:
-                obj.filename = filename
-            return data
+        raise NotImplementedError
+        # with open(filename, encoding="utf-8") as file:
+        #     data = cls.from_str(file.read())
+        #     for obj in data.objects:
+        #         obj.filename = filename
+        #     return data
```
janosh, line 2734: was this change accidental? i believe it breaks the `test_relax_maker` in [`atomate2`](https://github.com/materialsproject/atomate2/blob/563ba2a62c5e8351995e3e775518db7cc7650657/tests/cp2k/jobs/test_core.py#L38)


```py
    response = job.run(store=store)
               ^^^^^^^^^^^^^^^^^^^^
  src/jobflow/core/job.py", line 598, in run
    response = function(*self.function_args, **self.function_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  src/atomate2/cp2k/jobs/core.py", line 402, in make
    return super().make.original(self, structure, prev_dir)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  src/atomate2/cp2k/jobs/base.py", line 176, in make
    task_doc = TaskDocument.from_directory(Path.cwd(), **self.task_document_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  src/atomate2/cp2k/schemas/task.py", line 350, in from_directory
    calc_doc, cp2k_objects = Calculation.from_cp2k_files(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  src/atomate2/cp2k/schemas/calculation.py", line 375, in from_cp2k_files
    cp2k_objects.update(_get_basis_and_potential_files(dir_name))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  src/atomate2/cp2k/schemas/calculation.py", line 482, in _get_basis_and_potential_files
    data[Cp2kObject.BASIS] = BasisFile.from_file(str(dir_name / "BASIS"))  # type: ignore[index]
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  pymatgen/io/cp2k/inputs.py", line 2754, in from_file
    raise NotImplementedError
NotImplementedError
```

DanielYang59: Ah! Looks like a huge mistake by me. This certainly is not right. Sorry I don't know why this happened by please revert this.

DanielYang59: I just had another look: https://github.com/materialsproject/pymatgen/blob/55869a11c6d755e57b38969b34b9b63f0cc257e6/pymatgen/io/cp2k/inputs.py#L2745-L2764

And it doesn't look like intended by me. `from_file` would call `from_str` method which is not implemented, therefore `mypy` would complain about the return type. Therefore I inserted an `NotImplementedError`

DanielYang59: @janosh @JaGeo  Any comment on this? I'm not seeing much difference from this change. Maybe I'm getting anything wrong?

janosh: i'm not sure i understand the question. `from_str` can be implemented by a subclass of `DataFile` which would then make `from_file` work correctly if the commented out code is restored

DanielYang59: You're right. Its child class could overwrite the `from_str` method and make `from_file` work. I would open a PR to fix this now.

Sorry for the trouble cause (and wondering why this is not captured by tests).

janosh: `pymatgen` has imperfect test coverage. new tests are always appreciated!

DanielYang59: I guess no project has perfect test 😄 I really wish to help but I don't have any experience with `CP2K`. Sorry.

kavanase:
Hi! Thanks for your work on `pymatgen`! 😃 
Just to flag, this update seems to be causing some failures on our GitHub Actions tests (for [`ShakeNBreak`](https://shakenbreak.readthedocs.io/en/latest/) and [`doped`](https://doped.readthedocs.io/en/latest/)) with the latest `pymatgen` version. I'm not sure if it just needs `typing_extensions` to be added to the requirements?

```
Run pytest -vv -m "not mpl_image_compare" tests  # all non-plotting tests
============================= test session starts ==============================
platform linux -- Python 3.11.9, pytest-8.1.1, pluggy-1.4.0 -- /opt/hostedtoolcache/Python/3.11.9/x64/bin/python
cachedir: .pytest_cache
Matplotlib: 3.8.4
Freetype: 2.6.1
rootdir: /home/runner/work/ShakeNBreak/ShakeNBreak
configfile: pyproject.toml
plugins: mpl-0.1[7](https://github.com/SMTG-Bham/ShakeNBreak/actions/runs/8663869495/job/23758934155#step:6:8).0
collecting ... collected 121 items / 1 error / 29 deselected / 92 selected

==================================== ERRORS ====================================
___________________ ERROR collecting tests/test_analysis.py ____________________
ImportError while importing test module '/home/runner/work/ShakeNBreak/ShakeNBreak/tests/test_analysis.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/_pytest/python.py:520: in importtestmodule
    mod = import_path(
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/_pytest/pathlib.py:5[8](https://github.com/SMTG-Bham/ShakeNBreak/actions/runs/8663869495/job/23758934155#step:6:9)4: in import_path
    importlib.import_module(module_name)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1204: in _gcd_import
    ???
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1147: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:6[9](https://github.com/SMTG-Bham/ShakeNBreak/actions/runs/8663869495/job/23758934155#step:6:10)0: in _load_unlocked
    ???
/opt/hostedtoolcache/Python/3.[11](https://github.com/SMTG-Bham/ShakeNBreak/actions/runs/8663869495/job/23758934155#step:6:12).9/x64/lib/python3.11/site-packages/_pytest/assertion/rewrite.py:178: in exec_module
    exec(co, module.__dict__)
tests/test_analysis.py:[14](https://github.com/SMTG-Bham/ShakeNBreak/actions/runs/8663869495/job/23758934155#step:6:15): in <module>
    from shakenbreak import analysis
shakenbreak/analysis.py:23: in <module>
    from shakenbreak import input, io
shakenbreak/input.py:30: in <module>
    from doped.generation import DefectsGenerator, name_defect_entries
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/doped/generation.py:32: in <module>
    from pymatgen.transformations.advanced_transformations import CubicSupercellTransformation
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pymatgen/transformations/advanced_transformations.py:36: in <module>
    from pymatgen.transformations.standard_transformations import (
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pymatgen/transformations/standard_transformations.py:[17](https://github.com/SMTG-Bham/ShakeNBreak/actions/runs/8663869495/job/23758934155#step:6:18): in <module>
    from pymatgen.analysis.elasticity.strain import Deformation
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pymatgen/analysis/elasticity/__init__.py:5: in <module>
    from .elastic import (
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pymatgen/analysis/elasticity/elastic.py:21: in <module>
    from pymatgen.analysis.elasticity.stress import Stress
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pymatgen/analysis/elasticity/stress.py:11: in <module>
E   ModuleNotFoundError: No module named 'typing_extensions'
```

janosh:
@kavanase thanks for reporting. those imports should have been guarded by a `if TYPE_CHECKING:` so as not to be executed at run time. luckily a simple fix in #3752. we'll make a hot fix release tomorrow.

kavanase:
Great, thanks @janosh!

kavanase:
Hi! Sorry to be bugging again 😅 but just flagging something that was causing a few test failures for us.
The format of `Kpoints.kpts` has been changed to be a list of `tuple`s rather than a list of `list`s previously, but this change has only been made for certain cases, while others are still in `list` format. 
Maybe this is intended (if we can update our tests to accept both), but I think it mightn't be, and best to have a consistent format?

kavanase:
https://github.com/materialsproject/pymatgen/commit/6a06f3cb0a8b551ed782d86ed5963cd8aaed6674#r140937732

PR #3707, opened 2024-03-25
`Self` return type on `Lattice` methods

janosh:
define new type `PbcLike = tuple[bool, bool, bool]`. improve tests for `typing` module

### DIFF
--- a/pymatgen/core/lattice.py
+++ b/pymatgen/core/lattice.py
@@ -9,7 +9,7 @@ import operator
 import warnings
 from fractions import Fraction
 from functools import reduce
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 import numpy as np
 from monty.dev import deprecated
@@ -32,6 +32,8 @@ __copyright__ = "Copyright 2011, The Materials Project"
 __maintainer__ = "Shyue Ping Ong"
 __email__ = "shyuep@gmail.com"
 
+PbcType = tuple[bool, bool, bool]
+
 
 class Lattice(MSONable):
     """A lattice object. Essentially a matrix with conversion matrices. In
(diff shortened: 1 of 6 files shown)


pymatgen/core/lattice.py
```diff
@@ -207,12 +212,12 @@ def d_hkl(self, miller_index: ArrayLike) -> float:
         Returns:
             d_hkl (float)
         """
-        gstar = self.reciprocal_lattice_crystallographic.metric_tensor
+        g_star = self.reciprocal_lattice_crystallographic.metric_tensor
         hkl = np.array(miller_index)
-        return 1 / ((np.dot(np.dot(hkl, gstar), hkl.T)) ** (1 / 2))
+        return 1 / ((np.dot(np.dot(hkl, g_star), hkl.T)) ** (1 / 2))
 
-    @staticmethod
```
DanielYang59, line 214: Good catch! I notice some other methods should not be `staticmethods` as well (sorry I got too many `mypy` error to fix at this moment and could not help much on this.).

For example:
https://github.com/materialsproject/pymatgen/blob/05d3f9f5643c9585e15b297a58697432c9de4dcf/pymatgen/alchemy/transmuters.py#L297-L313

https://github.com/materialsproject/pymatgen/blob/05d3f9f5643c9585e15b297a58697432c9de4dcf/pymatgen/analysis/graphs.py#L199-L290

https://github.com/materialsproject/pymatgen/blob/05d3f9f5643c9585e15b297a58697432c9de4dcf/pymatgen/analysis/graphs.py#L1630-L1737

DanielYang59: I might open an issue to keep things organized. Instead of having issues scattered in comments. Sorry for any possible inconvenience.

janosh: no worries, there's no rush with #3705. i'll take care of any merge conflicts

DanielYang59: > no worries, there's no rush with #3705. i'll take care of any merge conflicts

Deeply appreciate that.

> frankly, `mypy` can be very annoying and if there are some errors that you can't fix, i'm happy to take a look but also feel free to add `type: ignore`s

Yes I would try my best to fix as many as I could, and ping you whenever necessary (grateful for this). I would try to avoid suppress error unless we have a good reason (to make the code as robust as we can).

PR #3714, opened 2024-03-26
Correct units of Element.atomic_orbitals

esoteric-ephemera:
The docstr for `pymatgen.core.Element.atomic_orbitals` indicates that the units of the LDA energies is eV. However the data in `periodic_table.json` was never converted to eV from Hartree, so `element.atomic_orbitals` prints energies in Hartree. (This is also true for pymatgen versions before the periodic table data was updated in [#3542](https://github.com/materialsproject/pymatgen/pull/3542).)

This was brought up in a [matsci post](https://matsci.org/t/unit-of-atomic-orbitals-energy/54325).

This PR just adds an `Element.atomic_orbitals_eV` property which converts the atomic orbitals energies to eV, and ensures no breaking changes.

## Summary by CodeRabbit

- **New Features**
	- Added a new property to display atomic orbital energies in electronvolts (eV) for clarity and consistency.
	- Enhanced `atomic_orbitals_eV` property to include a calculation based on `Ha_to_eV` conversion.
	- Added new unit conversion constants (`Ha_to_eV`, `Ry_to_eV`, `bohr_to_angstrom`, `amu_to_kg`) and tests for unit conversions.


janosh (approved):
oh dear 😵‍💫 thanks for the fix @esoteric-ephemera 👍

pymatgen/core/periodic_table.py
```diff
@@ -254,6 +255,18 @@ def __getattr__(self, item: str) -> Any:
             return val
         raise AttributeError(f"Element has no attribute {item}!")
 
+    @property
+    def atomic_orbitals_eV(self) -> dict[str, float]:
```
janosh, line 259: actually, i just noticed this method is not inherited by `Element`, perhaps due to being an `Enum`?

PR #3728, opened 2024-03-30
Replace general `raise Exception` and add missing `raise` keyword

DanielYang59:
## Summary

Major changes:

- Some `Exception` is missing the `raise` keyword (thought there would be some as I did notice some in previous clean up, but luckily just found one in fae74b023c07a5805b36faf7871f45ab897c4464) (tried `re` pattern: `^\s?(?!raise)\s+\w+error\(\w?`, `^\s?(?!raise)\s+\w?exception\(\w?`)
- Replace too general `raise Exception` with specific `Error`

Reverted for now:
- Suppress `Exception` with `contextlib.suppress` for simplicity (use `re` pattern: `except\s+.*:\n\s+pass`)

## Summary by CodeRabbit

- **Refactor**
	- Improved error handling across various modules by replacing generic exceptions with more specific ones (e.g., `RuntimeError`, `KeyError`, `ValueError`) for clearer error messaging and debugging.
	- Changed exception types in multiple functions for better error indication and handling.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -975,7 +975,7 @@ class Vasprun(MSONable):
                     eigenvals[Spin.up] = up_eigen
             else:
                 if "" in kpoint_file.labels:
-                    raise Exception(
+                    raise ValueError(
                         "A band structure along symmetry lines "
                         "requires a label for each kpoint. "
                         "Check your KPOINTS file"
@@ -2749,8 +2749,7 @@ class Outcar:
                 self.er_bp_tot = self.er_bp[Spin.up] + self.er_bp[Spin.down]
 
         except Exception:
-            self.er_ev_tot = self.er_bp_tot = None
-            raise Exception("IGPAR OUTCAR could not be parsed.")
+            raise RuntimeError("IGPAR OUTCAR could not be parsed.")
 
     def read_internal_strain_tensor(self):
         """
@@ -2779,7 +2778,7 @@ class Outcar:
             elif match.group(1).lower() == "z":
                 index = 2
             else:
-                raise Exception(f"Couldn't parse row index from symbol for internal strain tensor: {match.group(1)}")
+                raise IndexError(f"Couldn't parse row index from symbol for internal strain tensor: {match.group(1)}")
             results.internal_strain_tensor[results.internal_strain_ion][index] = np.array(
                 [float(match.group(i)) for i in range(2, 8)]
             )
@@ -2951,7 +2950,7 @@ class Outcar:
             self.piezo_tensor = self.piezo_tensor.tolist()
 
         except Exception:
-            raise Exception("LEPSILON OUTCAR could not be parsed.")
+            raise RuntimeError("LEPSILON OUTCAR could not be parsed.")
 
     def read_lepsilon_ionic(self):
         """
@@ -3068,7 +3067,7 @@ class Outcar:
             self.piezo_ionic_tensor = self.piezo_ionic_tensor.tolist()
 
         except Exception:
-            raise Exception("ionic part of LEPSILON OUTCAR could not be parsed.")
+            raise RuntimeError("ionic part of LEPSILON OUTCAR could not be parsed.")
 
     def read_lcalcpol(self):
         """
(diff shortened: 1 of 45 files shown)


janosh (commented):
thanks @DanielYang59. it's always nice to reduce code lines but `contextlib.suppress` [is  slower than `try/except`](https://docs.astral.sh/ruff/rules/suppressible-exception/#why-is-this-bad). not sure if it's enough to be noticeable but i think remember reading somewhere it can be quite a lot especially if used in a `for` loop. perhaps you can find a loop in your code changes and do some before/after run time testing and post the results here?

DanielYang59:
Sure I would have a test and let you know (happy to revert these if `contextlib` is significantly slower). I'm currently double-checking the `re` pattern for identify missing `raise` keywords (as I did notice some)

DanielYang59:
Could confirm `contextlib` is "significantly" slower than `try/except`:
```python
import timeit

common_setup_code = """
import contextlib
test_list = [0, 1, 2]
"""

method1_code = """
try:
    test_list[10]
except IndexError:
    pass
"""

method2_code = """
with contextlib.suppress(IndexError):
    test_list[10]
"""

# Execute method 1 and measure runtime
method1_time = timeit.timeit(stmt=method1_code, setup=common_setup_code, number=100000)

# Execute method 2 and measure runtime
method2_time = timeit.timeit(stmt=method2_code, setup=common_setup_code, number=100000)

print("Method 1 (except block) time:", method1_time)
print("Method 2 (contextlib.suppress) time:", method2_time)
```

Gives:
```
Method 1 (except block) time: 0.0103234
Method 2 (contextlib.suppress) time: 0.040782500000000006
```

janosh:
> Could confirm `contextlib` is "significantly" (4x time) slower than `try/except`:

probably doesn't matter given we're talking about microseconds here but i'm still a bit ambivalent about this change. also, a slightly cleaner test would be to not include the `import contextlib` for `try/except` since it's not needed there and to only import it once in the `contextlib` case since it's then cached for the rest of the Python session

DanielYang59:
> probably doesn't matter given we're talking about microseconds here but i'm still a bit ambivalent about this change. 

I'm quite unsure about this change too. `contextlib.suppress` is much cleaner though being slow.

> also, a slightly cleaner test would be to not include the `import contextlib` for `try/except` since it's not needed there and to only import it once in the `contextlib` case since it's then cached for the rest of the Python session

I actually included the import block intentionally such that `import` would not be the cause of time difference, here is the updated script:
```python
import timeit

setup_code = """
test_list = [0, 1, 2]
"""

method1_code = """
try:
    test_list[10]
except IndexError:
    pass
"""

method2_code = """
import contextlib
with contextlib.suppress(IndexError):
    test_list[10]
"""

# Execute method 1 and measure runtime
method1_time = timeit.timeit(stmt=method1_code, setup=setup_code, number=100000)

# Execute method 2 and measure runtime
method2_time = timeit.timeit(stmt=method2_code, setup=setup_code, number=100000)

print("Method 1 (except block) time:", method1_time)
print("Method 2 (contextlib.suppress) time:", method2_time)
```

gives:
```
Method 1 (except block) time: 0.009449200000000001
Method 2 (contextlib.suppress) time: 0.0490635000000000
```

DanielYang59:
Well I guess it's better to focus on changes "we really need" at this moment (like type annotation and bug fixes)? I would reverse this until we really need it.

pymatgen/io/vasp/outputs.py
```diff
@@ -2750,8 +2750,7 @@ def p_ion(results, match):
                 self.er_bp_tot = self.er_bp[Spin.up] + self.er_bp[Spin.down]
 
         except Exception:
-            self.er_ev_tot = self.er_bp_tot = None
```
DanielYang59, line 2753: Not sure we really need this?

DanielYang59:
@coderabbitai review

DanielYang59:
@janosh Please review. I'm 100% confident I was using the proper type of `Error`, please help me double-check. Thanks!

pymatgen/io/vasp/outputs.py
janosh: probably not. looks like it's set to that initially anyway

janosh (approved):
thanks @DanielYang59! big fan of more specific error classes. 👍

DanielYang59:
Thanks for reviewing. There seems to be quite a lot catching of general exceptions as well (`except Exception`), might be good to address this as well.

pymatgen/io/vasp/outputs.py
DanielYang59: Thanks for confirming!

Issue #3684, opened 2024-03-10
[Dev] Multiple dependencies for CI missing


DanielYang59:
Multiple external dependencies are missing from test setup, including:
- [x] `mcsqs`: not planned (https://github.com/materialsproject/pymatgen/issues/3684#issuecomment-1988998112)
- [x] `icet`: not planned (https://github.com/materialsproject/pymatgen/issues/3684#issuecomment-1987687507)
- [x] `emmet`: not planned (https://github.com/materialsproject/pymatgen/issues/3684#issuecomment-1987690470)
- [x] `Open Babel`: added in #3729
- [x] `enum_lib`: Added for Ubuntu only in #3729
- [x] `gulp`: not planned (libgfortran3 not supported by Ubuntu20 afterwards)
- [ ] `graphviz`
- [ ] `critic2`: https://aoterodelaroza.github.io/critic2/installation/
- [ ] `OBAlign`
- [ ] `vampire`
- [ ] `x_trans`
- [ ] `boltztrap2`
- [ ] `zeo`

Is this intended or should we fix it?

janosh:
i believe missing `icet` was deliberate as @esoteric-ephemera [added it only to `requirements-optional.txt`](https://github.com/materialsproject/pymatgen/pull/3593/files#diff-d52eaab4a615f02e5a1b6e6dd9e17be0c5f0519bcdb433cd28c45705b569da45R14) which is not used in CI.

though i could see how that is confusing which is why i've advocated for removing the requirements files in the past. @shyuep prefers to keep. i fwiw, i think the `atomate2` approach is better of having a [`strict` optional deps section in pyproject.toml](https://github.com/materialsproject/atomate2/blob/9f6438150d63706d6f9bf9b63f143932be27b82b/pyproject.toml#L72)

janosh:
`emmet` is purposefully not installed as it comes with a host of transitive deps.

as for the others, i think it would be good to install and test their respective `pymatgen` integrations in CI

esoteric-ephemera:
> i believe missing icet was deliberate as @esoteric-ephemera [added it only to requirements-optional.txt](https://github.com/materialsproject/pymatgen/pull/3593/files#diff-d52eaab4a615f02e5a1b6e6dd9e17be0c5f0519bcdb433cd28c45705b569da45R14) which is not used in CI.

[...]

PR #3729, opened 2024-03-31
[WIP] Add a pymatgen.io.openff module

orionarcher:
## Summary

This PR adds basic integration with the `openff.toolkit` Molecule representation. It is intended to be use in the `classical_md` workflows in `atomate2` and `emmet`.

Currently, this functionality is downstream in a PR to [atomate2](https://github.com/materialsproject/atomate2/pull/782). As basic IO functionality, I would like to migrate it upstream to `pymatgen`. Especially because some functions will also be used in the `classical_md` `emmet` `Builder`.

Major changes:

- Several utility functions that support IO between `pymatgen.Molecule`/`pymatgen.MoleculeGraph` <-> `openff.toolkit.Molecule`

## Todos

`openff.toolkit` is only installable via conda. I am not sure the best way to add it to CI. I'd appreciate guidance from the core developers on this.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


janosh (commented):
cool, excited to see this! 👍 

is the issue here that [`openff` is only installable](https://docs.openforcefield.org/projects/toolkit/en/stable/installation.html#installing-from-source) with `conda`/`mamba`?
we could try installing from source in CI

orionarcher:
> we could try installing from source in CI

Sounds good! I'll try adding that to the CI.

orionarcher:
OpenBabel is a core dependency of the OpenFF toolkit, and I don't believe it is pip installable. I believe that's the reason openff doesn't support pip). I believe we'd either have to build it from source OR use conda to install it. 

I know this is a problem in other parts of the MP ecosystem and I am curious how you'd recommend handling it?

DanielYang59:
> OpenBabel is a core dependency of the OpenFF toolkit, and I don't believe it is pip installable. 

I'm interested in hearing about this as well, as `OpenBabel` is not installed in CI as well https://github.com/materialsproject/pymatgen/issues/3684. Feel free to ping me if there is anything I could help.

janosh:
we used to install `openbabel` in CI with `conda` but it was removed in mid 2022 due to causing frequent issues. see https://github.com/materialsproject/pymatgen/commit/b13c65bd59b1f2b55b769c928a8ab8af7e539181. i'm open to giving it another try but i'd prefer not to install with `conda` and instead follow the [build from source instructions](https://open-babel.readthedocs.io/en/latest/Installation/install.html#basic-build-procedure). `conda` is slow and not currently used in pymatgen CI so would have to be installed first just to install a single dep

orionarcher:
I'll try building from source.

We could conda install openbabel, packmol, and the three required packages from the openff toolkit, which would be a little more worthwhile. A lower overhead option could be [micromamba](https://github.com/marketplace/actions/setup-micromamba) but I think that would require larger changes to the build script. Just a thought!

orionarcher:
Installing `openff-toolkit` from source requires separately installing all dependencies, some of which are only available on conda. I don't think there is a clean way to make testing work in CI without using either `conda` or `micromamba`.

janosh:
let's give `micromamba` a try then? interested to see how it performs here given it's meant to be faster than `conda`

orionarcher:
I switched over the test suite to use micromamba and it seems to be ~20-30% faster on ubuntu and comparable on windows.

This switch enabled me to install openbabel, which was not previously included in testing. I discovered several failing tests that were flying under the radar and that are now causing CI to fail.
- errors in `test_babel.py` which I fixed
- errors in `analysis/test_graphs.py` and `io/qchem/test_outputs.py` that I am not sure how to debug. These are failing both locally and in the ubuntu CI. It's not an issue in the windows CI because openbabel is not being installed.

@janosh, what do you recommend as the path forward here?

Should we:
a) merge as is with tests failing?
b) disable the tests and contact the maintainers?
c) have someone (preferably other than me) try to debug the issue in this PR?

janosh:
luckily fixing the 3 failing tests was fairly easy. fixing the third surfaced a 4 issue so there's still 1 failing test

> b) disable the tests and contact the maintainers?

i think we can skip `TestQCOutput.test_all` for now and ping @samblau and @espottesmith. if you guys have time to take a look, feel free to suggest a fix

janosh (approved):
thanks @orionarcher. excited about the new OpenFF integration!

and really appreciate that you reinstated the `openbabel` tests!

PR #3734, opened 2024-04-03
Fix `ChemicalPotentialDiagram` 2D plot not respecting `formal_chempots` setting

uliaschauer:
## Summary

When a ChemicalPotentialDiagram is generated with more than 2 elements and then shown as a 2D plot via an elements argument with two entries, the formal_chempots=False setting of the initial ChemicalPotentialDiagram is not passed on, resulting in formal chemical potentials in all cases.

Major changes:

- added a formal_chempots attribute to the ChemicalPotentialDiagram class
- passed this attribute in get_plot() when creating the subset ChemicalPotentialDiagram object

## Todos

none

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to handle formal chemical potentials in the Chemical Potential Diagram feature.
	- Added a new test method to compare plots with formal and non-formal data.


janosh (changes requested):
thanks, could you add a test for this?

PR #3739, opened 2024-04-07
Avoid using `eval`, replace manual offset in `enumerate` and rename single letter variables

DanielYang59:
## Summary

- Replace `eval` with safer alternatives
- Replace manual offset in `enumerate` with the `start` argument
- Minor format tweaks

## Summary by CodeRabbit

- **Refactor**
	- Improved loop logic and simplified conditions in magnetism analysis methods.
	- Enhanced safety by replacing `eval` with `literal_eval` in magnetism analysis.
	- Streamlined code for electronic structure calculations, including conditional assignments and property access.
	- Optimized parameter type checking and assignment operations in VASP inputs handling.
- **Documentation**
	- Updated docstring for better clarity in VASP inputs handling.

### DIFF
--- a/pymatgen/analysis/chemenv/utils/graph_utils.py
+++ b/pymatgen/analysis/chemenv/utils/graph_utils.py
@@ -268,7 +268,7 @@ class SimpleGraphCycle(MSONable):
         order in the list.
         """
         if edges_are_ordered:
-            nodes = [e[0] for e in edges]
+            nodes = [edge[0] for edge in edges]
             if not all(e1e2[0][1] == e1e2[1][0] for e1e2 in zip(edges, edges[1:])) or edges[-1][1] != edges[0][0]:
                 raise ValueError("Could not construct a cycle from edges.")
         else:
--- a/pymatgen/analysis/chemenv/utils/graph_utils.py
+++ b/pymatgen/analysis/chemenv/utils/graph_utils.py
@@ -490,8 +490,8 @@ def get_all_elementary_cycles(graph):
         edge_idx += 1
     cycles_matrix = np.zeros(shape=(len(cycle_basis), edge_idx), dtype=bool)
     for icycle, cycle in enumerate(cycle_basis):
-        for in1, n1 in enumerate(cycle):
-            n2 = cycle[(in1 + 1) % len(cycle)]
+        for in1, n1 in enumerate(cycle, start=1):
+            n2 = cycle[(in1) % len(cycle)]
             iedge = all_edges_dict[(n1, n2)]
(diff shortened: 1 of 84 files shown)


janosh (approved):
looks good to me.

sorry for hijacking this PR but i'd be interested to get your opinion on these variable renames. single-letter variables are a pet peeve of mine for hurting code readability so been trying to get rid of them for a while

DanielYang59:
> sorry for hijacking this PR 

No worries at all. I'm going off topic a bit too (Just cannot resist the pulse to change them, I understand that).

> but i'd be interested to get your opinion on these variable renames. single-letter variables are a pet peeve of mine for hurting code readability so been trying to get rid of them for a while

I would fully support such improvements, which makes the code much more readable (especially without reading the full code block). But for some common cases, I sometimes think it might not be uncessary (like `i/j` for index, `k/v` for `key/value`), but for more complex cases, using a more detailed var name is much better (especially when similar name could occur, which could be very confusing, for example the `core.surface` I'm currently cleaning up, I need to guess `c` stands for cluster or c-coordinate).

pymatgen/analysis/chemenv/utils/graph_utils.py
```diff
@@ -268,7 +268,7 @@ def from_edges(cls, edges, edges_are_ordered: bool = True) -> Self:
         order in the list.
         """
         if edges_are_ordered:
-            nodes = [e[0] for e in edges]
```
DanielYang59, line 271: Like this, this may or may not be necessary. I don't think people would be confused by this.

DanielYang59:
@janosh Please review. Thanks!

janosh:
`ruff` has had [limited auto-typing ability](https://docs.astral.sh/ruff/rules/missing-return-type-special-method) for a while.

i just ran

```sh
ruff check . --select ANN204 --unsafe-fixes
Found 674 errors.
[*] 550 fixable with the --fix option.
```

would you be interested in submitting a PR with those changes as a followup to #3705?

DanielYang59:
> `ruff` has had [limited auto-typing ability](https://docs.astral.sh/ruff/rules/missing-return-type-special-method) for a while.
> 
> i just ran
> 
> ```shell
> ruff check . --select ANN204 --unsafe-fixes
> Found 674 errors.
> [*] 550 fixable with the --fix option.
> ```
> 
> would you be interested in submitting a PR with those changes as a followup to #3705?

Wow this sounds very interesting (always amazed how much you know about all kinds of stuff). I would have a go at this after I finish cleaning up `core.surface`. Thanks for the useful input.

PR #3740, opened 2024-04-07
Add type annotations for `io.vasp.inputs/optics`

DanielYang59:
## Summary

- Add type annotations for `io.vasp.inputs/optics`
- Replace `str` as path type with `PathLike`
- Docstring tweaks

### Originally 

Proposed by @janosh in https://github.com/materialsproject/pymatgen/pull/3739#issuecomment-2041429466.

- Run `ruff check . --select ANN204 --unsafe-fixes` to auto-fix "ANN204", to experiment on auto typing fix with `ruff`.
- 135 `mypy` errors were dug out and need to be fixed. 

[`ANN204`](https://docs.astral.sh/ruff/rules/missing-return-type-special-method/) refers to: 
> Checks that "special" methods, like `__init__`, `__new__`, and `__call__`, have return type annotations.

## Summary by CodeRabbit

- **New Features**
  - Added a 60-second timeout to HTTP requests across various modules to improve system reliability during data fetching.

- **Refactor**
  - Enhanced code clarity by specifying explicit types for various attributes and variables across multiple modules.

- **Bug Fixes**
  - Adjusted handling of subprocess output and method argument defaults to prevent errors and improve data handling.


DanielYang59:
@janosh I re-visited this PR and realized it might not be a good idea to work on a single method globally like this, we should push the type annotation effort module by module. 

Though I agree `ruff` auto-fix is looking promising but we should not run it globally.

My reason being: by working like this, we're just adding types for a single (or very few) function among many other un-typed functions. But as the types within a module heavily depend on others, it would either dig out huge amount of extra work, or make the type very loose.

I would prefer close this and push our type-annotation effort module by module as discussed previously (starting for `core`, `io.vasp`, `electronic`, `phonon` and such). What do you think?

janosh:
going folder by folder sounds good. no need to close this PR though, we can just start with `pymatgen/io/vasp` in this one.

i just reverted the initial 

```sh
ruff check . --select ANN204 --unsafe-fixes --fix
```

and force pushed

```sh
ruff check pymatgen/io/vasp --select ANN204 --unsafe-fixes --fix
```

instead and keeping your manual fixes. you just need to run `git reset --hard @{u}` to get back to the tip of this branch

DanielYang59:
> going folder by folder sounds good. no need to close this PR though, we can just start with `pymatgen/io/vasp` in this one.

Good point. Let's do it.

> instead and keeping your manual fixes. you just need to run `git reset --hard @{u}` to get back to the tip of this branch

There isn't much valuable fixes though. Thanks for the tip!

pymatgen/io/vasp/inputs.py
```diff
@@ -362,55 +376,57 @@ def from_str(cls, data, default_names=None, read_velocities=True) -> Self:
...
-        cart = pos_type[0] in "cCkK"
-        n_sites = sum(n_atoms)
+        cart: bool = pos_type[0] in "cCkK"
+        n_sites: int = sum(n_atoms)
 
         # If default_names is specified (usually coming from a POTCAR), use
         # them. This is in line with VASP's parsing order that the POTCAR
         # specified is the default used.
-        if default_names:
+        if default_names is not None:
```
DanielYang59, line 405: The default value of `default_names` seems inconsistent in: https://github.com/materialsproject/pymatgen/blob/7f96ef2c6c4e753a5dafbac4fde6f5a136c0aec0/pymatgen/core/structure.py#L2860-L2863

Changed to `None` in 1a1752a38a526f39b2c12279f9f15fbad593a73d.

DanielYang59:
I'm trying to relocate the class methods such that dunder methods and properties come first. But this seems to pollute git history. Wondering if there is a way to relocate codes without such side effect?

DanielYang59:
@coderabbitai review

pymatgen/io/vasp/outputs.py
```diff
@@ -1957,8 +1957,8 @@ def __init__(self, filename):
         else:
             mag = mag_x
 
-        # data from beginning of OUTCAR
-        run_stats["cores"] = None
+        # Data from beginning of OUTCAR
+        run_stats["cores"]  = -1
```
DanielYang59: I changed the start value for number of cores from `None` to `-1` for type compatibility, let me know if this could potentially break anything, or there is any better way.

Similarly from memory usage:
```python
                if time_patt.search(line):
                    tok = line.strip().split(":")
                    try:
                        # try-catch because VASP 6.2.0 may print
                        # Average memory used (kb):          N/A
                        # which cannot be parsed as float
                        run_stats[tok[0].strip()] = float(tok[1].strip())
                    except ValueError:
                        run_stats[tok[0].strip()] = -1
                    continue
```

DanielYang59:
Can you please review this @janosh? I decided to separate the work for different modules as there are too many changes. Some return types are removed for completely un-typed __init__ in io.vasp.outputs to avoid `mypy` checking for now, I would work on it later.

There is one minor fix for the default value of `default_names` in https://github.com/materialsproject/pymatgen/pull/3740#discussion_r1569868926.

And there are two requests for changing the default values for type compatibility in https://github.com/materialsproject/pymatgen/pull/3740#discussion_r1570594077.

Other than these two, I don't expect further functional changes (the unit test failure is related to the requested default value change, I would need your confirmation before I could change the unit test :) ). Thanks!

pymatgen/io/vasp/outputs.py
janosh: i think it's safer to stick with `run_stats["cores"] = None`. can we not update the type to `int | None` and explicitly guard against `None` where necessary?

janosh:
> Wondering if there is a way to relocate codes without such side effect?

not really. have a look at [this video](https://youtu.be/qYJWBJvHqj0) which doesn't solve your problem but is slightly relevant

pymatgen/io/vasp/outputs.py
DanielYang59: Yes! Looks we could allow `None` by explicitly set type as `run_stats: dict[str, float | None] = {}`

Changed in 16d26857a6f7f9a433001f9048bd692e932e88a4

DanielYang59:
> > Wondering if there is a way to relocate codes without such side effect?
> 
> not really. have a look at [this video](https://youtu.be/qYJWBJvHqj0) which doesn't solve your problem but is slightly relevant

Thanks. It's always great to know more tips :)

Should be ready for reviewing then. I reverted the default values to `None`. Thanks!

pymatgen/analysis/chemenv/coordination_environments/coordination_geometries.py
```diff
@@ -908,7 +908,7 @@ def __init__(self, permutations_safe_override=False, only_symbols=None):
 
         self.minpoints = {}
         self.maxpoints = {}
-        self.separations_cg = {}
+        self.separations_cg: dict = {}
```
janosh, line 911: can we make the types more specific? `list`/`dict`/`set`/... alone don't add any information for the reader. `dict[str, CoordinationGeometry]`, `list[Structure]`, etc. is more helpful

pymatgen/analysis/structure_prediction/substitution_probability.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, lambda_table=None, alpha=-5):
 
         # create Z and px
         self.Z = 0
-        self._px = defaultdict(float)
+        self._px: defaultdict = defaultdict(float)
```
janosh, line 82:
```suggestion
        self._px: dict[str, float] = defaultdict(float)
```

pymatgen/apps/battery/plotter.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, xaxis="capacity", hide_negative=False):
             - frac_x: the atomic fraction of the working ion
             hide_negative: If True only plot the voltage steps above zero.
         """
-        self._electrodes = {}
+        self._electrodes: dict = {}
```
janosh, line 31:
```suggestion
        self._electrodes: dict[str, AbstractElectrode] = {}
```

janosh (changes requested):

pymatgen/analysis/chemenv/coordination_environments/coordination_geometries.py
DanielYang59: > can we make the types more specific? `list`/`dict`/`set`/... alone don't add any information for the reader. `dict[str, CoordinationGeometry]`, `list[Structure]`, etc. is more helpful

Yes this is very true. But in this PR I was working on adding types for `io.vasp`, sometimes `mypy` would complain about other modules (`need annotation for some line`) which I haven't yet got time to look closer into (you know the situation, when there are hundreds of errors to fix). But I would keep this in mind and try to more specific types :)

pymatgen/analysis/structure_prediction/substitution_probability.py
DanielYang59: I'm pretty sure I added some more types as `defaultdict` elsewhere. I would fix these together.

DanielYang59: This looks like `self._px: dict[Species, float] = defaultdict(float)` to me?
```python
        # create Z and px
        self.Z = 0
        self._px: dict[Species, float] = defaultdict(float)
        for s1, s2 in itertools.product(self.species, repeat=2):
            value = math.exp(self.get_lambda(s1, s2))
            self._px[s1] += value / 2
            self._px[s2] += value / 2
            self.Z += value
```

janosh: you're right!

janosh (approved):
thanks so much @DanielYang59! 🙏 

given the generic type annotations like `var_name: list = []` are easy to find and fix later, we can merge this now

DanielYang59:
Thanks for reviewing!

> given the generic type annotations like `var_name: list = []` are easy to find and fix later

Yes I don't know how `mypy` made those decisions, but it seems to mark some unrelated and un-typed parts and require types annotations for those. Such general types were added to stop such `mypy` complaints :)

PR #3741, opened 2024-04-09
Update ENCUT type to float in  incar_parameters.json

yuuukuma:
ENCUT in VASP INCAR settings should be float

## Summary by CodeRabbit

- **Refactor**
	- Updated the data type for the `ENCUT` parameter to enhance precision in calculations.


mkhorton:
Thanks @yuuukuma!

PR #3743, opened 2024-04-09
Replace unnecessary usage of `len` for iterable existence check

DanielYang59:
## Summary

Replace unnecessary usage of `len` for iterable existence check:

- `if len(x) > 0` replaced with `if x`
- `if len(x) < 1` replaced with `if not x`

## Summary by CodeRabbit

- **Refactor**
	- Streamlined logic in lattice module for improved efficiency.
	- Enhanced readability by replacing length checks with boolean evaluations.
	- Simplified conditional checks in structure module for better readability.
	- Improved initialization logic for bond valence calculations, enhancing compatibility with specific species.
	- Simplified conditional checks in mixing scheme module for streamlined logic.
	- Simplified condition checks for data presence in CIF module.
	- Updated method in CIF module to use boolean evaluation for argument extraction.
	- Appended various warnings based on data processing conditions.
- **Tests**
	- Simplified conditions in test cases for oxidation states and file line processing for clarity.

### DIFF
--- a/pymatgen/analysis/bond_valence.py
+++ b/pymatgen/analysis/bond_valence.py
@@ -160,7 +160,7 @@ class BVAnalyzer:
         forbidden_species = [get_el_sp(sp) for sp in forbidden_species] if forbidden_species else []
         self.icsd_bv_data = (
             {get_el_sp(specie): data for specie, data in ICSD_BV_DATA.items() if specie not in forbidden_species}
-            if len(forbidden_species) > 0
+            if forbidden_species
             else ICSD_BV_DATA
         )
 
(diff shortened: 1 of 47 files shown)


janosh (commented):
i think `if len(lst) > 0:` is more explicit and can be more readable than `if lst:` but don't feel strongly about that.

would be interested to hear other's opinions. maybe @mkhorton wants to chime in?

JaGeo:
> i think if len(lst) > 0: is more explicit and can be more readable than if lst: but don't feel strongly about that.
> 
> would be interested to hear other's opinions. maybe @mkhorton wants to chime in?

I like the first one more. (Sorry not sure how I ended up editing the old post 😬. Sometimes, I am a bit distracted.) Also funny how there are two exactly opposite recommendations by code rabbit 😅

DanielYang59:
It's not strongly held by me too (though stronger for the "`if len(x) < 1` -> `if not x`" replacement). I think the readability of `if var` strongly depends on how descriptive the variable name is. 

Feel free to close this PR if you think it's not necessary or adding ambiguity.

DanielYang59:
I would close this for now until someone decide it's necessary. Sorry for the confusion 😄

PR #3746, opened 2024-04-11
Collect test files (csv/json/cif) into folders

DanielYang59:
## Collect test files (csv/json/cif) into their own folders

- The `re` pattern used for moving `json` file in VS code: source `(f"\{TEST_FILES_DIR\})(/[^/]+\.json(\.gz)?")`, target `$1/json$2`


DanielYang59:
@coderabbitai ignore

janosh (changes requested):
how about instead of subfolders based on filetype, we split test files by the corresponding pymatgen module?

so a lot of the `tests/files/json` would go in `tests/files/entries` instead if they contain serialized phase diagram or Pourbaix entries.
CIF files and some JSON files would go in `tests/files/structures`.

other folders could be `tests/files/electronic_structure` (or just `tests/files/electronic`) for band structures and DOS, `tests/files/phonons` for phonon BS/DOS, `tests/files/lobster` (or perhaps  `tests/files/io/lobster`), etc.

DanielYang59:
> how about instead of subfolders based on filetype, we split test files by the corresponding pymatgen module?

It would be good to make the test file folders follow the same hierarchy as the code, but I'm afraid there would be huge overlap (and confusion) at least at this stage? (Currently we don't have a one-to-one mapping between test and test files) Multiple module might require access to the same format of file (or even the same file). For example `electronic_structure` might require `DOSCAR` from `vasp`, how can we decide where `DOSCAR` should go then?

On the other hand, separating files by format seems much easier to access for developers? (though people might need to access multiple test file folders if multiple formats are involved within the test)

DanielYang59:
> how does it become easier? the file extension already makes searching by file type easy. i just type `.json` in the file picker to get a list. putting it in a JSON folder doesn't confer additional meaning on what the file was meant to be used for. putting it in `tests/files/phonons` gives more information. 

I think my intention for cleaning up test files is not to add more information, but to clean up the "root directory" of test file folder 😄 . Currently we have quite a lot `json` and `cif` laid flat under the test file dir, and it would flood `ls` command (of course if you do a more explicit search there would be no difference).

> we just have to decide on a location. e.g. in your `DOSCAR` example, i think it should definitely go in `tests/files/io/vasp`

Yes that what I'm concerned about, other would need to "learn" about this agreement then. People would not need to read a `naming_convention` file or `README` file to know cif files would reside in the `cif` dir I assume?

janosh (changes requested):
i think the `cif/` folder could make sense but i'm less sure about the JSON folder since it's quite a lot more generic, containing files of very different purpose. how about we leave the `cif/` folder as you currently have it but split the `json/` based on what `pymatgen` module the test file belongs to? that way we minimize `git diff` from repeatedly moving any files around in subsequent PRs

DanielYang59:
>  how about we leave the `cif/` folder as you currently have it but split the `json/` based on what `pymatgen` module the test file belongs to? 

That's a great idea! Also sounds like a good start towards organizing test files by modules. I would get my hands on this next week.

DanielYang59:
@coderabbitai ignore

DanielYang59:
@janosh Please review this as well. I wasn't thinking that we could just do everything in one PR, but luckily here we are :) All test files sorted by module.

DanielYang59:
> it used to be ~1GB so 800MB is already an improvement. 

Great to know!

> don't think there's such a thing as garbage collection for `git`. history just keeps growing. 

There actually is https://git-scm.com/docs/git-gc. Not sure if it would be run automatically or not.

> you can decrease clone time by only getting the last commit
> 
> ```shell
> git clone https://github.com/materialsproject/pymatgen --depth 1
> ```

Thanks for the tip. Just saying though, I don't clone very often so it's not a problem :)

PR #3748, opened 2024-04-12
Clarify argument `shift` for `SlabGenerator.get_slab`

DanielYang59:
## Summary 

A follow-up PR for #3691. 
- Clarify misleading arg name `shift` for **private** method `SlabGenerator.get_slab`
- Fix typing of `slab` in function `center_slab` (should be `Structure` instead of `Slab`)

### Rationale 

- This method should be a private method (as per the docstring).
- The arg name `shift` for `SlabGenerator.get_slab` corresponds to the desired `termination` coordinate (instead of the shift vector to get the `Slab`), which could be misleading: https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/surface.py#L1087
Which might cause confusion in understanding the implementation:
https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/surface.py#L1119

image

#### Clarify `shift` attribute of `Slab` 

Similarly, the description of `shift` attribute of `Slab` is incorrect, it is NOT the shift but the **negative** shift.

https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/surface.py#L118-L119

<details>
    <summary> TODO for future PRs </summary>

## In future PRs

1. Function `center_slab`: https://github.com/materialsproject/pymatgen/blob/ad6eafee2474718813cba57aa8e1752f1181b639/pymatgen/core/surface.py#L759-L788
    
    **a.** During the `for site in slab:` loop, site coordinates would change.
    **b.** This method to put the `Slab` to the center seems overly complex, can we just calculate the z-center of the `Slab`, and move it to z=0.5 (fractional)?
    **c.** If not **b**, Sites with z<0 is not considered (it would only move `nn[1] >= slab.lattice.c`).
    **d.** A magic number for determining near neighbors by cutoff radius: `cutoff_radius = bond_dists[0] * 3`. 

2. The `move_to_other_side` method of `SlabGenerator`: https://github.com/materialsproject/pymatgen/blob/ad6eafee2474718813cba57aa8e1752f1181b639/pymatgen/core/surface.py#L1459-L1473

    https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1555181912
 
3. Currently there are three very similar methods/functions that need to **clarify in docstring**:
    a. `get_slab` of `SlabGenerator`: The actual method to generate `Slab`, should be a private method (as per the docstring, but have already been used externally).
    b. `get_slabs` of `SlabGenerator`: A wrapper method on top of `get_slab` to generate possible shifts and then call `get_slab`.
    c. `generate_all_slabls`: A wrapper function on top of `get_slabs`, which generate slabs up to a given Miller index.

## Next up: Fix some `Slab` related issues
Don't think I could resolve them in one shot, but good to collect them:
- #2591
- #3411
- #3412
- #3682 
</details>


pymatgen/core/surface.py
```diff
@@ -214,6 +214,44 @@ def surface_area(self) -> float:
         matrix = self.lattice.matrix
         return np.linalg.norm(np.cross(matrix[0], matrix[1]))
 
+    def is_symmetric(self, symprec: float = 0.1) -> bool:
```
DanielYang59, line 217: `is_symmetric` and `is_polar` look very much like a `property`. Wondering if there is a way to mark it as `property` and take an argument at the same time. (maybe carry a global `tolerance` attrib? As many downstream operations like `get_tasker2_slabs`, `get_slab`/`get_slabs` from `SlabGenerator` all need a tolerance)

pymatgen/core/surface.py
```diff
@@ -1104,10 +1103,8 @@ def get_slab(self, shift: float = 0, tol: float = 0.1, energy: float | None = No
         species = self.oriented_unit_cell.species_and_occu
 
         # Shift all atoms
-        # DEBUG(@DanielYang59): shift value in Angstrom inconsistent with frac_coordis
         frac_coords = self.oriented_unit_cell.frac_coords
-        # DEBUG(@DanielYang59): suspicious shift direction (positive for downwards shift)
-        frac_coords = np.array(frac_coords) + np.array([0, 0, -shift])[None, :]
+        frac_coords = np.array(frac_coords) + np.array([0, 0, shift])[None, :]
```
DanielYang59, line 1107: https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1553711017 migrated here. Clarified two issues:
- The unit for `shift` should be "fractional" instead of "Angstrom", docstring has been updated.
- (Breaking) Reversed the default shifting direction, previously a positive shift vector refers to shift "downwards".

pymatgen/core/surface.py
```diff
@@ -1231,8 +1228,7 @@ def gen_possible_shifts(ftol: float) -> list[float]:
 
             # Clustering does not work when there is only one atom
             if n_atoms == 1:
-                # TODO (@DanielYang59): why this magic number 0.5?
-                shift = frac_coords[0][2] + 0.5
+                shift = -frac_coords[0][2] + 0.5  # shift to center
```
DanielYang59, line 1231: https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1554518586 migrated here.

This looks like a typo, if the `shift` (in fractional coordinate) for single atom is `-frac_coords[0][2] + 0.5`, the atom would then be shifted to the center.

DanielYang59:
I would need discussion/confirmation/suggestion on these proposed changes, as they more or less seem to introduce functional changes. @janosh @shyuep 

Also trying to ping the original author (I guess) @CifLord in case you have any comment 😃

pymatgen/core/surface.py
janosh: if we change the shifting direction, that needs to be very prominently communicated in the changelog and the [recent breakage](https://pymatgen.org/compatibility.html#recent-breaking-changes) docs. could there be some intuition we're not thinking of why positive shift should decrease the z-axis? just trying to weigh the pros and cons of breaking. i also would expected z to increase but maybe explaining in the doc string why it's this way could warn users and avoid breaking.

pymatgen/core/surface.py
janosh: did this change affect any tests?

pymatgen/core/surface.py
janosh: i'm usually not in favor of extra state like `SlabGenerator.tolerance` unless really necessary. state tends to make code more confusing

i don't think you can make these into `@property`s without a breaking change either. perhaps best to leave as is?

janosh (approved):
thanks @DanielYang59 and sorry for the slow reply. this all looks good. [just had 1 question](https://github.com/materialsproject/pymatgen/pull/3748/files#r1583984225)

pymatgen/core/surface.py
DanielYang59: > i'm usually not in favor of extra state like `SlabGenerator.tolerance` unless really necessary. state tends to make code more confusing

Okay let's keep it as is. There're more urgent issues to address for `surface` than this.

> i don't think you can make these into `@property`s without a breaking change either. perhaps best to leave as is?

Yes sure. I would avoid breaking changes unless necessary.

pymatgen/core/surface.py
DanielYang59: > if we change the shifting direction, that needs to be very prominently communicated in the changelog and the [recent breakage](https://pymatgen.org/compatibility.html#recent-breaking-changes) docs. 

Okay! I would keep it in mind.

> could there be some intuition we're not thinking of why positive shift should decrease the z-axis? just trying to weigh the pros and cons of breaking. i also would expected z to increase but maybe explaining in the doc string why it's this way could warn users and avoid breaking.

I read through the entire module earlier and didn't notice much good reason for this. Also this counter-intuitive behavior is not explained in docstring/comment either. Tried to ping the original author (@CifLord I believe) but no reply yet.

pymatgen/core/surface.py
DanielYang59: Yes it does, two tests were broken but I forgot exactly which one because [previous reversing of shifting direction breaks another two tests](https://github.com/materialsproject/pymatgen/pull/3748#discussion_r1561953247).

Even taking previous "positive shift value to move atoms downwards along z-axis" into account, this seems a typo to me.

DanielYang59:
> thanks @DanielYang59 and sorry for the slow reply. 

No problem at all @janosh! 

> this all looks good. [just had 1 question](https://github.com/materialsproject/pymatgen/pull/3748/files#r1583984225)

Sorry it seems I wasn't making my intention clear enough, as I haven't started actually making changes yet. I proposed [several issues to address](https://github.com/materialsproject/pymatgen/pull/3748#issue-2238929115) but I want to hear about your comments/suggestions before getting my hands on it , because most of them seem more or less breaking (breaking in a way to fix issues, not for enhancement).

janosh:
> Sorry it seems I wasn't making my intention clear enough, as I haven't actually started making changes yet. I proposed https://github.com/materialsproject/pymatgen/pull/3748#issue-2238929115 but I want to hear about your comments/suggestions before getting my hands on it, because most of them seem more or less breaking (breaking in a way to fix issues, not for enhancement).

that i understood. but i was tentatively advising against all but one of the breaking changes you suggested (the one you clearly identified as a bug). how much does that leave to change here?

DanielYang59:
> that i understood. but i was tentatively advising against all but one of the breaking changes you suggested (the one you clearly identified as a bug). how much does that leave to change here?

Surely, let's focus on bug fixes then. Not sure if you mean "site positions change during`center_slab` loop"? This is clearly a bug, and the `get_d` unbound bug should already have been fixed during previous `pyright` fixes. Not sure about the change, should not be a lot, let's see.

pymatgen/core/surface.py
```diff
@@ -762,34 +745,19 @@ def center_slab(slab: Slab) -> Slab:
...
-
-    # If structure is case 2, shift all the sites
-    # to the other side until it is case 1
-    for site in slab:  # DEBUG (@DanielYang59): Slab position changes during loop?
-        # DEBUG (@DanielYang59): sites below z=0 is not considered (only check coord > c)
-        if any(nn[1] >= slab.lattice.c for nn in slab.get_neighbors(site, cutoff_radius)):
-            # TODO (@DanielYang59): the magic offset "0.05" seems unnecessary,
-            # as the Slab would be centered later anyway
-            shift = 1 - site.frac_coords[2] + 0.05
-            slab.translate_sites(all_indices, [0, 0, shift])
```
DanielYang59, line 785: I think this function is overly complicated, we just need to:
1. Check the slab region to make sure there is only one continuous slab layer (with `get_slab_regions` function).
2. If the `slab` has multiple slab regions, it would be quite tricky as we're not sure if the slab region is broken by periodic boundary (a good `get_slab_regions` should handle this case though) or really being non-continuous.
3. Move the center of the slab to z=0.5 (fractional coordinate).

pymatgen/core/surface.py
```diff
@@ -732,108 +829,36 @@ def get_equi_sites(slab: Slab, sites: list[int]) -> list[int]:
 def center_slab(slab: Slab) -> Slab:
```
DanielYang59, line 829: There is a huge pitfall here. The `center_slab` function actually takes a `slab` (`Structure` typed) instead of a real `Slab`. But `Slab` takes a few additional args.

pymatgen/core/surface.py
janosh: feel free to refactor! not sure how good the test coverage is. if poor, it's often good to write additional tests first before refactoring to make sure no previous behavior was changed. or if changes are necessary, that we document all of them in the changelog

DanielYang59: Fold this for now, would address in a future PR (to avoid too many breaking changes in a single PR.)

pymatgen/core/surface.py
```diff
@@ -1314,15 +1295,13 @@ def get_z_ranges(bonds: dict[tuple[Species | Element, Species | Element], float]
                                     z_ranges.extend([(0, z_range[1]), (z_range[0] + 1, 1)])
 
                                 # Neglect overlapping positions
-                                elif z_range[0] != z_range[1]:
-                                    # TODO (@DanielYang59): use the following for equality check
-                                    # elif not isclose(z_range[0], z_range[1], abs_tol=tol):
+                                elif not isclose(z_range[0], z_range[1], abs_tol=tol):
```
DanielYang59, line 1298: Ping myself: I'm pretty sure this breaks some unit test.

DanielYang59: Would revert this and fix this a separate PR.

DanielYang59:
@janosh Please review this. 

Everything is looking good and simple with the argument name clarified. Please refer to the head of this PR for details. Thanks!

DanielYang59:
@janosh Sorry for pinging you everywhere. Can you please comment on this one as well?

And sorry for proposing another potentially breaking change, but in this case `get_slab` should be a private method (though we're seeing it being call outside). So I assume a one year deprecation year should be pretty safe.

Or if you prefer not to rename this, we could clarify in the docstring :)

DanielYang59:
@janosh I decide to minimize (potentially) breaking changes and just clarify in the docstring.

janosh (approved):
thanks @DanielYang59, lgtm. just a conflict to resolve

DanielYang59:
@janosh thanks for reviewing. Resolved

pymatgen/core/surface.py
```diff
@@ -1231,25 +1222,25 @@ def get_slabs(
                 sorted by the number of bonds broken.
         """
 
-        def gen_possible_shifts(ftol: float) -> list[float]:
-            """Generate possible shifts by clustering z coordinates.
+        def gen_possible_terminations(ftol: float) -> list[float]:
```
janosh, line 1225: @DanielYang59 this rename looks like a breaking change? am i missing something?

DanielYang59: This is an internal (private) function of `get_slabs`, I don't expect this to be breaking :)

janosh: ah, i was on my phone and missed the double indent. always better to review on a big screen :)

DanielYang59: All good. But I'm not sure what caused all CI runs to fail? Perhaps some issue with the latest `chgnet`?

PR #3757, opened 2024-04-13
`pyright` fixes for `ext/io/phonon/symmetry/transformations/util/vis/dev_scripts` and improve `io.lobster`

DanielYang59:
## Summary

Started by @ml-evs in #3646:
- Install `pyright` to CI (delayed until all chunks are finished)
- Related `pyright` fixes
- Improvement on `io.lobster` by @JaGeo 
- Format tweaks (`sourcery`), I should not have done this (make it hard to view difference)

### Need confirmation on the following:
- [ ] https://github.com/materialsproject/pymatgen/pull/3757#discussion_r1564052381

####  `pyright` is commented out until all chunks are finished, otherwise everyone's CI would fail.

There are too many (713) `pyright` errors to fix in a single PR, I'm planning to chop up into chunks. First chunk would fix the following modules:
- [x] vis
- [x] util
- [x] transformations
- [x] symmetry
- [x] phonon
- [x] io
- [x] ext
- [x] dev_scripts

### Side note

- [ ] Unit test of `io.lobster.inputs` only failed on my MacOS machine: https://github.com/materialsproject/pymatgen/pull/3757#discussion_r1564074174
- [ ] `io.lobster` `diff` method could potentially be simplified: https://github.com/materialsproject/pymatgen/pull/3757#discussion_r1565120000

## Summary by CodeRabbit

- **New Features**
  - Enhanced the logic for setting calculation parameters based on symmetry and time reversal considerations.
  - Introduced better error handling and contextual exception messages across various modules.
  - Improved file handling with UTF-8 encoding for consistent data processing.

- **Bug Fixes**
  - Fixed issues with string formatting and variable initialization that affected data parsing and output consistency.
  - Corrected logical errors in conditional statements that impacted computational routines.

- **Refactor**
  - Streamlined string formatting using f-strings for better readability and performance.
  - Simplified and optimized various methods and functions for enhanced maintainability and clarity.

- **Documentation**
  - Updated method descriptions and error messages to be more informative and user-friendly.

- **Style**
  - Standardized variable naming and method calls to adhere to project coding standards.

- **Tests**
  - Adjusted file encoding in test setups to ensure compatibility and correctness during testing.


pymatgen/io/pwscf.py
```diff
@@ -334,10 +335,12 @@ def input_mode(line):
                         coords_are_cartesian = True
                     elif mode[1] == "crystal":
                         coords_are_cartesian = False
+
         structure = Structure(
             Lattice(lattice),
             species,
             coords,
+            # DEBUG (@DanielYang59): need input on coords_are_cartesian
```
DanielYang59, line 343: Need help on `io.pwscf` @janosh (I never use `pwscf`): `coords_are_cartesian` is only defined in one of the many branches, and is unbound elsewhere.

pymatgen/io/lobster/inputs.py
```diff
@@ -551,10 +555,9 @@ def write_KPOINTS(
                 kpts.append(f)
                 weights.append(0.0)
                 all_labels.append(labels[k])
-        ISYM = isym
-        comment = f"{ISYM=}, grid: {mesh} plus kpoint path" if line_mode else f"{ISYM=}, grid: {mesh}"
+        comment = f"{isym=}, grid: {mesh} plus kpoint path" if line_mode else f"{isym=}, grid: {mesh}"
 
-        KpointObject = Kpoints(
+        kpoint_object = Kpoints(
```
DanielYang59, line 560: I renamed a few internal variable to snake_case @JaGeo, let me know if you prefer not to change them :)

DanielYang59: It's weird, two of the `io.lobster.inputs` tests failed on my local test but passed in CI:
```python
(venv) yang@Yang-MacBook pymatgen % pytest tests/io/lobster/test_inputs.py             
........................F........F...........................                                                              [100%]
============================================================ FAILURES ============================================================
_____________________________________________ TestLobsterin.test_dict_functionality ______________________________________________

self = <lobster.test_inputs.TestLobsterin testMethod=test_dict_functionality>

    def test_dict_functionality(self):
>       assert self.Lobsterinfromfile.get("COHPstartEnergy") == -15.0
E       AssertionError

tests/io/lobster/test_inputs.py:1773: AssertionError
____________________________________________ TestLobsterin.test_read_write_lobsterin _____________________________________________

self = <lobster.test_inputs.TestLobsterin testMethod=test_read_write_lobsterin>

    def test_read_write_lobsterin(self):
        outfile_path = tempfile.mkstemp()[1]
        lobsterin1 = Lobsterin.from_file(f"{TEST_FILES_DIR}/cohp/lobsterin.1")
        lobsterin1.write_lobsterin(outfile_path)
>       lobsterin2 = Lobsterin.from_file(outfile_path)

tests/io/lobster/test_inputs.py:1793: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'pymatgen.io.lobster.inputs.Lobsterin'>, lobsterin = '/var/folders/nn/rmcc0j095h53jgz4gj734q900000gn/T/tmpf82ed5jw'

    @classmethod
    def from_file(cls, lobsterin: str) -> Self:
        """
        Args:
            lobsterin (str): path to lobsterin.
    
        Returns:
            Lobsterin object
        """
        with zopen(lobsterin, mode="rt") as file:
            data = file.read().split("\n")
        if len(data) == 0:
            raise OSError("lobsterin file contains no data.")
        Lobsterindict: dict[str, Any] = {}
    
        for datum in data:
            # Remove all comments
            if not datum.startswith(("!", "#", "//")):
                pattern = r"\b[^!#//]+"  # exclude comments after commands
                if matched_pattern := re.findall(pattern, datum):
                    raw_datum = matched_pattern[0].replace("\t", " ")  # handle tab in between and end of command
                    key_word = raw_datum.strip().split(" ")  # extract keyword
                    if len(key_word) > 1:
                        # check which type of keyword this is, handle accordingly
                        if key_word[0].lower() not in [datum2.lower() for datum2 in Lobsterin.LISTKEYWORDS]:
                            if key_word[0].lower() not in [datum2.lower() for datum2 in Lobsterin.FLOAT_KEYWORDS]:
                                if key_word[0].lower() not in Lobsterindict:
                                    Lobsterindict[key_word[0].lower()] = " ".join(key_word[1:])
                                else:
                                    raise ValueError(f"Same keyword {key_word[0].lower()} twice!")
                            elif key_word[0].lower() not in Lobsterindict:
>                               Lobsterindict[key_word[0].lower()] = float(key_word[1])
E                               ValueError: could not convert string to float: 'None'

pymatgen/io/lobster/inputs.py:599: ValueError
```

pymatgen/io/abinit/abiobjects.py
```diff
@@ -720,14 +720,9 @@ def __init__(
...
-                kptopt = 1
-            if not use_symmetries and use_time_reversal:
-                kptopt = 2
-            if not use_symmetries and not use_time_reversal:
-                kptopt = 3
-            if use_symmetries and not use_time_reversal:
-                kptopt = 4
+            kptopt = 1 if use_symmetries else 2
+            if not use_time_reversal:
+                kptopt += 2
```
JaGeo, line 725: I think this is wrong now.
If use_symmetries and not use_time_reversal: 4 in old verdion but 3 in the new one.

pymatgen/io/abinit/pseudos.py
```diff
@@ -843,13 +844,15 @@ def tm_header(filename, ppdesc):
...
+        proj_info = []
+        idx = None
         for idx in range(2 * (lmax + 1)):
             line = lines[idx]
             if idx % 2 == 0:
                 proj_info = [
                     line,
                 ]
-            if idx % 2 == 1:
+            elif idx % 2 == 1:
```
JaGeo, line 855:
```suggestion
            elif idx % 2 == 1:
```

I am confused by this change as well. I don't think elif is correct here

pymatgen/io/abinit/abiobjects.py
DanielYang59: Yes you are right! Thanks for catching this. I would revert this now.

pymatgen/io/lobster/inputs.py
JaGeo: Not sure why they are failing now. Why is there suddenly a 'None' where there should be a float.

pymatgen/io/abinit/pseudos.py
DanielYang59: Thanks for catching this! 

This looks correct but unnecessary to me? (`idx % 2` could be either 0 or 1) Can you elaborate on your reason?

 I would remove the redundant check.

pymatgen/io/lobster/inputs.py
DanielYang59: I don't know... I have to assume there is something wrong with either my machine or there is some platform-related issue (as I just tried to run test on the unmodified master branch and still got the same error, so I assume this is not related to this PR). Is test failing on your local machine?

Also I tried to add explicit encoding to both `io.lobster.inputs` and its unit tests, but didn't fix the problem.

pymatgen/io/lobster/inputs.py
```diff
@@ -156,6 +157,7 @@ def __setitem__(self, key, val):
     def __getitem__(self, item):
         """Implements getitem from dict to avoid problems with cases."""
         found = False
```
JaGeo, line 159:
```suggestion
        new_key = item
        for key_here in self:
            if item.strip().lower() == key_here.lower():
                new_key = key_here
```

Can you replace the while function with this? My code way overly complicated and the None is unnecessar.

pymatgen/io/fiesta.py
```diff
@@ -173,7 +173,7 @@ def bse_run(self):
                 stdout=fout,
             )
 
-        if self.folder != os.getcwd():
+        if self.folder != init_folder:
```
DanielYang59, line 176: This is not right. I would correct it.

pymatgen/io/lobster/inputs.py
```diff
@@ -182,6 +184,7 @@ def diff(self, other):
         similar_param = {}
         different_param = {}
         key_list_others = [element.lower() for element in other]
+        new_key = None
```
JaGeo, line 187:
```suggestion
        new_key = None
```

I am a bit worried about those. I think we should rather name them differently

pymatgen/io/lobster/inputs.py
JaGeo: Okay, I see. Yes, could be platform related. Will check again on my machine but maybe tomorrow. 😅 I also feel some of the code needs a bit of refactoring that I don't have time for at the moment.

pymatgen/io/abinit/pseudos.py
JaGeo: Ah. Sorry. I also misunderstood something here. Yes, you are right.

pymatgen/io/lobster/inputs.py
DanielYang59: You could just push your fixes here, no problem at all and I would deeply appreciate your help!

Currently I'm just trying to assign some of the possibly undefined variables with a default `None` value to prevent variables from being unbound (which doesn't alter the code logic by any means), which means if the code is designed in a way that could exit the loop without being assigned at all, the code would still fail.

For example:
```python
idx = None  # added in this PR

for idx in range(value):
    if idx % 2 == 0:
        do something

print(lines[index])  # might throw an error 

```

If is an irrational is assigned to `value`(for example `-1`), the loop would not be entered and `idx` would still be `None`. If we don't assign `idx = None`, it would then become a `NameError`. 

Seemingly this doesn't improve the robustness of the code at all, but my hope is this would be the first step, afterwards we would add type annotations and checks to ensure `idx` is a `int`, if not the code should exist gracefully. Sorry it's just too much work 😅, so I'm not planning to refine the code logic at least in this PR .

pymatgen/io/abinit/pseudos.py
DanielYang59: No worries at all. Thanks for asking and you still help me find a small improvement for this 😄

pymatgen/io/lobster/inputs.py
JaGeo: I don't have this error on Linux at all. Not sure how this happens. Could it be related to the files?

DanielYang59: I haven't got time to look closer into the reason sorry. I would ping you if I find anything. 

I just tried to run tests on CodeSpace (Linux runner) and no tests fail. I would assume there's something wrong with my machine but I don't know what exactly is. 

Actually I have met with similar issues where the `numpy` is missing and `ImportError` is suppressed, but unfortunately it isn't this time. I would have a look soon and see if I could find anything :)

pymatgen/io/lobster/inputs.py
DanielYang59: Thanks for fixing. Things look much cleaner now :)

pymatgen/io/lobster/inputs.py
JaGeo: Thanks!
Maybe a change in the test files format?

pymatgen/io/lobster/inputs.py
JaGeo: Yeah. This was written a very long time ago without thinking too much about the coding style 🙃. We will find similar things all over the code.
Once this is merged, I will check if any of this breaks LobsterPy (e.g., renaming of some ValueError messages). 
Thanks for working on these issues!

JaGeo: Maybe, we have to go through this one as well. I guess the initialisation here with None is probably also not necessary.

pymatgen/io/lobster/inputs.py
```diff
@@ -272,6 +260,7 @@ def write_lobsterin(self, path="lobsterin", overwritedict=None):
                     self[key] = entry
 
         filename = path
+        new_key = None
```
JaGeo, line 263: Same here. I am a bit worried about side effects from this initialisation

pymatgen/io/lobster/inputs.py
DanielYang59: No problem at all!

> Maybe a change in the test files format?

I tried to explicitly set the encoding in both `lobster.inputs` and its test which unfortunately doesn't help.

I prefer not to make too many guesses before I actually look into it, to avoid causing confusion to you and everyone else who might be watching. 

For some reason I always notice unit test fails on my local machine (OS set to English, though I speak Mandarin but I work/study in English) but passed in CI, it's super weird as I thought MacOS is also Unix-based, but could also suggest some inconsistency. And that's the reason I always add explicit encoding as `utf-8` to `with open()`, which is also suggested by `pylint`. I had a discussion with Janosh on this before, as the default varies from platform to platform. From the [official docs](https://docs.python.org/3/library/functions.html#open):
> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used. See the [codecs](https://docs.python.org/3/library/codecs.html#module-codecs) module for the list of supported encodings.

and 

>locale.getencoding()
    Get the current [locale encoding](https://docs.python.org/3/glossary.html#term-locale-encoding):
        On Android and VxWorks, return "utf-8".
        On Unix, return the encoding of the current [LC_CTYPE](https://docs.python.org/3/library/locale.html#locale.LC_CTYPE) locale. Return "utf-8" if nl_langinfo(CODESET) returns an empty string: for example, if the current LC_CTYPE locale is not supported.
        On Windows, return the ANSI code page.

I'm not sure if it's encoding or else now. But I would investigate soon and let you know :)

JaGeo: Thanks!

DanielYang59: No worries at all. 

I just did a quick (not thorough) investigation and found there is issue with the unit test, and wondering why this didn't show up in Unix/Windows runners at all...

The buggy test is the following (I removed unrelated part):
```python
class TestLobsterin(TestCase):
    def test_dict_functionality(self):
        Lobsterinfromfile = Lobsterin.from_file(f"{TEST_FILES_DIR}/cohp/lobsterin.1")
        assert Lobsterinfromfile.get("COHPstartEnergy") == -15.0
```

But the `from_file` method (wondering why this looks so complex) actually `lower()` all keys:
```python
    @classmethod
    def from_file(cls, lobsterin: str) -> Self:
        """
        Args:
            lobsterin (str): path to lobsterin.

        Returns:
            Lobsterin object
        """
        with open(lobsterin, mode="rt") as file:
            data = file.read().split("\n")
        if len(data) == 0:
            raise OSError("lobsterin file contains no data.")
        Lobsterindict: dict[str, Any] = {}

        for datum in data:
            # Remove all comments
            if not datum.startswith(("!", "#", "//")):
                pattern = r"\b[^!#//]+"  # exclude comments after commands
                if matched_pattern := re.findall(pattern, datum):
                    raw_datum = matched_pattern[0].replace("\t", " ")  # handle tab in between and end of command
                    key_word = raw_datum.strip().split(" ")  # extract keyword
                    if len(key_word) > 1:
                        # check which type of keyword this is, handle accordingly
                        if key_word[0].lower() not in [datum2.lower() for datum2 in Lobsterin.LISTKEYWORDS]:
                            if key_word[0].lower() not in [datum2.lower() for datum2 in Lobsterin.FLOAT_KEYWORDS]:
                                if key_word[0].lower() not in Lobsterindict:
                                    Lobsterindict[key_word[0].lower()] = " ".join(key_word[1:])
                                else:
                                    raise ValueError(f"Same keyword {key_word[0].lower()} twice!")
                            elif key_word[0].lower() not in Lobsterindict:
                                Lobsterindict[key_word[0].lower()] = float(key_word[1])
                            else:
                                raise ValueError(f"Same keyword {key_word[0].lower()} twice!")
                        elif key_word[0].lower() not in Lobsterindict:
                            Lobsterindict[key_word[0].lower()] = [" ".join(key_word[1:])]
                        else:
                            Lobsterindict[key_word[0].lower()].append(" ".join(key_word[1:]))
                    elif len(key_word) > 0:
                        Lobsterindict[key_word[0].lower()] = True

        return cls(Lobsterindict)
```

I'm sure you should have noticed the problem now, in unit test it stills get the value with the original key: `assert Lobsterinfromfile.get("COHPstartEnergy") == -15.0`, and it's thus default to `None`. I'm now just wondering why this doesn't show up elsewhere?

JaGeo: I rewrote the get part, right? It should use the lower version of it? I am not sure I can follow.

JaGeo: Probably one can simplify quite a lot here as well....

DanielYang59: >  It should use the lower version of it? 

Yes the unit test should use the lowered key `Lobsterinfromfile.get("cohpstartenergy")`, otherwise it would get the default `None`. 

> Probably one can simplify quite a lot here as well....

I would be more than happy to help with this but not in this PR. I'm not intending to include any functional changes (perhaps including code simplication) here, which might make the whole PR hard to track and review 😄

DanielYang59: > This is actually a test to test this?

Yes, see here: https://github.com/materialsproject/pymatgen/blob/37d1066a8f45b24dff8cf8c18b2549648897afce/tests/io/lobster/test_inputs.py#L1772-L1787

JaGeo: No, it should not use the lower version of it. This is wrong.

JaGeo: My guess is that this part here is not working on your machine:
https://github.com/DanielYang59/pymatgen/blob/9ec81411c35b86fbd3b352bd2232139eb30a29db/pymatgen/io/lobster/inputs.py#L151

I am not sure why.

JaGeo: In any case, I have to go through this implementation to make sure everything is consistent. I feel a few things aren't. Nevertheless, this unit test should work.

DanielYang59: I'm still using the code from the master branch, I'm not sure if you have made any correction or not.

I minimized the unit test:
```python
class TestLobsterin(TestCase):
    def test_dict_functionality(self):
        Lobsterinfromfile = Lobsterin.from_file(f"{TEST_FILES_DIR}/cohp/lobsterin.1")
        print(f"{Lobsterinfromfile=}")
        print(Lobsterinfromfile.get("COHPstartEnergy"))
        assert Lobsterinfromfile.get("COHPstartEnergy") == -15.0
```

And it returns:
```python
=========================================================== FAILURES ============================================================
_____________________________________________ TestLobsterin.test_dict_functionality _____________________________________________

self = <lobster.test_inputs.TestLobsterin testMethod=test_dict_functionality>

    def test_dict_functionality(self):
        Lobsterinfromfile = Lobsterin.from_file(f"{TEST_FILES_DIR}/cohp/lobsterin.1")
        print(f"{Lobsterinfromfile=}")
        print(Lobsterinfromfile.get("COHPstartEnergy"))
>       assert Lobsterinfromfile.get("COHPstartEnergy") == -15.0
E       AssertionError

tests/io/lobster/test_inputs.py:18: AssertionError
----------------------------------------------------- Captured stdout call ------------------------------------------------------
Lobsterinfromfile={'cohpstartenergy': -15.0, 'cohpendenergy': 5.0, 'basisset': 'pbeVaspFit2015', 'gaussiansmearingwidth': 0.1, 'basisfunctions': ['Fe 3d 4p 4s', 'Co 3d 4p 4s'], 'skipdos': True, 'skipcohp': True, 'skipcoop': True, 'skippopulationanalysis': True, 'skipgrosspopulation': True}
None
===================================================== slowest 30 durations ======================================================

(3 durations < 0.005s hidden.  Use -vv to show these durations.)
1 failed in 1.64s
```

JaGeo: Yeah, but only on your machine.

JaGeo: I did not make any modifications. The whole unit test passes for me.

DanielYang59: Yes it's weird and I don't know why. I saw you have `__get/setitem__` methods to handle case sensitivity but seem not working on my side. Maybe @janosh would have some insights?

In any case, we might enjoy the rest of the Sunday and come back tomorrow with clearer eyes. Thanks for trying to fix this!

JaGeo: Yes 😅. I will just make the del consistent. There is an issue with the logic there and then we check next week 

Thanks!

DanielYang59: Thanks!

I would try to give a hand on simplifying the logic here when we finish cleaning up unbound vars on this one.

Enjoy the weekend!

pymatgen/io/lobster/inputs.py
```diff
@@ -186,42 +180,35 @@ def diff(self, other):
...
-                        }
-                elif v1 != other[new_key]:
-                    different_param[k1.upper()] = {
+                different_param[k1.lower()] = {"lobsterin1": v1, "lobsterin2": None}
+            elif isinstance(v1, str):
+                if v1.strip().lower() != other[k1lower].strip().lower():
+                    different_param[k1.lower()] = {
                         "lobsterin1": v1,
-                        "lobsterin2": other[new_key],
+                        "lobsterin2": other[k1lower],
```
JaGeo, line 188: Have to check this lart again. I think I overlooked something here.

JaGeo: Fixed.

DanielYang59:
I just changed some error types for `io.lobster` in 72981f1d7bd6f4a62d2fa063652e5588ae3e0049. I'm not very good at deciding the type of errors as well, but it seems most usage of `OSError` is inappropriate in this case. From the [official docs](https://docs.python.org/3/library/exceptions.html):
> exception OSError(errno, strerror[, filename[, winerror[, filename2]]])

> This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found” or “disk full” (not for illegal argument types or other incidental errors).

It should be raised in cases of OS-related issues (like unable to read/write file, unable to start/close a process or such). In our case the file is read just fine, but the file content is empty. Perhaps we should raiss `RuntimeError`, or perhaps write a custom `EmptyInputFileError`?

pymatgen/io/lobster/inputs.py
DanielYang59: Thanks for fixing! 

I'm currently getting the following `ruff` error:
```python
pymatgen/io/lobster/inputs.py:210:13: SIM102 Use a single `if` statement instead of nested `if` statements
```
from https://github.com/materialsproject/pymatgen/blob/aee9bc49b4fe95857b11d23eb279fb0a60d8bc60/pymatgen/io/lobster/inputs.py#L209-L213

I would correct it if it's not intended.

Meanwhile, I didn't try myself so I could be wrong, do we need to hard-code a `diff` method ourselves? Have you considered using the [symmetric_difference](https://docs.python.org/3/library/stdtypes.html#frozenset.symmetric_difference) to calculate the intersection and difference?

JaGeo:
> I just changed some error types for `io.lobster` in [72981f1](https://github.com/materialsproject/pymatgen/commit/72981f1d7bd6f4a62d2fa063652e5588ae3e0049). I'm not very good at deciding the type of errors either, but it seems most usage of `OSError` is inappropriate in this case. From the [official docs](https://docs.python.org/3/library/exceptions.html):
> 
> > exception OSError(errno, strerror[, filename[, winerror[, filename2]]])
> 
> > This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found” or “disk full” (not for illegal argument types or other incidental errors).
> 
> It should be raised in cases of OS-related issues (like unable to read/write file, unable to start/close a process or such).
> 
> In our case the file is read just fine, but the file content is empty. Perhaps we should raise a general `RuntimeError`, or perhaps write a custom `EmptyInputFileError` which could then be reused around the entire code base?

That's fine. I am also fine with `RuntimeError`. I would just like to avoid spending too much time on it.

pymatgen/io/lobster/inputs.py
JaGeo: It's again implemented to take care of cases.

I am not sure if this is fine. I would need to have a careful look.

JaGeo: Yes, it's fine!

tests/io/lobster/test_inputs.py
```diff
@@ -2590,7 +2588,7 @@ def test_raises(self):
             self.hamilton_matrices = LobsterMatrices(filename=f"{TEST_FILES_DIR}/cohp/Na_hamiltonMatrices.lobster.gz")
 
         with pytest.raises(
-            OSError,
+            RuntimeError,
```
janosh, line 2591: i think `IOError` would be the most fitting built-in exception class here

janosh (commented):
thanks for this concerted clean up effort @DanielYang59 @JaGeo! 👍 

this all looks good. just one minor comment re appropriate exception class

tests/io/lobster/test_inputs.py
DanielYang59: Hi Janosh. Thanks for reviewing! However I don't quite agree on the usage of `IOError` in this case, similar to my reason against `OSError` here https://github.com/materialsproject/pymatgen/pull/3757#issuecomment-2054362292.

Maybe I don't understand it correctly, but I think `IOError` should be used in cases where there is a IO issue. For some reason the official docs doesn't give details on this, but from [this](https://www.tutorialspoint.com/How-to-catch-IOError-Exception-in-Python):
> IOError Exception

> It is an error raised when an input/output operation fails, such as the print statement or the open() function when trying to open a file that does not exist. It is also raised for operating system-related errors.

In our case, the file opens just fine but the content is empty, so I don't think it's actually an IO issue.

pymatgen/io/lobster/inputs.py
DanielYang59: > It's again implemented to take care of cases.

Yes I begin to understand your intention in `io.lobster` now (mostly to handle the case sensitivity issue).

> It looks fine on first sight.

I would experiment on this later and tell you how everything goes (maybe a bit later, I'm currently taking up too much work).

Thanks!

tests/io/lobster/test_inputs.py
janosh: fair enough, i'm definitely ok with `RuntimeError`

DanielYang59:
My last bit of concern about this PR would be the default coordinate system of `pwscf` here https://github.com/materialsproject/pymatgen/pull/3757#discussion_r1564052381. Please help me confirm this. Thanks!

tests/io/lobster/test_inputs.py
DanielYang59: All good!

pymatgen/io/pwscf.py
janosh: wow, so many code branches and only one of them was actually tested?

https://github.com/materialsproject/pymatgen/blob/37d1066a8f45b24dff8cf8c18b2549648897afce/tests/io/test_pwscf.py#L360-L367

this seems in urgent need of more tests. i think defaulting `coords_are_cartesian=False` seems right. i also never used this code so it's hard to know for sure. apparently no one does or this would have been reported long ago

DanielYang59: > so many code branches and only one of them was actually tested?

Yes, `coords_are_cartesian` is defined in only one of the few branches and is unbound all elsewhere 😓, meaning the other branches were likely never entered.

DanielYang59:
Thanks for reviewing @janosh and great collaboration on this one @JaGeo.

Sorry for hijacking this topic @ml-evs.

ml-evs:
> Sorry for hijacking this topic @ml-evs.

No problem, this is great work! It looks this was fairly comprehensive; I'll try reopening my PR that adds the CI runs that triggered this.

JaGeo:
> > Sorry for hijacking this topic @ml-evs.
> 
> No problem, this is great work! It looks this was fairly comprehensive; I'll try reopening my PR that adds the CI runs that triggered this.

This is only for some modules. More work is needed

DanielYang59:
> This is only for some modules. More work is needed

Yes! It's only like 1/3 of the all work needed. I would ping you once it's all ready. @ml-evs

janosh:
> Yes! It's only like 1/3 of the all work needed. I would ping you once it's all ready. @ml-evs

@DanielYang59 thanks a lot for donating your time to work on this! 👍

DanielYang59:
> @DanielYang59 thanks a lot for donating your time to work on this! 👍

Glad I can help :)

DanielYang59:
@janosh I'm still experimenting on an issue I met with @JaGeo in https://github.com/materialsproject/pymatgen/pull/3757#discussion_r1564074174. I would appreciate it if you could give me some advice 😄 

I just created a minimal code to recreate this issue:
```python
from collections import UserDict

class TestClass(UserDict):

    def __init__(self, settingsdict: dict):
        super().__init__()
        self.update(settingsdict)

    def __setitem__(self, key, val):
        """Remove case sensitivity."""
        new_key = next((key_here for key_here in self if key.strip().lower() == key_here.lower()), key)

        super().__setitem__(new_key, val)

    def __getitem__(self, item):
        """Implements getitem from dict to avoid problems with cases."""
        new_item = next((key_here for key_here in self if item.strip().lower() == key_here.lower()), item)

        return super().__getitem__(new_item)


if __name__ == "__main__":
    test_dict = {"ConTent": "hello", }
    test_instance = TestClass(test_dict)

    print(f"{test_instance=}")
    print(test_instance.get("content"), test_instance["content"])
```

On this same machine (MacOS Sonoma 14.4.1, Apple Silicon and Python managed by miniconda3).

With Python 3.10 and 3.11, I got:
```python
(base) yang@Yang-MacBook test % python3 -V
Python 3.10.14
(base) yang@Yang-MacBook test % python3 recreate.py 
test_instance={'ConTent': 'hello'}
hello hello
```

```python
(base) yang@Yang-MacBook test % python3 -V
Python 3.11.8
(base) yang@Yang-MacBook test % python3 recreate.py 
test_instance={'ConTent': 'hello'}
hello hello
```

While with Python 3.12, the two methods get different results:
```python
(base) yang@Yang-MacBook test % python3 -V               
Python 3.12.2
(base) yang@Yang-MacBook test % python3 recreate.py      
test_instance={'ConTent': 'hello'}
None hello
```

JaGeo:
Ah, interesting, Python version 😅. 
This is why it has not shown up on CI.

DanielYang59:
> @DanielYang59 I think I found it: [python/cpython#105524](https://github.com/python/cpython/issues/105524)

👍 So speedy. Yes looks like exactly what we're experiencing. I guess it's not intended?

The `get` method:
```python
    def get(self, key, default=None):
        if key in self:
            return self[key]
        return default
```

Which indeed does not use `__getitem__`

JaGeo:
I think we recently switched to UserDict for other issues.

DanielYang59:
> there's only 1 use of `UserDict` in all of `pmg` i think. we can try the [proposed fix](https://github.com/python/cpython/issues/105524#issuecomment-1610750842) or try to make it work without `UserDict`. the [docs weakly recommend to subclass `dict` directly](https://docs.python.org/3/library/collections.html#userdict-objects) instead of `UserDict`

I would keep it in mind. I'm now [secretly working on simplifying `io.lobster.inputs`](https://github.com/DanielYang59/pymatgen/tree/tweak-lobster) (prefer not to push it here for now because I'm now having much time to work on this).

PR #3758, opened 2024-04-14
Convert `kpts` in `Kpoints` to `Sequence[tuple]` and set it as `property`

DanielYang59:
## Summary

As discussed in https://github.com/materialsproject/pymatgen/commit/6a06f3cb0a8b551ed782d86ed5963cd8aaed6674#r140944147.

- Convert `kpts` in `Kpoints` to `tuple[float, float, float]`, and related type fixes
- Set `kpts` as `property` and add its `setter`
- Add unit tests for this property
- Relocate custom types (Vector3D Matrix3D SitePropsType) to `util.typing`


DanielYang59:
I just opened this early PR @janosh. Seems a lot of work and I'm quite concerned this might break many others (in typing, I don't expect much functional change thought, except for mutability).

Would you suggest `tuple[float, float, float]` or `tuple[int, int, int]` for each single kpoint (compatibility for 0-decimal float? )?

janosh:
I would prefer typing as int since float will still work and only cause a type error

DanielYang59:
> I would prefer typing as int since float will still work and only cause a type error

I agree with you on this, we should not do cater too much to edge case (0-decimal `float`) in this case, `int` is much clearer for `Kpoint`.

Shouldn't we add another custom type to `util.typing` for simplicity?
```python
Kpoint = tuple[int, int, int]
```

Maybe do the same to coordinates in the near future?
```python
Coordinate = tuple[float, float, float]
```

Meanwhile, maybe we should consider using a more general `Sequence[Kpoint]` instead of `Tuple[Kpoint]`? Currently we just want to hint developers/users that `kpts` is a "collection" of `Kpoint`, and using `tuple` for `Kpoint` could provide fixed length. Do we have a good reason for immutability for the collection itself (I thought it might be natural to add/remove `Kpoint` from the collection)?

janosh:
> Shouldn't we add another custom type to `util.typing` for simplicity (which also ensures that we don't need to change types everywhere like we're doing now, in case we need to change the type of Kpoint in the future)?

good idea!

> Maybe do the same to coordinates in the near future?

yes, i forgot to add could you move the `Vector3D` type to `util/typing.py`? not sure about renaming to `Coordinate`. `Vector3D` sounds more general to me.

https://github.com/materialsproject/pymatgen/blob/37d1066a8f45b24dff8cf8c18b2549648897afce/pymatgen/core/trajectory.py#L30

> Meanwhile, maybe we should consider using a more general Sequence[Kpoint] instead of tuple[Kpoint]? Currently we just want to hint developers/users that kpts is a "collection" of Kpoint, and using tuple for Kpoint could provide fixed length. Do we have a good reason for immutability for the collection itself (I thought it might be natural to add/remove Kpoint from the collection)?

that's a good point. let's do it your way

DanielYang59:
> yes, i forgot to add could you move the `Vector3D` type to `util/typing.py`? not sure about renaming to `Coordinate`. `Vector3D` sounds more general to me.
>
> https://github.com/materialsproject/pymatgen/blob/37d1066a8f45b24dff8cf8c18b2549648897afce/pymatgen/core/trajectory.py#L30

Yes sure. `Vector3D` sounds better (more general), and could be also used on types like velocity which is not actually a "coordinate".

> > Meanwhile, maybe we should consider using a more general Sequence[Kpoint] instead of tuple[Kpoint]? Currently we just want to hint developers/users that kpts is a "collection" of Kpoint, and using tuple for Kpoint could provide fixed length. Do we have a good reason for immutability for the collection itself (I thought it might be natural to add/remove Kpoint from the collection)?
> 
> that's a good point. let's do it your way

Great. I would try implement this. Thanks for the feedback.

tests/io/vasp/test_inputs.py
```diff
@@ -810,10 +810,10 @@ class TestKpoints:
     def test_init(self):
         filepath = f"{VASP_IN_DIR}/KPOINTS_auto"
         kpoints = Kpoints.from_file(filepath)
-        assert kpoints.kpts == [[10]], "Wrong kpoint lattice read"
+        assert kpoints.kpts == [(10,)], "Wrong kpoint lattice read"
         filepath = f"{VASP_IN_DIR}/KPOINTS_cartesian"
         kpoints = Kpoints.from_file(filepath)
-        assert kpoints.kpts == [[0.25, 0, 0], [0, 0.25, 0], [0, 0, 0.25]], "Wrong kpoint lattice read"
+        assert kpoints.kpts == [(0.25, 0, 0), (0, 0.25, 0), (0, 0, 0.25)], "Wrong kpoint lattice read"
```
DanielYang59, line 816: Currently the unit test used `float` for testing. 

Seems some inconsistency needs to be cleaned up, as the `Kpoint` under "line-mode" should actually be a coordinate.

pymatgen/io/vasp/inputs.py
```diff
@@ -1112,11 +1114,11 @@ def __init__(
...
                 Kpoints.supported_modes enum types.
-            kpts (2D array): 2D array of kpoints. Even when only a single
+            kpts (2D array): Array of kpoints. Even when only a single
                 specification is required, e.g. in the automatic scheme,
                 the kpts should still be specified as a 2D array. e.g.,
-                [[20]] or [[2,2,2]].
-            kpts_shift (3x1 array): Shift for Kpoints.
+                [(20,),] or [(2, 2, 2),].
+            kpts_shift (3x1 array): Shift for kpoints.
             kpts_weights: Optional weights for kpoints. Weights should be
```
DanielYang59, line 1122: This docstring differs (type of list[int]) from the implementation: https://github.com/materialsproject/pymatgen/blob/ff94faa5bb490354c420f188f1abeac17cb40931/pymatgen/io/vasp/inputs.py#L1524

DanielYang59:
Please review this if you have time @janosh :)

I ended up adding type alias as `Kpoint = Union[tuple[float, float, float], tuple[int,]]`:
- `tuple[int,]` for "automatic" mode: https://github.com/materialsproject/pymatgen/blob/ff94faa5bb490354c420f188f1abeac17cb40931/pymatgen/io/vasp/inputs.py#L1185-L1200
- I have to use `float` for `tuple[float, float, float]` to consider the line mode.

Any suggestion would be appreciated :)

pymatgen/util/typing.py
```diff
@@ -5,30 +5,33 @@
...
     from pymatgen.analysis.cost import CostEntry  # type: ignore[attr-defined]
     from pymatgen.analysis.phase_diagram import GrandPotPDEntry, PDEntry, TransformedPDEntry
     from pymatgen.entries import Entry
     from pymatgen.entries.computed_entries import ComputedEntry, ComputedStructureEntry, GibbsComputedStructureEntry
     from pymatgen.entries.exp_entries import ExpEntry
 
 
-PathLike = Union[str, OsPathLike]
-PbcLike = tuple[bool, bool, bool]
+PathLike: TypeAlias = Union[str, OsPathLike]
```
janosh, line 24: [docs](https://docs.python.org/3/library/typing.html#typing.TypeAlias) recommend not to use `TypeAlias`:

> Deprecated since version 3.12: [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is deprecated in favor of the [type](https://docs.python.org/3/reference/simple_stmts.html#type) statement

pymatgen/io/vasp/inputs.py
```diff
@@ -1151,6 +1152,62 @@ def __init__(
         self.tet_weight = tet_weight
         self.tet_connections = tet_connections
 
+    def __eq__(self, other: object) -> bool:
```
janosh, line 1155: can we add tests for this?

pymatgen/io/vasp/inputs.py
janosh, line 1: are there any new tests that would make sense to add with the changes in this file?

janosh (changes requested):
thanks @DanielYang59, this looks good! just 2 minor comments 👍

pymatgen/util/typing.py
DanielYang59: Thanks! I would fix this!

DanielYang59: I noticed `TypeAlias` is deprecated after Python 3.12 (which we haven't touched yet).

And `type` is new for Python 3.12 too? Maybe remove this type annotation altogether?

pymatgen/io/vasp/inputs.py
DanielYang59: Yes! I added this to the TODO but totally forgot about this. We should have a unit test for this property.

DanielYang59: Added in 596d256f3620cd080621bddbbec60609e77fd74d. Please let me know if you have more suggestion :)

Helped avoided a pitfall: `str` is also a `Sequence` (a sequence of char like C I guess?).

janosh (approved):
great stuff as usual, thanks @DanielYang59! 👍

you're writing excellent tests 🧪

DanielYang59:
> you're writing excellent tests 🧪

Thanks for reviewing and saying that! Appreciate it.

DanielYang59:
@janosh Also worth noting it appears `uv` is less reliable than native `pip`, as you might have noticed in tests for previous several commits where tests failed randomly. I'm not sure how much speedup `uv` gives over native `pip`, but if not that much, I would personally prefer a "slower but more reliable" method than a "faster but less reliable" method (because I assume CI is not performance critical?).

I was always wondering, does GitHub provide some dedicated container for CI tests (like codespace for CI)? As I notice environment preparation time is almost as much as actual test running time, it would be good to avoid building a new test env every time when there is no dependency change?

janosh:
> @janosh Also worth noting it appears uv is less reliable than native pip,

are you referring to this error? 

```py
E       _tkinter.TclError: Can't find a usable tk.tcl in the following directories: 
```

That also used to happen with `pip` occasionally. I wasn't keeping a close eye on CI last week so it's possible i missed other issues but didn't notice anything specific to `uv` other than the `pytorch` installation time out.

> I was always wondering, does GitHub provide some dedicated container for CI tests (like codespace for CI)? 

The `setup-python` GH action, supports caching which we used before with `pip`. But [I read in here](https://github.com/actions/setup-python/issues/822#issuecomment-1963717925) that it's slower than installing from scratch with `uv`. I didn't test this myself, but I also noticed that the post-setup step for creating the `pip` cache could be quite slow so it's possible that for `uv` it's slower than not using a cache

janosh:
> ```sh
> >> Run micromamba activate pmg
> /home/runner/work/_temp/225b20ec-6d19-4bb8-b69d-17e4bb02f7d0.sh: line 2: pytest: command not found
> ```

you're right, i did see that one as well. are you sure this is a `uv` issue? could also be a `micromamba` or a `uv`-`micromamba`  interop issue

Andrew-S-Rosen:
FYI that this is a slight breaking change re: kpts no longer being a list but a tuple, although you likely are aware of this due to the change in the test suite.

DanielYang59:
Hi @Andrew-S-Rosen, yes we're fully aware it might be breaking to change `Kpoint` from a `list` to a `tuple`. Janosh said he prefer it be to immutable wherever possible, and I agree using `tuple` would bring more convenience when typing (`list[float, float, float]` would not work).

But frankly I'm not quite sure if there is any unexpected side effect. Ideally things should only rely on the values of `Kpoint` instead of the type. Any suggestion would be appreciated.

Andrew-S-Rosen:
I don't think there will be too many practical side effects. I just noticed it failing my CI runs because I was checking for exact matches on `[kpt, kpt, kpt]` instead of `(kpt, kpt, kpt)`.

DanielYang59:
Yes the original unit tests were doing the same and we have to change the types too. Hopefully there would not be much operations that replies on `Kpoint` to be exactly `list` or mutable. Thanks for the input!

PR #3759, opened 2024-04-15
Add a CI run that only tests non-optional deps (optimistically)

ml-evs:
New version of #3646.

This PR adds a separate test suite run for the case that only mandatory dependencies have been installed (directly via `pip install -e .[dev]`), attempting to mimic what a user will actually have when they just `pip install pymatgen`. This will then hopefully catch errors where either optional imports are not properly guarded, or main functionality is rendered broken by optional dependency updates.

Not sure if this is something that you would actually want to merge given the potential maintenance overhead, but this might be able to spot more issues before they arise now that #3757 is merged.

Someone will need to carefully look at how I am invoking the tests, as I'm not sure if there is any coverage/durations magic going on that needs to be separated between the two runs (the comments to that effect at the top of `test.yml` might now be out of date).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,26 +66,38 @@ jobs:
         run: |
           sudo cp cmd_line/gulp/Linux_64bit/* /usr/local/bin/
 
+      - name: Install pymatgen and mandatory testing dependencies only
+        run: |
+          micromamba activate pmg
+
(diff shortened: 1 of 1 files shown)


ml-evs:
@coderabbit disable

ml-evs:
Superseded by #3857

PR #3765, opened 2024-04-18
Update `vasprun.converged_ionic` logic when `EDIFFG=0`

matthewkuner:
a new test for this has been added as well.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the convergence detection logic in relaxation runs to include scenarios where `EDIFFG=0`.
- **Tests**
	- Added a new test to ensure accurate handling of specific convergence scenarios in VASP simulations.


matthewkuner:
@janosh this is ready for merging

esoteric-ephemera:
Just for context / documentation: the specific use case for `EDIFFG = 0` is to ensure that a relaxation runs for `NSW` steps. In this case, the user isn't really worried about convergence wrt forces or energy differences, and @matthewkuner's solution prevents custodian from trying to correct the calc because `Vasprun.converged_ionic = False`. You can think about this as the non-AIMD way to generate a relaxation trajectory with DFT

janosh (approved):
thanks @matthewkuner for this change and @esoteric-ephemera for the context! 👍

janosh:
something wrong with the new test file it seems. can't be found for unknown reasons

matthewkuner:
closed in favor of #3783

PR #3771, opened 2024-04-19
Fix LobsterSet

naik-aakash:
# Issue

The POTCARs used to generate the input set are incorrect and do not match `PBE_54` POTCARs. Currently, the `potcar_symbols` variable stores element symbols, which leads to the input set trying to find potcars matching element symbols only.

# Fix

Correctly refer to potcars that should be used based on specified functional using  `potcar_symbols`  property of its base `VaspInputSet` class

# Todo

- [x]  Write a test to avoid such issues in future


JaGeo:
> is there a way to test this change with the scrambled POTCARs added by @esoteric-ephemera?

Oh, yeah, we definitely need a test 😅

naik-aakash:
Thanks @janosh ,for the suggestion, I was indeed planning to do that 😃 , will add the test in sometime

janosh:
thanks @naik-aakash 👍

PR #3776, opened 2024-04-21
Add type annotations for `io.vasp.outputs`

DanielYang59:
## Summary

- Add type annotations for `io.vasp.outputs`, a follow-up PR of #3740.
- Use `PathLike` as type over `str` for all paths.
- Add missing docstrings and clean up docstrings.
- Lazily import `Vasprun/Xdatcar` in `Trajectory` to be consistent with other part of the class.


DanielYang59:
@janosh While you are working on improving dcostrings, can you please skip this `io.vasp.outputs` part to avoid overlapping? I would help fixing some docstring issues while adding types. Thanks!

pymatgen/io/vasp/outputs.py
```diff
@@ -766,33 +786,39 @@ def run_type(self) -> str:
         if self.parameters.get("LUSE_VDW", False):
             run_type += "+rVV10"
         elif self.incar.get("IVDW") in IVDW_TYPES:
-            run_type += "+vdW-" + IVDW_TYPES[self.incar.get("IVDW")]
+            run_type += f"+vdW-{IVDW_TYPES[self.incar.get("IVDW", 0)]}"
```
DanielYang59, line 789: No sure if we really need the default `IVDW` as `0` (for no correction), see [VASP wiki](https://www.vasp.at/wiki/index.php/IVDW)

pymatgen/io/vasp/outputs.py
```diff
@@ -786,7 +786,7 @@ def run_type(self) -> str:
         if self.parameters.get("LUSE_VDW", False):
             run_type += "+rVV10"
         elif self.incar.get("IVDW") in IVDW_TYPES:
-            run_type += f"+vdW-{IVDW_TYPES[self.incar.get("IVDW")]}"
```
DanielYang59: I'm not quite sure why [this f-string failed in windows](https://github.com/materialsproject/pymatgen/actions/runs/8959057873/job/24604009920)?

```python
E     File "D:\a\pymatgen\pymatgen\pymatgen\io\vasp\outputs.py", line 789
E       run_type += f"+vdW-{IVDW_TYPES[self.incar.get("IVDW")]}"
E                                                      ^
E   SyntaxError: f-string: unmatched '('
```

pymatgen/io/vasp/outputs.py
```diff
@@ -1191,11 +1191,11 @@ def get_trajectory(self) -> Trajectory:
             structs.append(struct)
         return Trajectory.from_structures(structs, constant_lattice=False)
 
-    def update_potcar_spec(self, path: PathLike) -> None:
+    def update_potcar_spec(self, path: bool) -> None:
         """Update the specs based on the POTCARs found.
 
         Args:
-            path: path to search for POTCARs
+            path: path to search for POTCARs  # TODO: (@DanielYang59) misleading arg name
```
DanielYang59, line 1198: This arg (`path: bool`) name is misleading, the same applies to the `update_charge_from_potcar` method next to it.

This also leads a confusing type in: https://github.com/materialsproject/pymatgen/blob/a16ac3926c61b2e12b14b2bceb475da70cc0d755/pymatgen/io/vasp/outputs.py#L1087-L1122

Both methods look like private methods to me, but need confirmation. Looks like you are the VASP expert, can you please give me some advice? @esoteric-ephemera

pymatgen/io/vasp/outputs.py
```diff
@@ -3628,28 +3703,32 @@ def from_file(cls, filename: str) -> Self:
         return cls(poscar, data, data_aug=data_aug)
 
     @property
-    def net_magnetization(self):
-        """Net magnetization from Chgcar"""
+    def net_magnetization(self) -> float | None:
+        """Net magnetization from Chgcar."""
+        # TODO: (DanielYang59): default to 0.0 for easier typing?
         return np.sum(self.data["diff"]) if self.is_spin_polarized else None
```
DanielYang59, line 3709: Should we use `0.0` as default magnetic moment for easier typing?

tests/io/vasp/test_outputs.py
```diff
@@ -1708,7 +1708,6 @@ def test_standard(self):
         b = np.array([np.cross(a[1, :], a[2, :]), np.cross(a[2, :], a[0, :]), np.cross(a[0, :], a[1, :])])
         b = 2 * np.pi * b / vol
 
-        assert wavecar.filename == f"{VASP_OUT_DIR}/WAVECAR.N2"
```
DanielYang59, line 1711: I don't think it's quite necessary for `Wavecar` to carry a `filename` attribute (It's only used for unit test), is there any use case that I'm not aware of?

DanielYang59:
I'm now facing significant resistance from the "default value as `None` behavior" (most `mypy` errors are more or less related to possible default `None` values), similar to #3724.

For the following example:
```diff
poscar_dict = {"lattice": Lattice, "pos": Position}  # a terrible example dict

- comment: str | None = poscar_dict.get("comment", None)
+ comment: str = poscar_dict.get("comment", "")

# More downstream operations that only works for `str`
clean_comment = comment.strip()
```

Using empty str `""` could avoid the type being expanded to `str | None`.

Or perhaps: https://github.com/materialsproject/pymatgen/blob/a16ac3926c61b2e12b14b2bceb475da70cc0d755/pymatgen/io/vasp/outputs.py#L3630-L3633

Such default values as `None` would expand the type from `type` to `type | None`, meaning all **downstream operations** (or child objects) need to explicitly check for `None`.

Such cases are happening everywhere across the code base (for all kinds of types) and need to insert huge amount of checks `if var is None`. I'm not sure how to resolve this gracefully (I cannot even describe this issue properly because it's happening everywhere and for different types). Any suggestion would be appreciated @janosh.

pymatgen/io/vasp/outputs.py
esoteric-ephemera: Please be cognizant of downstream effects, especially in the materials project software stack. If there is no reason to remove something, then it may be best to leave as-is in case something downstream depends on it

pymatgen/io/vasp/outputs.py
esoteric-ephemera: `path` is a `str`, `path`, or `bool`. See the definition of `Vasprun.get_potcars`. It is unclear but I really don't want to change the core behavior of functions if they are working currently and there is no need to

I would prefer to just improve the docstr of these methods

pymatgen/io/vasp/outputs.py
DanielYang59: I'm not quite sure about downstream effect, but I assume default of `0` instead of `None` might align better with VASP behavior. I would be happy to revert this though, if you decide it might bring more risk than benefits :)

pymatgen/io/vasp/outputs.py
DanielYang59: Yes you are correct. I was mislead by the type of `parse_potcar_file` (the docstring is correct): https://github.com/materialsproject/pymatgen/blob/46b018ddc66bc8b60c9528a040bf7fce010f2dca/pymatgen/io/vasp/outputs.py#L223-L235

I would fix this type now. Thanks for the wonderful suggestions!

pymatgen/io/vasp/outputs.py
DanielYang59: Make sense to me. I'm not quite experienced with non-collinear calculations so let's leave it as is :)

tests/io/vasp/test_outputs.py
DanielYang59: Thanks!

pymatgen/io/vasp/outputs.py
DanielYang59: Perhaps we need to revisit this change? 

With current implementation, if the `IVDW` tag is absent from `INCAR`, `self.incar.get('IVDW')` would return `None`, and `f"+vdW-{IVDW_TYPES[self.incar.get('IVDW')]}"` would break by using `None` as the index.

Also I think setting the default value of `IVDW` is a rational default which agrees with the [VASP wiki](https://www.vasp.at/wiki/index.php/IVDW):
>IVDW = 0 | 1 | 10 | 11 | 12 | 13 | 2 | 20 | 21 | 202 | 263 | 3 | 4
Default: IVDW = 0 (no correction) 
Description: IVDW specifies a vdW (dispersion) correction.

DanielYang59:
@janosh Can you please revise this? 

For most `xml.etree` or `re` related searches, I chose to suppress `mypy` error instead of check for `None` (for example 0346b92c3e61fb3ff540eb7d21d1a918d002bd78), because I think it's fairly safe to assume the source file is properly formatted and match should normally not return `None`.

Only change and need discussion is the change of default `IVDW` to agree with VASP in c990b018af9fd3aa19d63cbfec6f1889e947b40d and discussions in https://github.com/materialsproject/pymatgen/pull/3776#discussion_r1618544768.

I notice [CI runs are cancelled randomly](https://github.com/materialsproject/pymatgen/actions/runs/9276409234/job/25523410105) from time to time?

PR #3777, opened 2024-04-21
`pyright` fixes

DanielYang59:
## Summary

- `pyright` fixes, a follow up PR of #3757.
- Enable `pyright` in linting workflow and `pre-commit`.

PR #3778, opened 2024-04-21
bugfix: chargemol runtime directory and `os.path.exists` fix

chiang-yuan:
## Summary

- Improve path logic
- Allow Chargemol to run in different directory (previously the code has implemented but is not working)
- Fix `os.path.isfile` bug. Change it into `os.path.exists`

## Todos

- add test

## Checklist

### DIFF
--- a/pymatgen/command_line/chargemol_caller.py
+++ b/pymatgen/command_line/chargemol_caller.py
@@ -46,6 +46,7 @@ import os
 import subprocess
 import warnings
 from glob import glob
+from pathlib import Path
 from shutil import which
 from typing import TYPE_CHECKING
 
@@ -57,8 +58,6 @@ from pymatgen.io.vasp.inputs import Potcar
 from pymatgen.io.vasp.outputs import Chgcar
 
 if TYPE_CHECKING:
-    from pathlib import Path
-
     from pymatgen.core import Structure
 
 __author__ = "Martin Siron, Andrew S. Rosen"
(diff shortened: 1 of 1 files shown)


pymatgen/command_line/chargemol_caller.py
```diff
@@ -170,34 +168,34 @@ def _execute_chargemol(self, **job_control_kwargs):
...
+            lines = self._write_jobscript_for_chargemol(
+                chgcar_path=os.path.join(temp_dir, "CHGCAR"),
+                **job_control_kwargs,
+            )
+            with open(os.path.join(temp_dir, "job_control.txt"), mode="w") as file:
+                file.write(lines)
+
+            response = subprocess.run(CHARGEMOL_EXE, capture_output=True, check=False)
+            _stdout = response.stdout.decode("utf-8")
+            response.stderr.decode("utf-8")
```
janosh, line 196: do these two lines serve a purpose? i don't think these raise when `response.returncode != 0`? i think we need a test for the previously raised `RuntimeError`

janosh (changes requested):
thanks @chiang-yuan for the `os.path.isfile` fix! 👍

pymatgen/command_line/chargemol_caller.py
chiang-yuan: maybe we can just use `check=True`? We don't even need to capture response/stdout here as it is not used?

janosh: yes, let's use `check=True`. it's a different error type [`CalledProcessError`](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) which I think is better than `RuntimeError` anyway but we should still add a test for this behavior.

chiang-yuan:
Try to fix the missing `command_lin` in the test file paths but it is unsuccessful due to the pyright issue. 
I will leave it a bit until the problem is solved globally #3777

shyuep:

PR #3781, opened 2024-04-23
Fix wrong error type in tests for `io.vasp.sets`

DanielYang59:
## Summary

Fix wrong error type in tests for `io.vasp.sets` in #3771, should be `RuntimeError` instead of `OSError`


DanielYang59:
@coderabbitai ignore

DanielYang59:
It seems like this PR is forgotten to be merged. Closed as it has been fixed in #3783.

PR #3786, opened 2024-04-26
Install some optional dependencies in CI

### LINKED ISSUE #3684: [Dev] Multiple dependencies for CI missing

DanielYang59:
## Summary

- Install some optional dependencies in CI, to fix #3684.
- Fix patch for `bader_caller` (setting `PATH` to empty is easier than changing the `patch` location, and serves the same result where `bader` cannot be found in `PATH`)

### Partially installed on Ubuntu

- `boltztrap2`: installable only with `pip` (`uv` wouldn't work and [suggest no `numpy` found when there is](https://github.com/materialsproject/pymatgen/actions/runs/9920394484/job/27407228711?pr=3786))
- `pygraphviz`: https://pygraphviz.github.io/documentation/stable/install.html#macos 

### Perhaps for a follow up PR

- [ ] `x_trans`: [http://www.icams.de/content/research/software-development/boltztrap/ (Page not found, #3921)](https://www.tuwien.at/en/tch/tc/theoretical-materials-chemistry/boltztrap)
- [ ] `critic2`: https://aoterodelaroza.github.io/critic2/installation/
- [ ] `OBAlign`: should come with `openbabel`?
- [ ] `vampire`: https://vampire.york.ac.uk/download/
- [ ] `zeo/pyzeo`: http://www.zeoplusplus.org/download.html (last update 2017)


janosh:
do we want to merge this as partially complete and leave some of the todos for later?

DanielYang59:
Thanks for asking. I was facing some resistance while trying to install them (maybe that's the reason they're not installed in CI). And what make things worse is that I never use any of these.

I'm currently working on the `vasp.outputs` type PR and haven't really look into this one yet. But I might try to install as much as I can in this PR and I would ping you then.

Also some help would be appreciated:
- `graphviz` seems to install fine but is still skipped in CI for some reason.
- `boltztrap2` is installed in Ubuntu, and revealed several broken unit tests 😭

DanielYang59:
@janosh Perhaps we could merge this as partially finished (as you suggested) and I would work towards the remaining parts shortly. The CI test failure is reported in #3969 and would be fixed by #3972, thanks!

janosh (approved):
thanks @DanielYang59! 👍

.github/workflows/test.yml
```diff
@@ -91,12 +93,6 @@ jobs:
           # suggesting no NumPy when there is
           pip install BoltzTraP2
 
-      - name: Install pymatgen
-        run: |
-          micromamba activate pmg
-
-          uv pip install --editable '.[${{ matrix.config.extras }}]' --resolution=${{ matrix.config.resolution }}
```
DanielYang59, line 83: Current we cannot put this before `pip install BoltzTraP2` (`BoltzTraP2` is included as an optional dependency) because `uv` cannot install `BoltzTraP2` for some reason.

Meanwhile I'm not quite sure why `pip install BoltzTraP2` failed in Windows either, I would look closer into this and fix it in a follow up PR if possible :)

Meanwhile I think we should not include 57d34e2e297a0669534692cc9154dde1bc66dadf here as a solution has been proposed in #3972 and we shouldn't just hijack his contribution :)

janosh: ah, hadn't seen https://github.com/materialsproject/pymatgen/pull/3972 yet.


can you move `pip install BoltzTraP2` into the preceding step then and add an inline comment explaining order matters?

also, let's use `matplotlib!=3.9.1` instead of `matplotlib<3.9.1` in pyproject

DanielYang59: Sure thing.

DanielYang59: But things are getting quite tricky at this moment (surrounding `BoltzTraP2`), as it would require `cython` to be built properly (I suggested to the original author to distribute wheels but wasn't making any progress so far). Perhaps we can comment out `BoltzTraP2` from optional dependency for now?

janosh: i think that's fine? we can also move `uv pip install cython setuptools wheel` further up

DanielYang59: Beautiful. That's better :)

DanielYang59: Meanwhile want to know if it's possible to insert in-line OS condition (Ubuntu only)? I googled earlier and didn't have much luck. `BoltzTraP2` is only installable on Ubuntu at this moment.

janosh: > I now cannot think of another way to install those dependencies other than the one prior to https://github.com/materialsproject/pymatgen/commit/155cbff5e01ab2108a9bf831ec6f73004b85fb8a. Perhaps you could push that directly? :)

maybe it's not possible then. sorry for messing up your PR. i'll take another look later

> Meanwhile want to know if it's possible to insert in-line OS condition (Ubuntu only) in test.yml? I googled earlier and didn't have much luck.

have a look at this [SO](https://stackoverflow.com/a/8597411)

DanielYang59:
Regarding the installation of `BolzTraP2` on Windows, I'm afraid [it wouldn't be an easy fix owing to the difference in `cmake` behavior](https://gitlab.com/sousaw/BoltzTraP2/-/merge_requests/20#note_2032779686). Let's install it on Linux only for now until we have a solution. 

I would look into why `uv` cannot install it later.

src/pymatgen/analysis/graphs.py
```diff
@@ -933,7 +933,7 @@ def draw_graph_to_file(
                     d["label"] = f"{d['weight']:.2f} {units}"
 
             # update edge with our new style attributes
-            g.edges[u, v, k] |= d
+            g.edges[u, v, k].update(d)
```
DanielYang59, line 936: @janosh Forgot to mention this, I have to revert your `.update -> |=` change here c67abbc6678ebeee6a324440570aa38cc0c66adb because it seems to [break unit test](https://github.com/materialsproject/pymatgen/actions/runs/10246364887/job/28343300325). 

I haven't got a chance to look closer yet, perhaps the operator is not defined for `edges`

janosh: yeah i noticed that, that was sloppy refactoring on my part. thanks for fixing!

> perhaps the operator is not well defined for edges

that's very likely correct

DanielYang59: > that was sloppy refactoring on my part.

`pygraphviz` was not installed in CI at that time, so there's no way for you to catch that :)

janosh: no, but i could have noticed that `g.edges[u, v, k]` is unlikely to be a `dict` in which the first thought is to check whether the class in question implements `__or__`. (if not, `|=` is undefined behavior)

.github/workflows/test.yml
DanielYang59: I just [pushed a MR (merge request in GitLab) to BoltzTraP2](https://gitlab.com/sousaw/BoltzTraP2/-/merge_requests/21) which should fix the installation issue with `uv`, and the original author is working on [fixing the installation on Windows as well](https://gitlab.com/sousaw/BoltzTraP2/-/issues/5). This would take some time but at least there is hope :)

janosh: nice work! that's what i call community effort 👍 😄

PR #3789, opened 2024-04-28
Exclude `gulp` from package data

DanielYang59:
## Exclude `gulp` from package data

- See discussion in #3788, the binaries and libs of super outdated `gulp` (11 yrs old) should not be included in the package data. And it's taking up 40% of the package size:
  ```bash
  pymatgen-2024.4.13 % du -sh * | sort -hr
   15M	pymatgen
   10M	cmd_line                   # only `gulp` inside `cmd_line`
   56K	pymatgen.egg-info
   16K	PKG-INFO
   12K	README.md
  8.0K	tests
  8.0K	setup.py
  8.0K	pyproject.toml
  4.0K	setup.cfg
  4.0K	LICENSE
  ```

After removal, the package size decreased from 6.9 MB to 3.0 MB.

- Should we even include them inside the code base?

- Two test files are included but I'm not seeing them inside `setup.py`?:
  ```
  pymatgen-2024.4.13 % ls tests 
  test_cli.py test_pkg.py
  ```


DanielYang59:
@coderabbitai ignore

janosh (approved):
> Should we even include them inside the code base?

i think it would be much better to remove and point people at the download page for GULP in a message to the user if pymatgen can't find the binary. but needs to be confirmed with @shyuep

DanielYang59:
> > Should we even include them inside the code base?
>
> i think it would be much better to remove and point people at the download page for GULP in a message to the user if pymatgen can't find the binary. 

Yes I agree with this. But the original link does not work on my side for some reason?
https://github.com/materialsproject/pymatgen/blob/ff94faa5bb490354c420f188f1abeac17cb40931/pymatgen/command_line/gulp_caller.py#L1-L4

But this website from Curtin University seems to work (I'm not sure if these two point to the same package though): https://gulp.curtin.edu.au/index.html

.github/workflows/test.yml
```diff
@@ -61,11 +60,6 @@ jobs:
       - name: Install uv
         run: micromamba run -n pmg pip install uv
 
-      - name: Copy GULP to bin
```
DanielYang59, line 64: I removed tests for gulp because it's not being run in CI anyway: https://github.com/materialsproject/pymatgen/blob/e04444e3c0f30fa362b31b371c1eff04bc995d06/tests/command_line/test_gulp_caller.py#L33-L36

pymatgen/command_line/gulp_caller.py
```diff
@@ -1,5 +1,5 @@
 """Interface with command line GULP.
-http://projects.ivec.org
+https://gulp.curtin.edu.au/index.html
```
DanielYang59, line 2: This website update need confirmation: 
- Is the original url not accessible at all, or just temporarily, or just something with my network?
- Is the new website pointing to the same package?

Meanwhile it might be hard to automate the download process because the new website requires a form being filled.

janosh: > Is the original url not accessible at all, or just temporarily, or just some issue with my network?

i also can't access http://projects.ivec.org. very likely it's permanently down.

> Is the new website pointing to the same package?

yep, that's the one

> Meanwhile it might be hard to automate the download process because the new website requires a form being filled.

i think it's fine to ask users to install manually. this is not core `pymatgen` functionality

janosh (approved):
thanks so much @DanielYang59. this is excellent!! will result in shorter install times from the next `pymatgen` release onwards thanks to no longer downloading outdated `gulp` files.

btw, there's [this nice tool](https://inspector.pypi.io/project/pymatgen/2024.4.13/packages/a4/7a/69c17fd3e364b6beb5b37a7da2dca2105f609bf83bd80fea5895b609a5b2/pymatgen-2024.4.13-cp310-cp310-macosx_11_0_arm64.whl/) that allows you to get a list of files in a pypi package

DanielYang59:
> thanks so much @DanielYang59. this is excellent!! will result in shorter install times from the next `pymatgen` release onwards thanks to no longer downloading outdated `gulp` files.

Thanks for reviewing. Not sure about the installation time, but the package is reduced by half :) The `GULP` package has been sneaky inside the package for such a long time.

> btw, there's [this nice tool](https://inspector.pypi.io/project/pymatgen/2024.4.13/packages/a4/7a/69c17fd3e364b6beb5b37a7da2dca2105f609bf83bd80fea5895b609a5b2/pymatgen-2024.4.13-cp310-cp310-macosx_11_0_arm64.whl/) that allows you to get a list of files in a pypi package

Thanks for sharing, it would be very helpful.

>> Meanwhile it might be hard to automate the download process because the new website requires a form being filled.

> i think it's fine to ask users to install manually. this is not core pymatgen functionality

I'm more concerned about autonomous package downloading in CI tests, but we would see when we need it because it's not supported by Ubuntu 20.04 afterwards anyway.

PR #3790, opened 2024-04-28
Format doc strings

janosh:
plus minor formatting and 1 deprecation of `NRANDOM` -> `n_random` in `coordination_geometry_symmetry_measures_fallback_random`


esoteric-ephemera:
@janosh : this broke some code in pymatgen-analysis-diffusion and thus also in emmet, particularly the change from `symmop` to `symm_op` in `Structure.apply_operation`. Can we announce breaking changes or avoid them if they're made for code style?

shyuep:
@janosh Seriously, stop breaking stuff with this kind of cosmetic stuff. I am going to insist that you maintain backwards compatibility. Otherwise, I am revoking your write privileges and you will submit PRs like everyone else.

janosh:
@esoteric-ephemera sorry about the hassle and thanks for taking the time to fix!

@shyuep i'll be more careful. i realized that this change could be breaking and did a search for `symmop` being passed as keyword in `pymatgen` and `atomate2` but ofc that leaves a lot of places where it could matter. just to clarify though, i did submit a PR for this like everyone else and CI passed so I doubt a reviewer would have noticed.

PR #3792, opened 2024-04-30
Imperative doc strings

janosh:
follow up to #3790


pymatgen/core/composition.py
```diff
@@ -186,15 +192,20 @@ def __eq__(self, other: object) -> bool:
...
+
+        Should ONLY be used for defining a sort order (the behavior is probably not what you'd expect).
         """
         if not isinstance(other, Composition):
             return NotImplemented
 
         for el in sorted(set(self.elements + other.elements)):
             if other[el] - self[el] >= Composition.amount_tolerance:
                 return False
+            # TODO @janosh 2024-04-29: is this a bug? why would we return True early?
```
janosh, line 208: @DanielYang59 @JaGeo curious if anyone else thinks this early `return True` in `comp1 >= comp2` looks strange? how can we return before having looped through the whole list of elements?

DanielYang59: Looks like a bug to me. Only one of the elements is "greater than" usually doesn't mean the `Composition` (set of elements) is greater than.

pymatgen/analysis/chemenv/coordination_environments/coordination_geometries.py
```diff
@@ -73,8 +73,7 @@ def __str__(self):
 
 
 class ExplicitPermutationsAlgorithm(AbstractChemenvAlgorithm):
-    """
-    Class representing the algorithm doing the explicit permutations for the calculation of
+    """Algorithm doing the explicit permutations for the calculation of
```
DanielYang59, line 76: Thanks for these docstring improves though. I was think the same. 

The keyword `class` should already be self-documenting and no need to repeat it's "Class for some purpose."

Also as per [PEP-8](https://peps.python.org/pep-0008/) the docstring should usually start with a brief high-level summary on the first line (the line with ```), so I totally support your changes here!

pymatgen/analysis/diffraction/tem.py
```diff
@@ -107,7 +107,7 @@ def generate_points(coord_left: int = -10, coord_right: int = 10) -> np.ndarray:
     def zone_axis_filter(
         self, points: list[tuple[int, int, int]] | np.ndarray, laue_zone: int = 0
     ) -> list[tuple[int, int, int]]:
-        """Filters out all points that exist within the specified Laue zone according to the zone axis rule.
+        """Filter out all points that exist within the specified Laue zone according to the zone axis rule.
```
DanielYang59, line 110: Thanks for reinforcing this, I fixed some similar issues (and thought `pymatgen` has a different convention than others). PEP-8 also seems to support this style.

DanielYang59:
I hope I'm not caring too much about nitty gritty details, but I have a feeling that the `Returns` prefix in a `property` should be removed too? It should be a description of the method as a `property`, not as a function, right?

https://github.com/materialsproject/pymatgen/blob/9964c4a335b69623f1c51d13472cff086c0f0c72/pymatgen/core/composition.py#L290-L340

pymatgen/core/composition.py
janosh: thanks for confirming. we can fix it in a subsequent PR to keep this one restricted to doc strings

janosh:
> but I have a feeling that the `Returns` prefix in the docstring for a `property` should be removed too?

i agree

DanielYang59:
> > but I have a feeling that the `Returns` prefix in the docstring for a `property` should be removed too?
> 
> i agree

I'm afraid there is some misunderstanding. I thought the docstring of a `property` should be describing the `property` itself, instead of the `method`.

For example:
```diff
@property
    def labels(self):
-        """Get a dictionary containing kink information:
+        """A dictionary containing kink information:
```

What do you think?

janosh:
> I'm afraid there is some misunderstanding. I thought the docstring of a property should be describing the property itself, instead of the method.

no misunderstanding, that's my goal as well. there may have been some mass find and replace changes where this was not the end result. but i'll get to those in the 2nd pass

DanielYang59:
> no misunderstanding, that's my goal as well. there may have been some mass find and replace changes where this was not the end result. but i'll get to those in the 2nd pass

Beautiful! Thanks for the improvements.

Meanwhile maybe can we set the CodeRabbit to review only when we explicitly call it (like `@CodeRabbit review`)? It seems to be badly polluting conversation history.

janosh:
>  It seems to be badly polluting conversation history.

very much so! i asked @shyuep if we could disable or configure it differently. e.g. would be nice if it ignored PRs from maintainers

shyuep:
There is. Coderabbit.yaml file that can be edited in the repo.

PR #3798, opened 2024-05-01
Fix overlayed subplots in `BSPlotterProjected.get_projected_plots_dots()`

### LINKED ISSUE #3760: 'get_projected_plots_dots()' and 'get_projected_plots_dots_patom_pmorb()' do not work properly
### Python version

3.11.7

### Pymatgen version

2024.4.13

### Operating system version

Ubuntu 22.04.3 LTS

### Current behavior

I try to get band structures with specific elements and orbitals with `get_projected_plots_dots()`.  While a figure with 3 subfigures is expected, the program returns a figure with only one subfigure and the place where the other two subfigures are supposed to be is blank. 
image
However, when I downgrade pymatgen to 2023.8.10, no such problems occur.
`get_projected_plots_dots_patom_pmorb()` also does not work properly.
Moreover, there are extra axes in the figure.
`matplotlib` version is 3.8.0.

### Expected Behavior

A figure with several subfigures displays normally

### Minimal example

```Python
vasprun = Vasprun("vasprun.xml",parse_projected_eigen=True)
bs_data = vasprun.get_band_structure(line_mode=True)
plt_3=BSPlotterProjected(bs=bs_data)
plt_3.get_projected_plots_dots({'K':['s'],'Ti':['d'],'O':['p']})
plt.show()
```

_No response_

janosh:
closes #3760

**before**

before-3760-fix

**after**

after-3760-fix

there were multiple issues in `BSPlotterProjected.get_projected_plots_dots`: all subplots were plotted into the 1st axes. the large axis labels from `pretty_plot` were applied to the whole figure, not the individual subplots

ran out of time on this fix but noticed that the fat lines in the `Cu d` still look off. like it's still overlaying multiple band structures. maybe someone else has time to take a look @DanielYang59 @JaGeo? also @HERTZCAI could you comment on what output you're expecting and share the `vasprun.xml` file you were plotting in #3760

### DIFF
--- a/pymatgen/apps/borg/queen.py
+++ b/pymatgen/apps/borg/queen.py
@@ -79,7 +79,7 @@ class BorgQueen:
             valid_paths.extend(self._drone.get_valid_paths((parent, subdirs, files)))
         data: list[str] = []
         total = len(valid_paths)
-        for idx, path in enumerate(valid_paths, 1):
+        for idx, path in enumerate(valid_paths, start=1):
             new_data = self._drone.assimilate(path)
             self._data.append(new_data)
             logger.info(f"{idx}/{total} ({idx / total:.1%}) done")
--- a/pymatgen/electronic_structure/bandstructure.py
+++ b/pymatgen/electronic_structure/bandstructure.py
@@ -810,13 +810,13 @@ class BandStructureSymmLine(BandStructure, MSONable):
         """
         to_return = []
         for idx in self.get_equivalent_kpoints(index):
-            for b in self.branches:
-                if b["start_index"] <= idx <= b["end_index"]:
+            for branch in self.branches:
+                if branch["start_index"] <= idx <= branch["end_index"]:
                     to_return.append(
                         {
-                            "name": b["name"],
-                            "start_index": b["start_index"],
-                            "end_index": b["end_index"],
+                            "name": branch["name"],
+                            "start_index": branch["start_index"],
+                            "end_index": branch["end_index"],
                             "index": idx,
                         }
                     )
--- a/pymatgen/electronic_structure/bandstructure.py
+++ b/pymatgen/electronic_structure/bandstructure.py
@@ -243,25 +243,22 @@ class BandStructure:
             returns an empty dict
         """
         result = {}
-        structure = self.structure
         for spin, v in self.projections.items():
             result[spin] = [[defaultdict(float) for i in range(len(self.kpoints))] for j in range(self.nb_bands)]
             for i, j, k in itertools.product(
                 range(self.nb_bands),
                 range(len(self.kpoints)),
-                range(len(structure)),
+                range(len(self.structure)),
             ):
-                result[spin][i][j][str(structure[k].specie)] += np.sum(v[i, j, :, k])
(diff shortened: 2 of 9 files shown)


DanielYang59:
> ran out of time on this fix but noticed that the fat lines in the Cu d still look off. like it's still overlaying multiple band structures. maybe someone else has time to take a look @DanielYang59 @JaGeo? 

Thanks for ping me. I would definitely be happy to have a look later.

> also @HERTZCAI could you comment on what output you're expecting and share the vasprun.xml file you were plotting in https://github.com/materialsproject/pymatgen/issues/3760

Yes the desired output and a vasprun.xml to create would certainly be helpful (I currently don't have a file to test at hand).

JaGeo:
@janosh i don't really have the time at the moment. Sorry!

DanielYang59:
> @janosh i don't really have the time at the moment. Sorry!

I could give a hand on this then 😄 Thanks!

janosh:
@DanielYang59 you should be able to load any JSON bandstructure file with projections the way the current test does

https://github.com/materialsproject/pymatgen/blob/a9667f7e236a1e2120f7b905529e653c994161e8/tests/electronic_structure/test_plotter.py#L183-L191

if you search the code base for `"projections": {"` you find several.

DanielYang59:
Tweaked a bit (orbital name subscript, and hide empty orbitals), but the `Cu_d` orbital still is a mess.
Figure_1

JaGeo:
@DanielYang59 I have one example here with an expected outcome and files. Maybe, this could help? It's based on LOBSTER output but uses the same plotting options.

DanielYang59:
> @DanielYang59 I have one example here with an expected outcome and files. Maybe, this could help? It's based on LOBSTER output but uses the same plotting options.
> 
> https://github.com/materialsvirtuallab/matgenb/blob/master/notebooks/2019-01-11-How%20to%20plot%20and%20evaluate%20output%20files%20from%20Lobster.ipynb

Thanks for sending that through. Sorry to say but that seems to look even worse than what we have now 😓. The title and tick labels badly overlapped. Maybe we should fix that as well?

janosh:
> It seems to be using the same Class, maybe we should rerun the plotter to see if things improved?

that would be good. could migrate the `matgenb` repo to `ruff` for linting and formatting while at it

> The title and tick labels badly overlapped. Maybe we should fix that as well?

yes please!

given that the example @JaGeo linked also has fat bands, i'm starting to think the `Cu d` is looking as it's supposed to in which case after fixing the overlapping tick labels, we're done here

Screenshot 2024-05-02 at 09 30 57

JaGeo:
@janosh I think the point sizes are chosen in a way that it looks a bit messy...

And yes, fixing the overlap of labels would be awesome.

JaGeo:
Could we also restrict the plot to the actual path and not show white space before and after?

DanielYang59:
I just tweaked the code to allow user choosing marker size themselves, which seem to improve the overlapping a bit.
Figure_1

DanielYang59:
> Could we also restrict the plot to the actual path and not show white space before and after?

Absolutely!
Figure_1

PR #3804, opened 2024-05-03
Add `is_radioactive` property to Element class

AntObi:
## Summary

Major changes:
* This PR adds an `is_radioactive` property to the `Element` class which allows users to check if an Element is radioactive.
* The `contains_element_type` method in the `Composition` class has radioactive as a possible category added, so users can check if a composition contains radioactive elements.

Feature motivation:
Motivated by a recent publication (https://pubs.acs.org/doi/10.1021/acs.chemmater.4c00643) which identifies that some of the discovered compounds in the GNoME database (https://github.com/google-deepmind/materials_discovery) contain radioactive elements, it would be useful to be able to screen out compositions containing radioactive elements or to be able to characterise how many radioactive compositions exist in a dataset of compositions/structures.

## Todos
* Agree upon which elements are considered radioactive. Currently, my PR has the following elements considered radioactive: Tc, Pm, and all elements with atomic number greater than 83 (Z>83) i.e. Po-Lr (Og).

## Checklist

### DIFF
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -508,7 +508,7 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
             category (str): one of "noble_gas", "transition_metal",
                 "post_transition_metal", "rare_earth_metal", "metal", "metalloid",
                 "alkali", "alkaline", "halogen", "chalcogen", "lanthanoid",
-                "actinoid", "quadrupolar", "s-block", "p-block", "d-block", "f-block"
+                "actinoid", "radioactive", "quadrupolar", "s-block", "p-block", "d-block", "f-block"
 
         Returns:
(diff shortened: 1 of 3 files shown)


janosh (approved):
thanks for adding this! 👍 

> Agree upon which elements are considered radioactive. Currently, my PR has the following elements considered radioactive: Tc, Pm, and all elements with atomic number greater than 83 (Z>83) i.e. Po-Lr (Og).

looks good to me. is this not widely agreed upon?

AntObi:
> thanks for adding this! 👍 
> 
> > Agree upon which elements are considered radioactive. Currently, my PR has the following elements considered radioactive: Tc, Pm, and all elements with atomic number greater than 83 (Z>83) i.e. Po-Lr (Og).
> 
> looks good to me. is this not widely agreed upon?

I would say yes they are.

PR #3811, opened 2024-05-06
Copy `matgenb` repo into new `pymatgen` `examples/` folder

janosh:
motivation: current pymatgen docs are purely class/method and function doc strings. pretty dry stuff which mostly helps you if you already have a good idea of what you want to do and just need to be reminded of what parameters are available or what their defaults are. to get familiar with a new part of pymatgen, you often have to go straight to reading the source code which can be a significant learning curve (a high activation barrier if you like 😄), esp. for beginners.

luckily `matgenb` has provided a decent collection of usage example notebooks for years. given that i only discovered it over a year after starting to contribute to pymatgen (not the start of using it mind you), it's likely unknown to a large part of the pymatgen user base.

merging those Jupyter notebooks into the `pymatgen` repo will greatly increase visibility. it also allows us to subject them to `pymatgen` higher code quality standards and potentially even integrate them into CI to ensure they don't break over time.

### To do

decide what format we want the examples to have. currently jupyter notebooks but could be converted to simple python scripts with VSCode style `# %%` cell delimeters. these can be converted to `.ipynb` format on the fly and from there converted to HTML to be rendered in the `pymatgen` docs at some point. this simplifies running the scripts in CI for testing but has the significant disadvantage that cell outputs cannot be saved

---

pinging @DanielYang59 as we're planning to pool our effort on this PR. also pinging @JaGeo @mkhorton since it ties in with private communication about lowering the barrier to entry to the MP ecosystem


janosh:
@coderabbit ignore

JaGeo:
@janosh I think it is a great idea to bring this back to main pymatgen. 
There might be several failing examples now with all the recent code changes. I also never updated the second LOBSTER notebook and wrote a completely new one.

DanielYang59:
@janosh. Can we set CodeRabbit to review only when we explicitly call it? It's badly polluting conversations.

mkhorton:
Concur, I think this is a great idea! I’m not sure the genesis of matgenb, but I know that many people do not know it exists. I’ve been asked many times for example notebooks from people, and having some example notebooks directly in the pymatgen repo would definitely help with discoverability.

shyuep:
The reason is because matgenb is not just about pymatgen. There are examples for pymatgen as well as non-pymatgen packages. Instead of unilaterally doing such stuff, I suggest you first *ask* and then do. It is a simple question or Issue. All this is leading to wasted effort. If you feel discoverability is an issue, by all means add a link to matgenb from pymatgen/examples. Also, matgenb is already mentioned multiple times in the docs. I can't help if people don't read the docs.

janosh:
@shyuep i'll address your points one by one.

> The reason is because matgenb is not just about pymatgen. There are examples for pymatgen as well as non-pymatgen packages.

i don't see that as being an issue. `pymatgen` is an ecosystem package. many people would come here to look for examples on how to use pymatgen in combination with other codes

> If you feel discoverability is an issue, by all means add a link to matgenb from pymatgen/examples. Also, matgenb is already mentioned multiple times in the docs

like you said, there are multiple `matgenb` links in the docs and discoverability is clearly _still_ an issue. i definitely spent time reading the docs when i started using and contributing to PMG and yet i wasn't aware of `matgenb` for the longest time. see my OP:

> >  given that i only discovered `matgenb` over a year after starting to contribute to pymatgen 

finally, i would like push back on your last statement:

> I can't help if people don't read the docs.

this is not true. you could simply not unilaterally close PRs to improve the docs, esp. if they attracted broad stakeholder support like this one! this too creates wasted effort by requiring me to argue what i believe are the obvious merits of this PR

shyuep:
CleanShot 2024-05-15 at 07 15 37@2x
This is what the front page looks like now. There are links not only in the SECOND paragraph, but also the auxiliary links at the top. There is also an examples folder that has a README pointing people to matgenb.

janosh:
discoverability is only one of several problems. have you tried running the example notebooks? many of them are broken. by blocking this PR, you're preventing them from becoming better maintained and continuously tested. see again my OP:

> it also allows us to subject them to pymatgen's higher code quality standards and potentially even integrate them into CI to ensure they don't break over time.

shyuep:
By all means add CI to matgenb. Again, the answer cannot be "just move everything to the pymatgen repo to make life easier for the maintainer". By that argument, we should have one giant repo for all the codes.

janosh:
> the answer cannot be "just move everything to the pymatgen repo to make life easier for the maintainer"

why not? that's exactly what should be done if it's the easiest way of improving usage examples

we're talking about moving docs for pymatgen into the pymatgen repo... just like other packages have their example notebooks in the same repo.

> By that argument, we should have one giant repo for all the codes.

that's basically what pymatgen is already! plus that's exactly how even huge companies like Google handle it internally: one giant monorepo. i really don't see how this is a counterargument

JaGeo:
My outside perspective: it would certainly be nice to ensure that these examples are still working in the future. Automatic execution in CI would be great.

PR #3813, opened 2024-05-07
Re-enable some useful `ruff` rules

DanielYang59:
## Re-enable some useful `ruff` rules

-  `COM812`: [missing-trailing-comma](https://docs.astral.sh/ruff/rules/missing-trailing-comma/): no change to code base
- `ISC001`: [single-line-implicit-string-concatenation](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/): might conflict with other `ruff` rules sometimes
- `RUF012`: [mutable-class-default](https://docs.astral.sh/ruff/rules/mutable-class-default/)
- `PLC1901`: [compare-to-empty-string](https://docs.astral.sh/ruff/rules/compare-to-empty-string/): no change to code base
- `PYI024`[collections-named-tuple](https://docs.astral.sh/ruff/rules/collections-named-tuple/): make typing of `namedtuple` possible

**REVERTED** 
- `C408`: [unnecessary-collection-call](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/) (confirmed no comment is dropped)


DanielYang59:
@coderabbitai ignore

pymatgen/io/abinit/abiobjects.py
```diff
@@ -357,7 +356,10 @@ class SpinMode(namedtuple("SpinMode", "mode nsppol nspinor nspden"), AbivarAble,
         - spinor_nomag (non-collinear, no magnetism)
     """
 
-    __slots__ = ()
+    mode: Any
+    nsppol: Any
+    nspinor: Any
+    nspden: Any
```
DanielYang59, line 362: I was unable to guess the types of these.

pymatgen/util/provenance.py
```diff
@@ -68,10 +67,12 @@ class HistoryNode(namedtuple("HistoryNode", ["name", "url", "description"])):
     Attributes:
         name (str): The name of a code or resource that this Structure encountered in its history.
         url (str): The URL of that code/resource.
-        description (dict): A free-form description of how the code/resource is related to the Structure.
+        description (str): A free-form description of how the code/resource is related to the Structure.
```
DanielYang59, line 70: The type of `description` looks incorrect?

janosh (commented):
i actually often prefer the `dict(foo=bar)` version over `{"foo": bar}` since it's more compact due to not needing so many quotes. this sometimes avoids a line break which imo helps readability. i'd vote to keep [C408](https://docs.astral.sh/ruff/rules/unnecessary-collection-call) in the ignore list

DanielYang59:
> i actually often prefer the `dict(foo=bar)` version over `{"foo": bar}` since it's more compact due to not needing so many quotes. this sometimes avoids a line break which imo helps readability. 

`dict(foo=bar)` is 13 characters and `{"foo": bar}` is 12 (including the whitespace). So I assume there is no reason `dict()` is more compact than `{}`?

And it seems [`{}` should be faster than `dict()`](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/):
> It's unnecessary to call, e.g., dict() as opposed to using an empty literal ({}). The former is slower because the name dict must be looked up in the global scope in case it has been rebound.

DanielYang59:
It should be ready for review now @janosh. Thanks. Let me know if want to enable any more/less `ruff` rules.

I tried to enable: https://github.com/materialsproject/pymatgen/blob/46b018ddc66bc8b60c9528a040bf7fce010f2dca/pyproject.toml#L72

But it seems to force all kind of list builder into the form of list comprehension (especially some nested conditions list), which would greatly reduce readability.

janosh:
> dict(foo=bar) is 13 characters and {"foo": bar} is 12 (including the whitespace).

that inverts once you have 2 or more keys in your dict. at that point the collection call is shorter

DanielYang59:
> > dict(foo=bar) is 13 characters and {"foo": bar} is 12 (including the whitespace).
> 
> that inverts once you have 2 or more keys in your dict. at that point the collection call is shorter

That's true. But if you look into the commit 38a5aa7c800e7d048922d4d80271dcbe28f7c761, it appears among all changes, this case you mentioned (where a line is broken by extra characters) only happened once: https://github.com/materialsproject/pymatgen/blob/46b018ddc66bc8b60c9528a040bf7fce010f2dca/pymatgen/io/vasp/sets.py#L2219

And all other changes didn't introduce an extra line.

But anyway, feel free to revert that commit if you really prefer not to change that (it's all automatically changed by `ruff` so wasn't any effort from me). But at least we should search and replace some `list()` or `dict()`.

janosh (approved):
thanks @DanielYang59, lgtm! 👍

DanielYang59:
Beautiful. Thanks for reviewing @janosh  Appreciate it.

pyproject.toml
```diff
@@ -61,23 +61,17 @@ lint.ignore = [
...
   "C408",    # unnecessary-collection-call
-  "COM812",  # missing trailing comma
   "D105",    # Missing docstring in magic method
   "D205",    # 1 blank line required between summary line and description
   "D212",    # Multi-line docstring summary should start at the first line
-  "ISC001",
-  "NPY002",  # replace legacy numpy.random with numpy.random.Generator
   "PD901",   # pandas-df-variable-name
   "PERF203", # try-except-in-loop
   "PERF401", # manual-list-comprehension (TODO fix these or wait for autofix)
```
DanielYang59, line 69: I forgot to remove this `TODO` tag, would do it in another PR.

I'm pretty sure we don't really want to enable this rule:
> it seems to force all kind of list builder into list comprehensions (especially some nested conditions list), which would greatly reduce readability.

DanielYang59:
I just look into [this speed comparison](https://switowski.com/blog/dict-function-vs-literal-syntax/) between `dict()` and `{}`, and it surprised me as the speed improvement seems quite significant (could be more than 2x the time for `dict()`).

PR #3814, opened 2024-05-08
Add types for `core.periodic_table/bonds/composition/ion/lattice/libxcfunc`, new type `MillerIndex` and fix Lattice hash

DanielYang59:
### Add types annotations for:
- [x] `core.periodic_table`
- [x] `core.bonds`
- [x] `core.composition`
- [x] `core.ion`
- [x] `core.lattice`
- [x] `core.libxcfunc`

### New custom type:
- New custom type `MillerIndex = tuple[int, int, int]`

### Bug fixes:
1. Seem like a typo, I assume it should check if `HO` exists and replace them with `OH`. So the condition should be `if (self.composition.get("H") and self.composition.get("O")) is not None:`, fixed in b4a70eef7666743459a6e242a0efc18762ff3b49: https://github.com/materialsproject/pymatgen/blob/2e1c3016742d930cbdc5f4129e4036819e7755d3/pymatgen/core/ion.py#L144-L145
2. Magic number for hash of `Lattice` fixed in 88e4edad5655eac909737c568114ceb33416e95e: https://github.com/materialsproject/pymatgen/blob/2e1c3016742d930cbdc5f4129e4036819e7755d3/pymatgen/core/lattice.py#L837-L838

### Need confirmation:
1. In 710cb910e3a1322b057c868592d88e8627c5a98e, access to `Lattice._pbc` has been redirected to `Lattice.pbc`, with a `setter` method. Wondering if there is any reason for the previous implementation?
3. Is the following needed by anything? https://github.com/materialsproject/pymatgen/blob/2e1c3016742d930cbdc5f4129e4036819e7755d3/pymatgen/core/libxcfunc.py#L502-L504


DanielYang59:
@coderabbitai ignore

DanielYang59:
@janosh Can you please review this at your convenience? Feel free to revert 24b1546e98ad7d8d9a1a515b1f6d2663f7a920dc though (I personally don't like to have dunder method scattered around the class randomly, but such change badly pollute Git Blame). Thanks!

DanielYang59:
I tried to help fix https://github.com/materialsproject/pymatgen/pull/3792#discussion_r1584003750 in case you don't have time for this @janosh @JaGeo . 

But the updated implementation breaks the following test `assert c3 < c1`:
https://github.com/materialsproject/pymatgen/blob/6c696cdb327621d5d16c675478f4ed0e024feea1/tests/core/test_composition.py#L504-L516

I think the comparison between two `Composition` with different `Element` does not really make much sense (even the atomic number of `S` is greater than `O`, but then there is `SO2` > `O3` which doesn't make sense personally). 

How should we update this behavior? Should we return `False` for such cases to indicate `S` is not greater than `O` and at the same `O` is also not greater than `S` because they are not comparable (perhaps with a warning), or just raise an error?

pymatgen/core/ion.py
```diff
@@ -141,13 +178,14 @@ def get_reduced_formula_and_factor(self, iupac_ordering: bool = False, hydrates:
         el_amt_dict = {k: int(round(v)) for k, v in comp.get_el_amt_dict().items()}
         formula, factor = reduce_formula(el_amt_dict, iupac_ordering=iupac_ordering)
 
+        # DEBUG (@DanielYang59): what does the following condition check?
         if self.composition.get("H") == self.composition.get("O") is not None:
```
DanielYang59, line 182: Looks like a typo: https://github.com/materialsproject/pymatgen/blob/6c696cdb327621d5d16c675478f4ed0e024feea1/pymatgen/core/ion.py#L144-L145

pymatgen/core/libxcfunc.py
```diff
@@ -485,20 +482,17 @@ def is_hyb_mgga_family(self) -> bool:
...
         return cls[dct["name"]]
 
-    def to_json(self):
-        """Get a json string representation of the MSONable object."""
+    def to_json(self) -> str:
+        """Get a json string representation of the LibxcFunc."""
         return json.dumps(self.as_dict(), cls=MontyEncoder)
-
-
-if __name__ == "__main__":
```
DanielYang59, line 502: Not sure if we still need this.

pymatgen/core/lattice.py
```diff
@@ -67,43 +72,45 @@ def __init__(self, matrix: ArrayLike, pbc: PbcLike = (True, True, True)) -> None
...
-        matrix, lengths = self._matrix, self.lengths
-        angles = np.zeros(3)
-        for dim in range(3):
-            jj = (dim + 1) % 3
-            kk = (dim + 2) % 3
-            angles[dim] = np.clip(np.dot(matrix[jj], matrix[kk]) / (lengths[jj] * lengths[kk]), -1, 1)
-        angles = np.arccos(angles) * 180.0 / np.pi
-        return tuple(angles.tolist())  # type: ignore
+    def __hash__(self) -> int:
+        return 7  # DEBUG (DanielYang): magic hash number
```
DanielYang59, line 110: Another magic hash

pymatgen/core/lattice.py
```diff
@@ -67,43 +72,45 @@ def __init__(self, matrix: ArrayLike, pbc: PbcLike = (True, True, True)) -> None
...
         self._lll_matrix_mappings: dict[float, tuple[np.ndarray, np.ndarray]] = {}
         self._lll_inverse = None
-        if len(pbc) != 3 or {*pbc} - {True, False}:
-            raise ValueError(f"pbc must be a tuple of three True/False values, got {pbc}")
 
-        # don't import module-level, causes circular import with util/typing.py
-        from pymatgen.util.typing import PbcLike
+        if len(pbc) != 3:
+            raise ValueError(f"pbc must be a tuple of three, got {pbc}.")
+        if any(item not in {True, False} for item in pbc):
```
DanielYang59, line 78: TODO: merge this two checks

DanielYang59:
Please review this one as well @janosh. Appreciate your time!

janosh (approved):
thanks @DanielYang59, lgtm! 👍

DanielYang59:
Thanks for reviewing @janosh! 

Do you have any comments on the "Need Confirmation" section of the PR summary https://github.com/materialsproject/pymatgen/pull/3814#issue-2285421346?

Regarding (1), I noticed multiple part of `core.lattice` access `self._matrix` instead of `self.matrix` where `matrix` is a `property`. I'm not sure if this is intended? Does it offer any benefit that I'm not aware of?


For example (not exhaustive):
https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/lattice.py#L111-L112
https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/lattice.py#L201-L204

Is it related to: https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/lattice.py#L44

janosh:
using the attribute directly instead of the property method doesn't add to the call stack and so saves small amount of work.

but you can't lazy-load an attribute like you can a property. same with caching. that's probably what "Properties lazily generated for efficiency. " is referring to

> In https://github.com/materialsproject/pymatgen/commit/710cb910e3a1322b057c868592d88e8627c5a98e, access to Lattice._pbc has been redirected to Lattice.pbc, with a setter method. Wondering if there is any reason for the previous implementation?

not that i'm aware. your way makes more sense imo

janosh:
> 2\. Is the following needed by anything? https://github.com/materialsproject/pymatgen/blob/2e1c3016742d930cbdc5f4129e4036819e7755d3/pymatgen/core/libxcfunc.py#L502-L504

i think that's safe to remove

DanielYang59:
Thanks for the input. It's very helpful. I would do them in a future PR.

Also worth north is the `regen_libxcfunc.py` dev script seems to depend on the comments to work properly. This feel so weird and might be good to improve this as well.
https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/dev_scripts/regen_libxcfunc.py#L107-L111
https://github.com/materialsproject/pymatgen/blob/578d29ca2e27393c4afe38454b5915a3a984a544/pymatgen/core/libxcfunc.py#L34-L408

PR #3819, opened 2024-05-11
Improve unphysical (greater than 1) occupancy handling in `CifParser` and add missing site label `if not check_occu`

### LINKED ISSUE #3816: Argument `check_occu` is not working as intended (in `_get_structure` and `parse_structures`)
### Python version

3.11.8

### Pymatgen version

2024.2.8

### Operating system version

_No response_

### Current behavior

Hi all,
I was trying to deal with a crystal structure (in .cif format) that had warnings on having occupancy issues:
```
"UserWarning: Some occupancies ([1.0, ... (x876)]) sum to > 1! If they are within the occupancy_tolerance, they will be rescaled. The current occupancy_tolerance is set to: 1.0
  warnings.warn(msg)
```
I am unsure if it is the structure's error given that many other programs (ase, VESTA, etc) were not reporting error, so I tried to turn off the occupancy check and check myself.  When I called CifParser function `parse_structures` with `check_occu` = False, the same warning and errors were logged. 
Checking the source code I think there is something implemented that is inconsistent with the algorithm, by a quick glance:
Under `_get_structures()`:
```
 # If check_occu is True or the occupancy is greater than 0, create comp_d
            if not check_occu or occu > 0:
                coord = (x, y, z)
                match = get_matching_coord(coord)
                comp_dict = {el: max(occu, 1e-8)}
```
I think `not check_occu` is not consistent with the algorithm commented above.
Meanwhile, the warning I'm receiving comes from the following code snippet that is not switched off by "check_occu":
```
        if any(occu > 1 for occu in sum_occu):
            msg = (
                f"Some occupancies ({sum_occu}) sum to > 1! If they are within "
                "the occupancy_tolerance, they will be rescaled. "
                f"The current occupancy_tolerance is set to: {self._occupancy_tolerance}"
            )
            warnings.warn(msg)
            self.warnings.append(msg)
```

My temporarily hacky solution to my structure is to raise the occupancy_tolerance to 1000 to get the structure.

### Expected Behavior

Sites that are too close to each other are kept without raising warning and errors that prevent getting the structures.

### Minimal example

_No response_

_No response_

DanielYang59:
## Summary

To fix #3816.

- Raise `ValueError` if occupancy exceeds `occupancy_tolerance` and if `check_occu`, with a more descriptive error message.
-  [**Need Confirmation**] Allow any occupancy (ignore `occupancy_tolerance`) in cif when `not check_occu` in 7cf5ba93ed57b1167171aee7ab35043c74da3037, a warning would still be raised.
- Fix a bug where site label is lost `if not check_occu` by 05c11d4fbe95222eb4582b9986200dab42b36a00.

### A more descriptive error message for unphysical sites

With currently implementation (with `check_occu = False`), `ValueError` for occupancy beyond tolerance would be raised by the checker inside `Structure` (repored in #3816), with a misleading error message (`Some occupancies ([2.0, 2.0, 2.0, 1.0, 1.0]) sum to > 1`), where it should not be `> 1`, but `> occupancy_tolerance` instead. 

### Clarify `check_occu` 

https://github.com/materialsproject/pymatgen/blob/bb68c780834fea588f35420e1835ed3b21894a44/pymatgen/io/cif.py#L1263-L1266

The name of `check_occu` does not reflect its functionality. With current implementation, occupancy would be checked regardless of the value of `check_occu`. 

https://github.com/materialsproject/pymatgen/blob/bb68c780834fea588f35420e1835ed3b21894a44/pymatgen/io/cif.py#L1067-L1081

We would need to clarify its behavior in docstring.


tests/io/test_cif.py
```diff
@@ -735,10 +735,10 @@ def test_bad_cif(self):
         filepath = f"{TEST_FILES_DIR}/cif/bad_occu.cif"
         parser = CifParser(filepath)
         with pytest.raises(
-            ValueError, match="No structure parsed for section 1 in CIF.\nSpecies occupancies sum to more than 1!"
```
DanielYang59, line 738: Previous error message might be confusing, as this fails because the occupancy is greater than `1 + tolerance`, not just `1`.

pymatgen/io/cif.py
```diff
@@ -1196,8 +1203,7 @@ def parse_structures(
                 "in the CIF file as is. If you want the primitive cell, please set primitive=True explicitly.",
                 UserWarning,
             )
-        if not check_occu:  # added in https://github.com/materialsproject/pymatgen/pull/2836
-            warnings.warn("Structures with unphysical site occupancies are not compatible with many pymatgen features.")
```
DanielYang59, line 1200: This warning should not be raised just because `if not check_occu`, it should really check the occupancy and only warn if the occupancy is "unphysical".

DanielYang59:
@janosh. Can you please review (and comment) on this? Thanks.

pymatgen/io/cif.py
```diff
@@ -1149,7 +1149,10 @@ def get_matching_coord(
             all_species_noedit = all_species.copy()  # save copy before scaling in case of check_occu=False, used below
             for idx, species in enumerate(all_species):
                 total_occu = sum(species.values())
-                if 1 < total_occu <= self._occupancy_tolerance:
+                if check_occu and total_occu > 1 + self._occupancy_tolerance:
```
KunhuanLiu, line 1152: I think the original design of `occupancy_tolerance` is easier to understand. Changing to `1 + occupancy_tolerance` can cause more confusion.
As we discussed in #3816 the original implementation of occupany_tolerance should be fine!

DanielYang59: Thanks for your comment!

To me, the valid value range being `[expected, expected + tolerance]` instead of `[expected, tolerance]` seems more consistent with other parts of the code.

But I might revert this change now that we're trying to minimize breaking changes, clarify this in the docstring should suffice.

pymatgen/io/cif.py
```diff
@@ -314,18 +314,18 @@ class CifParser:
     def __init__(
         self,
         filename: PathLike | StringIO,
-        occupancy_tolerance: float = 1.0,
+        occupancy_tolerance: float = 0.0,
```
KunhuanLiu: I think original design here makes more sense; occupancy_tolerance defaults to 1.0, which is the expected highest occupancy of an atom site that could go. By default, when occupancy is > 1, throws an error.
Waiting on someone more experienced to comment on cases where `check_occu = False`

DanielYang59: Yes I think this is not a core concern here, and I have reverted this in 188b2d6ab6328ec2fd8e131d845c200f019e43d7

pymatgen/io/cif.py
```diff
@@ -1149,7 +1149,10 @@ def get_matching_coord(
             all_species_noedit = all_species.copy()  # save copy before scaling in case of check_occu=False, used below
             for idx, species in enumerate(all_species):
                 total_occu = sum(species.values())
-                if 1 < total_occu <= self._occupancy_tolerance:
+                if check_occu and total_occu > self._occupancy_tolerance:
+                    raise ValueError(f"Occupancy {total_occu} exceeded tolerance.")
```
DanielYang59, line 1153: This would help capture `occupancy > tolerance`. Without this, if an occupancy is greater than `tolerance`, it would not be scaled, and lead to the following error message which doesn't show the reason for failure:

```
Some occupancies ([2.0, 2.0, 2.0, 1.0, 1.0]) sum to > 1! If they are within the occupancy_tolerance, they will be rescaled. The current occupancy_tolerance is set to: 1.0
No structure parsed for section 1 in CIF.
...
in CifParser.parse_structures(self, primitive, symmetrized, check_occu, on_error)
   1219     warnings.warn("Issues encountered while parsing CIF: " + "\n".join(self.warnings))
   1221 if len(structures) == 0:
-> 1222     raise ValueError("Invalid CIF file with no structures!")
   1223 return structures

ValueError: Invalid CIF file with no structures!
```

PR #3822, opened 2024-05-12
Add types annotations for `core.interface`

DanielYang59:
## Summary

- Add types annotations for `core.interface`
- Use more specific types for `ClassVar`

### New custom type aliases
```
Tuple3Ints = tuple[int, int, int]
Tuple3Floats = tuple[float, float, float]
```


DanielYang59:
Not working on this for now, would reopen it later.

pymatgen/core/xcfunc.py
```diff
@@ -101,13 +101,13 @@ class type_name(NamedTuple):
         (xcf.GGA_X_B88, xcf.GGA_C_LYP): type_name("GGA", "BLYP"),
     }
 
-    del type_name
+    # del type_name
```
DanielYang59, line 104: Delete `type_name` right after usage seem to break `TYPE_CHECKING` for `ClassVar`, need to confirm if this would incur any significant memory penalty.

pymatgen/io/abinit/pseudos.py
```diff
@@ -1024,7 +1024,7 @@ class ppdesc(NamedTuple):
         10: ppdesc(10, "HGHK", "NC", None),
     }
 
-    del ppdesc
+    # del ppdesc
```
DanielYang59, line 1027: Same to https://github.com/materialsproject/pymatgen/pull/3822/commits/4484d34a28d26874708b1f708a05a2227b7a55e5#r1623746333

pymatgen/core/interface.py
```diff
@@ -392,7 +397,7 @@ def gb_from_parameters(
         ab_shift: tuple[float, float] = (0, 0),
         normal: bool = False,
         ratio: list[int] | None = None,
-        plane: list[int] | None = None,
+        plane: tuple[int, int, int] | None = None,
```
DanielYang59, line 400: Similar to the kpoints type, I have to change `plane` to `tuple[int, int, int]` in order to track its length. I'm hoping no code would reply on its type.

DanielYang59:
@janosh Would appreciate it if you could review this at your convenience, thanks!

tests/core/test_interface.py
```diff
@@ -38,11 +38,11 @@ def setUp(self):
...
         assert self.Cu_GB1.rotation_angle == approx(123.74898859588858)
         assert self.Cu_GB1.vacuum_thickness == approx(1.5)
-        assert self.Cu_GB2.rotation_axis == [1, 2, 3]
+        assert_allclose(self.Cu_GB2.rotation_axis, [1, 2, 3])
         assert_allclose(self.Cu_GB1.ab_shift, [0.0, 0.0])
         assert_allclose(self.Cu_GB2.ab_shift, [0.2, 0.2])
-        assert self.Cu_GB1.gb_plane == [1, 3, 1]
-        assert self.Cu_GB2.gb_plane == [1, 2, 3]
+        assert_allclose(self.Cu_GB1.gb_plane, [1, 3, 1])
+        assert_allclose(self.Cu_GB2.gb_plane, [1, 2, 3])
```
janosh, line 45: i would prefer the old code. no need to use `assert_allclose` with integers. better to have more sensitive tests that fail if we expect something to be integers and somehow they become floats

janosh (changes requested):
thanks @DanielYang59! this all looks good. just 1 minor comment regarding the test change

tests/core/test_interface.py
DanielYang59: Okay sure f3cabe8d1a7017ef38050d4620ae359783b36c98. But then we need to change its type to `tuple` :)

pymatgen/core/xcfunc.py
janosh: i wouldn't worry about mem overhead here. the `type_name` variable goes out of scope and will be garbage-collected on line 121 anyway

pymatgen/core/interface.py
janosh: given it's always length 3, a tuple would have been the better type here from the start

tests/core/test_interface.py
janosh: yes, feel free to remove the `del type_name` and `del ppdesc` lines

DanielYang59: Thanks for the confirmation :)

pymatgen/core/xcfunc.py
DanielYang59: Thanks!

pymatgen/core/interface.py
DanielYang59: Thanks for confirming, and the same applies to several other variables:
```python
    @staticmethod
    def get_trans_mat(
        r_axis: tuple[int, int, int] | tuple[int, int, int, int],
        angle: float,
        normal: bool = False,
        trans_cry: NDArray | None = None,
        lat_type: str = "c",
        ratio: list[int] | None = None,
        surface: tuple[int, int, int] | tuple[int, int, int, int] | None = None,
        max_search: int = 20,
        quick_gen: bool = False,
    ):
```

I'm wondering however, is there a more efficient way to express the length of the `tuple` when it could be one of several possible lengths, other than the super long `r_axis: tuple[int, int, int] | tuple[int, int, int, int],`?

janosh: you could define type variables:

```py
Tuple3Ints = tuple[int, int, int]
Tuple4Ints = tuple[int, int, int, int]
Tuple3Ints | Tuple4Ints
```

DanielYang59: Yes this is a more concise way :) typing `tuple[int, ...]` is making the line length break the roof.
Would implement this in this PR

DanielYang59: Yep, we're thinking the same :)

DanielYang59:
@janosh Just ping you again in case it got forgotten :) No rush

pymatgen/io/pwmat/outputs.py
```diff
@@ -190,12 +190,12 @@ def __init__(self, filename: PathLike):
...
         self._kpts, self._kpts_weight, self._hsps = self._parse_kpt()
 
-    def _parse_band(self) -> tuple[int, int, int]:
+    def _parse_band(self) -> Tuple3Ints:
         """Parse REPORT file to obtain spin switches, the number of kpoints
         and the number of bands.
 
         Returns:
-            tuple[int, int, int]: containing:
+            Tuple3Ints: containing:
```
janosh, line 198: i think for doc strings it's better to keep the explicit type since they're rendered as [HTML in the API docs](https://pymatgen.org/pymatgen.html) where people would need to go hunting for the definition of that type variable to understand what it means

janosh (changes requested):
thanks for the consistent use of `tuple` type aliases!

pymatgen/io/pwmat/outputs.py
DanielYang59: Done in 64924f3e8f3b66d5b5a9c9d0184e1b6b38a06e7f. Thanks for the comment :)

pymatgen/core/interface.py
```diff
@@ -22,18 +24,21 @@
...
 
-    from numpy.typing import ArrayLike
+    from numpy.typing import ArrayLike, NDArray
     from typing_extensions import Self
 
-    from pymatgen.util.typing import CompositionLike, Vector3D
+    from pymatgen.core import Element
+    from pymatgen.util.typing import CompositionLike, Matrix3D, MillerIndex, Tuple3Floats, Vector3D
+
+Tuple4Ints = tuple[int, int, int, int]
```
DanielYang59, line 39: I put it here because it doesn't seem to be needed by other code, `tuple[int, int, int, int]` doesn't seem to be commonly used. But we could relocate it if needed

janosh (approved):
thanks @DanielYang59! 👍

pymatgen/core/interface.py
janosh: ah, didn't see that. i would relocate right away

DanielYang59:
Thanks for reviewing! Appreciate that

PR #3826, opened 2024-05-13
Revert to `pytorch` installation with `uv` in CI

DanielYang59:
### Revert to install `torch` with `uv`

Also link to #4063, **`pip install torch` is taking up 70% of the total dep install time**

Revert to `pytorch` installation with `uv` in CI, 
- [x] It appears https://github.com/astral-sh/uv/issues/1921 has been resolved.
 - [ ] [The following intermittent error](https://github.com/materialsproject/pymatgen/actions/runs/9051480913/job/24868047255) causes false negative CI failures, reported in https://github.com/astral-sh/uv/issues/2586:
```
error: Failed to download distributions
  Caused by: Failed to fetch wheel: torch==2.2.1
  Caused by: Failed to extract archive
  Caused by: error decoding response body
  Caused by: request or response body error
  Caused by: error reading a body from connection
  Caused by: end of file before message length reached
```
- [ ] Also https://github.com/astral-sh/uv/issues/4402.


janosh (approved):
thanks @DanielYang59! 👍

DanielYang59:
> we're getting
> 
> ```
> pytest: command not found
> ```
> 
> which is not a torch related error message?

My bad, I didn't said it clear enough. If you look more closely into the log, the failure actually comes from the "Install pymatgen and dependencies" stage (for some reason it still passed).

image

janosh:
at least it's no longer failing due to timeouts. looks like now we're hitting this issue https://github.com/astral-sh/uv/issues/2586

DanielYang59:
> at least it's no longer failing due to timeouts. looks like now we're hitting this issue [astral-sh/uv#2586](https://github.com/astral-sh/uv/issues/2586)

Look like this one. I guess we would need to still install `torch` with `pip`, and somehow prevent `uv` from trying to install `torch` later (not sure how to achieve this though, I would check its manual).

DanielYang59:
Also link to #4063, `pip install torch` is taking up 70% of the total dep install time

.github/workflows/test.yml
```diff
@@ -74,15 +74,16 @@ jobs:
           micromamba install -n pmg -c conda-forge bader enumlib openff-toolkit packmol pygraphviz tblite --yes
 
       - name: Install pymatgen and dependencies via uv
+        env:
+          UV_REQUEST_TIMEOUT: 300  # torch could easily timeout
```
DanielYang59, line 78: Add this timeout from https://github.com/astral-sh/uv/pull/1694

.github/workflows/test.yml
```diff
@@ -74,15 +74,16 @@ jobs:
           micromamba install -n pmg -c conda-forge bader enumlib openff-toolkit packmol pygraphviz tblite --yes
 
       - name: Install pymatgen and dependencies via uv
+        env:
+          UV_HTTP_TIMEOUT: 300  # torch could easily timeout
```
DanielYang59, line 78: https://docs.astral.sh/uv/configuration/environment/

> HTTP_TIMEOUT (or UV_HTTP_TIMEOUT): If set, uv will use this value (in seconds) as the timeout for HTTP reads (default: 30 s).

DanielYang59:
Closing this one in #4100, reason: currently `chgnet` and `matgl` doesn't support Python 3.13, so `torch` would not be required in CI for now.

PR #3827, opened 2024-05-13
Guard `TYPE_CHECKING` only imports

DanielYang59:
## Summary

- Guard `TYPE_CHECKING` only imports
- Format docstring to Google style in `io.aims.sets` 

### Can we automate this check (and fix)?

`flake8` provides [`TYP001` rule](https://github.com/asottile/flake8-typing-imports) to check this, but doesn't seem able to correct it automatically. Have to run `flake8 --select=TYP001` and fix manually.

`ruff` might [have this implemented](https://github.com/astral-sh/ruff/issues/2096) but I didn't find it in the [available rules](https://docs.astral.sh/ruff/rules/).


DanielYang59:
@janosh. Please review this, thanks.

Wondering if we could automate this checking?

janosh:
> Wondering if we could automate this checking?

`ruff` has an open issue about implementing this rule which you could upvote and/or leave a comment

https://github.com/astral-sh/ruff/issues/2302

DanielYang59:
I would resolve it now :)

DanielYang59:
Resolved.

janosh:
is the main benefit here meant to be code clarity? that type-checking-time only imports are clearly separated? i don't think we get any speedup since we're importing the `typing` module anyway to get `TYPE_CHECKING`

DanielYang59:
Thanks for reviewing. And sorry I'm unable to help completely guard all imports by `TYPE_CHECKING`

PR #3831, opened 2024-05-15
Move test structures out of `util` directory

DanielYang59:
## Summary

- Move test structures out of `util` directory

We might need a better name for the directory?


janosh (approved):
thanks @DanielYang59! 👍 i don't recall why those structures weren't in the `tests/files` folder to begin with

DanielYang59:
Thanks for reviewing.

Meanwhile I would say `uv install torch` seems to have a higher and higher failure rate for some reason. We have always been seeing it to fail. Should we consider revert to slower but more reliable `pip`?

janosh:
we're already installing it with `pip`. maybe there's a flag in `uv pip install` to tell to ignore packages which are already installed. just to confirm, currently `uv` tries to install `pytorch` again?

https://github.com/materialsproject/pymatgen/blob/daa5c75573862aae3fbc0fc1a1ed421791dc603f/.github/workflows/test.yml#L73

DanielYang59:
> we're already installing it with `pip`. 

Yes we are.

> maybe there's a flag in `uv pip install` to tell it to ignore packages which are already installed? just to confirm, currently `uv` tries to install `pytorch` again?

I'm not aware of such arg, but I would do some research and let you know. 

I don't know how `uv/pip` works under the hood, but to detect installed packages, `uv` seems to need to same search process just like install, and once this failed, the install stage would fail. 

So you are right, we need to explicitly tell `uv` to ignore `torch`, we cannot let it detect itself

PR #3838, opened 2024-05-18
[Deprecation] Correct cases for memory units in `core.units` and fix `unit_type` detection of `FloatWithUnit.from_str`

### LINKED ISSUE #3833: Incorrect letter cases for memory in `core.units`
It appears the case for [multiple-byte units](https://en.wikipedia.org/wiki/Byte) for memory is incorrect. And we should be aware that `b/B` refer to `bit/byte` respectively.

https://github.com/materialsproject/pymatgen/blob/bb68c780834fea588f35420e1835ed3b21894a44/pymatgen/core/units.py#L77-L87

DanielYang59:
## Summary

- Correct cases for memory units in `core.units` to fix #3833
- Fix `FloatWithUnit.from_str` to correctly detect `unit_type` from string with space between the value and the unit 

### Correct cases for memory units in `core.units`

I assume this is another breaking fix, so I marked the incorrect units as deprecated until `2025-01-01`. Any suggestion would be appreciated.

Currently the case of `byte` is incorrect in `core.units`, where [`B` refers to "byte" and `b` refers to "bit"](https://en.wikipedia.org/wiki/Byte). 
Also note `1 byte = 8 bit` so we cannot use them interchangeably.

https://github.com/materialsproject/pymatgen/blob/bb68c780834fea588f35420e1835ed3b21894a44/pymatgen/core/units.py#L77-L83

#### Case sensitivity for memory units

I assume there is no good reason to remove case sensitivity for `memory` **alone** (case sensitivity of other units is maintained), especially when the case of `memory` has its meaning:
https://github.com/materialsproject/pymatgen/blob/14b7357d373c1eae88a3ca014fb14ffba44df75f/pymatgen/core/units.py#L90-L91

### Fix `FloatWithUnit.from_str` to correctly detect `unit_type` from string with space between value and unit 

Previously for a string with space (which adheres to the [SI Unit rules](https://physics.nist.gov/cuu/Units/checklist.html)) like `Memory.from_str("1 MB")`, the string would be split into `1` and ` MB`, and the space in ` MB` prevents the `unit_type` being detected properly and ended up being `None`.

```python
from pymatgen.core.units import Memory


mem_0 = Memory.from_str("1MB")

print(mem_0.unit_type)  # >>> "memory"

mem_1 = Memory.from_str("1 MB")

print(mem_1.unit_type)  # >>> None
```


tests/core/test_units.py
```diff
@@ -94,11 +94,11 @@ def test_length(self):
         assert str(y.unit) == "ang^3"
 
     def test_memory(self):
-        mega = Memory(1, "Mb")
+        mega = Memory(1, "MB")
         assert mega.to("byte") == 1024**2
         assert mega == Memory(1, "mb")
 
-        same_mega = Memory.from_str("1Mb")
+        same_mega = Memory.from_str("1MB")
```
DanielYang59, line 101: For some reason `Memory.from_str("1 MB")` failed to detect `unit_type`, which works for other units.

As per the [SI Unit rules and style conventions](https://physics.nist.gov/cuu/Units/checklist.html):
> There is a space between the numerical value and unit symbol, even when the value is used in an adjectival sense, except in the case of superscript units for plane angle.

tests/core/test_units.py
```diff
@@ -94,11 +94,11 @@ def test_length(self):
         assert str(y.unit) == "ang^3"
 
     def test_memory(self):
-        mega = Memory(1, "Mb")
+        mega = Memory(1, "MB")
         assert mega.to("byte") == 1024**2
         assert mega == Memory(1, "mb")
 
-        same_mega = Memory.from_str("1Mb")
+        same_mega = Memory.from_str("1 MB")
```
DanielYang59, line 101: As per the [SI Unit rules and style conventions](https://physics.nist.gov/cuu/Units/checklist.html):
> There is a space between the numerical value and unit symbol, even when the value is used in an adjectival sense, except in the case of superscript units for plane angle.

DanielYang59:
@shyuep Can I have your advice on this please? This should be very similar to the change to `Element`, and I scheduled the same grace period until `2025-01-01`

DanielYang59:
@shyuep Can you please review this? It should be very similar to #3817.

The rationale for this change is that something like `kB` (kilobyte)  and `kb`(kilobit) refer to different units for memory (or storage in general), and as such it should not be case insensitive.

shyuep:
This is fine. I am enabling auto-merge. There was a failing test that needs to be fixed.

DanielYang59:
Thanks. I notice that too https://github.com/materialsproject/pymatgen/commit/e572d34b206f48a43fe102a1c46e0532a60ff7de#r143073539, perhaps we also want to remove the `id` argument to avoid shadowing the builtin in https://github.com/materialsproject/pymatgen/commit/e572d34b206f48a43fe102a1c46e0532a60ff7de#r143131235?

shyuep:
Thanks. I renamed id to id_

DanielYang59:
Thanks!

PR #3848, opened 2024-05-30
Simplify case handling in `lobster.inputs`

DanielYang59:
## Simplify case handling in `lobster.inputs`

- [**Potentially breaking**] Rename class variable `LISTKEYWORDS` to `LIST_KEYWORDS` for consistency.

#### Use `{lowercase: camalCase}` to simplify key case handling, for the following methods:
Now lower-case keys are used internally, and Camel case keys are reserved in case of writing (good looking) output files.
- magic `dict` methods
- `from_file`
- `write_lobsterin`: Also use it to recover the key case.
- `diff`

#### TODOs
- [x] Update available `lobsterin` keywords


pymatgen/io/lobster/inputs.py
```diff
@@ -116,74 +116,80 @@ class Lobsterin(UserDict, MSONable):
...
+            raise KeyError(f"{key=} is not available") from exc
 
     def __contains__(self, key) -> bool:
-        """Implements getitem from dict to avoid problems with different key casing."""
-        normalized_key = next((k for k in self if key.strip().lower() == k.lower()), key)
+        """To avoid cases sensitivity problems."""
+        return key.lower().strip() in self.data
 
-        key_is_unknown = normalized_key.lower() not in map(str.lower, Lobsterin.AVAILABLE_KEYWORDS)
-        return not key_is_unknown and normalized_key in self.data
```
DanielYang59, line 166: This doesn't look necessary, user cannot set a key that is not in `AVAILABLE_KEYWORDS` because it has already been guarded by `__setitem__`

pymatgen/io/lobster/inputs.py
```diff
@@ -116,176 +115,140 @@ class Lobsterin(UserDict, MSONable):
...
             else:
-                similar_param[k1.lower()] = v1
+                same_params[k1] = v1
 
+        # Check other
         for k2, v2 in other.items():
-            if (
-                k2.lower() not in similar_param
-                and k2.lower() not in different_param
-                and k2.lower() not in [key.lower() for key in self]
```
DanielYang59, line 221: This check logic looks a bit over-complicated. I hope I didn't misunderstand here.

If `k2` is not in `self`, it then cannot be in the union `same_params`.

DanielYang59:
@JaGeo Can you please review this? And let me know if there is any concern, thanks :)

JaGeo:
Maybe next week. Sorry!

DanielYang59:
> Maybe next week. Sorry!

All good. No worries at all, whenever you have time :)

QuantumChemist:
> @JaGeo Can you please review this? And let me know if there is any concern, thanks :)

I could also review it on the weekend :)

QuantumChemist:
> @QuantumChemist Sure. Thank you!!!
> 
> I would love to have a final look before we merge it as it is so important for all our LOBSTER workflows and to avoid breakages here 😅

Of course. My review would be additionally :)

DanielYang59:
> I could also review it on the weekend :)

That would be hugely appreciated :) Let me know if you find anything suspicious.

pymatgen/io/lobster/inputs.py
```diff
@@ -160,21 +160,21 @@ def __setitem__(self, key, val: Any) -> None:
 
         super().__setitem__(key, val.strip() if isinstance(val, str) else val)
 
-    def __getitem__(self, key) -> Any:
+    def __getitem__(self, key: str) -> Any:
         """To avoid cases sensitivity problems."""
         try:
-            return self.data[key.strip().lower()]
+            return super().__getitem__(key.strip().lower())
```
DanielYang59, line 166: I'm not sure if we have a reason to bypass the method resolution order (especially when we chose to inherit from `UserDict` here), let me know if there's anything that I'm not aware of :)

pymatgen/io/lobster/inputs.py
```diff
@@ -1,6 +1,6 @@
-"""
-Module for reading Lobster input files. For more information
-on LOBSTER see www.cohp.de.
+"""Module for reading Lobster input files.
```
QuantumChemist, line 1:
```suggestion
"""Module for reading LOBSTER input files.
```

pymatgen/io/lobster/inputs.py
```diff
@@ -54,25 +54,24 @@
 
 
 class Lobsterin(UserDict, MSONable):
-    """
-    This class can handle and generate lobsterin files
-    Furthermore, it can also modify INCAR files for lobster, generate KPOINT files for fatband calculations in Lobster,
+    """Handle and generate lobsterin files.
+    Furthermore, it can also modify INCAR files for LOBSTER, generate KPOINTS files for fatband calculations in Lobster,
```
QuantumChemist, line 58:
```suggestion
    Furthermore, it can also modify INCAR files for LOBSTER, generate KPOINTS files for fatband calculations in LOBSTER,
```

pymatgen/io/lobster/inputs.py
```diff
@@ -54,25 +54,24 @@
...
     and generate the standard primitive cells in a POSCAR file that are needed for the fatband calculations.
     There are also several standard lobsterin files that can be easily generated.
-    """
 
-    # reminder: lobster is not case sensitive
+    Reminder: LOBSTER is not case sensitive.
```
QuantumChemist, line 62:
```suggestion
    Reminder: lobsterin input commands are not case sensitive.
```

pymatgen/io/lobster/inputs.py
```diff
@@ -54,25 +54,24 @@
...
+    _FLOAT_KEYWORDS: tuple[str, ...] = (
         "COHPstartEnergy",
         "COHPendEnergy",
         "gaussianSmearingWidth",
         "useDecimalPlaces",
         "COHPSteps",
     )
-    # one of these keywords +endstring can be used in file
-    STRING_KEYWORDS = (
+    # One of these keywords + string
```
QuantumChemist, line 73: is the "can be used in file ..." missing here?

pymatgen/io/lobster/inputs.py
```diff
@@ -82,8 +81,8 @@ class Lobsterin(UserDict, MSONable):
         "kSpaceCOHP",
         "EwaldSum",
     )
-    # the keyword alone will turn on or off a function
-    BOOLEAN_KEYWORDS = (
+    # The keywords alone will turn on or off a function
```
QuantumChemist, line 84:
```suggestion
    # These keywords can trigger additional functionalities
```

pymatgen/io/lobster/inputs.py
```diff
@@ -116,176 +115,139 @@ class Lobsterin(UserDict, MSONable):
...
-    AVAILABLE_KEYWORDS = FLOAT_KEYWORDS + STRING_KEYWORDS + BOOLEAN_KEYWORDS + LISTKEYWORDS
+    # Generate {lowered: original} mappings
+    FLOAT_KEYWORDS: ClassVar = {key.lower(): key for key in _FLOAT_KEYWORDS}
+    STRING_KEYWORDS: ClassVar = {key.lower(): key for key in _STRING_KEYWORDS}
+    BOOLEAN_KEYWORDS: ClassVar = {key.lower(): key for key in _BOOLEAN_KEYWORDS}
+    LIST_KEYWORDS: ClassVar = {key.lower(): key for key in _LIST_KEYWORDS}
 
-    def __init__(self, settingsdict: dict):
+    # All known keywords
+    AVAILABLE_KEYWORDS: ClassVar = {**FLOAT_KEYWORDS, **STRING_KEYWORDS, **BOOLEAN_KEYWORDS, **LIST_KEYWORDS}
```
QuantumChemist, line 132: is this up-to-date with lobster5?

pymatgen/io/lobster/inputs.py
```diff
@@ -116,176 +115,139 @@ class Lobsterin(UserDict, MSONable):
...
-        leading and trailing white spaces. Similar to INCAR class.
-        """
-        # due to the missing case sensitivity of lobster, the following code is necessary
-        new_key = next((key_here for key_here in self if key.strip().lower() == key_here.lower()), key)
-
-        if new_key.lower() not in [element.lower() for element in Lobsterin.AVAILABLE_KEYWORDS]:
-            raise KeyError("Key is currently not available")
-
-        super().__setitem__(new_key, val.strip() if isinstance(val, str) else val)
+        Necessary due to the missing case sensitivity of LOBSTER.
```
QuantumChemist, line 150:
```suggestion
        Necessary due to the missing case sensitivity of input keywords in LOBSTER.
```

pymatgen/io/lobster/inputs.py
```diff
@@ -116,176 +115,139 @@ class Lobsterin(UserDict, MSONable):
...
+        """To avoid cases sensitivity problems."""
+        super().__delitem__(key.lower().strip())
 
-    def diff(self, other):
-        """
-        Diff function for lobsterin. Compares two lobsterin and indicates which parameters are the same.
-        Similar to the diff in INCAR.
+    def diff(self, other: Self) -> dict[str, dict[str, Any]]:
+        """Compare two Lobsterin and find which parameters are the same.
+        Similar to the diff method of Incar.
```
QuantumChemist, line 181:
```suggestion
        Similar to the diff method of INCAR.
```

pymatgen/io/lobster/inputs.py
```diff
@@ -639,28 +642,27 @@ def _get_potcar_symbols(POTCAR_input: str) -> list:
...
+        INCAR_input: PathLike = "INCAR",
+        POTCAR_input: PathLike | None = None,
+        Vasprun_output: PathLike = "vasprun.xml",
         dict_for_basis: dict | None = None,
         option: str = "standard",
-    ):
-        """
-        Will generate Lobsterin with standard settings.
+    ) -> Self:
+        """Generate Lobsterin with standard settings.
```
QuantumChemist, line 652:
```suggestion
        """Generate lobsterin with standard settings.
```

pymatgen/io/lobster/inputs.py
```diff
@@ -688,13 +691,13 @@ def standard_calculations_from_vasp_files(
             "onlycohpcoop",
             "onlycohpcoopcobi",
             "onlymadelung",
-        ]:
+        }:
             raise ValueError("The option is not valid!")
 
         lobsterin_dict: dict[str, Any] = {
-            # this basis set covers most elements
+            # This basis set covers most elements
```
QuantumChemist, line 698:
```suggestion
            # This basis set covers elements up to Lr, Z = 103
```

pymatgen/io/lobster/inputs.py
```diff
@@ -728,114 +731,120 @@ def standard_calculations_from_vasp_files(
...
         if dict_for_basis is not None:
-            # dict_for_basis={"Fe":'3p 3d 4s 4f', "C": '2s 2p'}
-            # will just insert this basis and not check with poscar
+            # dict_for_basis = {"Fe":"3p 3d 4s 4f", "C": "2s 2p"}
+            # Will just insert this basis and not check with poscar
             basis = [f"{key} {value}" for key, value in dict_for_basis.items()]
         elif POTCAR_input is not None:
-            # get basis from POTCAR
-            potcar_names = Lobsterin._get_potcar_symbols(POTCAR_input=POTCAR_input)
+            # Get basis from POTCAR
```
QuantumChemist, line 819:
```suggestion
            # Get basis functions from POTCAR
```

tests/io/lobster/test_inputs.py
```diff
@@ -1562,7 +1562,7 @@ def setUp(self):
         self.Lobsterin4 = Lobsterin.from_file(f"{TEST_DIR}/lobsterin.4.gz")
 
     def test_from_file(self):
-        # test read from file
+        # Test read from file
```
QuantumChemist, line 1565:
```suggestion
        # Test reading from file
```

QuantumChemist (commented):
My suggestions are more of cosmetic nature 🙈

pymatgen/io/lobster/inputs.py
DanielYang59: Oops... I tried to correct cases for `LOBSTER` (which refers to the package, instead of the class), and it seems I missed some. Thanks for catching this!

pymatgen/io/lobster/inputs.py
DanielYang59: This should not be changed, `Incar` (Camel case) referes to the `pymatgen` `Incar` class, not the VASP input file `INCAR`.

pymatgen/io/lobster/inputs.py
DanielYang59: Perhaps we need to change this for consistency as well: https://github.com/materialsproject/pymatgen/blob/2982968a44d0ff65392057488aee4871210bc847/pymatgen/io/lobster/inputs.py#L342-L348

pymatgen/io/lobster/inputs.py
DanielYang59: Yeps... But this comment and its siblings are a bit ambiguous to me?

https://github.com/materialsproject/pymatgen/blob/2982968a44d0ff65392057488aee4871210bc847/pymatgen/io/lobster/inputs.py#L66
https://github.com/materialsproject/pymatgen/blob/2982968a44d0ff65392057488aee4871210bc847/pymatgen/io/lobster/inputs.py#L74
https://github.com/materialsproject/pymatgen/blob/2982968a44d0ff65392057488aee4871210bc847/pymatgen/io/lobster/inputs.py#L85

Especially the last one: https://github.com/materialsproject/pymatgen/blob/2982968a44d0ff65392057488aee4871210bc847/pymatgen/io/lobster/inputs.py#L119

pymatgen/io/lobster/inputs.py
DanielYang59: I think the previous expression is already clear enough for this one... And I would like to hear about other comments too :)

pymatgen/io/lobster/inputs.py
DanielYang59: Thanks for clarifying this, perhaps it's better for us to use consistent wording here https://github.com/materialsproject/pymatgen/pull/3848#discussion_r1623648797? I would personally prefer "input keywords" over "input commands", because to me, "command" sounds more like something people might type in to the terminal :)

pymatgen/io/lobster/inputs.py
DanielYang59: I should trust you as a LOBSTER developer 😄 , but I still checked the manual and confirmed this: 
> If you wish to override the basis set selected by default, you
may do so with this keyword. It currently supports bunge
(for elements up to Xe), koga and pbeVaspFit2015 (up to
Lr; yes, that is Z = 103).

pymatgen/io/lobster/inputs.py
DanielYang59: I think no... Upon checking the manual, LOBSTER 5.0.0 might have introduced the following additional keywords:
> printLmosOnAtoms Followed by indices of targeted atoms, this keyword will
calculate embedded localized molecular orbitals as formu-
lated in M. Pauls, D. Schnieders, R. Dronskowski, J. Phys.
Chem. A 2023, 127, 6541–6551 and print those centered on
the requested atoms to file in a cube file format.

> printTotalSpilling This keyword will instruct LOBSTER to print the
absolute total spilling next to the absolute charge
spilling.

> customSTOforAtom This keyword allows to define custom STO basis
sets for atoms. The syntax is as follows:
useBasisSet custom
customSTOforAtom [ELEMENT] [BASIS_FILE]
Where [ELEMENT] needs to be replaced by the ele-
ment symbol and [BASIS_FILE] is a basis set file
residing within the working directory that defines
the basis. An example of such basis set file is
shipped with every copy of LOBSTER.

pymatgen/io/lobster/inputs.py
QuantumChemist: Oh sorry, that's my oversight!

pymatgen/io/lobster/inputs.py
QuantumChemist: I found the "alone" part strange and "turn on or off" a somewhat cumbersome expression.

pymatgen/io/lobster/inputs.py
QuantumChemist: this might be something worth to check carefully as e.g `cobibetween` is available since lobster 4.0.0

pymatgen/io/lobster/inputs.py
QuantumChemist: yup, I think these commands can be summarized/fused and then be put into the docstring, where it's easier to look up the usage.

pymatgen/io/lobster/inputs.py
QuantumChemist: "Keyword" and "command" is used interchangeably e.g. `List of keywords`.  🙈 But yeah, I think, although it should be clear what this is referring to, it can't be a bad idea to be language-wise concise. I think `lobsterin keywords` (without the "input" as I suggested above) is the most descriptive and concise.

pymatgen/io/lobster/inputs.py
QuantumChemist: I copied it from the manual haha

pymatgen/io/lobster/inputs.py
QuantumChemist: It might be a good idea to be language-wise concise here as well, to just prevent that people potentially confuse it with "basis set".

pymatgen/io/lobster/inputs.py
```diff
@@ -335,7 +336,7 @@ def get_basis(
         potcar_symbols: list[str],
         address_basis_file: PathLike | None = None,
     ) -> list[str]:
-        """Get the basis from given potcar_symbols, e.g., ["Fe_pv", "Si"].
+        """Get the basis functions from given potcar_symbols, e.g., ["Fe_pv", "Si"].
```
QuantumChemist, line 347: ah, great to see that you changed it here as well :D

pymatgen/io/lobster/inputs.py
DanielYang59: All good, just a single oversight among these many good ones :)

pymatgen/io/lobster/inputs.py
DanielYang59: Yep :) thanks

pymatgen/io/lobster/inputs.py
DanielYang59: > I found the "alone" part strange and "turn on or off" a somewhat cumbersome expression.

I guess `along` refers to no additional arguments (just the command), for example `skipdos` vs `COHPstartEnergy -35`, yes the `turn on or off` sounds a bit too verbose

pymatgen/io/lobster/inputs.py
DanielYang59: Sure I would change these

pymatgen/io/lobster/inputs.py
DanielYang59: > yup, I think these commands can be summarized/fused and then be put into the docstring, where it's easier to look up the usage.

I think this sounds like a task for the LOBSTER manual, not `pymatgen`? Perhaps we should somehow clarify these expressions here (or maybe include an example for each?)

pymatgen/io/lobster/inputs.py
QuantumChemist: I know what the "alone" was referring to. I'm just thinking of a way to say this more elegant and efficient.

pymatgen/io/lobster/inputs.py
QuantumChemist: ah now that I read it again, I think "+ float" and "+ string" is understandable.

pymatgen/io/lobster/inputs.py
DanielYang59: Yes I would trust you as the language expert :) Perhaps asking LLMs

JaGeo:
```suggestion
    # The keywords themselves (without suffix) can trigger additional functionalities.
```

pymatgen/io/lobster/inputs.py
```diff
@@ -54,25 +54,24 @@
...
+    _FLOAT_KEYWORDS: tuple[str, ...] = (
         "COHPstartEnergy",
         "COHPendEnergy",
         "gaussianSmearingWidth",
         "useDecimalPlaces",
         "COHPSteps",
     )
-    # one of these keywords +endstring can be used in file
-    STRING_KEYWORDS = (
+    # One of these keywords + string
```
JaGeo, line 73: Maybe something like this?

```suggestion
    # These keywords need an additional string suffix
```

pymatgen/io/lobster/inputs.py
JaGeo: I am not sure what you mean with bypass here?

pymatgen/io/lobster/inputs.py
JaGeo: Several might be missing. I did only update those that I did not need for the automation.

pymatgen/io/lobster/inputs.py
JaGeo: I actually don't think the commments should be removed. They refer to the way that we later on store the information in pymatgen. This has not much to do with the Lobster manual itself as the Lobster manual does not introduce any classification of their commands.

pymatgen/io/lobster/inputs.py
```diff
@@ -116,176 +117,145 @@ class Lobsterin(UserDict, MSONable):
...
+        same_params = {}
+        diff_params = {}
 
+        # Check self
         for k1, v1 in self.items():
-            k1_lower = k1.lower()
-            k1_in_other = next((key_here for key_here in other if key_here.lower() == k1_lower), k1_lower)
-            if k1_lower not in key_list_others:
-                different_param[k1.lower()] = {"lobsterin1": v1, "lobsterin2": None}
+            if k1 not in other:
```
JaGeo, line 202: I am actually not sure this will work for all cases. Did you check with case sensitivity?

pymatgen/io/lobster/inputs.py
JaGeo: I really went recently through this code and it all worked. I am very much afraid so many drastic chances might break a lot of things. This will take a lot of review time that I currently don't really have. We might need to leave this for a few weeks.

pymatgen/io/lobster/inputs.py
```diff
@@ -301,27 +271,53 @@ def from_dict(cls, dct: dict) -> Self:
...
 
         Args:
-            incar_input (str): path to input INCAR
-            incar_output (str): path to output INCAR
-            poscar_input (str): path to input POSCAR
-            isym (int): isym equal to -1 or 0 are possible. Current Lobster version only allow -1.
+            incar_input (PathLike): path to input INCAR
+            incar_output (PathLike): path to output INCAR
+            poscar_input (PathLike): path to input POSCAR
+            isym (int): ISYM equals -1 or 0 are possible. Current LOBSTER version only allow -1.
```
JaGeo, line 317:
```suggestion
            isym (int): ISYM equals -1 or 0 are possible.
```

The last part can be removed now.

pymatgen/io/lobster/inputs.py
```diff
@@ -301,27 +271,53 @@ def from_dict(cls, dct: dict) -> Self:
...
+        incar_output: PathLike = "INCAR.lobster",
+        poscar_input: PathLike = "POSCAR",
+        isym: Literal[-1, 0] = -1,
         further_settings: dict | None = None,
-    ):
-        """Will only make the run static, insert nbands, make ISYM=-1, set LWAVE=True and write a new INCAR.
-        You have to check for the rest.
+    ) -> None:
+        """Write INCAR file. Will only make the run static,insert NBANDS,
+        set ISYM=-1, LWAVE=True and you have to check for the rest.
```
JaGeo, line 311:
```suggestion
        set ISYM=0, LWAVE=True and you have to check for the rest.
```

pymatgen/io/lobster/inputs.py
QuantumChemist: Yeah, that's what I realized too when I was going through it again, see above (could have phrased that more clearly)

pymatgen/io/lobster/inputs.py
JaGeo: (sounds harsher than intended 🙈. I just hoped I could have a quick look 😅. In any case, thank you for the woek)

pymatgen/io/lobster/inputs.py
DanielYang59: The original implementation (access directly through the internal dict `self.data` would bypass the MRO), I have changed it to access through the `__getitem__` magic method to respect this.

pymatgen/io/lobster/inputs.py
DanielYang59: I think it should, because we now store all keys as lower-case internally, so when we try to get a key from a `Lobsterin` instance, we don't need to cast it to lower case again anymore.

Yes I would have a look at the unit test and add one if there isn't any test on this.

pymatgen/io/lobster/inputs.py
DanielYang59: All good :), there isn't as much change as it seems to be to the code itself (sorry I'm making quite some changes to the comments which might have flooded the difference).

The core change is the case handling surround `UserDict/dict` functionalities. Now it would store the `Lobsterin` keywords(keys) internally as lowered as we don't need to lower every time we try to `get` a key or such.

pymatgen/io/lobster/inputs.py
DanielYang59: If we were to update the default (recommended) `ISYM` value from `-1` to `0`, should we consider updating the default argument to `0` as well for consistency (`isym: Literal[-1, 0] = -1,`)? And issue a warning for such changes in behavior?

pymatgen/io/lobster/inputs.py
DanielYang59: OK!

pymatgen/io/lobster/inputs.py
DanielYang59: Sure!

pymatgen/io/lobster/inputs.py
DanielYang59: Thanks for the helpful discussion!

pymatgen/io/lobster/inputs.py
DanielYang59: Thanks!

pymatgen/io/lobster/inputs.py
```diff
@@ -116,176 +117,145 @@ class Lobsterin(UserDict, MSONable):
         "kpointwiseSpilling",
         "LSODOS",
     )
-    # several of these keywords + ending can be used in a lobsterin file:
-    LISTKEYWORDS = ("basisfunctions", "cohpbetween", "createFatband")
-
-    # all keywords known to this class so far
-    AVAILABLE_KEYWORDS = FLOAT_KEYWORDS + STRING_KEYWORDS + BOOLEAN_KEYWORDS + LISTKEYWORDS
+    # These keywords + ending can be used in a lobsterin file
```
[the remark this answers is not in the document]
DanielYang59: Thanks, updated!

pymatgen/io/lobster/inputs.py
JaGeo: Yes, it would be good to do this.

DanielYang59: > Yes, it would be good to do this.

I made this change in b351443aaf521b3b1fe729f288c0a6f6e91259cc.

> The effect for the user should be minimal as LOBSTER and VASP run for both settings.

I think this is even more dangerous than a breaking change, because user might not even notice the default value has changed. So I set the default to `None` to detect if the default value is being used, instead of directly changing `-1` to `0`.

QuantumChemist: I would set the default value to 0. The only difference between 0 and -1 is how long the nonSCF-VASP and LOBSTER calculations take. A user really needing -1 will be aware of it.

pymatgen/io/lobster/inputs.py
DanielYang59: Added in 08d48164bf2e21588d41c20bec8a75a25b1d38e1 and a11423243f4369d108700f3d0a905a154460b819. Let me know if anything is missing or I'm adding to the wrong section :)

JaGeo:
@DanielYang59 I don't have time at the moment. I would really appreciate that we merge this very carefully to avoid any breakages - especially also of Lobster workflows in atomate2.

DanielYang59:
Sure no rush at all. Let's leave it until you have time to review :)

JaGeo:
Thanks @DanielYang59 for understanding. I really appreciate your work here but cannot keep up with the pace. Sorry!

DanielYang59:
All good, I'm not in a rush at all :) Just tagging it to make everything organized. Thanks!

QuantumChemist:
> Do we have further concerns? @JaGeo @QuantumChemist If not I might ping Janosh for reviewing?

I don't see any more things that require changes. I could check if it breaks LobsterPy's functionality over the weekend (this or the next ones). Sorry, I'm also too busy to test the changes in depth atm.

DanielYang59:
> I don't see any more things that require changes. I could check if it breaks LobsterPy's functionality over the weekend (this or the next ones). Sorry, I'm also too busy to test the changes in depth atm.

No worries, that would very helpful too :)

JaGeo:
- [ ] Run atomate2 and lobsterpy tests with Python 3.12
- [ ] Run atomtae2 and lobsterpy tests with Python 3.11

JaGeo:
It's fine. The Lobster tests in atomate2 and LobsterPy tests are all passing. From my side, it is ready to merge.

DanielYang59:
Thanks for the test :) I would ping @janosh for final review.

janosh (approved):
lgtm, thanks @DanielYang59! 👍

tests/io/lobster/test_inputs.py
```diff
@@ -1841,14 +1839,14 @@ def test_dict_functionality(self):
         assert len_after == len_before - 1
 
         # Test case sensitivity of |= operator
-        self.Lobsterin |= {"skipCOHP": True}  # Camel case
+        self.Lobsterin["skipCOHP"] = True  # Camel case
```
DanielYang59, line 1842: This was intended for testing the case insensitivity of `|=` operator, should not have been changed :)

janosh: oops, sorry about the extra work. can you revert in your next PR?

DanielYang59: Sure :) No problem. And thanks for reviewing and the `PymatgenTest` replacement

PR #3752, opened 2024-04-12
Fix `typing_extension` `ImportError` in downstream packages

janosh:
closes #3751

## Summary by CodeRabbit

- **Refactor**
	- Improved code readability by consolidating imports in core modules.
	- Enhanced type hinting and conditional type checking across various modules.

- **Tests**
	- Updated test functions with specific error messages for better clarity.

- **Chores**
	- Updated revision of pre-commit hooks to ensure code quality.

PR #3852, opened 2024-05-31
Run CI with two different `uv` resolution strategies: `highest` and `lowest-direct`

janosh:
this ensures pymatgen is compatible with both the oldest and newest dependency versions it allows in `setup.py`

also fix `typing_extensions` `ImportError` reported in https://github.com/materialsproject/pymatgen/pull/3752#issuecomment-2142954988 by @kavanase @Andrew-S-Rosen

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,19 +25,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # maximize CI coverage of different platforms and python versions while minimizing the
+        # total number of jobs. We run all pytest splits with the oldest supported python
+        # version (currently 3.9) on windows (seems most likely to surface errors) and with
+        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix). We
+        # ignore mac-os, which is assumed to be similar to ubuntu.
+        config:
+          - { os: windows-latest, python: "3.9", resolution: highest }
+          - { os: ubuntu-latest, python: "3.12", resolution: lowest-direct }
         # pytest-split automatically distributes work load so parallel jobs finish in similar time
-        os: [ubuntu-latest, windows-latest]
-        version:
-          - { python: "3.9", resolution: highest }
-          - { python: "3.12", resolution: lowest-direct }
         split: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
-        # include/exclude is meant to maximize CI coverage of different platforms and python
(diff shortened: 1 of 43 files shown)


pymatgen/core/structure.py
```diff
@@ -60,6 +59,7 @@
     from ase.optimize.optimize import Optimizer
     from matgl.ext.ase import TrajectoryObserver
     from numpy.typing import ArrayLike, NDArray
+    from typing_extensions import Self
```
ml-evs, line 62: So we can only type check on Python 3.10? Is there no ruff rule that can be turned on that does

```suggestion
    try:
        from typing_extensions import Self
    except ImportError:
        from typing import Self
```

or something similar?

DanielYang59:
@shyuep Sorry for the trouble caused surrounding import of `Self`. As it seems to be causing trouble in the v2024.5.31, can I ask for yanking this release and issue a new one? Apologize again for this oversight

.github/workflows/test.yml
```diff
@@ -25,22 +25,18 @@ jobs:
...
       fail-fast: false
       matrix:
+        # maximize CI coverage of different platforms and python versions while minimizing the
+        # total number of jobs. We run all pytest splits with the oldest supported python
+        # version (currently 3.9) on windows (seems most likely to surface errors) and with
+        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix). We
+        # ignore mac-os, which is assumed to be similar to ubuntu.
+        config:
+          - { os: windows-latest, python: "3.9", resolution: highest }
+          - { os: ubuntu-latest, python: "3.12", resolution: lowest-direct }
```
DanielYang59, line 35: Is this a typo? Why "Python 3.12" is the "lowest-direct" while "3.9" is the "highest"?

.github/workflows/test.yml
```diff
@@ -25,22 +25,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # maximize CI coverage of different platforms and python versions while minimizing the
+        # total number of jobs. We run all pytest splits with the oldest supported python
+        # version (currently 3.9) on windows (seems most likely to surface errors) and with
+        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix). We
+        # ignore mac-os, which is assumed to be similar to ubuntu.
```
DanielYang59, line 32: I think this should be removed now that it's testing `macos-latest` now?

janosh: good catch!

PR #3853, opened 2024-06-01
Fix filter condition for warn msg of unphysical site occupancy in `io.cif`

DanielYang59:
## Summary

- Fix a mistake in #3819, where the `filter` condition is incorrect, and also cast `filter` to `list`.
- Add unit test for this error message


DanielYang59:
@janosh I pushed this hot fix to address the error message issue, and feel free to revert to `monty` version change if you want to get rid of duplicate versions once and for all :)

janosh (approved):
thanks for the timely fix @DanielYang59! 👍

PR #3858, opened 2024-06-04
Add new `.pmgrc.yaml` setting `PMG_VASP_PSP_SUB_DIRS: dict[str, str]`

janosh:
the current `PotcarSingle.functional_dir` hard-codes the subfolder structure of `PMG_VASP_PSP_DIR`. e.g. the subfolder containing the `PBE_64` POTCARs _must_ be named `POT_PAW_PBE_64`. this causes problems e.g. in cases where the POTCARs are shared across an HPC cluster and users don't have permissions to rename shared files (just happened to me).

https://github.com/materialsproject/pymatgen/blob/232fa1f842d1078dd69d8d1631b4dd19a9b35c2f/pymatgen/io/vasp/inputs.py#L1757-L1779

the new setting is read from `.pmgrc.yaml` and overrides the `PotcarSingle.functional_dir` names (leaving unspecified ones as is). example:

```yaml
# .pmgrc.yaml
PMG_VASP_PSP_DIR: /path/to/pseudo-potentials
PMG_VASP_PSP_SUB_DIRS:
  PBE_64: MY_PBE_64_DIR
  LDA_64: MY_LDA_64_DIR
```

this PR implement the new setting, adds tests for it and restores (uncomments) + fixes an existing `PotcarSingle` test
also improves `FileNotFoundError` messages on missing POTCARs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,8 +28,7 @@ jobs:
         # maximize CI coverage of different platforms and python versions while minimizing the
         # total number of jobs. We run all pytest splits with the oldest supported python
         # version (currently 3.9) on windows (seems most likely to surface errors) and with
-        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix). We
-        # ignore mac-os, which is assumed to be similar to ubuntu.
+        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix).
         config:
           - os: windows-latest
             python: "3.9"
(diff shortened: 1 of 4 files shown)


janosh:
pinging @esoteric-ephemera since i can't request a review from you but would be great to get your thoughts if any

mkhorton:
Seems reasonable to me. I am not familiar with the `W_HASH` variants.

esoteric-ephemera:
This looks good to me! This can also benefit those who are newer to VASP and only care about more recent generations of pseudopotentials.

@mkhorton: afaik, these are identical to the non `_W_HASH` variants (`{PBE,LDA}_52`, `{PBE,LDA}_54`) but with a SHA and possibly COPYR field added.

PR #3859, opened 2024-06-04
SpaceGroup changes

### LINKED ISSUE #3845: Incorrect SpaceGroup symbol attribute for 16 space group types
### Python version

3.10

### Pymatgen version

2024.5.1 or master

### Operating system version

_No response_

### Current behavior

output of minimal example:
```
48 Pnnn1
50 Pban1
59 Pmmn1
68 Ccce1
70 Fddd1
85 P4/n1
125 P4/nbm1
126 P4/nnc1
129 P4/nmm1
130 P4/ncc1
201 Pn-31
203 Fd-31
222 Pn-3n1
224 Pn-3m1
227 Fd-3m1
228 Fd-3c1
16
```

### Expected Behavior

For the minimal example provided below I would expect `len(err)` to be zero, meaning that every SpaceGroup symbol attribute provides a valid string that can be used to instantiate again a SpaceGroup object.
However, this is not the case for 16 space group types (see above). They all differ from their correct Hermann Mauguin symbol by an additionally appended "1" (resulting even in an impossible fourth symmetry direction / blickrichtung for most of them).

The symbol attribute is set here
https://github.com/materialsproject/pymatgen/blob/74e692b0186ec77881acd3adbf1bbefeea8ca096/pymatgen/symmetry/groups.py#L226

 I am not sure what the "universal" Hermann Mauguin symbol (`"universal_h_m"` key) and the additional `:1` represents and why the key `"hermann_mauguin"` is not used instead, but the current code gives wrong space group symbols in 16 cases.

### Minimal example

```Python
from pymatgen.symmetry.groups import SpaceGroup

err = []
for sg_num in range(1, 231):
    sg = SpaceGroup.from_int_number(sg_num)
    sg_symbol = sg.symbol
    try:
        SpaceGroup(sg_symbol)
    except ValueError as e:
        print(sg_num, sg_symbol)
        err.append(sg_num)

print(len(err))
```

_No response_

kaueltzen:
## Summary
Fixes #3845 , #3861 and #3862 .

Includes the following changes in the symmetry module:

- The SpaceGroup symbol attribute now refers to its Hermann-Mauguin symbol with underscore notation of screw axes.
- The setting of the 7 rhombohedral space group types is now stored in an additional hexagonal bool attribute which can be set explicitly or by appending the setting to the `int_symbol` string parameter with the setting appended as before. The latter overrides the former for backwards compatibility.
- The `SpaceGroup point_group` attribute now always refers to the point group in Hermann Mauguin notation.
- A class method `from_space_group` was added to `PointGroup` which initializes the 32 crystal classes from a SpaceGroup string in Hermann-Mauguin notation.
- Changes of the space group databases `symm_data.json` and `symm_ops.json` were made which are documented in [dev_scripts/update_spacegroup_data.py](https://github.com/materialsproject/pymatgen/pull/3859/files#diff-09c03a9345c6a677bacba3c15d2506f6467be798b3eabc05f7286a1a6dea9278) .
- If the full symbol is not available (for non-standard settings), a warning is now issued.
- Tests for the abovementioned points were added
- Notes on how to convert the old to the new symbol were added to docs/compatibility.md


pymatgen/symmetry/groups.py
```diff
@@ -223,13 +236,13 @@ def __init__(self, int_symbol: str) -> None:
...
                 ops = [SymmOp.from_xyz_str(s) for s in spg["symops"]]
-                self.symbol = re.sub(r":", "", re.sub(r" ", "", spg["universal_h_m"]))
+                self.symbol = spg["hermann_mauguin"]
                 if int_symbol in SpaceGroup.sg_encoding:
                     self.full_symbol = SpaceGroup.sg_encoding[int_symbol]["full_symbol"]
                     self.point_group = SpaceGroup.sg_encoding[int_symbol]["point_group"]
                 else:
-                    self.full_symbol = re.sub(r" ", "", spg["universal_h_m"])
-                    self.point_group = spg["schoenflies"]
+                    self.full_symbol = spg["hermann_mauguin"]  # TODO full symbol not available from spg->maybe not add?
```
janosh, line 244: can you elaborate on this TODO? does this mean the current `full_symbol` is incorrect or incomplete?

pymatgen/symmetry/groups.py
```diff
@@ -223,13 +236,13 @@ def __init__(self, int_symbol: str) -> None:
...
+                    self.point_group = spg["schoenflies"].split("^")[0]  # TODO map onto Hermann Mauguin notation?
```
janosh, line 245: let's come to a decision here before merging this PR. what would be the benefits of mapping onto Hermann Mauguin notation?

tests/symmetry/test_groups.py
```diff
@@ -217,6 +219,11 @@ def test_repr(self):
             symbol = ORDERED_SYMBOLS[num - 1]
             assert repr(sg) in f"SpaceGroup({symbol=})"
 
+    def test_valid_symbol(self):  # Added after issue #3845
+        for num in range(1, 231):
+            symbol = SpaceGroup.from_int_number(num).symbol
+            assert SpaceGroup(symbol).int_number == num
```
janosh, line 225: this is definitely a good test to have and an argument in favor of the changes in this PR

tests/symmetry/test_groups.py
```diff
@@ -198,14 +205,9 @@ def test_subgroup_supergroup(self):
         assert SpaceGroup("Pma2").is_subgroup(SpaceGroup("Pccm"))
         assert not SpaceGroup.from_int_number(229).is_subgroup(SpaceGroup.from_int_number(230))
 
-    def test_hexagonal(self):
-        for num in (146, 148, 155, 160, 161, 166, 167):
-            sg = SpaceGroup.from_int_number(num, hexagonal=False)
-            assert not sg.symbol.endswith("H")
```
janosh, line 204: why remove this test?

janosh (changes requested):
thanks @kaueltzen for submitting this fix! 👍  i left 3 questions

pymatgen/symmetry/groups.py
kaueltzen: Yes - as far as I can see, the full symbol (e.g., F4/m-32/m instead of Fm-3m) is only contained in SpaceGroup.sg_encoding, but not in SpaceGroup.SYMM_OPS. Both spg keys "hermann_mauguin" and "universal_h_m" do not contain this information. We could still set this attribute but issue a warning (maybe that's better than not setting it in these cases?)

pymatgen/symmetry/groups.py
kaueltzen: The information is the same, but in these cases we'd have different notations for space group (Hermann Mauguin) and point group (Schoenflies) within one object. Also, it might cause issues if one wants to compare the point groups for a larger dataset (and the same point group could be there in different representations).

In order to map it, we would either need a database or we could, in principle, extract the point group in Hermann Mauguin notation from the space group symbol (by removing the centering and translation parts of symmetry elements) - this would require additional testing, though.

tests/symmetry/test_groups.py
kaueltzen: I removed it as the representation ("H" as last character) is not part of the symbol attribute in any case now.
We could maybe still keep this test and replace the assertion with a test for the hexagonal attribute `assert sg.hexagonal == False` to check that this is given correctly to the SpaceGroup init.

pymatgen/symmetry/groups.py
kaueltzen: Maybe there could be a PointGroup class method from_spacegroup() that does exactly this?

tests/symmetry/test_groups.py
janosh: replacing the assertion with a test for the hexagonal attribute sounds good

pymatgen/symmetry/groups.py
janosh: that sounds like a good idea! would you prefer to do that in a separate PR (or just open an issue for it)?

kaueltzen: Great, i'd prefer to just open an issue for it and add it in this PR - will do both tomorrow (:

mkhorton:
Many thanks for this @kaueltzen!

In case it is helpful in future, adding a note that the `maggroups` module in pymatgen uses a different dataset (from ISO-MAG, used with permission). It may or may not be helpful to cross-check information, since it's a superset of the crystallographic groups.

**Important note for this PR:** this technically counts as a breaking change. I can imagine cases whereby users are generating and storing space group strings in a database, for example, with the previous logic. Therefore, can you add a note to `docs/compatibility.md` noting the change, and how affected users can replace their previous `symbol` with the new `symbol`?

JaGeo:
> **Important note for this PR:** this technically counts as a breaking change. I can imagine cases whereby users are generating and storing space group strings in a database, for example, with the previous logic. Therefore, can you add a note to `docs/compatibility.md` noting the change, and how affected users can replace their previous `symbol` with the new `symbol`?

Thanks @mkhorton ! We know it is breaking. We had asked in the issue how to handle this. Thanks for pointing out how to do it in the best way! Very helpful.

mkhorton:
No problem, apologies I’d missed the issue comment.

kaueltzen:
> Many thanks for this @kaueltzen!
> 
> In case it is helpful in future, adding a note that the `maggroups` module in pymatgen uses a different dataset (from ISO-MAG, used with permission). It may or may not be helpful to cross-check information, since it's a superset of the crystallographic groups.
> 
> **Important note for this PR:** this technically counts as a breaking change. I can imagine cases whereby users are generating and storing space group strings in a database, for example, with the previous logic. Therefore, can you add a note to `docs/compatibility.md` noting the change, and how affected users can replace their previous `symbol` with the new `symbol`?

Thanks @mkhorton for pointing that out! Will add a note.

kaueltzen:
Hey, while working on the new `PointGroup` class method `from_space_group()`, I found another bug in `SpaceGroup` referenced here: #3862 which is also important for this PR.
What are your thoughts on that?

kaueltzen:
Another minor issue: the space group type I2_12_12_1 is mapped onto I2_12_121 in `SpaceGroup.full_sg_mapping` , the wrong key is also there in `SpaceGroup.sg_encoding` . Will fix that as well.

pymatgen/symmetry/groups.py
kaueltzen: Added the new class method and also added this info directly to the `symm_ops.json` database so it does not have to be called every time when initializing a `SpaceGroup` object. The changes to `symm_ops.json` and `symm_data.json` are documented in `dev_scripts.update_space_group_data.py` - was not sure if this is the correct place, though!

pymatgen/symmetry/groups.py
kaueltzen: Added a warning and this seems to be only invoked when initializing non-standard settings (as expected). E.g., `SpaceGroup("Pmc2_1")` does not invoke a warning, but `SpaceGroup("P2_1ma")` does.

However, I struggle to write a test for that:
```python
    def test_full_symbol_warning(self):
        with pytest.warns(Warning,
                          match="Full symbol not available, falling back to short Hermann Mauguin symbol P2_1ma instead.") as warns:
            _ = SpaceGroup("P2_1ma")
        assert len(warns) == 1
```
This fails with `Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted. Emitted warnings: [].` and I don't understand why. Any ideas?

janosh: are you sure you weren't testing against the PyPI release of `pymatgen`? your test passes for me. just pushed it to this PR. i expect it'll pass here as well

pymatgen/symmetry/groups.py
janosh: thank you! this is great. it's super valuable to leave such a thorough paper trial for people to understand what happened later

pymatgen/symmetry/groups.py
kaueltzen: Thanks for figuring out this issue! https://github.com/materialsproject/pymatgen/pull/3859/commits/a33cc3cf92e471865d132ca981d995bf87c8a8be

kaueltzen:
Hey @mkhorton @shyuep this is ready to be reviewed.

JaGeo:
As far as I see it, all review requests have been addressed by @kaueltzen . Would be great if this could be merged if the changes are fine.

JaGeo:
Oh, now there are some conflicts 🫠

shyuep:
It is just a switch to a src based layout. Once you move the files, the conflicts should be simple to resolve.

kaueltzen:
> It is just a switch to a src based layout. Once you move the files, the conflicts should be simple to resolve.

Hey @shyuep @mkhorton the merge conflicts have been resolved. Could you please review this?

JaGeo:
@shyuep is there anything that still needs to be done here? We would be very happy if this could be merged and fixed.

shyuep:
Merged. Thanks,

JaGeo:
@shyuep thank you!

DanielYang59:
Looks like this PR breaks [one of the unit tests in windows](https://github.com/materialsproject/pymatgen/actions/runs/9930219406/job/27428698747)?
```
 _______________ TestSpacegroupAnalyzer.test_get_crystal_system ________________

self = <symmetry.test_analyzer.TestSpacegroupAnalyzer testMethod=test_get_crystal_system>

    def test_get_crystal_system(self):
        crystal_system = self.sg.get_crystal_system()
        assert crystal_system == "orthorhombic"
        assert self.disordered_sg.get_crystal_system() == "tetragonal"
    
        orig_spg = self.sg._space_group_data["number"]
>       self.sg._space_group_data["number"] = 0
E       TypeError: 'SpglibDataset' object does not support item assignment
```

Feel free to ping me if there's anything I could help :)

JaGeo:
@kaueltzen could you maybe take a look?

kaueltzen:
Hi @DanielYang59 i cannot see where this might be due to the SpaceGroup changes. it seems like it is due to the latest spglib release? https://spglib.readthedocs.io/en/stable/releases.html#v2-5-0-9-jul-2024

The output data is not longer in dict format but in classes like SpglibDataset. The dict-like interface is planned to be removed, so there will be a lot of necessary changes in pymatgen. would changing the requirements of spglib to <2.5.0 be ok for you for now?

DanielYang59:
@kaueltzen Apologize. Sorry I didn't look very closely into the failure (just noticed unit tests begin to fail after this PR, and this is related to spacegroup). Sorry for misleading you.

I could confirm using `spglig` lower than 2.5.0 fixed this meanwhile :)

kaueltzen:
> I could confirm using `spglig` lower than 2.5.0 fixed this meanwhile :)

Great :+1: and thanks for looking into this!

PR #3863, opened 2024-06-05
Replace hard coded class names, add types for `io.vasp.sets`

DanielYang59:
## Summary

- Improve types for `io.vasp.sets` and comment tweaks
- Replace hard coded class names
- Use `deadline` in `deprecated` decorator for easier tracking of deprecated functions

### DIFF
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -317,7 +317,7 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
 
     def copy(self) -> Self:
         """A copy of the composition."""
-        return Composition(self, allow_negative=self.allow_negative)
+        return type(self)(self, allow_negative=self.allow_negative)
 
     @property
     def formula(self) -> str:
@@ -338,7 +338,7 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
 
     @property
     def iupac_formula(self) -> str:
-        """A formula string, with elements sorted by the iupac
+        """A formula string, with elements sorted by the IUPAC
         electronegativity ordering defined in Table VI of "Nomenclature of
         Inorganic Chemistry (IUPAC Recommendations 2005)". This ordering
         effectively follows the groups and rows of the periodic table, except
@@ -354,7 +354,7 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
     @property
     def element_composition(self) -> Self:
         """The composition replacing any species by the corresponding element."""
-        return Composition(self.get_el_amt_dict(), allow_negative=self.allow_negative)
+        return type(self)(self.get_el_amt_dict(), allow_negative=self.allow_negative)
 
     @property
     def fractional_composition(self) -> Self:
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -25,7 +25,7 @@ from pymatgen.util.string import Stringify, formula_double_format
 
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterator
-    from typing import ClassVar
+    from typing import Any, ClassVar
 
     from typing_extensions import Self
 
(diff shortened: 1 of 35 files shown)


pymatgen/io/vasp/sets.py
```diff
@@ -362,7 +365,7 @@ def write_input(
             files_to_transfer=self.files_to_transfer,
         )
 
-    def as_dict(self, verbosity=2):
+    def as_dict(self, verbosity: int = 2) -> dict:  # TODO: redefine verbosity?
```
DanielYang59, line 368: `verbosity` doesn't really reflect what is does, perhaps deprecate it and rename to something like `exclude_structure`?

pymatgen/io/vasp/sets.py
```diff
@@ -309,12 +306,18 @@ def __post_init__(self):
...
+        self.prev_outcar: Outcar | None = None
+        self._ispin: int | None = None
+
+    def __str__(self) -> str:
+        return type(self).__name__
+
+    def __repr__(self) -> str:
+        return type(self).__name__
 
     @deprecated(message="get_vasp_input will be removed in a future version of pymatgen. Use get_input_set instead.")
```
DanielYang59, line 319: Perhaps add a `deadline`?

pymatgen/io/vasp/sets.py
janosh: `verbosity`, while misleading in many cases, is the canonical control parameter name in `monty` and so changing it would likely encounter push back

pymatgen/io/vasp/sets.py
janosh: please do. maybe 2 years?

DanielYang59: ok!

pymatgen/io/cp2k/inputs.py
```diff
@@ -2089,7 +2089,7 @@ def __init__(self, npoints: int, kpoints: Iterable, units: str = "B_VECTOR") ->
         )
 
 
-@deprecated(KpointSet, "Kpoint_Set has been deprecated in favor of KpointSet on 2024-03-16")
+@deprecated(KpointSet, "Kpoint_Set has been deprecated in favor of KpointSet on 2024-03-16", deadline=(2025, 3, 16))
```
DanielYang59, line 2092: I added an 1-year deadline here (consistent with its neighbors), free feel to change this.

pymatgen/io/vasp/sets.py
DanielYang59: I just notice this when going through `core.sites`, looks like it's intended to tweak the verbosity of dict representations.

But the verbosity as an `int` still is a bit ambiguous to me...https://github.com/materialsproject/pymatgen/blob/5c7889c65289352e053b99bab8ec4d17686abfa6/pymatgen/core/sites.py#L579-L585

pymatgen/core/structure.py
```diff
@@ -284,7 +284,7 @@ def species_and_occu(self) -> list[Composition]:
         return [site.species for site in self]
 
     @property
-    @deprecated(message="Use n_type_sp instead")
+    @deprecated(message="Use n_type_sp instead")  # DEBUG: no such property
```
DanielYang59, line 287: `SiteCollection` doesn't seem to have a `n_type_sp` `property`?

pymatgen/io/cp2k/inputs.py
janosh: that works for me

pymatgen/core/structure.py
janosh: yeah, looks like the deprecation message was meant to say "Use `n_elems` instead". i think we can just use `replacement=n_elems`

pymatgen/io/cp2k/inputs.py
DanielYang59: Great!

pymatgen/core/structure.py
DanielYang59: Yes. However I don't quite know how to refer to a method/property of a class in such a decorator, `self.n_elems` nor `SiteCollection.n_elems` seem to work?

pymatgen/core/composition.py
```diff
@@ -317,7 +317,7 @@ def is_element(self) -> bool:
 
     def copy(self) -> Self:
         """A copy of the composition."""
-        return type(self)(self, allow_negative=self.allow_negative)
+        return Composition(self, allow_negative=self.allow_negative)
```
DanielYang59: Need some help on this, replacing this seems to break tests for `core.ion`, I assume there might be some issue with the inheritance of `Ion`?

janosh: i assume you mean this error?

```py
E       TypeError: Ion.__init__() got an unexpected keyword argument 'allow_negative'
```

should be fixed by just forwarding any additional keywords in the `Ion.__init__` to `Composition.__init__` (see 7639b91)

DanielYang59: Yes, and thanks for the fix!

pymatgen/core/structure.py
DanielYang59: Solved, swapping the order fixed this :)
```
    @property
    def n_elems(self) -> int:
        """Number of types of atoms."""
        return len(self.types_of_species)

    @property
    @deprecated(message="Use n_type_sp instead", replacement=n_elems)
    def ntypesp(self) -> int:
        """Number of types of atoms."""
        return len(self.types_of_species)
```

pymatgen/io/cp2k/inputs.py
```diff
@@ -2173,7 +2173,7 @@ def pairwise(iterable):
         return cls(kpoint_sets=kpoint_sets, filename="BAND.bs")
 
 
-@deprecated(BandStructure, "Deprecated on 2024-03-29.", deadline=(2025, 3, 29))
```
DanielYang59: I removed the date here as I don't think user would care about the date it was marked as deprecated, but the data it would be removed

DanielYang59:
@janosh Ready for review now, thanks for your time :)

pymatgen/io/cp2k/inputs.py
janosh: as a developer, i definitely care when it was deprecated. of course that information can be retrieved from `git` history but is less convenient. it could be a code comment rather than part of the message but i'd prefer to keep a record of how long a deprecation period was decided

DanielYang59: Okay, let revert it then :)

janosh (approved):
thanks @DanielYang59! nice work! 👍

DanielYang59:
Thanks for reviewing, and I really appreciate your merging this before splitting `vasp.sets` :)

PR #3865, opened 2024-06-05
Split VASP input sets into submodules

janosh:
follow up PR to #3835, specifically [this exchange](https://github.com/materialsproject/pymatgen/pull/3835#issuecomment-2118090800), that splits VASP input sets (currently all in a 3.3k line `sets.py` file) into submodules `base.py`, `lobster.py`, `matpes.py`, `mit.py`, `mp.py`, `mvl.py`.

no breaking changes expected as all input sets are re-exported from the new `pymatgen/io/vasp/sets/__init__.py` and so are importable the same way they were before. i did not re-export some private functions like `_get_recommended_lreal` in `base.py` since they were already marked private and no other files in pymatgen were importing them

pinging @shyuep, @mkhorton, @esoteric-ephemera, @utf, @JaGeo for review


DanielYang59:
I don't have any objection for this PR, but could we "possibly" split after #3863? As I think splitting would be easier (cut and paste) than reapplying all changes?

mkhorton:
In favor, thanks @janosh! The `sets.py` file was getting far too unwieldly.

shyuep:
Given that I expressed my objections to the split earlier, this PR should not have been merged without my review. Also, this PR clearly broke something. See #3988

shyuep:
Further, I have no idea how this PR was done, but the PR changes showed the creation of the sets folder with the submodules, but it did not show the deletion of the sets.py file. As a result, reversion of this PR caused all kinds of breakages.

Issue #3780, opened 2024-04-23
Incompatibility Issue with ALGO Parameter in Incar and Vasprun Classes


yanghan234:
### Python version

Python 3.9

### Pymatgen version

Version: 2024.4.13

### Operating system version

Ubuntu 20.04 LTS

### Current behavior

There appears to be an inconsistency in handling the `ALGO` parameter between the `Incar` and `Vasprun` classes. Specifically, when setting the `ALGO` parameter to "GW0" using the `Incar` class, the value is automatically reformatted to "Gw0" (with only the first letter capitalized). 
```
ALGO = Gw0 
...
```

However, the `Vasprun` class expects "GW0" with all letters in uppercase when checking the parameter value: https://github.com/materialsproject/pymatgen/blob/442d740eb7794d360cb35f187e271b3debb7f047/pymatgen/io/vasp/outputs.py#L1167

In fact, if I manually change "Gw0" in the vasprun.xml to "GW0", the problem is solved. So, it seems to me that we only need to modify the line to 
```python
if potcar and self.incar.get("ALGO", "").upper() not in ["GW0", "G0W0", "GW", "BSE"]: 
```

### Expected Behavior

The Vasprun should parse the vasprun.xml generated with `ALGO=Gw0` without any error.

### Minimal example

```Python
# Note this script runs longer than 20 minutes
from pymatgen.core import Structure
from pymatgen.io.vasp.sets import MVLGWSet
from pymatgen.io.vasp import Vasprun
import os

# construct an FCC silicon structure
si_structure = Structure(
    lattice=[[0, 2.73, 2.73], [2.73, 0, 2.73], [2.73, 2.73, 0]],
    species=["Si", "Si"],
    coords=[[0, 0, 0], [0.25, 0.25, 0.25]],
)

static_job = MVLGWSet(si_structure, mode="STATIC")
static_job.write_input("test-static")
os.system("export OMP_NUM_THREADS=1 && cd test-static && mpirun -np 8 vasp_std && cd ..")

diag_job = MVLGWSet.from_prev_calc("test-static", mode="DIAG")
diag_job.write_input("test-diag")
os.system("export OMP_NUM_THREADS=1 && cd test-diag && mpirun -np 8 vasp_std && cd ..")

gw_job = MVLGWSet.from_prev_calc("test-diag", mode="GW")
gw_job.write_input("test-gw")
[...]

PR #3866, opened 2024-06-06
Bug fix: fixed the leading capitalized ALGO in Vasprun

yanghan234:
## Summary

Major changes:

- This PR solved the issue in #3780 

`pymatgen` automatically converts the input `ALGO` to the leading capitalized format, e.g. `GW0` will be converted to `Gw0` and written to `vasprun.xml`. In such case, `Vasprun` would fail to parse the results. In this PR, I fixed this bug. 

Tagging @mkhorton for awareness.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -1219,7 +1219,7 @@ class Vasprun(MSONable):
         """
         potcar = self.get_potcars(path)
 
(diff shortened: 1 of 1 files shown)


janosh (changes requested):
thanks @yanghan234! can you add a test for this fix?

mkhorton:
Thanks @yanghan234, I will merge this.

PR #3868, opened 2024-06-07
Bug fix: handle non-integer oxidation states in `Species`

esoteric-ephemera:
Simple bug fix for when elements have non-integer oxidation states assigned to them. Pymatgen can assign non-integer charges to sites, but when trying to initialize `Species` with non-integer oxidation state:
```
from pymatgen.core import Species
Species("C0.25+")._oxi_state
```
this raises an `AttributeError`.

This PR fixes that bug and adds test cases to `tests/core/test_periodic_table.test_symbol_oxi_state_str`.


janosh (approved):
thanks @esoteric-ephemera! 👍

PR #3869, opened 2024-06-08
Slightly more robust `MSONAtoms` handling

Andrew-S-Rosen:
Broader context: This PR basically impacts nobody and is just a marginally safer way to index the `MSONAtoms` info key when parsing the dictionary representation.

Technical details: As noted by @samblau, if for some reason your `MSONAtoms.as_dict()` representation no longer has an `atoms_info` key, then doing `MSONAtoms.from_dict()` will crash. In virtually all cases, the user will have an `atoms_info` key, but it's pretty trivial to fix this by doing `.get(dct["atoms_info"], {})` instead of `dct["atoms_info"]`, so I've done that here.

### DIFF
--- a/pymatgen/io/ase.py
+++ b/pymatgen/io/ase.py
@@ -72,7 +72,7 @@ class MSONAtoms(Atoms, MSONable):
         # to be used in a round-trip fashion and does not work properly with constraints.
         # See ASE issue #1387.
         mson_atoms = cls(decode(dct["atoms_json"]))
-        atoms_info = MontyDecoder().process_decoded(dct["atoms_info"])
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
Will re-open if needed.

janosh:
how come it's no longer needed? seems like a sensible change

Andrew-S-Rosen:
Feel free to merge. I was indifferent.

janosh:
i guess i'm still missing something. if it avoids an error that occurred in practice for @samblau, it's a keeper, no?

samblau:
I only encountered this error because I was dealing with task docs from an older version of QuAcc... this shouldn't be a common scenario. But it still seems better to have it than not.

PR #3870, opened 2024-06-09
Fix bug in VASP parameter parsing

mkhorton:
VASP 6.4.3 introduced additional white space in the values of some parameters in the vasprun.xml, for example ` <i type="string" name="GGA    ">PE</I>`. This leads to incorrect determination of `run_type`.

### DIFF
--- a/pymatgen/io/vasp/outputs.py
+++ b/pymatgen/io/vasp/outputs.py
@@ -722,6 +722,10 @@ class Vasprun(MSONable):
 
         Hubbard U terms and vdW corrections are detected automatically as well.
         """
+
+        # Care should be taken: if a GGA tag is not specified, VASP will default
+        # to the functional specified by the POTCAR. It is not clear how
+        # VASP handles the "--" value.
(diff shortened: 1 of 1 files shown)


pymatgen/io/vasp/outputs.py
```diff
@@ -722,6 +722,10 @@ def run_type(self) -> str:
 
         Hubbard U terms and vdW corrections are detected automatically as well.
         """
+
+        # Care should be taken: if a GGA tag is not specified, VASP will default
+        # to the functional specified by the POTCAR. It is not clear how
+        # VASP handles the "--" value.
```
janosh, line 728: > It is not clear how VASP handles the "--" value.

i think @esoteric-ephemera figured out what happens for `--`

esoteric-ephemera: It [still runs](https://github.com/materialsproject/custodian/issues/286#issuecomment-1699798494) but it's not a great practice for clarity of results

PR #3871, opened 2024-06-09
Enable more ruff rules including `A`, `DTZ`, `PGH` and more

### LINKED ISSUE #3791: [Dev] `datetime.datetime.utcnow()` deprecated and replacement breaks `strptime`
## `datetime.datetime.utcnow()` deprecated 

See discussion in https://github.com/materialsproject/pymatgen/pull/3705#discussion_r1537225117, `datetime.datetime.utcnow()` is marked as deprecated and might be [removed after Python 3.12](https://discuss.python.org/t/deprecating-utcnow-and-utcfromtimestamp/26221/15).

The same issue has been reported to `monty` as well: https://github.com/materialsvirtuallab/monty/issues/275.

But the new replacement returns slightly different output (an additional timezone info `+00:00` at the end), which breaks `strptime`:
```python
import datetime

start_time_old = datetime.datetime.utcnow()  # 2024-03-25 08:46:23.748342 

start_time_new = datetime.datetime.now(datetime.UTC)  # 2024-03-25 08:46:23.748472+00:00

print(start_time_old, start_time_new)  


dt_old = datetime.datetime.strptime(
    str(start_time_old), "%Y-%m-%d %H:%M:%S.%f"
)
print(dt_old)


dt_new = datetime.datetime.strptime(
    str(start_time_new), "%Y-%m-%d %H:%M:%S.%f"
)
print(dt_new)
```

Raises:
```python
  File "/Users/yang/developer/test/test.py", line 16, in <module>
    dt = datetime.datetime.strptime(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.2_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/_strptime.py", line 554, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "/opt/homebrew/Cellar/python@3.12/3.12.2_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/_strptime.py", line 336, in _strptime
    raise ValueError("unconverted data remains: %s" %
ValueError: unconverted data remains: +00:00
```

DanielYang59:
## Summary

- Enable more `ruff` rules
- Revert an accidental change in tests for `io.lobster.inputs` in https://github.com/materialsproject/pymatgen/pull/3848#discussion_r1632144875
- Other format tweaks

### Enable more [`ruff` rules](https://docs.astral.sh/ruff/rules/)

The following requires code change:
- [rule family `PGH`](https://docs.astral.sh/ruff/rules/#pygrep-hooks-pgh) for stricter typing
- [rule family `A`](https://docs.astral.sh/ruff/rules/#flake8-builtins-a) to avoid shadowing of builtins

> [!CAUTION]
> b1aa5bbe8e and 50e9b7dde changed multiple `ValueError` to more semantic `TypeError` to fix [`TRY004`](https://docs.astral.sh/ruff/rules/type-check-without-type-error/). this is breaking. however, downstream repercussions are expected to be small.

pinging @mkhorton @shyuep for approval

#### Future PR
- 7fac0b466bbb18372334b87c990c46a1096b6415 need to fix #3791
- Enable rule family `S` and `NPY`

### Other format tweaks
- Replace `match.group(x)` with `match[x]`
- Replace `array[0:x]` slicing with `array[:x]`
- Replace already multi-lined `dict()` constructor with `{}` [for speed](https://switowski.com/blog/dict-function-vs-literal-syntax/) (almost 2x-4x speedup).


JaGeo:
@DanielYang59 could you trigger the atomate tests?

DanielYang59:
@JaGeo Thanks for commenting. Can you let me how to achieve that? It should only be possible when pushing to `master` I guess?https://github.com/materialsproject/pymatgen/blob/624e6eb5cfefbf5736fe31b8910dd19c1042e94e/.github/workflows/test.yml#L94

JaGeo:
@DanielYang59 ah okay. Then, we will see after merging! Thanks!

pyproject.toml
```diff
@@ -50,10 +59,13 @@ select = [
   "RET",     # flake8-return
   "RSE",     # flake8-raise
   "RUF",     # Ruff-specific rules
+  # "S",     # flake8-bandit
```
DanielYang59, line 62: This [rule family](https://docs.astral.sh/ruff/rules/#flake8-bandit-s) provides some security related checks, it might be beneficial to enable it. But too many errors related to it, better to separate to another PR.

pyproject.toml
```diff
@@ -25,20 +25,29 @@ line-length = 120
 
 [tool.ruff.lint]
 select = [
+  "A",       # flake8-builtins
+  # "ANN",   # flake8-annotations
```
DanielYang59, line 29: This one is good, but it requires all arguments to be type annotated. We're not ready yet.

tests/analysis/chemenv/coordination_environments/test_read_write.py
```diff
@@ -127,8 +127,6 @@ def test_structure_environments_neighbors_sets(self):
             str(nb_set) == "Neighbors Set for site #6 :\n - Coordination number : 4\n - Voronoi indices : 1, 4, 5, 6\n"
         )
 
-        assert nb_set == nb_set
```
DanielYang59, line 130: It's so weird, there are multiple comparison with itself in 0ea568e48b9d0dbff7a8c39c43c048db41898500, does this serve any purpose that I'm not aware of?

DanielYang59:
@janosh Can you please advice on whether we want to fix [`A002`](https://docs.astral.sh/ruff/rules/builtin-argument-shadowing/)? It requires quite a lot of breaking argument renames.

Other than that, b1aa5bbe8e55461fa2e59672d5762272218f681b might have broken some unit test, and I would fix them :)

janosh:
> @janosh Can you please advice on whether we want to fix [A002](https://docs.astral.sh/ruff/rules/builtin-argument-shadowing/)? It requires quite a lot of breaking argument renames.

let's enable it to avoid builtin shadowing in the future but add `noqa: A002` for all existing cases to avoid breaking changes

pymatgen/vis/structure_vtk.py
```diff
@@ -629,11 +629,11 @@ def add_faces(self, faces, color, opacity=0.35):
             else:
                 raise ValueError("Number of points for a face should be >= 3")
 
-    def add_edges(self, edges, type="line", linewidth=2, color=(0.0, 0.0, 0.0)):
+    def add_edges(self, edges, type="line", linewidth=2, color=(0.0, 0.0, 0.0)):  # noqa: A002
```
DanielYang59, line 632: Does this unused `type` arg exist for a reason? If not, removing it would get rid of this `A002` error

janosh: i'd guess it's there to conform with the signature of the parent class?

forgot to mention earlier, `ruff` [can auto-insert](https://docs.astral.sh/ruff/tutorial/#adding-rules) `noqa`s with the ``--add-noqa` flag

DanielYang59:
@janosh Ready for review! Thanks for your time as always :)

pymatgen/vis/structure_vtk.py
DanielYang59: > i'd guess it's there to conform with the signature of the parent class?

Not sure, this looks like the base class: https://github.com/materialsproject/pymatgen/blob/624e6eb5cfefbf5736fe31b8910dd19c1042e94e/pymatgen/vis/structure_vtk.py#L35

> forgot to mention earlier, `ruff` [can auto-insert](https://docs.astral.sh/ruff/tutorial/#adding-rules) `noqa`s with the `--add-noqa` flag

Thanks for the tip!

tests/core/test_xcfunc.py
```diff
@@ -14,7 +14,6 @@ def test_xcfunc_api(self):
         ixc_1 = XcFunc.from_abinit_ixc(1)
         assert ixc_1.type == "LDA"
         assert ixc_1.name == "LDA_XC_TETER93"
-        assert ixc_1 == ixc_1
         assert ixc_1 == "LDA_XC_TETER93"
```
janosh, line 17: i think it makes sense to keep this line but move it into a separate method `def test_eq(self): ...`

in general, the name `test_xcfunc_api` is pretty meaningless and should be split into more descriptive method names, like `test_class_methods`, `test_serialization`, or something like that

tests/core/test_structure.py
```diff
@@ -114,7 +114,6 @@ def test_as_dataframe(self):
 
     def test_equal(self):
         struct = self.struct
-        assert struct == struct
         assert struct == struct.copy()
```
janosh, line 117: these lines are important to keep in `__eq__` tests. we want to make sure the magic method returns `True` as expected

tests/analysis/chemenv/coordination_environments/test_structure_environments.py
```diff
@@ -151,8 +149,6 @@ def test_light_structure_environments(self):
         assert len(nb_set) == 4
         assert hash(nb_set) == 4
 
-        assert nb_set == nb_set
-
         assert (
```
janosh, line 152: better to keep these as well though they should better be in separate test methods `def test_eq`, `test_equal` (the name of `__eq__` test methods is another thing to standardize across the code base)

tests/analysis/chemenv/coordination_environments/test_read_write.py
janosh: yes, these are there to check correctness of custom `__eq__` methods

janosh: it might make sense to ignore `PLR0124` specifically in test files rather than ignore this error on a case by case basis

tests/core/test_structure.py
DanielYang59: Thanks, I didn't realize that!

tests/analysis/chemenv/coordination_environments/test_read_write.py
DanielYang59: Thanks, I was about to add it to `[tool.ruff.lint.per-file-ignores]` too :)

tests/symmetry/test_analyzer.py
```diff
@@ -263,76 +263,75 @@ def test_get_ir_reciprocal_mesh_map(self):
 
     def test_get_conventional_standard_structure(self):
         structure = Structure.from_file(f"{TEST_FILES_DIR}/cif/bcc_1927.cif")
-        assert structure == structure
```
DanielYang59, line 266: Ping myself, revert this deletion

DanielYang59:
@janosh I replaced some already multi-lined dictionary constructor `dict()` with `{}` for speed in 47d72f3c3c1738f7926ab30e30cceb6d4c8ffae4, I know you don't want to introduce extra line, so I just replaced those that wouldn't do so. 

From my side, the [speed up](https://switowski.com/blog/dict-function-vs-literal-syntax/) is something I really want to have (2x - 4x, though being in the nanosecond level) :)

tests/core/test_xcfunc.py
DanielYang59: Done in 1fea2f43f02f189b6820ab604954d975b817dfae

DanielYang59:
@shyuep Can I have your approval on the following potentially breaking change (change some `ValueError` to `TypeError` where the type doesn't match)? Thanks!

> [b1aa5bb](https://github.com/materialsproject/pymatgen/commit/b1aa5bbe8e55461fa2e59672d5762272218f681b) and [50e9b7d](https://github.com/materialsproject/pymatgen/commit/50e9b7ddee9764324795d84bff532ca6bc500ae1) changed multiple `ValueError` to more semantic `TypeError` to fix [`TRY004`](https://docs.astral.sh/ruff/rules/type-check-without-type-error/). this is breaking. however, downstream repercussions are expected to be small.

shyuep:
If the errors are not the type that are meant to be caught, it is fine to change the error type. What I mean is that if the error is something that happens because of a bad use of code. Rather than something like catching an expected exception case and then handling it in some way.

DanielYang59:
Thanks for your reply!

Yes I could confirm those are exceptions we `raise` instead of catch (`try...except...`), see b1aa5bbe8e55461fa2e59672d5762272218f681b and 50e9b7ddee9764324795d84bff532ca6bc500ae1.

Mostly in the following pattern:
```diff
if not isinstance(variable, expected_type):
-    raise ValueError("variable should be in {expected_type}")
+    raise TypeError("variable should be in {expected_type}")
```

Still it could be breaking if any downstream operation reply on catching a `ValueError`, but I would consider this a fix as the wrong type of exception was raised in the first place.

shyuep:
Thanks I have merged them,.

DanielYang59:
Thanks!

PR #3877, opened 2024-06-12
Accounting for FHI-aims species' defaults 

ansobolev:
## Summary

This PR adds new SpeciesDefaults object to account for files with the default definitions of species-related keywords (integration grids, basis sets...) in FHI-aims, named _species' defaults_. This is done to abstract away the work with species' defaults from the `AimsControlIn` object, to bring the API more in line with VASP, and to enable running heterogeneous calculations that need different precision levels for different atoms.  

Major changes:

- An `AimsSpeciesFile` class representing the defaults for one species is introduced
- A `SpeciesDefaults` class representing a list of species' defaults is introduced; the `AimsControlIn` class now uses `SpeciesDefault`
- An `AIMS_SPECIES_DIR` configuration variable is introduced which should point to the root of `FHIaims/species_defaults` directory
-  Breaking change: the possibility to use default species' defaults directory is removed. Every input sets need explicit definition of the species' defaults directory relative to `FHIaims/species_defaults` or `FHIaims/species_defaults/defaults_2020`. 

## Todos

None

## Checklist


tpurcell90:
Would it be possible to reintroduce the default `species_dir` behavior? For some people using only one set of defaults is relatively okay

tpurcell90:
otherwise this looks good from an fhi-aims perspective. This will break atomate2 tests but we can fix that once it is merged

ansobolev:
> Would it be possible to reintroduce the default `species_dir` behavior? For some people using only one set of defaults is relatively okay

Yeah, it could be possible (just remove one check and add `species_dir` to `paths_to_try`, I guess), but based on my discussions over here people also want to freely change the defaults, and they don't like writing out the full path every time.

tpurcell90:
Okay that makes sense

ansobolev:
Hi, is there anything that should be done here?

JaGeo:
@ansobolev Thanks for contributing in any case! I alao hope one of the maintainers will soon have some time! 😃

shyuep:
Thanks for the contribution. I have merged it!

PR #3886, opened 2024-06-18
Correcting lll_reduce execution and allow setting ftol for termination searching making heterogeneous interfaces

jinlhr542:
Why setting the two structures the same before finding the mappings between them?

### DIFF
--- a/src/pymatgen/core/periodic_table.json
+++ b/src/pymatgen/core/periodic_table.json
@@ -33,7 +33,10 @@
         "Critical temperature": "no data K",
         "Density of solid": "10070 kg m<sup>-3</sup>",
         "Electrical resistivity": "no data 10<sup>-8</sup> &Omega; m",
-        "Electronic structure": "[Rn].6d<sup>1</sup>.7s<sup>2</sup>",
+        "Electronic structure": {
+            "0": "[Rn].6d<sup>1</sup>.7s<sup>2</sup>",
+            "3": "[Xe].4f<sup>14</sup>.5d<sup>10</sup>.6s<sup>2</sup>.6p<sup>6</sup>"
+        },
         "Ionic radii": {
             "3": 1.26
         },
@@ -191,7 +194,10 @@
         "Critical temperature": "no data K",
         "Density of solid": "10490 kg m<sup>-3</sup>",
         "Electrical resistivity": "1.63 10<sup>-8</sup> &Omega; m",
-        "Electronic structure": "[Kr].4d<sup>10</sup>.5s<sup>1</sup>",
+        "Electronic structure": {
+            "0": "[Kr].4d<sup>10</sup>.5s<sup>1</sup>",
+            "1": "[Kr].4d<sup>10</sup>"
+        },
         "ICSD oxidation states": [
             1,
             2,
@@ -375,7 +381,10 @@
         "Critical temperature": "no data K",
         "Density of solid": "2700 kg m<sup>-3</sup>",
         "Electrical resistivity": "2.7 10<sup>-8</sup> &Omega; m",
-        "Electronic structure": "[Ne].3s<sup>2</sup>.3p<sup>1</sup>",
+        "Electronic structure": {
+            "0": "[Ne].3s<sup>2</sup>.3p<sup>1</sup>",
+            "3": "1s<sup>2</sup>.2s<sup>2</sup>.2p<sup>6</sup>"
+        },
         "ICSD oxidation states": [
             3
         ],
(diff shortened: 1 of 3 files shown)


jinlhr542:
1

DanielYang59:
@janosh Sorry this indeed looks like an accidental bug I introduced.

jinlhr542:
Many thanks for the reviewing. I have swapped the order to rewrite struct into lll_reduced one after doing mapping.

I also added another property for the CoherentInterfaceBuilder class to allow to set the ftol searching for different terminations. Sometimes a very small ftol is needed to shift the termination into different atomic planes as shown here.
image

jinlhr542:
Also added some features to allow 2D-identical terminations which might be 3D-non-identical to be saved as shown in the examples below:
image
image
image
image
As can be seen, the first and second terminations will be merged into one in the old version. I added an extra index before the label of the slab termination so that they can be distinguished to avoid merging.

pymatgen/core/interface.py
```diff
@@ -2876,7 +2876,7 @@ def label_termination(slab: Structure) -> str:
 
     sp_symbol = SpacegroupAnalyzer(top_plane, symprec=0.1).get_space_group_symbol()
     form = top_plane.reduced_formula
-    return f"{form}_{sp_symbol}_{len(top_plane)}"
+    return f"{t_index+1}_{form}_{sp_symbol}_{len(top_plane)}"
```
DanielYang59, line 2879: First I need to make clear that I'm not a maintainer (just happen to contribute to this module). Some preliminary comments on this change to `label_termination`:
- I would be on board for giving this flexibility to control the clustering tolerance here, instead of hard-coding the tolerance to be `0.25`.
-  Can we possibly update the docstring to include these two new arguments here?
- I'm sure others would strongly suggest against such breaking enhancement, especially when we could make it backwards compatible (make this prefix an empty string by default for example).

jinlhr542: Many thanks for the comment. I have revised the code to make backwards usage compatible by setting up a label_index bool variable. I also updated the docstring.

DanielYang59: No problem. Instead of adding two args for a single behavior, perhaps we could reused the single `t_index` arg, and set it to `None` by default. And add a prefix when it is not None.

jinlhr542: Done. `t_index + 1`was used because I prefer to label the termination to start from 1. Indeed it is not necessary.

DanielYang59: > Done. 

Beautiful! Thanks for the change. 

The the docstring of `label_termination` is still missing?

> `t_index + 1`was used because I prefer to label the termination to start from 1. Indeed it is not necessary.

Even you prefer the index to start from 1, adding an offset **inside** `label_termination` might still be unexpected (as you pass `index` in, and `index + 1` comes out). 

I'm not suggesting you do this here, but if you really want the index counter to start from a value other than zero, you could use the `start` arg of `enumerate(iterable, start=0)`, e.g. using `enumerate(iterable, start=1)` would start the counter from `1` instead of `0`.

Another tip being, you could [install `pre-commit` locally](https://pre-commit.com/) (`pre-commit install`) to run `pre-commit` hooks prior to each commit.

pymatgen/core/interface.py
```diff
@@ -2839,7 +2839,7 @@ def from_slabs(
         return iface
 
 
-def label_termination(slab: Structure) -> str:
+def label_termination(slab: Structure, ftol: float = 0.25, t_index=None) -> str:
```
DanielYang59, line 2842: I mean the docstring of this func is still missing.

Also can you please add type annotation for `t_index` as well? (`t_index: int | None = None`)

You could add an extra `,` after `t_index` to wrap this `def` into multiple lines now that it's getting very long:
```diff
- def label_termination(slab: Structure, ftol: float = 0.25, t_index=None) -> str:
+ def label_termination(slab: Structure, ftol: float = 0.25, t_index=None,) -> str:
```
And `ruff` would automatically wrap it.

jinlhr542:
Dear developers it seems there are some conflicts cannot be resolved automatically, could you please have a check?

DanielYang59:
Sorry I'm unable to resolve conflicts for you as I don't have write access.

You might want to look at [this tutorial](https://code.visualstudio.com/docs/sourcecontrol/overview#_merge-conflicts) to resolve merge conflicts with VSCode with a GUI, and [this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github) from the web. 

@janosh Can you please review this?

jinlhr542:
It seems that I do not have write access to resolve the conflicts on the web. I am using the github desktop and I have resolved it locally before commiting this version. I have no idea why the check still reported conflicts...

DanielYang59:
I [didn't see any merge conflicts from my side](https://github.com/DanielYang59/pymatgen/tree/help-fix-3886) for some reason, when you merge (and resolve conflicts during merge), you would also need to push the merge as a commit. I didn't see the merge being pushed, can you please confirm this?

Also your fork is not up to date with remote, perhaps you should [consider sync it](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)?

jinlhr542:
Actually I have solved the conflict from my side before push it. I clicked syncing just now but it also says conflict remaining

PR #3892, opened 2024-06-23
Enable Ruff rule family "N" and "S"

DanielYang59:
## Summary

Follow up #3871:
- Enable [Ruff rule family "N"](https://docs.astral.sh/ruff/rules/#pep8-naming-n), fix violation of [NPY002](https://docs.astral.sh/ruff/rules/numpy-legacy-random/) (the use of legacy `np.random`)
    > NumPy recommends using a dedicated Generator instance rather than the random variate generation methods exposed directly on the random module, as the new Generator is both faster and has better statistical properties.
- Enable [Ruff rule family "S"](https://docs.astral.sh/ruff/rules/#flake8-bandit-s), for security related checks.
- Replace standard lib `random` with `np.random`


pymatgen/io/vasp/inputs.py
```diff
@@ -672,9 +672,10 @@ def set_temperature(self, temperature: float) -> None:
             temperature (float): Temperature in Kelvin.
         """
         # mean 0 variance 1
```
DanielYang59, line 674: This comment seems incorrect, `np.random.randn` (and the new Generator approach) generate "random floats in the half-open interval [0.0, 1.0).", which is not standard normal distribution.

pymatgen/core/structure.py
```diff
@@ -3642,11 +3641,12 @@ def get_boxed_structure(
         ):
             box_center = [(i + 0.5) * a, (j + 0.5) * b, (k + 0.5) * c]
             if random_rotation:
+                rng = np.random.default_rng()
                 while True:
                     op = SymmOp.from_origin_axis_angle(
                         (0, 0, 0),
-                        axis=np.random.rand(3),
-                        angle=random.uniform(-180, 180),
```
DanielYang59, line 3649: Usage of `random` module is removed altogether, as there is no good reason to reply on two different random generators (`numpy` and `random`) at the same time.

tests/symmetry/test_analyzer.py
```diff
@@ -574,8 +575,8 @@ def test_dihedral(self):
         assert pg_analyzer.sch_symbol == "Ih"
 
     def test_symmetrize_molecule1(self):
-        np.random.seed(77)
-        distortion = np.random.randn(len(C2H4), 3) / 10
+        rng = np.random.default_rng(77)
+        distortion = rng.standard_normal((len(C2H4), 3)) / 10
```
DanielYang59, line 579: For some reason the random distortion from the new Generator breaks the unit test (change the denominator to `20` for smaller distortion fixes this though, but need to confirm whether the tolerance is intended)

tests/io/vasp/test_sets.py
```diff
@@ -107,21 +107,21 @@ def test_sets_changed(self):
             with open(input_set) as file:
                 text = file.read().encode("utf-8")
                 name = os.path.basename(input_set)
-                hashes[name] = hashlib.sha1(text).hexdigest()
+                hashes[name] = hashlib.sha256(text).hexdigest()
```
DanielYang59, line 110: Replaced weak `sha1` with `sha256` to fix [S324](https://docs.astral.sh/ruff/rules/hashlib-insecure-hash-function/), I don't think this would break anything as this hash is just an internal sanity check, but still pin this for confirmation.

pymatgen/cli/pmg_config.py
```diff
@@ -143,17 +143,17 @@ def setup_potcars(potcar_dirs: list[str]):
                     dest = os.path.join(base_dir, os.path.basename(fname))
                     shutil.copy(fname, dest)
                     ext = fname.split(".")[-1]
-                    if ext.upper() in ["Z", "GZ"]:
-                        with subprocess.Popen(["gunzip", dest]) as process:
+                    if ext.upper() in {"Z", "GZ"}:
+                        with subprocess.Popen(["/usr/bin/gunzip", dest]) as process:
```
DanielYang59, line 147: Commit 26ccb8591b3363783b15f224c6da09311f81d5cc need careful confirmation. 

[Ruff rule S607](https://docs.astral.sh/ruff/rules/start-process-with-partial-path/) prevents "partial executable path (relative path)" which might be hijacked by modifying the `PATH`.

However I'm concerned about the consistency of absolute paths for these binaries, as I'm not 100% sure they would reside in the same position across different distributions (not to mention different OS, like Ubuntu vs MacOS).

pymatgen/command_line/mcsqs_caller.py
```diff
@@ -178,7 +178,11 @@ def _parse_sqs_path(path) -> Sqs:
     detected_instances = len(list(path.glob("bestsqs*[0-9]*.out")))
 
     # Convert best SQS structure to CIF file and pymatgen Structure
-    with Popen("str2cif < bestsqs.out > bestsqs.cif", shell=True, cwd=path) as process:  # noqa: S607
+    with (
+        open(os.path.join(path, "bestsqs.out")) as input_file,
+        open(os.path.join(path, "bestsqs.cif"), "w") as output_file,
+    ):
+        process = Popen(["str2cif"], stdin=input_file, stdout=output_file, cwd=path)  # noqa: S607
```
DanielYang59, line 185: Need to really carefully check this as well, as `mcsqs_caller` is not covered by unit test.

pymatgen/io/vasp/outputs.py
```diff
@@ -19,6 +18,7 @@
 from typing import TYPE_CHECKING, cast
 
 import numpy as np
+from defusedxml import ElementTree as ET
```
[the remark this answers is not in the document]
shyuep: Unless it is an actual significant speed upgrade (at least 10x), please do not introduce new dependencies. The XML parsing is a small part of pymatgen and we expect that it might be even less significant with the new vasprun.h5 format.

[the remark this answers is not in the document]
DanielYang59: Thanks for the input :) The `xml` -> `defusedxml` replacement is a [security patch](https://docs.astral.sh/ruff/rules/suspicious-xml-element-tree-usage/):
> Many XML parsers are vulnerable to XML attacks (such as entity expansion), which cause excessive memory and CPU usage by exploiting recursion. An attacker could use such methods to access unauthorized resources.

> Consider using the defusedxml package when parsing untrusted XML data, to protect against XML attacks.


And yes I'm not sure about whether we want to worry about security issues for `pymatgen`, especially when potential side effect is unknown. I might ignore this S314 rule until we really want to patch this at some point.

pymatgen/util/provenance.py
```diff
@@ -256,7 +256,7 @@ def __init__(
         if not all(sys.getsizeof(h) < MAX_HNODE_SIZE for h in history):
             raise ValueError(f"One or more history nodes exceeds the maximum size limit of {MAX_HNODE_SIZE} bytes")
 
-        self.created_at = created_at or datetime.datetime.utcnow()
+        self.created_at = created_at or datetime.datetime.now(datetime.timezone.utc)
```
DanielYang59, line 259: Need to track https://github.com/materialsvirtuallab/monty/pull/688 until a new release is issued.

src/pymatgen/electronic_structure/dos.py
```diff
@@ -822,7 +822,7 @@ def get_band_filling(
 
         # Only consider up to Fermi level in numerator
         energies = dos.energies - dos.efermi
-        return np.trapz(dos_densities[energies < 0], x=energies[energies < 0]) / np.trapz(
+        return np.trapezoid(dos_densities[energies < 0], x=energies[energies < 0]) / np.trapezoid(
```
DanielYang59, line 825: pre-commit seems to update the `np.trapz` with new `np.trapezoid` in NumPy 2, and would break CI tests

janosh (approved):
thanks @DanielYang59, great stuff as always! 👍

DanielYang59:
Thanks for saying that and reviewing.

Can I get a quick comment on https://github.com/materialsproject/pymatgen/pull/3892#discussion_r1649879457 please? I don't expect a unit test to pass only on some certain random seed :)

tests/symmetry/test_analyzer.py
janosh: oh yeah, i was actually going to ask if you updated the tests to work with new random behavior moving from std lib `random` to `np.random`. i think it's fine for a test to depend on the right random seed. can you update the tests?

DanielYang59: > oh yeah, i was actually going to ask if you updated the tests to work with new random behavior moving from std lib `random` to `np.random`. 

Okay I would do that very soon :)

tests/io/vasp/test_sets.py
```diff
@@ -107,21 +107,21 @@ def test_sets_changed(self):
...
-            "MPHSERelaxSet.yaml": "0d0d96a620461071cfd416ec9d5d6a8d2dfd0855",
-            "MPRelaxSet.yaml": "f2949cdc5dc8cd0bee6d39a5df0d6a6b7c144821",
-            "MPSCANRelaxSet.yaml": "167668225129002b49dc3550c04659869b9b9e47",
-            "MVLGWSet.yaml": "104ae93c3b3be19a13b0ee46ebdd0f40ceb96597",
-            "MVLRelax52Set.yaml": "4cfc6b1bd0548e45da3bde4a9c65b3249da13ecd",
-            "PBE54Base.yaml": "ec317781a7f344beb54c17a228db790c0eb49282",
-            "PBE64Base.yaml": "480c41c2448cb25706181de268090618e282b264",
-            "VASPIncarBase.yaml": "19762515f8deefb970f2968fca48a0d67f7964d4",
-            "vdW_parameters.yaml": "04bb09bb563d159565bcceac6a11e8bdf0152b79",
+            "MVLGWSet.yaml": "eba4564a18b99494a08ab6fdbe5364e7212b5992c7a9ef109001ce314a5b33db",
```
janosh, line 113: pinging @mkhorton to let you know @DanielYang59 switched to a new hash algo for the VASP input sets (presumably recommended by some `ruff` rule?) and updated the `known_hashes` accordingly

tests/test_cli.py
```diff
@@ -20,28 +20,28 @@ def cd_tmp_path(tmp_path: Path, monkeypatch: MonkeyPatch):
 
 
 def test_pmg_analyze(cd_tmp_path: Path):
-    exit_status = os.system(f"pmg analyze {TEST_FILES_DIR}/io/vasp/fixtures/scan_relaxation")
+    exit_status = os.system(f"pmg analyze {TEST_FILES_DIR}/io/vasp/fixtures/scan_relaxation")  # noqa:S605
```
janosh, line 23: maybe the right time to switch to `subprocess.run` here over `os.system`? don't have a strong opinion either way

src/pymatgen/io/vasp/inputs.py
```diff
@@ -659,7 +659,7 @@ def from_dict(cls, dct: dict) -> Self:
     def set_temperature(self, temperature: float) -> None:
         """
         Initialize the velocities based on Maxwell-Boltzmann distribution.
-        Removes linear, but not angular drift (same as VASP).
+        Remove linear, but not angular drift (same as VASP).
```
janosh, line 662: i would keep non-imperative "Removes"

src/pymatgen/cli/pmg_config.py
```diff
@@ -202,8 +202,8 @@ def build_bader(fortran_command="gfortran"):
     cwd = os.getcwd()
     state = True
     try:
-        urlretrieve(bader_url, "bader.tar.gz")
-        subprocess.call(["tar", "-zxf", "bader.tar.gz"])
+        urlretrieve(bader_url, "bader.tar.gz")  # noqa: S310
+        subprocess.call(["/usr/bin/tar", "-zxf", "bader.tar.gz"])
```
janosh, line 206: not sure we can hard-code `/usr/bin/tar` here. leaving it at `tar` will search everything on the user's path, not just 1 directory

janosh (changes requested):

tests/io/vasp/test_sets.py
DanielYang59: Yes https://github.com/materialsproject/pymatgen/pull/3892#discussion_r1650885177

src/pymatgen/cli/pmg_config.py
DanielYang59: I was quite concern about these changes as well, https://github.com/materialsproject/pymatgen/pull/3892#discussion_r1650911892, as security is not a core concern for `pymatgen`. Explicitly use the absolute path could avoid PATH hijacking.

src/pymatgen/io/vasp/inputs.py
DanielYang59: I think I changed this to keep consistency with the former sentence:
> Initialize the velocities based on Maxwell-Boltzmann distribution.

DanielYang59: Okay I misunderstand this here. Would revert this

tests/test_cli.py
DanielYang59: Okay, I would prefer the flexibility from `subprocess` too. Also noticed a violation of [S602](https://docs.astral.sh/ruff/rules/subprocess-popen-with-shell-equals-true/) in `io.packmol` too, I would fix that one as it looks quite dangerous.

src/pymatgen/io/packmol.py
```diff
@@ -1,20 +1,17 @@
...
+- PackmolSet provides a "run" method to run PACKMOL locally.
 
-PackmolBoxGen(InputGenerator) provides a recipe for packing molecules into a
-box, which returns a PackmolSet object.
+- PackmolBoxGen provides "get_input_set" for packing molecules into a box,
+which returns a PackmolSet object.
 
-For the run() method to work, you need to install the packmol package
-See http://m3g.iqm.unicamp.br/packmol or
-http://leandro.iqm.unicamp.br/m3g/packmol/home.shtml
```
DanielYang59, line 13: The second URL doesn't seem to work (at least on my side). 

Also removed some implementation details from the docstring (from explain the parent class `PackmolBoxGen(InputGenerator)` and "This adopts the minimal core I/O interface (see pymatgen/io/core).").

src/pymatgen/io/packmol.py
```diff
@@ -199,24 +196,17 @@ def get_input_set(
...
-
-            if mol is None:
-                raise ValueError("Molecule cannot be None.")
+            else:
+                raise ValueError("Molecule is not provided.")  # noqa: TRY004
 
             fname = f"packmol_{dct['name']}.xyz"
             mapping[fname] = mol.to(fmt="xyz")
             if " " in str(fname):
-                # NOTE - double quotes are deliberately used inside the f-string here, do not change
```
DanielYang59, line 217: @rkingsbury Can I get some confirmation here, thanks:)
1. Didn't see double quotes being used here, so I assume it's safe to remove this note. (The other note at line 172 is valid)
2. I believe it's safe to remove the `black` tag `fmt: on/off` here? As I'm not seeing `pre-commit` formatting this file here (running `black` directly would indeed wrap those lines).

tests/test_cli.py
DanielYang59: Done in fdf52f16f707fbc0ec782b1af9b8c57d927188f0, just did a global search and these look like the only usage of `os.system` across the code base.

src/pymatgen/io/packmol.py
janosh: this was probably carelessly refactored by me at some point as `!r` results in single quotes. thanks for pointing this out. let's go back to what i think was the original state

```py
f'structure "{fname}"\n'
```

janosh (approved):
thanks @DanielYang59! 👍

src/pymatgen/io/packmol.py
rkingsbury: At the time I added this, `packmol` was VERY pecuiliar about things like single vs. double quotes in the input file. I believe I opened an issue about it at the time, and they fixed the behavior to be a little less finicky. In principle as long as the packmol unit tests pass I think we should be OK, but it might be wise to see if someone can do some additional manual testing as well.

DanielYang59: Thanks a lot for the additional input and confirmation. And ping me at anytime if anything doesn't look right, thanks!

src/pymatgen/cli/pmg_config.py
DanielYang59: @janosh I'm still pretty concerned about commit 26ccb8591b3363783b15f224c6da09311f81d5cc which includes using the absolute path for the following:
- /usr/bin/gunzip
- /usr/bin/gzip
- /usr/bin/make
- /usr/bin/tar
- /bin/cp

There are pretty basic packages but owing to the huge diversities of Linux distro, I'm unsure if there would be any exception. According to ChatGPT 4o:

> the Filesystem Hierarchy Standard (FHS) provides guidelines and standards for directory structures and file placements in Unix-like operating systems, including Linux. The FHS is maintained by the Linux Foundation.

>Here are some relevant points from the FHS regarding the locations of these binaries:

>    /bin: This directory contains essential command binaries that are required for single-user mode and for all users (e.g., cp, ls, cat).

>    /usr/bin: This directory contains the majority of user commands. Binaries that are not essential for system boot or repair but are intended for use by all users are placed here. This includes binaries like gzip, gunzip, make, and tar.

It should be quite safe IMO.

PR #3894, opened 2024-06-25
build against NPY2

njzjz:
## Summary

Major changes:

- fix 1: Build against NPY2 instead of NPY1. This will support both NPY1 and NPY2. See https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice.
- fix 2: Replace `np.int_t` (missing in NPY2) with `np.int64_t` and `np.float_` with `np.float64`.
- fix 3: Replace `long` with `np.int64_t`, as "The default integer type on Windows is now int64 rather than
int32, matching the behavior on other platforms" per the change log of NPY2.

## Todos

- feature 2: ...
- fix 2:

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


njzjz:
Unluckily, there are still several errors. Some are related to the upstream chgnet and scipy and cannot be fixed here.

janosh:
@njzjz sorry for the lack of reply here! looks like this is almost finished. thanks for the effort! feel free to skip the failing chgnet tests for now. re `scipy`, we'll have to pin to `<1.14` as the [newest release dropped Python 3.9 support](https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html). but 1.13.0 already supports `numpy` v2

DanielYang59:
I really appreciate the pioneer work @njzjz! Meanwhile in case anyone missed this, [`ruff` now has a dedicated rule `NPY201`](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html) to handle API migration automatically :)
> `ruff check path/to/code/ --select NPY201`

njzjz:
For cross reference, I've submitted the SciPy issue to https://github.com/scipy/scipy/issues/21052

njzjz:
I do not know how to fix the rest of the errors and need help.

janosh:
@njzjz feel free to skip the two failing `chgnet` tests. we'll migrate it to `numpy` v2 once `pymatgen` has. for `test_delta_func`, you can conditionally skip it like this:

```py
@pytest.mark.skipif(sys.platform == "win32" and np.__version__ > "2.0.0", reason="Fails on Windows with numpy > 2.0.0, awaiting https://github.com/scipy/scipy/issues/21052 resolution")
def test_delta_func():
    x = np.array([0, 1, 2, 3, 4, 5])
```

not sure about `TestWavecar.test_get_parchg`. that might need a closer look. maybe @esoteric-ephemera has some ideas on what's going wrong?

janosh:
@njzjz this line is failing because of hundreds of new numpy warnings. we can make it more robust:

https://github.com/materialsproject/pymatgen/blob/4ffec03c93a513a58a0e48814d1f1eef7ce1e59b/tests/analysis/elasticity/test_elastic.py#L223

```py
assert sum("No eq state found" in str(warn) for warn in warns) == 1
assert sum("Extra strain states in strain-" in str(warn) for warn in warns) == 1
```

pyproject.toml
```diff
@@ -223,6 +225,7 @@ ignore = [
     "TRY003",  # Avoid specifying long messages outside the exception class
     "TRY300",  # Checks for return statements in try blocks
     "TRY301",  # Checks for raise statements within try blocks
+    "NPY201",  # Numpy 2.0 deprecation
```
DanielYang59, line 228: I think we don't need to include `NPY201` here, `pymatgen` now enables all ruff rules by default, and this is the "ignore" section, not the "select" section (where rules are enabled).

njzjz: reverted

DanielYang59: Thanks! I would probably be able to help looking at the remaining failing tests later once I have access to the Windows PC at home :)

DanielYang59:
Just did some investigation on the failure of `TestWavecar.test_get_parchg`, unfortunately I wasn't able to come to a definite conclusion, but still share my findings in case it's helpful.

I believe this failure has something to do with the float point precision change after NumPy 2.0 (though I don't know exactly how), with the following script on Windows Python 3.12:
```python
import numpy as np

from pymatgen.io.vasp import Poscar, Wavecar


TEST_DIR = "tests/files/io/vasp"
poscar = Poscar.from_file(f"{TEST_DIR}/inputs/POSCAR")
w = Wavecar(f"{TEST_DIR}/outputs/WAVECAR.H2.ncl")

w.coeffs.append([np.ones((2, 100))])
c = w.get_parchg(poscar, -1, 0, phase=False, spinor=None)

print(np.min(c.data["total"]))  # 0.0 in NumPy 1.x, 8.738907122657083e-33 in NumPy 2.x
assert not np.all(c.data["total"] > 0.0)  # AssertionError
```

Note the min value in `c.data["total"]` is very close to zero ([the min positive value with `float32` is around 1e-38](https://en.wikipedia.org/wiki/Single-precision_floating-point_format)).

From the ["NumPy 2.0 migration guide"](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html), there seems to be two changes related to this:
1. The default Windows integer was `int32` for NumPy 1.x, and is now `int64` (confirmed on my Windows PC), for Linux/Mac it's always `int64`
2. [The change in the scalar promotion rules](https://numpy.org/neps/nep-0050-scalar-promotion.html#nep50)

However, with these changes, I would expect the precision on Windows to be consistent with Linux on NumPy 2.x. Perhaps someone more familiar with the internal workings of `Wavecar` would know the reason :)

janosh:
re `test_get_parchg`, i suggest we simply invert the condition for now with a comment explaining why. sth like

```py
# this condition was inverted from `assert not np.all(...)` as it started failing during the
# numpy v2 migration on 2024-08-06. unclear what it was testing in the first place.
assert np.all(chgcar.data["total"] > 0.0)
```

janosh (approved):
thanks a lot @njzjz for this work and esp. for reporting multiple upstream issues! 🥇 and thanks @DanielYang59 for supporting! good team effort to get the many failing tests here sorted out.

DanielYang59:
The failing tests for `tests/io/abinit/test_abiobjects.py` looks like another randomly failing test. I cannot recreate it on my Windows PC.

src/pymatgen/optimization/linear_assignment.pyx
```diff
@@ -50,7 +50,8 @@ class LinearAssignment:
     """
 
     def __init__(self, costs, epsilon=1e-13):
-        self.orig_c = np.array(costs, dtype=np.float_, copy=False, order="C")
+        # https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
```
DanielYang59, line 53: @janosh This doesn't look like something we should keep inside the code (it's pretty easy to find)? What do you think?

janosh: i'm generally in favor of linking relevant docs. even if easy to find, people might not think to look for them. but no strong opinion in this case

janosh:
> The failing tests for tests/io/abinit/test_abiobjects.py looks like another randomly failing test. I cannot recreate it on my Windows PC.

probably something to report to https://github.com/Unidata/netcdf4-python, even if just to say the error message is self-contradictory...

> E   OSError: Error loading charset-233f44715e5aacaf3a688c2faff5ddf7.dll; The operation completed successfully.

src/pymatgen/optimization/linear_assignment.pyx
DanielYang59: This one looks more like a "disposable" migration guide which we would only do once :)

DanielYang59:
> > The failing tests for tests/io/abinit/test_abiobjects.py looks like another randomly failing test. I cannot recreate it on my Windows PC.
> 
> probably something to report to https://github.com/Unidata/netcdf4-python, even if just to say the error message is self-contradictory...

Agreed, just for my own record, it failed in another PR randomly https://github.com/materialsproject/pymatgen/actions/runs/10268736299/job/28412461926?pr=3892

njzjz:
> E OSError: Error loading charset-233f44715e5aacaf3a688c2faff5ddf7.dll; The operation completed successfully.

This part of the code is generated by [delvewheel](https://github.com/adang1345/delvewheel) [here](https://github.com/adang1345/delvewheel/blob/e74686253154416dcba7c0cfd644f3b847bf267c/delvewheel/_wheel_repair.py#L74). I think I can report there.

According to the [Windows documentation], this error message means no error. So it may be incorrect to throw an error.

> ERROR_SUCCESS
> 0 (0x0)
> The operation completed successfully.

See also https://stackoverflow.com/questions/7524142/what-does-windows-error-0-error-success-mean

njzjz:
> E OSError: Error loading charset-233f44715e5aacaf3a688c2faff5ddf7.dll; The operation completed successfully.

Submitted to https://github.com/adang1345/delvewheel/issues/51

janosh:
@njzjz can you pin pyproject to the fixed [`delvewheel>=1.7.4`](https://pypi.org/project/delvewheel/1.7.4/)

njzjz:
> @njzjz can you pin pyproject to the fixed [`delvewheel>=1.7.4`](https://pypi.org/project/delvewheel/1.7.4/)

This may not work, as delvewheel is a build-time dependency, not a runtime dependency.

janosh:
> This may not work, as delvewheel is a build-time dependency, not a runtime dependency.

ah, didn't realize that. so the place to pin that would be here

https://github.com/materialsproject/pymatgen/blob/1dfc9e4f7bbddaf5d93fe9b09b425ca1be6a58f4/pyproject.toml#L1-L4

but looks like we're in the green anyway so would be good to remove the temp install command in `test.yml`

janosh:
thanks again @njzjz! 👍

DanielYang59:
Just recorded another failure in https://github.com/materialsproject/pymatgen/actions/runs/10319015504/job/28566565139:
> OSError: Error loading libxml2-14375db763c77d38e40bbe5c028add84.dll; The operation completed successfully.

pyproject.toml
```diff
@@ -88,7 +90,7 @@ ase = ["ase>=3.23.0"]
 # don't depend on tblite above 3.11 since unsupported https://github.com/tblite/tblite/issues/175
 tblite = ["tblite[ase]>=0.3.0; python_version<'3.12'"]
 vis = ["vtk>=6.0.0"]
-abinit = ["netcdf4>=1.6.5"]
+abinit = ["netcdf4>=1.7.1"]
```
DanielYang59, line 93: @janosh Janosh can I have a quick comment on this bump of `netcdf4` version here from commit 1dfc9e4f7bbddaf5d93fe9b09b425ca1be6a58f4?

It seems to cause CI failure (more details on https://github.com/materialsproject/pymatgen/pull/4128#discussion_r1810203743) in Ubuntu CI runner for `netcdf4>=1.7.1.post1` (1.7.1 is yanked), `1.6.5` seems to work though

I also tried to [install `delvewheel` but doesn't seem to help](https://github.com/materialsproject/pymatgen/pull/4128/commits/75e23b13b113f06e28bc6513a14485b0e0a0bb28)

There seem to be [a lot of discussion around this error](https://github.com/Unidata/netcdf4-python/issues?q=is%3Aissue%20state%3Aopen%20NetCDF%3A%20HDF%20error) but didn't see a solid solution yet. And the issue for me is that I cannot recreate this with my local Ubuntu machine (with exactly the same dependency versions).

janosh: `netcdf` needed to be bumped in #3894 to fix the following CI errors:

```py
_____________ ERROR collecting tests/io/abinit/test_abiobjects.py _____________
tests\io\abinit\test_abiobjects.py:10: in <module>
    from pymatgen.io.abinit.abiobjects import (
src\pymatgen\io\abinit\__init__.py:5: in <module>
    from .netcdf import (
src\pymatgen\io\abinit\netcdf.py:25: in <module>
    import netCDF4
C:\Users\runneradmin\micromamba\envs\pmg\lib\site-packages\netCDF4\__init__.py:28: in <module>
    _delvewheel_patch_1_7_0()
C:\Users\runneradmin\micromamba\envs\pmg\lib\site-packages\netCDF4\__init__.py:25: in _delvewheel_patch_1_7_0
    raise OSError('Error loading {}; {}'.format(lib, ctypes.FormatError(ctypes.get_last_error())))
E   OSError: Error loading charset-[23](https://github.com/materialsproject/pymatgen/actions/runs/10265440264/job/28401647005#step:8:24)3f44715e5aacaf3a688c2faff5ddf7.dll; The operation completed successfully.
```

if it's causing problems again and this time downgrading helps, by all means let's do it :)

DanielYang59: Current `pymatgen` dependency declare has duplicate (would be fixed #4128), you must have been mislead too :)

The `netcdf4` bumped is not really the one installed in CI (the one in `optional` section is the one actually installed in CI, surely you know):
https://github.com/materialsproject/pymatgen/blob/3ee17e2c346dea6e9248eb04c989e58880914650/pyproject.toml#L95

https://github.com/materialsproject/pymatgen/blob/3ee17e2c346dea6e9248eb04c989e58880914650/pyproject.toml#L100-L113

So [it still installed](https://github.com/materialsproject/pymatgen/actions/runs/10307213885/job/28532023809):  `+ netcdf4==1.6.5`


I [looked into the error log again](https://github.com/materialsproject/pymatgen/actions/runs/10265440264/job/28401647005?pr=3894), and turns out it's `+ netcdf4==1.7.1.post1` causing the following error:
```
 ______________ ERROR collecting tests/io/abinit/test_pseudos.py _______________
tests\io\abinit\test_pseudos.py:11: in <module>
    from pymatgen.io.abinit.pseudos import Pseudo, PseudoTable
src\pymatgen\io\abinit\__init__.py:5: in <module>
    from .netcdf import (
src\pymatgen\io\abinit\netcdf.py:25: in <module>
    import netCDF4
C:\Users\runneradmin\micromamba\envs\pmg\lib\site-packages\netCDF4\__init__.py:28: in <module>
    _delvewheel_patch_1_7_0()
C:\Users\runneradmin\micromamba\envs\pmg\lib\site-packages\netCDF4\__init__.py:25: in _delvewheel_patch_1_7_0
    raise OSError('Error loading {}; {}'.format(lib, ctypes.FormatError(ctypes.get_last_error())))
E   OSError: Error loading charset-233f44715e5aacaf3a688c2faff5ddf7.dll; The operation completed successfully.
```

So I might revert that pin to `<1.7.1.post1` and ([1.7.1.post2 also triggers error](https://github.com/materialsproject/pymatgen/actions/runs/11471574769/job/31922707829?pr=4128)), [same for latest 1.7.2](https://github.com/materialsproject/pymatgen/actions/runs/11471711256/job/31923083563?pr=4128)

PR #3895, opened 2024-06-26
[Breaking] Fix comparision for `core.Composition`

### LINKED ISSUE #3815: Bug in `core.composition` comparison
## Bug in `core.composition` comparison

Open this in case it got forgotten. 

As found out by @janosh in https://github.com/materialsproject/pymatgen/pull/3792#discussion_r1584003750, there is an obvious bug with `Composition` comparison where `True` is returned too early:
https://github.com/materialsproject/pymatgen/blob/6c696cdb327621d5d16c675478f4ed0e024feea1/pymatgen/core/composition.py#L195-L212

On top of this obvious bug, there are a few questions:
- The docstring adds to ambiguity: `Should ONLY be used for defining a sort order (the behavior is probably not what you'd expect).`.
- It appears that we don't need to sort this union set?

### How to define the comparison behavior
How should we define the behavior when an `Element` is only in one of the `Composition` and this `Composition` is not a superset of another? For example:
1. `Composition({"O": 1, "N": 1})` > `Composition({"O": 1})`  # This makes sense
2. `Composition({"N": 1})` > `Composition({"O": 1})`  # This feels a bit weird though the atomic number is indeed greater
3. `Composition({"O": 2, "N": 1})` > `Composition({"O": 2, "C": 1})`  # This doesn't seem to make sense? But this could be inferred from (2) 

So should we prevent `Composition` with different `Element` from being compared altogether?
  
https://github.com/materialsproject/pymatgen/blob/6c696cdb327621d5d16c675478f4ed0e024feea1/tests/core/test_composition.py#L504-L516

DanielYang59:
## Summary

- Fix #3815.

### Unchanged behavior
```python
# Single element
print(Composition({"O": 2}) > Composition({"O": 1}))   # True

# Multiple shared elements
print(Composition({"Fe": 3, "O": 4}) > Composition({"Fe": 2, "O": 3}))   # True

# Superset
print(Composition({"C":1, "O": 2}) > Composition({"O": 2}))   # True
```

### Breaking change (fix)

As per the docstring: https://github.com/materialsproject/pymatgen/blob/fbf8ec0a5a9c0fb649975d66dedb31b17bcc9dfc/pymatgen/core/composition.py#L195-L202

```python
# The followings would now be False (with a warning)
# Single element
print(Composition({"S": 1}) > Composition({"O": 1}))   # False, with a warning
print(Composition({"O": 1}) > Composition({"S": 1}))   # False, with a warning

# Multiple elements but not superset
print(Composition({"S": 1, "O": 2}) > Composition({"C": 1, "O": 2}))   # False, with a warning
print(Composition({"C": 1, "O": 2}) > Composition({"S": 1, "O": 2}))   # False, with a warning
```

### DIFF
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -203,7 +203,7 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
         if not isinstance(other, type(self)):
             return NotImplemented
 
-        if not set(self.elements).issuperset(other.elements):
+        if not set(self.elements).issuperset(set(other.elements)):
             # TODO: revise warning message
             warnings.warn("Elements is not a superset.")
             return False
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -197,16 +197,16 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
         if all elements in B are in A and the amount of each element in A is
         greater than or equal to the amount of the element in B within
         Composition.amount_tolerance.
-
-        Should ONLY be used for defining a sort order (the behavior is probably not what you'd expect).
         """
         if not isinstance(other, type(self)):
             return NotImplemented
 
+        # if not set(self.elements).issuperset(other.elements):
+        #     return False
+
         for el in sorted(set(self.elements + other.elements)):
             if other[el] - self[el] >= type(self).amount_tolerance:
                 return False
-            # TODO @janosh 2024-04-29: is this a bug? why would we return True early?
             if self[el] - other[el] >= type(self).amount_tolerance:
                 return True
         return True
--- a/pymatgen/core/composition.py
+++ b/pymatgen/core/composition.py
@@ -197,18 +197,20 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
         if all elements in B are in A and the amount of each element in A is
         greater than or equal to the amount of the element in B within
         Composition.amount_tolerance.
+
+        Should ONLY be used for sorting purpose (the behavior is probably not what you'd expect).
         """
         if not isinstance(other, type(self)):
             return NotImplemented
 
-        # if not set(self.elements).issuperset(other.elements):
-        #     return False
+        if not set(self.elements).issuperset(other.elements):
+            # TODO: revise warning message
+            warnings.warn("Elements is not a superset.")
+            return False
 
         for el in sorted(set(self.elements + other.elements)):
             if other[el] - self[el] >= type(self).amount_tolerance:
                 return False
-            if self[el] - other[el] >= type(self).amount_tolerance:
-                return True
         return True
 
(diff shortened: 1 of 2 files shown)


DanielYang59:
Can I have your advice on defining the comparability here? Thanks a lot in advance. @shyuep @mkhorton @janosh

shyuep:
There is no meaning to comparing compositions.

DanielYang59:
Thanks for your quick response. I assume you mean we don't define comparison to be chemically meaningful, but just for sorting purpose (as per the docstring)? Am I understanding that correct?
https://github.com/materialsproject/pymatgen/blob/fbf8ec0a5a9c0fb649975d66dedb31b17bcc9dfc/pymatgen/core/composition.py#L195-L202

shyuep:
I would say unless there is some good reason, ie some broken randomness in the comparison or sorting, I would leave it be. This is purely a non meaningful sorting thing and there is no reason to cause any upstream breakage by changing it.

DanielYang59:
Thanks for your reply, I really appreciate that. Here are currently two reasons:
1. `True` is returned early for multi-element `Composition`:
  https://github.com/materialsproject/pymatgen/blob/fbf8ec0a5a9c0fb649975d66dedb31b17bcc9dfc/pymatgen/core/composition.py#L206-L212

  And lead to the following false positive:
  ```python
  print(Composition({"Fe": 4, "S": 2}) > Composition({"Fe": 3, "S": 4}))  # True
  ```

2. For the single-element case, but currently the implementation doesn't agree with the docstring: 
  ```python
  print(Composition({"S": 1}) > Composition({"O": 1}))   # True
  ```
  And as per the docstring, 
  > We consider compositions A >= B if all elements in B are in A and the amount of each element in A is greater than or equal to the amount of the element in B
  
  This does not satisfy the condition that "all elements in B are in A" and should be `False` instead.

What do you think?

shyuep:
Well, there really isn't any meaning to it. An element not present in a composition has an implied amount of 0. The intent is to sort all elements by electronegativity order. We compare the most electropositive element first. If one amount is greater than the other, we make a conclusion. If not, we compare the next element. In your example, we will only consider S if the amounts of Fe are the same within tolerance.

In the second case, there are two elements to be compared - S and O. Since S is more electropositive, it is compared first. The first one has 1 S and the second one has 0 S. So it is >.

The docstring can be clarified to be clearer. But I don't think there is much to do in terms of actual code since there is no rigorous way to define comparisons for Compositions. It is what it is.

shyuep:
I push a docstring correction. Since this is a nonsense comparison to begin with, let's not spend too much time dwelling on it.

DanielYang59:
Okay sure. Thanks for the explanation and clarifying the docstring. 

Perhaps we want to remove this TODO tag as well?
https://github.com/materialsproject/pymatgen/blob/f10ace8809765ce573fb334269ee199d59b2dbb8/pymatgen/core/composition.py#L209

PR #3897, opened 2024-06-26
Bug fix for ComputedEntry from_dict

esoteric-ephemera:
In `pymatgen.entries.computed_entries`, `ComputedEntry` takes three optional fields: `energy_adjustments`, `parameters`, `data`, which are not correctly read in by its `from_dict` method. This is because, for example:
```
parameters = {k: MontyDecoder().process_decoded(v) for k, v in (dct.get("parameters", {})}
```
will throw `AttributeError: 'NoneType' object has no attribute 'items'` if the input `ComputedEntry` dict has `parameters = None`.

Because the typing of `ComputedEntry` suggests that these fields can be `None`, this PR ensures that the `from_dict` can properly handle null fields and also ensures correct param defaults (`energy_adjustments` incorrectly defaults to an empty `dict` instead of `list`).

Also added a test for this behavior. cc @tsmathis who found this behavior

### DIFF
--- a/pymatgen/entries/computed_entries.py
+++ b/pymatgen/entries/computed_entries.py
@@ -482,27 +482,29 @@ class ComputedEntry(Entry):
         Returns:
             ComputedEntry
         """
-        # the first block here is for legacy ComputedEntry that were
-        # serialized before we had the energy_adjustments attribute.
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks for the fix.

Issue #3889, opened 2024-06-22
dftd4 support.


hongyi-zhao:
### Feature Requested

VASP has implemented support for [dftd4](https://github.com/dftd4/dftd4#building-vasp-with-support-for-d4), as stated on [VASP wiki](https://www.vasp.at/wiki/index.php/DFT-D4).

On the other hand, pymatgen still doesn't support dftd4, as shown [here](https://github.com/materialsproject/pymatgen/blob/master/pymatgen/io/vasp/vdW_parameters.yaml). So, I think it should be better if we can add support for dftd4 in pymatgen.

### Proposed Solution

According to the description [here](https://www.vasp.at/wiki/index.php/DFT-D4), it seems adding the following into https://github.com/materialsproject/pymatgen/blob/master/pymatgen/io/vasp/vdW_parameters.yaml should be enough:
```
dftd4:
    IVDW: 13
```
If there are any omissions or considerations I have overlooked, please feel free to point them out.

### Relevant Information

The related packages, such as [this file](https://github.com/materialsproject/atomate2/blob/50ae3c05daa5d6e2deaeda4c37a4277e7b7ac4f5/src/atomate2/vasp/sets/base.py#L277) in the atomate2, should also be updated accordingly.

esoteric-ephemera:
Hi @hongyi-zhao, for the time being: you can pass `{"IVDW" : 13}` to the `user_incar_settings` of any VASP input set and this will accomplish the same thing. The `vdW_parameters.yaml` file is more of a convenience thing than a required one

But this should be a quick add to pymatgen. The base sets file you linked in atomate2 is [being overhauled to inherit directly from pymatgen](https://github.com/materialsproject/atomate2/pull/854) so no changes will be needed there

hongyi-zhao:
> But this should be a quick add to pymatgen.

So, a quick PR is appropriate, in this case.

PR #3899, opened 2024-06-26
Add VASP setting for the dftd4 vdW functional.

hongyi-zhao:
Add VASP setting for the dftd4 vdW functional, as discussed in https://github.com/materialsproject/pymatgen/issues/3889.


shyuep:
This PR is fine but there are failing tests.

JaGeo:
@shyuep  these failing tests seem to be "tests" that make sure the input sets stay the same. Some hashes are out of date. 

They say one should notify all input set owners: 
@mkhorton @janosh and you as well @shyuep

mkhorton:
Thanks for the PR @hongyi-zhao!

Indeed, the hash failures as @JaGeo are a safety mechanism to make sure that we are intentional when changing input sets. Since this is an additive change this is all good. I have committed a change to update the test.

Once the tests pass I would be happy to merge.

PR #3900, opened 2024-06-28
Improve Keep Redundant Spaces algorithm for PatchedPhaseDiagram

CompRhys:
## Summary
- modularize the keep redundant spaces code, and fix the algorithm which previously didn't get rid of all the redundant spaces.

## Checklist

### DIFF
--- a/src/pymatgen/analysis/phase_diagram.py
+++ b/src/pymatgen/analysis/phase_diagram.py
@@ -1619,29 +1619,21 @@ class PatchedPhaseDiagram(PhaseDiagram):
         # Add the elemental references
         inds.extend([min_entries.index(el) for el in el_refs.values()])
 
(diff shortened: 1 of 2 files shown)


janosh (approved):
~`PatchedPhaseDiagrams` pickled to disk with this change should result in smaller files as fewer sub phase diagrams need to be stored~

@CompRhys just corrected me, all this is saving are some memory IDs which are on the order of a few bytes so no meaningful decrease in file size expected

mkhorton:
Thanks @CompRhys!

I think some comment strings need to be updated: for example, there is a claim that `PatchedPhaseDiagram` uses the same `as_dict()` as `PhaseDiagram` but this is not the case (`computed_data`, i.e. including the qhull entries, are not stored). Otherwise this looks great.

janosh (approved):
@CompRhys thanks for addressing the outdated doc strings!

PR #3901, opened 2024-07-01
prettify periodic_table.json for readability

rkingsbury:
Prettify `periodic_table.json` to improve its readability. (using the Prettify JSON VS Code extension)

I considered the alternative of converting to YAML so that comments could be inserted (see #3851 ), but I found the following way back in the git log from 2011:

```
--
Date:   Mon Nov 14 13:19:54 2011 -0500

    Move periodicTable back to json format for actual loading.  Loading times ~ 200x
--
Date:   Thu Nov 10 12:07:13 2011 -0500

    Further sanitize yaml for periodic table to be proper sentence case keys.
--
Date:   Thu Nov 10 10:40:16 2011 -0500

    Switch periodic table data format to YAML for easier maintenance.
```

I'm not up to date on whether in modern times YAML is still much slower to read than .json, but I thought it best to stick with .json for now.


shyuep:
Pls test the loading speeds on the pretty vs non pretty version. This is something that is loaded often. So every millisecond counts. I would not prioritize being able to read the JSON directly since people seldom do it and it can always be pretty printed if needed.

rkingsbury:
@shyuep good point; the impact does not appear significant. See timing results below. I will note that I repeated these tests multiple times manually, with multiple kernel restarts, etc. and the results were noisy, with load times always within +/- 1.5 ms of each other. Sometimes the pretty version loaded faster, sometimes the non-pretty version loaded faster.

Run on python 3.12.0 on a windows laptop (WSL2) on battery power, restarting a fresh kernel before running.
```
%%timeit -n 10 -r 500
import json

# Step 1: Load the JSON file
with open('/home/ryan/mambaforge/envs/pbx/code/pymatgen/src/pymatgen/core/periodic_table.json', 'r') as file:
    data = json.load(file)
```

Results (output of 3 successive runs for each condition is shown)
**Current `master` (not pretty)**:
`3.51 ms ± 790 µs per loop (mean ± std. dev. of 500 runs, 10 loops each)`
```
The slowest run took 4.67 times longer than the fastest. This could mean that an intermediate result is being cached.
2.77 ms ± 622 µs per loop (mean ± std. dev. of 500 runs, 10 loops each)
```
```
The slowest run took 5.98 times longer than the fastest. This could mean that an intermediate result is being cached.
3.93 ms ± 1.45 ms per loop (mean ± std. dev. of 500 runs, 10 loops each)
```



**This PR (pretty)**
`3.61 ms ± 718 µs per loop (mean ± std. dev. of 500 runs, 10 loops each)`
```
The slowest run took 4.39 times longer than the fastest. This could mean that an intermediate result is being cached.
3.25 ms ± 868 µs per loop (mean ± std. dev. of 500 runs, 10 loops each)
```
`3.64 ms ± 875 µs per loop (mean ± std. dev. of 500 runs, 10 loops each)`

shyuep:
Thanks

PR #3915, opened 2024-07-10
Fix chemical system method for different oxidation states

danielzuegner:
## Summary

Currently, when the same element occurs in multiple oxidation states in a structure (e.g., `Composition({"Na+": 1, "Na2+": 1, "Cl-": 1})`), `structure.composition.chemical_system` yields that element repeated, e.g., `Composition({"Na+": 1, "Na2+": 1, "Cl-": 1})`), `structure.composition.chemical_system == "Cl-Na-Na"`. My hypothesis is that this was introduced by the fix to this issue: https://github.com/materialsproject/pymatgen/issues/2248.

This PR fixes the bug by using the newly introduced `chemical_system_set` property of `Composition`, and adds the line
```python
assert Composition({"Na+": 1, "Na2+": 1, "Cl-": 1}).chemical_system == "Cl-Na"
```
to `test_composition.py`.

## Checklist


mkhorton:
Thanks for reporting @danielzuegner, this seems like a clear bug. We should get this fixed :)

mkhorton (approved):

PR #3768, opened 2024-04-18
WIP: Adding support for reading vaspwave.h5 in Wavecar and Chgcar class

Shibu778:
Dear Maintainers,
This is my first contribution to `pymatgen`. I am relatively new to opensource and would appreciate your feedback and guidance in improving the added features and this PR. In further commits, I will add `read_from_hdf5` functionality to `Chgcar` and write all the tests.

## Summary

Major changes:

- feature 1: Added a method `read_from_hdf5` in `Wavecar` class to read `WAVECAR` related information from `vaspwave.h5` file (refer to [vaspwiki](https://www.vasp.at/wiki/index.php/Vaspwave.h5)).
- fix 1: made a function to read the legacy `WAVECAR` format that was previously implemented in `Wavecar` class.
- fix 2: added appropriate comments and doc-string
- fix 3: Some part of the `pymatgen/io/vasp/outputs.py` file got formatted by black formatter

## Todos

This is work in progress, following needs to be done

- feature 2: Add the same feature in `Chgcar` class
- fix 4: Write appropriate test to read the `hdf5` files in using `Wavecar` and `Chgcar` class

## Pre-commit result
`pre-commit run --all-files` passed all everything

```shell
~\Github\pymatgen> pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
check yaml...............................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
mypy.....................................................................Passed
codespell................................................................Passed
cython-lint..............................................................Passed
double-quote Cython strings..............................................Passed
blacken-docs.............................................................Passed
markdownlint.............................................................Passed
nbstripout...............................................................Passed
```
## Checklist

- [x] Google format doc strings added. Check with `ruff`.
- [x] Type annotations included. Check with `mypy`.
- [x] The tests for fixes in `Wavecar` class remains unchanged to read legacy `WAVECAR` file 
- [x] Tests added for new features.

## Summary by CodeRabbit

- **Refactor**
	- Improved code readability and formatting in various functions.
	- Updated function signatures for clarity.
  
- **Documentation**
	- Enhanced function descriptions with new comments and docstrings.

PR #3916, opened 2024-07-10
Add new MP input sets / concurrently update MatPES POTCARs

esoteric-ephemera:
This PR adds new MP input sets that resulted from a benchmarking effort over the past year. For the benchmarks, refer to:
- https://github.com/materialsproject/foundation/pull/26
- [Reference living document](https://drive.google.com/file/d/1fUUx0wrrtMRcSss5yv3NiQuC7J5IiEKL/view)

Critical updates include:
- POTCARs
- `KSPACING` adjustment by bandgap
- `LREAL = False`
- `LMAXMIX = 6`

The updated KSPACING function is slightly different from what was discussed with the MP foundation initially. This is to ensure that calculations with the highest $k$-point density (KSPACING = 0.22) would not be deprecated by this PR. Previously I recommended that the minimum KSPACING = 0.2, but this is effectively the same as 0.22.

~Leaving this up as a WIP pending further community discussion and writing tests~

**`MatPESStaticSet` Users:** please note that the recommended POTCARs for MatPES (PBE64 generally) have updated. These updates are consistent with the data generation for MatPES, please use the 12 recommended updated POTCARs


src/pymatgen/io/vasp/MP24RelaxSet.yaml
```diff
@@ -0,0 +1,129 @@
...
+  ENCUT: 680
+  IBRION: 2
+  ISIF: 3
+  ISMEAR: 0 # included to have some reasonable default
+  ISPIN: 2
+  KSPACING: 0.22 # included to have some reasonable default
+  LAECHG: True
+  LASPH: True
+  LCHARG: True
+  LELF: False # LELF = True restricts calculation to KPAR = 1
```
mkhorton, line 19: I believe it restricts to NPAR = 1 not KPAR = 1. I am not sure where this restriction originates.

There is some logic that expensive-to-compute quantities should be saved by default? (e.g., to my knowledge the only way to recover the KE density from a VASP calculation is to store the ELF)

I see the sense in leaving this as `False` but I wonder what should be stored by default? I imagine there is a trade-off with I/O and post-processing time too, especially for simple calculations.

esoteric-ephemera: The [manual says NPAR = 1](https://www.vasp.at/wiki/index.php/LELF), but the actual error message I get when I run VASP 6.4.2 (just checked) is: `ELF: KPAR>1 not implemented, sorry.` Either way, restricts the parallelization

You can reconstruct the KED from the orbitals in WAVECAR, since those should just be the plane wave coefficients + wave vectors (think we have tools in PMG to parse this but haven't looked thoroughly)

I'm good with `LELF = True` for the final static in the MP workflow, since this is usually very cheap (it starts from the WAVECAR of the r2SCAN relaxation)

mkhorton: So we hard-code `KPAR: 1` for the final static?

Makes sense with the WAVECAR, I am genuinely wondering whether one day we should encourage storage of this by default if storage costs continue to reduce. I am encountering more and more post-processing tools that would really benefit if a large dataset of wave functions were available. I wonder how compressible they are with modern methods too, something I will not have time to investigate unfortunately!

esoteric-ephemera: A little hesitant to do that, especially if we want to treat larger systems, but it'll require some testing for me to see if `KPAR = 1` is a real practical issue

Saving the [vaspwave.h5 file](https://www.vasp.at/wiki/index.php/Vaspwave.h5) might be better in the long run. There's currently an [open PR](https://github.com/materialsproject/pymatgen/pull/3768) for this, I also need to finish the vaspout.h5 PR

matthewkuner:
@esoteric-ephemera is this PR ready for merging?

esoteric-ephemera:
@matthewkuner  debugging some potential issues with the Ba pseudo (updated to the GW one) - hope to have an update soon

mkhorton:
Copying a suggestion here:

> @esoteric-ephemera a late suggestion, but can we set GGA_COMPAT to False? See [here](https://www.vasp.at/wiki/index.php/GGA_COMPAT) for discussion.

esoteric-ephemera:
@mkhorton: Sorry I forgot to answer! Yes shouldn't be an issue to add this, makes sense to me
@shyuep: Should  be ready for review

src/pymatgen/io/vasp/MP24RelaxSet.yaml
shyuep: I don't think this is an issue. A user can always override KPAR with user incar settings.

Andrew-S-Rosen: @esoteric-ephemera, @mkhorton: Just a head's up that the VASP team does strongly suggest using NPAR = 1 for LELF = True. I asked about it on the forum recently. Here was their response: https://www.vasp.at/forum/viewtopic.php?t=20020.

src/pymatgen/io/vasp/MP24RelaxSet.yaml
```diff
@@ -0,0 +1,31 @@
+# Default VASP settings for calculations in the Materials Project
+# using the Regularized-Restored Strongly Constrained and Appropriately
+# Normed functional (r2SCAN).
+PARENT: PBE64Base
+INCAR:
+  ALGO: Normal
+  EDIFF: 1.e-05
+  EDIFFG: -0.02
+  ENAUG: 1360
```
Andrew-S-Rosen, line 9: Linking this to https://github.com/materialsproject/pymatgen/issues/4301.

PR #3917, opened 2024-07-10
added cp2k_output cube to pymatgen volume parser

Sudo-Raheel:
## Summary
added a function parse_cube() in pymatgen.io.cp2k.output
slight change in parse_initial_structure() ... cp2k output different ...i used cp2k/cp2k-git-feb-2022 to generate outputs....
I have benchmarked the planar average using pymatgen against cubecruncher(https://www.cp2k.org/tools:cubecruncher)... 
These results are in shown in "https://github.com/Sudo-Raheel/cp2k_to_pymatgen/tree/main/cube_file"
Few more points 
1) the documentation should mention that print level should be at least medium in cp2k ..else it wont be able to parse the structure from cp2k.out
2) allow using the phrase "poisson_periodicity" to find whether is_molecule() works only when default poisson solver in cp2k is used .

### DIFF
--- a/src/pymatgen/io/cp2k/outputs.py
+++ b/src/pymatgen/io/cp2k/outputs.py
@@ -26,6 +26,7 @@ from pymatgen.io.cp2k.inputs import Keyword
 from pymatgen.io.cp2k.sets import Cp2kInput
 from pymatgen.io.cp2k.utils import natural_keys, postprocessor
 from pymatgen.io.xyz import XYZ
+from pymatgen.io.common import VolumetricData
 
 __author__ = "Nicholas Winner"
 __version__ = "2.0"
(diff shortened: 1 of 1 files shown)


Sudo-Raheel:
@DanielYang59  any chance you can help me with this

DanielYang59:
Sorry I have never used CP2K... But I could probably answer some general questions. So is there anything I could help with?

@JaGeo Any chance you know who is maintaining the CP2K module? Thanks! @nwinner maybe (from the `__author__`)?

JaGeo:
@DanielYang59 at least, @nwinner wrote a cp2k workflow for atomate2 😀

nwinner:
I originally wrote the cp2k module and the cube parser. Yes, I also included an atomate2 workflow.

Here is the constructor for cube parsing into VolumetricData.

https://github.com/materialsproject/pymatgen/blob/27d90b777cedc80593d93b6e8b66dee04c167460/src/pymatgen/io/common.py#L356

Is there a reason to implement a separate parser now in the cp2k module?

Any other questions you have for me?

Issue #3928, opened 2024-07-17
Installation failed using self-forked version


jinlhr542:
### Python version

3.11

### Pymatgen version

2024.06.10

### Operating system version

Mac

### Current behavior

![image](https://github.com/user-attachments/assets/f932b172-fdd8-4ba6-955e-938b2210c03b)
![image](https://github.com/user-attachments/assets/bb4a4c91-4fad-40f8-8030-2300a2d47473)
I am trying to contribute to pymatgen but as you can see when I install my forked pymatgen by 'pip install .' The packages are not installed

### Expected Behavior

\

DanielYang59:
I'm guess you might have a `pymatgen.py` shadowing the real `pymatgen`, let me know if it's not the case :)

jsukpark:
Same issue here. Python 3.12, Pymatgen version 2024.6.10, CentOS Linux 7. Trying to run

```
from pymatgen.io.pwscf import PWInput
```

after installing my forked pymatgen gives

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'pymatgen.io'
```

Tried using both `pip install .` in the top directory and `pip install -U git+https://github.com/jsukpark/pymatgen` for installing my forked pymatgen, same issue happens in both.

DanielYang59:
Hi @jsukpark thanks for providing the information. I just tried to clone your fork and things seem to work fine on my side. 

Can you post the output of `pip install .` so that I could have a closer look?

jsukpark:
```
(torch-v2) login2.frontera(1027)$ pip install .
Processing /work2/08542/epark/frontera/pymatgen
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: matplotlib>=3.8 in /work2/08542/epark/frontera/miniconda3/envs/torch-v2/lib/python3.12/site-packages (from pymatgen==2024.6.10) (3.9.0)
Requirement already satisfied: monty>=2024.5.24 in /work2/08542/epark/frontera/miniconda3/envs/torch-v2/lib/python3.12/site-packages (from pymatgen==2024.6.10) (2024.7.12)
Requirement already satisfied: networkx>=2.2 in /work2/08542/epark/frontera/miniconda3/envs/torch-v2/lib/python3.12/site-packages (from pymatgen==2024.6.10) (3.3)
[...]

PR #3934, opened 2024-07-18
[Hot Fix] Fix `setuptools` for pymatgen packaging

DanielYang59:
### Summary

- A hot fix for #3928, where the `setuptools` in not configured to include the complete `pymatgen` package (all subpackages are missing) after moving to the src layout.
- Turn off `include-package-data`, which [is `true` by default](https://setuptools.pypa.io/en/latest/userguide/datafiles.html), meaning `package-data` is not having any effect and all data files would be included into the package.
- Remove resolved skip tag: https://github.com/materialsproject/pymatgen/blob/02341828b94a001b4f3cbdae6b24eaeeb0df7d7d/tests/core/test_structure.py#L1771

#### Site note

To reliably recreate this issue, one might need to [completely clear build cache](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#caching-and-troubleshooting) (removing `build/`, `dist/` and `*.egg-info`), otherwise you might find yourself (me for example) modifying the `pyproject.toml` file but the wheel remains the unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,7 +133,7 @@ get_environment = "pymatgen.cli.get_environment:main"
 
 [tool.setuptools.packages.find]
 where = ["src"]
-include = ["pymatgen"]
+include = ["pymatgen", "pymatgen.*"]
 
 [tool.setuptools.package-data]
 "pymatgen.analysis" = ["*.yaml", "*.json", "*.csv"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,6 +131,9 @@ feff_plot_cross_section = "pymatgen.cli.feff_plot_cross_section:main"
 feff_plot_dos = "pymatgen.cli.feff_plot_dos:main"
 get_environment = "pymatgen.cli.get_environment:main"
 
+[tool.setuptools]
+include-package-data = false
+
 [tool.setuptools.packages.find]
 where = ["src"]
 include = ["pymatgen", "pymatgen.*"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -131,9 +131,12 @@ feff_plot_cross_section = "pymatgen.cli.feff_plot_cross_section:main"
 feff_plot_dos = "pymatgen.cli.feff_plot_dos:main"
 get_environment = "pymatgen.cli.get_environment:main"
 
+[tool.setuptools]
+include-package-data = false
+
 [tool.setuptools.packages.find]
```
DanielYang59, line 137: We can completely get rid of this `[tool.setuptools.packages.find]` section completely owing to the [`automatic discovery` feature of `setuptools`](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html), but I guess explicitly declaring it might be clearer?

DanielYang59:
@shyuep @janosh @mkhorton Can you please review this hot fix? Thanks!

pyproject.toml
```diff
@@ -136,7 +136,6 @@ include-package-data = false
 
 [tool.setuptools.packages.find]
 where = ["src"]
-include = ["pymatgen", "pymatgen.*"]
```
DanielYang59: Perhaps just get rid of `include` altogether to use auto discover. Is there any sub-package we want to exclude?

PR #3936, opened 2024-07-18
Adding Abinit magmoms from netCDF files to Structure.site_properties

gbrunin:
## Summary

In recent developments in Abinit, the variables `intgden` and `ratsph` have been added in the netCDF output files. These correspond to the computed integrated density around atoms and the integration radii. The former can be interpreted as the local magnetic moment on each atom. Since this is a useful quantity for, e.g., CHGNet model training, I have added them as `site_properties` of the `Structure` that is extracted from the netCDF file. If `intgden` is not found in the .nc file, nothing is added. 

While testing with abipy, I also found that recent commits in io.abinit.pseudos were backward incompatible for PAW pseudopotentials. I reverted the changes and added a test.

Tagging @gmatteo to check this.


gbrunin:
This is ready to be reviewed. I don't know why ruff is failing, I did not make changes in io.vasp, io.aims and analysis.

gbrunin:
Hi @janosh, @shyuep, @mkhorton, whenever you have some time, this is ready to be reviewed. It would be great for us if these changes could be included in the next release, to facilitate developments in abipy (since this PR is fixing breaking changes). Thanks!

tests/files/io/abinit/Fe_magmoms_collinear_GSR.nc
janosh, line 1: `tests/files/io/abinit/Fe_magmoms_collinear_GSR.nc`  and `tests/files/io/abinit/Fe_magmoms_noncollinear_GSR.nc` are big files. are they actually needed for the test? if so, they should be compressed.

janosh (changes requested):

tests/files/io/abinit/Fe_magmoms_collinear_GSR.nc
gbrunin: Hi, yes the files are necessary because they included the latest changes in the netCDF files in Abinit. I also totally understand the need to keep files as small as possible. I have compressed them, I hope it's enough.

tests/files/io/abinit/28ni.paw
janosh, line 1: wow, the GSR files compressed to much smaller than expected. can we compress this file as well? sorry, should have mentioned this file in the first review

gbrunin: done

janosh (approved):
thanks for this contribution @gbrunin! 👍

PR #3941, opened 2024-07-19
`is_subgroup()` modifications in `SpaceGroup` and `PointGroup`

### LINKED ISSUE #3937: Group-subgroup relationship bugs in `SpaceGroup` and `PointGroup` class
### Python version

3.10

### Pymatgen version

master

### Operating system version

_No response_

### Current behavior

Hey, two remarks regarding the `is_subgroup` method of `SpaceGroup`:

First, the current implementation does not allow isomorphic group-subgroup relationships (same space group type) because of 
https://github.com/materialsproject/pymatgen/blob/02341828b94a001b4f3cbdae6b24eaeeb0df7d7d/src/pymatgen/symmetry/groups.py#L537
For example, I would expect the following code to not raise an error:
`assert SpaceGroup("P3").is_subgroup(SpaceGroup("P3"))`
The `if j not in all_groups` needs to be removed in the first iteration.

Second, the current implementation disregards klassengleiche group-supergroup relationships where the overall translational symmetry is increased, but not through centering. For example, I would expect this to not raise an error:
`assert SpaceGroup("Fm-3m").is_subgroup(SpaceGroup("Pm-3m"))`

However, because of https://github.com/materialsproject/pymatgen/blob/02341828b94a001b4f3cbdae6b24eaeeb0df7d7d/src/pymatgen/symmetry/groups.py#L528 this returns False. I assume it is to save resources? I would either remove it completely (safest) or restrict it to certain cases, if possible(?).

-------------------

Also, I had a look at the uncommented `PointGroup` subgroup test:
https://github.com/materialsproject/pymatgen/blob/02341828b94a001b4f3cbdae6b24eaeeb0df7d7d/tests/symmetry/test_groups.py#L59
This test is failing because the -3 is defined along (111) in m-3m, but along (001) in -3m (therefore, the matrices are different). Generally, this affects all group-subgroup relationships if the groups are in crystal systems with different crystallographic directions / blickrichtungen. For example, this is also failing,
`assert PointGroup("2").is_subgroup(PointGroup("4"))`
because the rotation axis is along *c* in the tetragonal, but is notated along *b* in the monoclinic system (in this database).
I would specify this in the warning https://github.com/materialsproject/pymatgen/blob/02341828b94a001b4f3cbdae6b24eaeeb0df7d7d/src/pymatgen/symmetry/groups.py#L79
In the future, it would probably be best to have a database similar to the maximal subgroup database of space group types.

---------------------

If you agree, I can implement and test the SpaceGroup changes.

### Expected Behavior

I would expect my code snippets from above to not raise an error.

### Minimal example

_No response_

_No response_

kaueltzen:
## Summary
Resolves #3937 .

### SpaceGroup
- enabled isomorphic group subgroup relationships, added test
- Constrained https://github.com/materialsproject/pymatgen/blob/02341828b94a001b4f3cbdae6b24eaeeb0df7d7d/src/pymatgen/symmetry/groups.py#L528 to non-klassengleiche group subgroup relationships so cases like `assert SpaceGroup("Fm-3m").is_subgroup(SpaceGroup("Pm-3m"))` pass, added test
- fixed bug introduced by me in #3859 (Apologies!) of wrong short Hermann Mauguin symbol computation for multiple trigonal groups
- modified `point_group` entries of `symm_data.json` to be possible `symbol` attributes of `PointGroup` (also see comment below)

### PointGroup
- added `crystal_system` attribute to `PointGroup`
- enabled `init` from other settings than in database and from full symbol, added test
- added notes in docs that axes of `PointGroup` object may differ from given `int_symbol` in `init` / given space group symbol in `from_space_group` as currently only one setting per point group is available
- added crystallographic direction issue check for `is_subgroup` of `PointGroup` -> a `NotImplementedError` is now raised if the blickrichtungen / crystallographic directions of the crystal systems of the two groups differ


src/pymatgen/symmetry/groups.py
```diff
@@ -166,6 +173,44 @@ def get_orbit(self, p: ArrayLike, tol: float = 1e-5) -> list[np.ndarray]:
                 orbit.append(pp)
         return orbit
 
+    def is_subgroup(self, supergroup: PointGroup) -> bool:
```
kaueltzen, line 176: Added the constraints due to crystallographic direction issue (I could think of) but would prefer a database solution similar to `SpaceGroup`. For example, it is an option to contact the Bilbao Crystallographic Server and ask for permission to use their database?

JaGeo: If they release it with a license that allows us to use it, I guess it is an option.

kaueltzen: Not sure about that.

kaueltzen:
Hey @shyuep @mkhorton @janosh this is ready to be reviewed.

src/pymatgen/symmetry/groups.py
```diff
@@ -88,7 +109,10 @@ def is_supergroup(self, subgroup: SymmetryGroup) -> bool:
         Returns:
             bool: True if this group is a supergroup of the supplied group.
         """
-        warnings.warn("This is not fully functional. Only trivial subsets are tested right now. ")
+        warnings.warn(
+            "This is not fully functional. Only trivial subsets are tested right now. "
```
janosh, line 113: typo: `Only trivial (subsets->supergroup)...`

maybe extract these duplicate messages into a `trivial_group_warning` format string with a `which_set` parameter that you fill with `subgroup`/`supergroup` in `def is_(sub|super)group`

janosh (changes requested):
looks great! thanks @kaueltzen. just a minor suggestion

src/pymatgen/symmetry/groups.py
kaueltzen: Hi, I'm not sure what you mean. A function like this?
```
def trivial_group_warning(subgroup: str, supergroup: str) -> str:
    return (
        f"This is not fully functional. Only trivial subsets are tested right now."
        f"This will not work if the crystallographic directions of subgroup "
        f"{subgroup} and supergroup {supergroup} are different."
    )
```
that is called in `is_(sub|super)group`?
```
warnings.warn(trivial_group_warning(subgroup=self.symbol, supergroup=supergroup.symbol))
```

dev_scripts/update_spacegroup_data.py
```diff
@@ -75,6 +77,11 @@ def remove_identity_from_full_hermann_mauguin(symbol: str) -> str:
 
 SYMM_DATA["space_group_encoding"] = new_symm_data
 
+for sg_symbol in SYMM_DATA["space_group_encoding"]:
+    SYMM_DATA["space_group_encoding"][sg_symbol]["point_group"] = PointGroup.from_space_group(
```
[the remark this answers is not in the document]
janosh: is this still waiting on external input? looks like this PR went stale so maybe need to ping someone?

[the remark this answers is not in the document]
kaueltzen: Hey @janosh , originally, i was hoping for input from you, @mkhorton  or @shyuep . However, if there are no strong opinions on this, these changes can be merged.

src/pymatgen/symmetry/groups.py
kaueltzen: Hey @janosh can you maybe elaborate on your comment from Aug 2? Thanks!

PR #2772, opened 2022-12-13
Added methods to compute and compare DOS fingerprints

naik-aakash:
##  Compute and compare fingerprints of project and total density of states. 

This functionality could be really useful to compare the DOS of materials from different softwares  (eg:- LOBSTER and VASP ) or versions (Eg:- VASP v5 and VASP v6)

- Added method to compute fingerprints 
- Added method to compare two DOS fingerprints (tanimoto index and dot product)
- Added test cases for the corresponding methods

PR #3946, opened 2024-07-22
DOS Fingerprints enhancements

naik-aakash:
# Summary

1.  Add possibility to compare phonon dos , implementation is similar to electronic dos as implemented in https://github.com/materialsproject/pymatgen/pull/2772
2. Add `Wasserstein `distance as metric besides existing Tanimoto index

# Todo's

- [x] Add methods to compute fingerprints and compare phonon dos
- [x] Move methods to compute fingerprints and compare phonon dos to `PhononDos ` class from `CompletePhononDos` class (Phonon workflow returns `PhononDos` objects in the taskdoc)
- [x] Update doc-strings where needed
- [x] Add tests for phonon dos comparisons
- [x] Adapt electronic dos fingerprint comparison to include Wasserstein metric 
- [x] Add tests for updated electronic dos comparisons
- [x] Address review comments


naik-aakash:
Hi @shyuep , @janosh , @mkhorton  and @JaGeo , this PR is ready for review.

src/pymatgen/phonon/dos.py
```diff
@@ -411,6 +413,162 @@ def get_last_peak(self, threshold: float = 0.05) -> float:
...
+                Default is True.
+            min_f (float): The minimum mode frequency to include in the fingerprint (default is None)
+            max_f (float): The maximum mode frequency to include in the fingerprint (default is None)
+            n_bins (int): Number of bins to be used in the fingerprint (default is 256)
+            normalize (bool): If true, normalizes the area under fp to equal to 1. Default is True.
+
+        Returns:
+            PhononDosFingerprint: The phonon density of states fingerprint
+                of format (frequencies, densities, n_bins, bin_width)
+        """
```
JaGeo, line 437: is this code very similar to the one for the electronic structure? if so, can one combine?

naik-aakash: It is similar, but we do not have option to get fingerprints for projected dos in case of phonons. 

I think, If we want to combine both than we need to move these methods out of electronic and phonon dos class and possibly create another class just for getting fingerprints and comparisons which can accept PhononDos / Electronic dos objects, that can probably reside in `pymatgen.analysis` , maybe in fingerprints.py?

JaGeo: Let's wait what the other reviewers think

src/pymatgen/electronic_structure/dos.py
```diff
@@ -1197,29 +1212,36 @@ def fp_to_dict(fp: NamedTuple) -> dict:
...
     def get_dos_fp_similarity(
-        fp1: NamedTuple,
-        fp2: NamedTuple,
+        fp1: DosFingerprint,
+        fp2: DosFingerprint,
         col: int = 1,
         pt: int | str = "All",
         normalize: bool = False,
-        tanimoto: bool = False,
+        metric: str | None = "Tanimoto",
```
janosh, line 1220: let's make this type annotation a `Literal` with `tanimoto`, `wasserstein` and `dot-product` (or `cosine-sim`). i think using `None` for dot product is confusing here

src/pymatgen/electronic_structure/dos.py
```diff
@@ -1197,29 +1212,36 @@ def fp_to_dict(fp: NamedTuple) -> dict:
...
         Raises:
-            ValueError: If both tanimoto and normalize are set to True.
+            NotImplementedError: If metric other than Tanimoto, Wasserstien and None is requested.
+            ValueError:  If normalize is set to True along with the metric.
 
         Returns:
             float: Similarity index given by the dot product
         """
+        if metric not in [None, "Tanimoto", "Wasserstein"]:
+            raise NotImplementedError(
```
janosh, line 1241: `ValueError` is more semantic here imo

src/pymatgen/phonon/dos.py
janosh: i'm easy. you know best how much redundancy there is and much how over-engineering (if any) it might involve to extract a common `fingerprint.py` module out of the e_dos and ph_dos modules.

tests/phonon/test_dos.py
```diff
@@ -126,6 +126,57 @@ def test_get_last_peak(self):
         peak_freq = self.dos.get_last_peak(threshold=0.5)
         assert peak_freq == approx(4.9662820761)
 
+    def test_get_dos_fp(self):
```
janosh, line 129: nice work, thanks for the thorough test coverage! 👍

janosh (changes requested):
this looks very useful, thanks @naik-aakash! 👍
excited to start testing it in https://github.com/janosh/ffonons

naik-aakash:
Thanks @janosh for the feedback.  I will try to address the comments next week :smile:

src/pymatgen/electronic_structure/dos.py
naik-aakash: Done

src/pymatgen/electronic_structure/dos.py
naik-aakash: Thanks . Have made the change now

naik-aakash:
Hi @JaGeo  and @janosh , I have now addressed review comments now and could be merged if there are no further comments.

PS: Regarding creating a separate module for fingerprints, I had another detailed look at the code and concluded it would overcomplicate the code  logic with minimal advantage of reducing few lines of code. So better to keep it in current place I feel as of now.

janosh:
thanks @naik-aakash! 👍

PR #3950, opened 2024-07-26
Fix MPRester tests and access phonon properties from the new API without having `mp-api` installed.

AntObi:
## Summary

Major changes:

* Add functions to get phonon properties (phonon band structures and phonon DOS) to `_MPResterBasic`, so that a non-power user can retrieve this properties without using `mp-api` making `_MPResterBasic` closer in line with `_MPResterLegacy` in terms of the properties that can be retrieved.
* Fix `get_summary_by_material_id`
* Fixes issue in `test_matproj.py` where tests are always skipped because `requests.get("https://materialsproject.org", timeout=600).status_code` always returns 403 and instead either checks the status the relevant Materials Project site depending on whether a legacy or next-gen API KEY is supplied
* Fix issue in `test_matproj.py` where `TestMPResterNewBasic` was failing because it didn't inherit from the `PymatgenTest` class and the `setUp` method was in the incorrect case.
* Uncomment phonon tests for for `TestMPResterNewBasic`
* Import the `MPRestError` from `pymatgen.ext.matproj_legacy` instead of `pymatgen.ext.matproj`

## Todos

* A future PR will incorporate accessing electronic structure data, but this will require more time due to issues I've found in the new API. Not needed to be done for this particular PR.

## Checklist


shyuep:
Many thanks for this contribution.

Issue #3888, opened 2024-06-21
CrystalNN bug with atom bonded to its own images


fxcoudert:
### Python version

3.11.2

### Pymatgen version

2024.6.10

### Operating system version

macOS 14.5

### Current behavior

Take a structure with only one atom per cell, which bonds to its own images in neighboring cells. Many simple metals do that, like mp-2647057.

Apply CrystalNN to get a bonded structure. The structure for mp-2647057 looks like this:

```
Structure Graph
Structure: 
Structure Summary
Lattice
    abc : 2.428889488469988 2.428889488469988 2.428889488469988
 angles : 109.47122063449069 109.47122063449069 109.47122063449069
 volume : 11.030656874268672
      A : -1.40232 1.40232 1.40232
      B : 1.40232 -1.40232 1.40232
      C : 1.40232 1.40232 -1.40232
    pbc : True True True
PeriodicSite: Co0 (Co) (0.0, 0.0, 0.0) [0.0, 0.0, 0.0]
Graph: bonds
from    to  to_image    
----  ----  ------------
   0     0  (1, 1, 1)   
   0     0  (1, 0, 0)   
   0     0  (0, 0, 1)   
   0     0  (0, 1, 0)   
```

The bonds graph is incorrect. The Co atom should be 8-connected, and there should be bonds to atoms at positions `(-1, -1, -1) (-1, 0, 0) (0, 0, -1) (0, -1, 0)`. These missing bonds have the exact same length and characteristics as the ones present, in fact they are equivalent by symmetry.

If you make a 2x2x2 supercell and apply CrystalNN, you get the correct coordination.

### Expected Behavior

The Co atom should be 8-coordinated.

### Minimal example

```Python

crystallnn = CrystalNN()
from pymatgen.analysis.local_env import CrystalNN
from pymatgen.ext.matproj import MPRester

with MPRester(apikey) as mpr:
    mp_data = mpr.materials.summary.search(
        material_ids=["mp-2647057"],
        fields=["material_id", "formula", "formula_pretty", "nelements", "structure"]
    )

structure = mp_data[0].structure
bonded_struct = crystallnn.get_bonded_structure(structure)
bonded_struct
```
```


### Relevant files to reproduce this bug

_No response_

fxcoudert:
Parts of CrystalNN get the right information. For example:

- `crystallnn.get_cn(structure, 0)` returns `8`, which is correct
[...]

PR #3954, opened 2024-07-26
Fix coordination number bug

jmmshn:
# `StructureGraph.get_coordination_of_site` gives wrong number for small unit cells

Refer to #3888 

The `StructureGraph` object assumes that periodic images of the edges will added twice in the graph so the coordination number algorithm assumes that all self-connections are double-counted.

However, we have updated the definition of the graph such that periodic images of the same edge will only appear once (ie. if `SiteA[0,0,0]--SiteA[0,0,1]` is an edge, we should not see another edge from  `SiteA[0,0,0]--SiteA[0,0,-1]`)

I have also removed the double counting from MoleculeGraph under the assumption that self-connection should never show up in a molecule.

## Modified test assertion value

The two of the edges added here are redundant:
https://github.com/materialsproject/pymatgen/blob/98c57888cc4f17ef4d8692d777ad0d1806ee7ae9/tests/analysis/test_graphs.py#L53-L56

But the coordination number should still be `4` since we have a square lattice.

### DIFF
--- a/src/pymatgen/analysis/graphs.py
+++ b/src/pymatgen/analysis/graphs.py
@@ -2478,8 +2478,7 @@ class MoleculeGraph(MSONable):
         Returns:
             int: the number of neighbors of site n.
         """
-        n_self_loops = sum(1 for n, v in self.graph.edges(n) if n == v)
-        return self.graph.degree(n) - n_self_loops
+        return self.graph.degree(n)
 
     def draw_graph_to_file(
         self,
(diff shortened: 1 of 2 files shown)


mkhorton:
Ah, so there was a dual bug, an error in the test too? Thanks for fixing.

jmmshn:
> Ah, so there was a dual bug, an error in the test too? Thanks for fixing.

Yeah, I'm 99% sure the test is wrong.

PR #3958, opened 2024-07-31
Fix Incar `check_params` for `Union` type 

### LINKED ISSUE #3957: Vasp Incar check_params() can't handle Union types
### Python version

3.11.9

### Pymatgen version

2024.6.10

### Operating system version

Ubuntu 22.04.4 LTS

### Current behavior

Error Message:
```
  File "/home/wladerer/github/abvio/tests/test_incar.py", line 356, in test_valid_model
    incar.check_params()
  File "/home/wladerer/.venv/base/lib/python3.11/site-packages/pymatgen/io/vasp/inputs.py", line 979, in check_params
    if param_type is not None and not isinstance(val, eval(param_type)):
                                                      ^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'Union' is not defined

Ran 1 test in 0.002s

FAILED (errors=1)
Finished running tests!
```

Incar.check_params() does not seem to handle Union types

Something strange is happening here where `isinstance(eval(param_type))`
```python
# Check value and its type
            param_type = incar_params[tag].get("type")
            allowed_values = incar_params[tag].get("values")

            if param_type is not None and not isinstance(val, eval(param_type)):
                warnings.warn(f"{tag}: {val} is not a {param_type}", BadIncarWarning, stacklevel=2)
```
Is unaware of the typing module Union type. I can see that it is not imported, so when I update the imports
```python
from typing import TYPE_CHECKING, Any, Literal, Union
```
the error goes away. 

```
test_valid_model (test_incar.TestIncarModel.test_valid_model) ... EDIFF tag type: <class 'str'> tag value <class 'float'>
ISMEAR tag type: <class 'str'> tag value <class 'int'>
LREAL tag type: <class 'str'> tag value <class 'bool'>
EDIFFG tag type: <class 'str'> tag value <class 'float'>
LWAVE tag type: <class 'str'> tag value <class 'bool'>
MAGMOM tag type: <class 'str'> tag value <class 'list'>
ok

Ran 1 test in 0.001s

OK
Finished running tests!
```

### Expected Behavior

No output 

### Minimal example

```Python
import unittest

from pymatgen.io.vasp import Incar 


class TestCheckIncarParams(unittest.TestCase):

    def test_dict_with_magmoms(self):

        incar_dict = {
            "SYSTEM": "Test",
            "ENCUT": 400,
            "ISMEAR": 0,
            "SIGMA": 0.1,
            "IBRION": 2,
            "NSW": 0,
            "LCHARG": False,
            "LWAVE": False,
            "LORBIT": 11,
            "LREAL": "Auto",
            "ISYM": 0,
            "MAGMOM": [0.6, 0.6, 0.2],
        }

        incar = Incar.from_dict(incar_dict)
        incar.check_params()
```

Can be run using the minimal example

DanielYang59:
### Summary

- Fix Incar `check_params` for `Union` type, to fix #3957, changed `incar_parameters.json` of Union type to use `isinstance` syntax directly. Otherwise Python 3.9 Windows don't support type checking with `isinstance(data, Union[str, float])`.

    **Would appreciate it if anyone has smarted way to resolve this!**

- Bump `palettable` version to remove dependency of `setuptools` https://github.com/jiffyclub/palettable/pull/46

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,10 +75,9 @@ jobs:
       - name: Install pymatgen and dependencies
         run: |
           micromamba activate pmg
-          # TODO1: remove temporary fix. added since uv install torch is flaky.
+          # TODO remove temporary fix. added since uv install torch is flaky.
           # track https://github.com/astral-sh/uv/issues/1921 for resolution
-          # TODO2: remove touch version pin upon new matgl release
-          pip install torch==2.2.1 --upgrade
+          pip install torch --upgrade
 
           uv pip install numpy cython
           uv pip install --editable '.[${{ matrix.config.extras }}]' --resolution=${{ matrix.config.resolution }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,9 +75,10 @@ jobs:
       - name: Install pymatgen and dependencies
         run: |
           micromamba activate pmg
-          # TODO remove temporary fix. added since uv install torch is flaky.
+          # TODO1: remove temporary fix. added since uv install torch is flaky.
           # track https://github.com/astral-sh/uv/issues/1921 for resolution
-          pip install torch --upgrade
+          # TODO2: remove touch version pin upon new matgl release
+          pip install torch==2.2.1 --upgrade
 
           uv pip install numpy cython
           uv pip install --editable '.[${{ matrix.config.extras }}]' --resolution=${{ matrix.config.resolution }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ classifiers = [
 ]
 dependencies = [
     "matplotlib>=3.8",
-    "monty>=2024.5.24",
+    "monty>=2024.7.29",
     "networkx>=2.2",
     'numpy>=1.25.0,<2.0 ; platform_system == "Windows"',
     'numpy>=1.25.0 ; platform_system != "Windows"',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ dependencies = [
     "networkx>=2.2",
     'numpy>=1.25.0,<2.0 ; platform_system == "Windows"',
     'numpy>=1.25.0 ; platform_system != "Windows"',
-    "palettable>=3.1.1",
+    "palettable>=3.3.3",
     "pandas>=2",
     "plotly>=4.5.0",
     "pybtex>=0.24.0",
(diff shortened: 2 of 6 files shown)


wladerer:
This might be a silly question, but is changing the "type" parameter in the json file possible? The following evaluates to typing.Union
 
```python
type(eval('bool | str'))
```

DanielYang59:
That's actually a genius solution IMO, now that using of `eval` is not risky in this scenario. 

However the only pitfall being that the [alias of `Union[x, y]` as `x | y` is introduced in Python 3.10](https://peps.python.org/pep-0604/), and as the min supported version for `pymatgen` is 3.9 at this moment, we still cannot implement that change. But still I think it's a cleaner solution than mine, perhaps we could import `Union` for now until we move to Python 3.10+.

DanielYang59:
Not sure why some errors (seemingly unrelated to this PR) popped up, fixed two of them but still get one left:
> src/pymatgen/util/provenance.py:49: in is_valid_bibtex
    parser = bibtex.Parser()
E   AttributeError: 'NoneType' object has no attribute 'Parser'

Perhaps owing to the import error: https://github.com/materialsproject/pymatgen/blob/720fdf86539236c737f12250ae71d68e3872c616/src/pymatgen/util/provenance.py#L15-L19

But it seems `pytex` has been correctly installed in CI:
>  + pybtex==0.24.0

DanielYang59:
@janosh Can you please review this? Thanks. I believe the CI failing is unrelated to this PR.

janosh:
> I believe the CI failing is unrelated to this PR.

i think the issue is due to installing the latest `torch` in `test.yml` while  `matGL` requires `torch=2.2.1` or older

https://github.com/materialsproject/pymatgen/blob/633e7bd7700fc96b407d882cce2b473a38a97c67/.github/workflows/test.yml#L80

can you try pinning to torch `2.2.1` in CI with a TODO to remove pin later?

DanielYang59:
Yes sure. Let me have a try.

DanielYang59:
Looks like `matgl` has removed the `torch` version limitation? https://github.com/materialsvirtuallab/matgl/commit/d0bf202ad4316d058e8e64a08f5c4fccb53aa0d9

janosh:
afaik `dgl` isn't compatible with `torch==2.3` yet, let alone 2.4 so don't think the `matgl` up pin should have been removed. i was advocating for an down pin in another thread. either way, let's revert the temp `torch==2.2.1` pin here and just skip the matGL tests for now to keep things moving

DanielYang59:
Sure! Thanks for the input, and I know nothing about `matgl` :)

janosh (approved):
thanks a lot @DanielYang59! 👍

DanielYang59:
Thanks for reviewing!

PR #3923, opened 2024-07-16
Migrate `spglib` to new `SpglibDataset` format with version 2.5.0

DanielYang59:
### Summary

- Migrate `spglib` to new data format as pointed out by @kaueltzen, to fix #3922
- Add types for `SpacegroupAnalyzer` and `PointGroupAnalyzer`

PR #3964, opened 2024-08-01
switch to attr access interface for transformation matrix

tsmathis:
an attr access interface change was missed in Tensor when spglib was upgraded to >=2.5.0 (#3923)

spglib throws the following deprecation warning currently:
`DeprecationWarning: dict interface (SpglibDataset['transformation_matrix']) is deprecated.Use attribute interface ({self.__class__.__name__}.{key}) instead`


janosh (approved):
thanks @tsmathis! 👍

Issue #3969, opened 2024-08-02
TestStructureGraph.test_draw TypeError: unhashable type: 'list'


drew-parsons:
There is an error in python3.12 in the analysis test_graph at
https://github.com/materialsproject/pymatgen/blob/940eb60cbc2848530b2560f75d56298a3c3f714c/tests/analysis/test_graphs.py#L376C9-L376C76

Output running tests is
```py
_________________________ TestStructureGraph.test_draw _________________________

self = <tests.analysis.test_graphs.TestStructureGraph testMethod=test_draw>

    @pytest.mark.skipif(
        pygraphviz is None or not (which("neato") and which("fdp")),
        reason="graphviz executables not present",
    )
    def test_draw(self):
        # draw MoS2 graph
        self.mos2_sg.draw_graph_to_file(f"{self.tmp_path}/MoS2_single.pdf", image_labels=True, hide_image_edges=False)
        mos2_sg = self.mos2_sg * (9, 9, 1)
        mos2_sg.draw_graph_to_file(f"{self.tmp_path}/MoS2.pdf", algo="neato")

        # draw MoS2 graph that's been successively multiplied
        mos2_sg_2 = self.mos2_sg * (3, 3, 1)
        mos2_sg_2 = mos2_sg_2 * (3, 3, 1)
        mos2_sg_2.draw_graph_to_file(f"{self.tmp_path}/MoS2_twice_mul.pdf", algo="neato", hide_image_edges=True)

        # draw MoS2 graph that's generated from a pre-multiplied Structure
        mos2_sg_premul = StructureGraph.from_local_env_strategy(self.structure * (3, 3, 1), MinimumDistanceNN())
        mos2_sg_premul.draw_graph_to_file(f"{self.tmp_path}/MoS2_premul.pdf", algo="neato", hide_image_edges=True)

        # draw graph for a square lattice
        self.square_sg.draw_graph_to_file(f"{self.tmp_path}/square_single.pdf", hide_image_edges=False)
        square_sg = self.square_sg * (5, 5, 1)
        square_sg.draw_graph_to_file(f"{self.tmp_path}/square.pdf", algo="neato", image_labels=True, node_labels=False)

        # draw graph for a body-centered square lattice
        self.bc_square_sg.draw_graph_to_file(f"{self.tmp_path}/bc_square_single.pdf", hide_image_edges=False)
        bc_square_sg = self.bc_square_sg * (9, 9, 1)
[...]

PR #3972, opened 2024-08-05
test_graph: add filename only to pdf list

drew-parsons:
The pdfs list must be constructed from filenames only without other path components, i.e. use last entry from split()

## Summary

Major changes:

- fix: fix list of pdf files in test_draw (test_graph)

## Checklist

- [x ] Tests added for new features/fixes (test already present).

Closes: #3969


DanielYang59:
Thanks for the fix. I'm still looking into the issue, as I just installed `pygraphviz` in CI and I was having a different error for that particular test for some reason (both in CI and locally) https://github.com/materialsproject/pymatgen/actions/runs/10246364887/job/28343300325?pr=3786.

janosh (approved):
thanks @drew-parsons!

PR #3933, opened 2024-07-17
Parallel Joblib Process Entries

CompRhys:
## Summary

Adds the option to use joblib to apply corrections in parallel, this depended on https://github.com/materialsvirtuallab/monty/pull/691 which was linked to https://github.com/materialsproject/pymatgen/issues/3898. Opening this PR incase maintainers think it's a useful addition to the code.

## Checklist

- [x] Google format doc strings added. Check with `ruff`.
- [x] Type annotations included. Check with `mypy`.
- [x] Tests added for new features/fixes.
- [ ] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

PR #3977, opened 2024-08-08
Fix `mypy` errors on master branch

### LINKED ISSUE #3976: Some type annotations errors when trying to merge master into my local development branch.
Today, I tried to merge pymatgen master into my local development branch, but met the following type annotations errors:

```shell
(datasci) werner@x13dai-t:~/Public/repo/github.com/materialsproject/pymatgen.git$ git commit -am 'merge master'
[INFO] Checking merge-conflict files only.
check yaml...........................................(no files to check)Skipped
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/pymatgen/symmetry/analyzer.py:1543: error: Incompatible types in assignment (expression has type "Collection[Any]", variable has type "Molecule")  [assignment]
src/pymatgen/io/vasp/sets.py:833: error: Argument 1 to "get_ir_reciprocal_mesh" of "SpacegroupAnalyzer" has incompatible type "tuple[float, float, float] | tuple[int]"; expected "tuple[int, int, int]"  [arg-type]
src/pymatgen/io/vasp/sets.py:867: error: Argument 1 to "get_ir_reciprocal_mesh" of "SpacegroupAnalyzer" has incompatible type "tuple[float, float, float] | tuple[int]"; expected "tuple[int, int, int]"  [arg-type]
src/pymatgen/entries/compatibility.py:558: error: Incompatible types in assignment (expression has type "ComputedEntry | None", variable has type "ComputedEntry")  [assignment]
src/pymatgen/entries/compatibility.py:614: error: Incompatible return value type (got "tuple[ComputedEntry | ComputedStructureEntry, bool]", expected "ComputedEntry | None")  [return-value]
src/pymatgen/entries/compatibility.py:1536: error: List item 0 has incompatible type "Any | list[ComputedEntry | ComputedStructureEntry]"; expected "ComputedEntry"  [list-item]
Found 6 errors in 3 files (checked 2 source files)

cython-lint..........................................(no files to check)Skipped
double-quote Cython strings..........................(no files to check)Skipped
markdownlint.........................................(no files to check)Skipped
nbstripout...........................................(no files to check)Skipped
pyright..................................................................Passed
```

The corresponding problematic code lines are shown below: 

```python
werner@x13dai-t:~/Public/repo/github.com/materialsproject/pymatgen.git$ sed -n '1543p' src/pymatgen/symmetry/analyzer.py
        new = eq["sym_mol"]
werner@x13dai-t:~/Public/repo/github.com/materialsproject/pymatgen.git$ sed -n '833p;867p' src/pymatgen/io/vasp/sets.py
                mesh = sga.get_ir_reciprocal_mesh(base_kpoints.kpts[0])
            mesh = sga.get_ir_reciprocal_mesh(zero_weighted_kpoints.kpts[0])
werner@x13dai-t:~/Public/repo/github.com/materialsproject/pymatgen.git$ sed -n '558p;614p;1536p' src/pymatgen/entries/compatibility.py
        entry = self._process_entry_inplace(entry, **kwargs)
        return entry, ignore_entry
            return [entries]
```

I tried some possible modifications, but none of them could pass the pre-commit check. I would be very grateful if you have any suggestions for fixes.

Regards,
Zhao

DanielYang59:
### Summary

- Fix `mypy` errors on master branch, to fix #3976
- Docstring tweaks

### DIFF
--- a/src/pymatgen/entries/compatibility.py
+++ b/src/pymatgen/entries/compatibility.py
@@ -1565,7 +1565,6 @@ class MaterialsProjectAqueousCompatibility(Compatibility):
         # Pre-process entries with the given solid compatibility class
         if self.solid_compat:
             entries = self.solid_compat.process_entries(entries, clean=True, inplace=inplace, n_workers=n_workers)
-            return [entries]  # DEBUG: incompatible return type
 
         # when processing single entries, all H2 polymorphs will get assigned the
         # same energy
--- a/src/pymatgen/entries/compatibility.py
+++ b/src/pymatgen/entries/compatibility.py
@@ -35,9 +35,9 @@ from pymatgen.util.joblib import set_python_warnings, tqdm_joblib
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
-    from typing import Literal
+    from typing import Any, Literal
 
-    from pymatgen.util.typing import CompositionLike
+    from pymatgen.util.typing import CompositionLike, PathLike
 
 
 __author__ = "Amanda Wang, Ryan Kingsbury, Shyue Ping Ong, Anubhav Jain, Stephen Dacek, Sai Jayaraman"
@@ -55,7 +55,7 @@ MP_COMPAT_CONFIG = loadfn(f"{MODULE_DIR}/MPCompatibility.yaml")
 # This was compiled by cross-referencing structures in Materials Project from exp_compounds.json.gz
 # used in the fitting of the MP2020 correction scheme, and applying the BVAnalyzer algorithm to
 # determine oxidation state. O and S are not included since these are treated separately.
-MP2020_ANION_OXIDATION_STATE_RANGES = {
+MP2020_ANION_OXIDATION_STATE_RANGES: dict[str, tuple[int, int]] = {
     "Br": (-1, -1),
     "Cl": (-1, -1),
     "F": (-1, -1),
@@ -144,7 +144,12 @@ class PotcarCorrection(Correction):
     'PAW_PBE O 08Apr2002'].
     """
 
-    def __init__(self, input_set: type[VaspInputSet], check_potcar: bool = True, check_hash: bool = False) -> None:
+    def __init__(
+        self,
+        input_set: type[VaspInputSet],
+        check_potcar: bool = True,
+        check_hash: bool = False,
+    ) -> None:
         """
         Args:
             input_set (InputSet): object used to generate the runs (used to check
@@ -170,6 +175,9 @@ class PotcarCorrection(Correction):
         self.input_set = input_set
         self.check_hash = check_hash
         self.check_potcar = check_potcar
+
+    def __str__(self) -> str:
+        return f"{self.input_set.__name__} Potcar Correction"
 
     def get_correction(self, entry: AnyComputedEntry) -> ufloat:
         """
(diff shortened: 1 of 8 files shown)


src/pymatgen/entries/compatibility.py
```diff
@@ -1522,18 +1554,18 @@ def process_entries(
...
+        # If not inplace, process entries on a copy
         if not inplace:
             entries = copy.deepcopy(entries)
 
-        # pre-process entries with the given solid compatibility class
+        # Pre-process entries with the given solid compatibility class
         if self.solid_compat:
             entries = self.solid_compat.process_entries(entries, clean=True, inplace=inplace, n_workers=n_workers)
-            return [entries]
+            return [entries]  # DEBUG: incompatible return type
```
DanielYang59, line 1568: Incompatible return type, `self.solid_compat.process_entries` returns `list[AnyComputedEntry]` and this method should return `list[AnyComputedEntry]` instead of `list[list[AnyComputedEntry]]`

janosh: maybe wrapping `entries` as `[entries]` was added before the above check?

```py
if isinstance(entries, ComputedEntry):
    entries = [entries]
```

do any tests break when you just `return entries`?

janosh: just noticed that it says "preprocess entries" in the comment above so probably wasn't meant to return at all on this line. tests pass without it. can you confirm @rkingsbury?

janosh (approved):
thanks @DanielYang59! 👍

DanielYang59:
Thanks for reviewing. There was one type error for `io.cp2k` I haven't got to fix, it's it failing on your side?

src/pymatgen/entries/compatibility.py
rkingsbury: Indeed, there should not have been a `return` statement on that line. I'm not sure when that was introduced (can't tell from the git blame) but that was actually a pretty serious bug. It was effectively preventing the aqueous compatibility scheme from doing anything when given default arguments.

DanielYang59: Looks like it was added two weeks ago in https://github.com/materialsproject/pymatgen/pull/3933 in 5c5b74d4c7b26967d8dc766eefb882496e9091e6. Great intuition @janosh and thanks a lot for the confirmation

janosh: @rkingsbury can you think of a good test to add that would have caught this breaking change?

CompRhys: > Indeed, there should not have been a `return` statement on that line. I'm not sure when that was introduced (can't tell from the git blame) but that was actually a pretty serious bug. It was effectively preventing the aqueous compatibility scheme from doing anything when given default arguments.

I have no recollection of why I would have added this return statement, sadly I think it's likely that I added it to do some sort of manual checking that it worked correctly when run with joblib and then failed to remove it. Sorry.

janosh: for once, `mypy` proves its worth 😄

in general, can't wait for materials science codes to be fully statically typed!

rkingsbury: > @rkingsbury can you think of a good test to add that would have caught this breaking change?

What I would have expected to happen is with default arguments, the `ComputedEntry` would be returned with the normal solid energy corrections (`MaterialsProject2020Compatibility`) but no aqueous corrections. I guess existing tests didn't catch it because many (even most) solids won't receive any Aqueous corrections when they are processed. It's only a few materials like diatomic gases and materials containing OH that would get the adjustment.

PR #3978, opened 2024-08-08
Check Inputs to Trajectory.

CompRhys:
## Summary

Checks the inputs to Trajectory following an issue encountered where `get_positions()` method can return more positions than sites depending on filters. 

## Checklist


janosh (approved):
thanks @CompRhys! 👍

src/pymatgen/core/trajectory.py
```diff
@@ -73,48 +73,61 @@ def __init__(
...
+                i-1), but not relative to the base_position.
             base_positions: shape (N, 3). The starting positions of all atoms in the
                 trajectory. Used to reconstruct positions when converting from
                 displacements to positions. Only needs to be specified if
-                `coords_are_displacement=True`. Defaults to the first index of
-                `coords` when `coords_are_displacement=False`.
+                coords_are_displacement=True. Defaults to the first index of
+                coords when coords_are_displacement=False.
         """
+        coords = np.asarray(coords)
```
janosh, line 112: for later reference, considered using `np.atleast3d` to normalize user input with insufficient dimensions but unfortunately when passed a 2d array, `atleast3d` inserts a singleton dim at the end rather than the start so probably more hassle than it's worth

PR #3979, opened 2024-08-08
Assign frame properties to molecule/structure when indexing trajectory

CompRhys:
## Summary

When indexing a trajectory the site properties would be assigned to the structure/molecule but the frame properties would not be propagated down to the structure/molecule in question.

## Checklist


janosh (approved):
good catch, thanks @CompRhys! 👍

PR #3981, opened 2024-08-08
Fix 98 cases of `ruff` `PLR6104`

janosh:
`ruff check --preview --fix --select PLR6104 --unsafe-fixes`


janosh:
closing as even more breaking fixes than expected 😅

PR #3984, opened 2024-08-09
Fix mypy errors for `io.cp2k`

DanielYang59:
### Summary

- Fix mypy errors for `io.cp2k`
- `mypy` now ignore `override` error globally

### DIFF
--- a/src/pymatgen/io/cp2k/inputs.py
+++ b/src/pymatgen/io/cp2k/inputs.py
@@ -1998,7 +1998,7 @@ class Kpoints(Section):
         )
 
     @classmethod
-    def from_kpoints(cls, kpoints: VaspKpoints, structure=None) -> Self:
+    def from_kpoints(cls, kpoints: VaspKpoints, structure: Structure | None = None) -> Self:
         """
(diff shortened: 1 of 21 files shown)


src/pymatgen/io/cp2k/inputs.py
```diff
@@ -2008,11 +2008,11 @@ def from_kpoints(cls, kpoints: VaspKpoints, structure: Structure | None = None)
                 do this automatically without spglib present at execution time.
         """
         kpts = kpoints.kpts
-        weights = kpoints.kpts_weights
+        weights: Sequence[float] | None = kpoints.kpts_weights
 
         if kpoints.style == KpointsSupportedModes.Monkhorst:
-            kpt: Kpoint | float = kpts[0]
-            x, y, z = (kpt, kpt, kpt) if isinstance(kpt, (int, float)) else kpt
```
DanielYang59: Might need additional check on 578687a085f9867a6c6f900b763e7ebbf7329b62. Now `Kpoint` cannot be a `float | int`, but `tuple[float, float, float] | tuple[int]` only

janosh (approved):
always nice to see those `type: ignore`s go! 👍

PR #3986, opened 2024-08-09
Minor change for step_size in CubicSupercellTransformation.apply_transformation 

QuantumChemist:
Recently, @JaGeo and I had made some changes to the CubicSupercellTransformation.apply_transformation function in https://github.com/materialsproject/pymatgen/pull/3938.

I would like to set the step_size more coarse for the "cubic" supercells without the force_90_degree constraints as the previous step_size has led to unreasonable long run times for me.

PR #3987, opened 2024-08-09
Revert "Minor change for step_size in CubicSupercellTransformation.apply_transformation "

JaGeo:
Reverts materialsproject/pymatgen#3986


JaGeo:
Cannot revert at the moment, as the connection to COD is failing. Will do this, as soon as possible.

PR #3992, opened 2024-08-10
Explicitly use `int64` in Numpy/cython code to avoid OS inconsistency

### LINKED ISSUE #3990: Incompatability with numpy <2 on Windows for `Lattice.find_points_in_spheres`
### Python version

3.11

### Pymatgen version

2024.8.9

### Operating system version

Windows

### Current behavior

As of pyamtgen 2024.8.8, there is an incompatibility with `Lattice.get_points_in_sphere` on Windows when using numpy < 2. Everything seems to run fine on numpy 2.0.1 and/or non-Windows machines.

```
pip install pymatgen==2024.8.9 numpy==1.26.4
```

```python
structure = Structure(
  lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],
  species=["Mg", "O"],
  )
structure.lattice.get_points_in_sphere(structure.frac_coords,(0,0,0),1)
```

Traceback:

```python
----> 1 structure.lattice.get_points_in_sphere(structure.frac_coords,(0,0,0),1)

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\lattice.py:1387, in Lattice.get_points_in_sph)
   1384 pbc = np.ascontiguousarray(self.pbc, dtype=int)
   1385 center_coords = np.ascontiguousarray([center], dtype=float)
-> 1387 _, indices, images, distances = find_points_in_spheres(
   1388     all_coords=cart_coords, center_coords=center_coords, r=float(r), pbc=pbc, lattice=latt_matrix, tol=1e-8
   1389 )
   1390 if len(indices) < 1:
   1391     # Return empty np.array (not list or tuple) to ensure consistent return type
   1392     # whether sphere contains points or not
   1393     return np.array([]) if zip_results else tuple(np.array([]) for _ in range(4))

File src\\pymatgen\\optimization\\neighbors.pyx:48, in pymatgen.optimization.neighbors.find_points_in_sp)

ValueError: Buffer dtype mismatch, expected 'const int64_t' but got 'long'
```

This influences many other functions in Pymatgen, such as the example below:

```python
from pymatgen.core.surface import generate_all_slabs
structure = Structure(
  lattice=[[0, 2.13, 2.13], [2.13, 0, 2.13], [2.13, 2.13, 0]],
  species=["Mg", "O"],
  )
generate_all_slabs(structure,1,10,10)
```

Traceback:

```python
File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\surface.py:1650, in generate_all_slabs(structure, max_index, min_slab_size, min_vacuum_size, bonds, tol, ftol, max_broken_bonds, lll_reduce, center_slab, primitive, max_normal_search, symmetrize, repair, include_reconstructions, in_unit_planes)
   1638 for miller in get_symmetrically_distinct_miller_indices(structure, max_index):
   1639     gen = SlabGenerator(
   1640         structure,
   1641         miller,
   (...)
   1648         in_unit_planes=in_unit_planes,
   1649     )
-> 1650     slabs = gen.get_slabs(
   1651         bonds=bonds,
   1652         tol=tol,
   1653         ftol=ftol,
   1654         symmetrize=symmetrize,
   1655         max_broken_bonds=max_broken_bonds,
   1656         repair=repair,
   1657     )
   1659     if len(slabs) > 0:
   1660         logger.debug(f"{miller} has {len(slabs)} slabs... ")

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\surface.py:1348, in SlabGenerator.get_slabs(self, bonds, ftol, tol, max_broken_bonds, symmetrize, repair, ztol)
   1345 matcher = StructureMatcher(ltol=tol, stol=tol, primitive_cell=False, scale=False)
   1347 final_slabs: list[Slab] = []
-> 1348 for group in matcher.group_structures(slabs):
   1349     # For each unique slab, symmetrize the
   1350     # surfaces by removing sites from the bottom
   1351     if symmetrize:
   1352         sym_slabs = self.nonstoichiometric_symmetrized_slab(group[0])

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\analysis\structure_matcher.py:795, in StructureMatcher.group_structures(self, s_list, anonymous)
    793 s_list = self._process_species(s_list)
    794 # Prepare reduced structures beforehand
--> 795 s_list = [self._get_reduced_structure(s, self._primitive_cell, niggli=True) for s in s_list]
    797 # Use structure hash to pre-group structures
    798 if anonymous:

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\analysis\structure_matcher.py:795, in <listcomp>(.0)
    793 s_list = self._process_species(s_list)
    794 # Prepare reduced structures beforehand
--> 795 s_list = [self._get_reduced_structure(s, self._primitive_cell, niggli=True) for s in s_list]
    797 # Use structure hash to pre-group structures
    798 if anonymous:

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\analysis\structure_matcher.py:947, in StructureMatcher._get_reduced_structure(cls, struct, primitive_cell, niggli)
    945 reduced = struct.copy()
    946 if niggli:
--> 947     reduced = reduced.get_reduced_structure(reduction_algo="niggli")
    948 if primitive_cell:
    949     reduced = reduced.get_primitive_structure()

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\structure.py:2252, in IStructure.get_reduced_structure(self, reduction_algo)
   2242 """Get a reduced structure.
   2243
   2244 Args:
   (...)
   2249     Structure: Niggli- or LLL-reduced structure.
   2250 """
   2251 if reduction_algo == "niggli":
-> 2252     reduced_latt = self._lattice.get_niggli_reduced_lattice()
   2253 elif reduction_algo == "LLL":
   2254     reduced_latt = self._lattice.get_lll_reduced_lattice()

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\lattice.py:1227, in Lattice.get_niggli_reduced_lattice(self, tol)
   1224 gamma = math.acos(Y / 2 / a / b) / math.pi * 180
   1225 lattice = type(self).from_parameters(a, b, c, alpha, beta, gamma)
-> 1227 mapped = self.find_mapping(lattice, e, skip_rotation_matrix=True)
   1228 if mapped is not None:
   1229     if np.linalg.det(mapped[0].matrix) > 0:

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\lattice.py:1010, in Lattice.find_mapping(self, other_lattice, ltol, atol, skip_rotation_matrix)
    974 def find_mapping(
    975     self,
    976     other_lattice: Self,
   (...)
    979     skip_rotation_matrix: bool = False,
    980 ) -> tuple[Lattice, np.ndarray | None, np.ndarray] | None:
    981     """Find a mapping between current lattice and another lattice. There
    982     are an infinite number of choices of basis vectors for two entirely
    983     equivalent lattices. This method returns a mapping that maps
   (...)
   1008         None is returned if no matches are found.
   1009     """
-> 1010     return next(self.find_all_mappings(other_lattice, ltol, atol, skip_rotation_matrix), None)

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\lattice.py:947, in Lattice.find_all_mappings(self, other_lattice, ltol, atol, skip_rotation_matrix)
    944 lengths = other_lattice.lengths
    945 alpha, beta, gamma = other_lattice.angles
--> 947 frac, dist, _, _ = self.get_points_in_sphere(
    948     [[0, 0, 0]], [0, 0, 0], max(lengths) * (1 + ltol), zip_results=False
    949 )
    950 cart = self.get_cartesian_coords(frac)
    951 # This can't be broadcast because they're different lengths

File ~\miniconda\envs\test\Lib\site-packages\pymatgen\core\lattice.py:1387, in Lattice.get_points_in_sphere(self, frac_points, center, r, zip_results)
   1384 pbc = np.ascontiguousarray(self.pbc, dtype=int)
   1385 center_coords = np.ascontiguousarray([center], dtype=float)
-> 1387 _, indices, images, distances = find_points_in_spheres(
   1388     all_coords=cart_coords, center_coords=center_coords, r=float(r), pbc=pbc, lattice=latt_matrix, tol=1e-8
   1389 )
   1390 if len(indices) < 1:
   1391     # Return empty np.array (not list or tuple) to ensure consistent return type
   1392     # whether sphere contains points or not
   1393     return np.array([]) if zip_results else tuple(np.array([]) for _ in range(4))

File src\\pymatgen\\optimization\\neighbors.pyx:48, in pymatgen.optimization.neighbors.find_points_in_spheres()

ValueError: Buffer dtype mismatch, expected 'const int64_t' but got 'long
```

DanielYang59:
### Summary

- Explicitly use `int64` in cython code to avoid OS inconsistency ([default int type changed in Windows](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html)) as follow up PR of #3894, to fix #3990
- [x] Run unit test on Windows with `NumPy < 2`
- [x] Revert NumPy<2 pin after tests pass

### Cython code need attention

-  `src/pymatgen/optimization/linear_assignment.pyx`
- `src/pymatgen/optimization/neighbors.pyx`
- `src/pymatgen/util/coord_cython.pyx`

### Pin some threads from `NumPy` for reference:
- https://github.com/numpy/numpy/pull/24224
- [NumPy roadmap- Change default int on windows to int64](https://github.com/orgs/numpy/projects/9/views/1?pane=issue&itemId=23800241)
- https://github.com/numpy/numpy/issues/9464


DanielYang59:
@njzjz Perhaps you could give me some advice or comment here because I'm not a cython expert? Thanks a ton in advance.

Looks like quite some code is broken when running on Windows system with NumPy < 2, because with 862c454bfd2e0428b27fdafe9e0a7d81654005b9 integer is explicitly declared as `int64` independent of the platform, however on Windows with NumPy < 2 arrays might still be `int32` (`long`) by default.

It looks like we should update all cython-related code to internally cast type as `int64` (instead of reverting the `long` -> `int64` change to a `long` -> `intp` change)?

tests/core/test_structure.py
```diff
@@ -1688,6 +1688,7 @@ def test_calculate_ase(self):
         assert not hasattr(calculator, "dynamics")
         assert self.cu_structure == struct_copy, "original structure was modified"
 
+    @pytest.mark.skip(reason="chgnet is failing with Numpy 1, see #3992")
```
DanielYang59, line 1691: Need a closer look, [`chgnet` tests are failing for NumPy 1 as well](https://github.com/materialsproject/pymatgen/actions/runs/10330613130/job/28599854900?pr=3992), I guess all ints are now `int64` in pymatgen, which might not be the case for `chgnet`'s cython code.

njzjz:
Fused types may help: http://docs.cython.org/en/stable/src/userguide/fusedtypes.html
https://stackoverflow.com/a/28112966/9567349

njzjz:
Also, it may be useful to run tests against both numpy 1.x and 2.x.

DanielYang59:
Thanks a lot for the input, also pin some threads from `NumPy` to guide our migration:
- https://github.com/numpy/numpy/pull/24224
- [NumPy roadmap- Change default int on windows to int64](https://github.com/orgs/numpy/projects/9/views/1?pane=issue&itemId=23800241)

DanielYang59:
@njzjz I decide perhaps still explicitly use `int64` to ensure consistency (not sure if that's the best solution, free feel to correct me). Would appreciate it if you have any comment. Thanks a ton!

@janosh I'm not a cython expert (important 😅 ) so please review all changes very carefully :) 

Added a `NumPy 1` workflow for windows as [recommended by NumPy doc](https://numpy.org/devdocs/dev/depending_on_numpy.html#build-time-dependency), but still not sure if it's worth it because I'm not expecting no new issues to pop up once we fixes these.

janosh (commented):
are you sure we need to change `np.array(dtype=int->np.int64)` across the whole code base? reading the numpy docs you linked, it sounds like the potential for platform inconsistency is specific to cython code? so might be enough to only `int->np.int64` in `.pyx` files?

DanielYang59:
> it sounds like the potential for platform inconsistency is specific to cython code?

Sorry perhaps I should tweak the title, current one might be misleading. After NumPy 2.x, the default NumPy int type for 64-bit windows system [changed from `int32` to `int64`](https://numpy.org/doc/stable/release/2.0.0-notes.html) (not just cython, sorry).

> The default integer type on Windows is now int64 rather than int32, matching the behavior on other platforms

Meaning if we pass a numpy array as `np.array(data, dtype=int)` into a cython function, the latter now would explicitly expect `int64`, and the dtype could be `int32` for the former in Numpy 1.x windows machines (therefore those CI failures). So I don't think just changing just the types in cython code suffice.

Meanwhile I think this is better than a fused type of `int32` and `int64`, because if we explicitly use `int64`, we then could be confident our code would generate consistent results independent of OS? Because personally I think `int64` might be the "mainstream" way to go (any NP2 system or Linux system)?

tests/core/test_structure.py
janosh: @AegisIK can you comment on this? is the fix on the `chgnet` side to ensure all ints in the atom/bond graph construction are `np.int64`?

pyproject.toml
```diff
@@ -115,6 +115,7 @@ optional = [
     # "openbabel>=3.1.1; platform_system=='Linux'",
 ]
 numba = ["numba>=0.55"]
+numpy1 = ["numpy>=1.25.0,<2"]  # Test NP1 on Windows (quite buggy ATM)
```
janosh, line 118: nice, just to check did you look at the CI dep install logs? just want to be sure it's actually installing `numpy<2`

janosh (approved):
thanks a lot for looking into this @DanielYang59! hugely appreciated! 👍

pyproject.toml
DanielYang59: [Positive](https://github.com/materialsproject/pymatgen/actions/runs/10537208035/job/29198324593?pr=3992):
>  + numpy==1.26.4

DanielYang59:
No problem. Also appreciate the input from @njzjz 

Perhaps we should add a note in next release informing downstream packages of such potentially breaking change (`chgnet` for example)?

janosh:
sure thing! would you like to draft a note?

DanielYang59:
Sure, and feel free to polish it as you like.

### [Breaking] Default NumPy/Cython integer type changed to `int64` on Windows 

**Change**: With NumPy 2.x, [the default integer type changed to `int64` on Windows 64-bit system](https://numpy.org/doc/stable/release/2.0.0-notes.html). Consequently, `pymatgen` now explicitly uses `np.int64` for all NumPy and Cython code instead of the platform-dependent `int` type.

**Recommendation**: Please explicitly declare `dtype=np.int64` when initializing a NumPy array if possible. You might also need to [test NumPy 1.x on Windows in CI pipelines](https://numpy.org/devdocs/dev/depending_on_numpy.html#build-time-dependency).

tests/core/test_structure.py
DanielYang59: Look like this test is passing for latest version of chgnet now!
```
 (venv312) venvyang@Yang-MacBook pymatgen % venv312/bin/pytest tests/core/test_structure.py::TestStructure::test_calculate_ase
.                                                                                                                                             [100%]
=============================================================== slowest 30 durations ================================================================
0.07s call     tests/core/test_structure.py::TestStructure::test_calculate_ase

(2 durations < 0.005s hidden.  Use -vv to show these durations.)
1 passed in 1.20s

(venv312) venvyang@Yang-MacBook pymatgen % pip freeze| grep chgnet
chgnet==0.4.0
```

PR #3994, opened 2024-08-12
Set kpoints in from_str method as integer in auto Gamma and Monkhorst modes

### LINKED ISSUE #3993: Kpoints Sampling Saved as Float When Writing Kpoints file
### Python version

Python 3.10.14

### Pymatgen version

Version: 2024.7.18

### Operating system version

Red Hat Enterprise Linux Ver 8.6

### Current behavior

When using the `Kpoints` class from `pymatgen.io.vasp.inputs` to read a Gamma-centered KPOINTS file and then write it back, the fourth line of the KPOINTS file is converted to a float string. This causes VASP5.4.4 calculations to fail when using this file. Specifically, in the `from_str` method of the Kpoints class, the fourth line of the KPOINTS file is cast to `kpts` as `Sequence[tuple[float, float, float]]`. The `kpts` attribute is then handled as a sequence of floats throughout the process, and it is written back as floats.

### Expected Behavior

When writting a Gamma-centered KPOINTS file, the number of K-point divisions should  be cast to a int, or it should be explicitly cast to an integer when writing the file if the style is Gamma-centered or  Monkhorst-Pack. 
Initially, the `kpts` attribute of the Kpoints class was handled as `list[int]` in `from_str` method, but it was changed to `Sequence[tuple[float, float, float]]` in #3758. While this change itself is not problematic, it would be more convenient if `kpts` were an integer sequence for Gamma-centered or Monkhorst-Pack.

### Minimal example

```Python
#The following source code is a simplified code that reproduces this problem.
#The fourth line of the KPOINTS file resulting from the output of this code is `3.0 3.0 3.0`.

from pymatgen.io.vasp.inputs import Kpoints

Kpoints_str = """\
An example
0
Gamma
3    3    3
0.0  0.0  0.0"""

kpoints = Kpoints.from_str(Kpoints_str)
kpoints.write_file("KPOINTS")
```

_No response_

DanielYang59:
### Summary
 
- Set kpoints in `Kpoints.from_str` method as integer in auto Gamma-centered and Monkhorst-Pack modes, to fix #3993

It turns out typing `Kpoint` as a sequence of `float` is giving us quite some headache and confusion: https://github.com/materialsproject/pymatgen/blob/b28c9371d684b8a605830cd13060bef9accf68cf/src/pymatgen/util/typing.py#L63

It was originally done this way for compatibility with line-mode and explicit kpoint mode (when typing, `float` acts as `Union[int, float]` instead of just `float`). Is there any other cases where `Kpoint` could be `float` that I forgot to mention?

**Inputs and comments are hugely appreciated.**


src/pymatgen/io/vasp/inputs.py
```diff
@@ -1239,7 +1239,7 @@ def style(self, style) -> None:
     def automatic(cls, subdivisions: int) -> Self:
         """
         Constructor for a fully automatic Kpoint grid, with
-        gamma centered Monkhorst-Pack grids and the number of subdivisions
```
DanielYang59, line 1242: It should be Gamma-centered according to [the VASP Wiki](https://www.vasp.at/wiki/index.php/KPOINTS)?
> The following KPOINTS file generates a regular Γ \Gamma -centered k-point.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -1250,6 +1250,8 @@ def automatic(cls, subdivisions: int) -> Self:
         Returns:
             Kpoints object
         """
+        warnings.warn("Please use INCAR KSPACING tag.", DeprecationWarning, stacklevel=2)
```
DanielYang59, line 1253: Not sure if we should follow [VASP Wiki's recommendation](https://www.vasp.at/wiki/index.php/KPOINTS):
> Deprecated: The [KSPACING](https://www.vasp.at/wiki/index.php/KSPACING) tag provides almost the same functionality. Preferably use that method instead.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -1424,7 +1424,7 @@ def automatic_density_by_lengths(
         lattice = structure.lattice
 
         abc = lattice.abc
-        num_div: Tuple3Ints = tuple(np.ceil(ld / abc[idx]) for idx, ld in enumerate(length_densities))
+        num_div: Tuple3Ints = tuple(math.ceil(ld / abc[idx]) for idx, ld in enumerate(length_densities))
```
DanielYang59, line 1427: Need confirmation here: `np.ceil` would return `float`, I think it should be `int` instead in this context.

DanielYang59:
It turns out typing `Kpoint` as a sequence of `float` instead of `int` is giving us quite some headache and confusion: https://github.com/materialsproject/pymatgen/blob/b28c9371d684b8a605830cd13060bef9accf68cf/src/pymatgen/util/typing.py#L63

It was original done this way to be compatible with line mode and explicit kpoints mode (when typing, `float` acts as a Union type of `int | float` instead of just `float`). I would need to think of a clearer way to do this.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -2604,12 +2604,6 @@ def __init__(
     def __str__(self) -> str:
         return "\n".join(str(potcar).strip("\n") for potcar in self) + "\n"
 
-    def __iter__(self) -> Iterator[PotcarSingle]:
```
DanielYang59, line 2613: I believe it should be safe to remove this now:
```python
from pymatgen.io.vasp.inputs import Potcar


potcar = Potcar.from_file("tests/files/io/vasp/inputs/POTCAR_C2.gz")

for psingle in potcar:
    print(type(psingle))   >>> <class 'pymatgen.io.vasp.inputs.PotcarSingle'>
```

src/pymatgen/io/vasp/inputs.py
JaGeo: To be backward compatible, this needs to stay in the input sets

DanielYang59: Thanks a lot for the comment! I'm not sure if `pymatgen` wants to broadcast this warning at all in the first place, because it's more a VASP thing.

> this needs to stay in the input sets.

I don't fully understand sorry. My intention was to emit a warning when `Kpoints.automatic()` is called (which agrees with the VASP manual), why should this be inside the input set (`VaspInputSet`?)?

DanielYang59: @JaGeo Can you perhaps elaborate on this comment please? Sorry I don't fully understand.

JaGeo: Sorry! i misunderstood something here! I don't have any further comments.

DanielYang59: All good! I would mark it as resolved then. Do feel free to comment if you have any!

DanielYang59:
@janosh Could you please comment and review this PR? 

Currently the type of Kpoint is a bit confusing (mainly around the existing of `tuple[float, float, float]` for line mode and explicit kpoint mode) and I tried to clarify as much as I could in this PR. Also is there any other cases where Kpoint could be a sequence of floats that I left out?

tests/io/vasp/test_inputs.py
```diff
@@ -888,29 +888,44 @@ def test_static_constructors(self):
         kpoints = Kpoints.gamma_automatic((3, 3, 3), [0, 0, 0])
         assert kpoints.style == Kpoints.supported_modes.Gamma
         assert kpoints.kpts == [(3, 3, 3)]
+        assert all(isinstance(kpt, int) for kpt in kpoints.kpts[0])
+
         kpoints = Kpoints.monkhorst_automatic((2, 2, 2), [0, 0, 0])
         assert kpoints.style == Kpoints.supported_modes.Monkhorst
         assert kpoints.kpts == [(2, 2, 2)]
-        kpoints = Kpoints.automatic(100)
+        assert all(isinstance(kpt, int) for kpt in kpoints.kpts[0])
```
janosh, line 896: could be refactored to `{*map(type, kpoints.kpts[0])} == {int}` but maybe less readable

src/pymatgen/io/vasp/inputs.py
janosh: lgtm

janosh (approved):
except for a minor nitpick, i think this is safe to merge

src/pymatgen/io/vasp/inputs.py
```diff
@@ -1516,10 +1518,10 @@ def from_str(cls, string: str) -> Self:
 
         # Automatic Gamma-centered or Monkhorst-Pack KPOINTS, with optional shift
         if style in {"g", "m"}:
-            _kpt: list[float] = [float(i) for i in lines[3].split()]
+            _kpt: list[int] = [int(i) for i in lines[3].split()]
             if len(_kpt) != 3:
                 raise ValueError("Invalid Kpoint length.")
-            kpt: Tuple3Floats = cast(Tuple3Floats, tuple(_kpt))
+            kpt: Tuple3Ints = cast(Tuple3Ints, tuple(_kpt))
```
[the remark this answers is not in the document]
DanielYang59: Yes it's indeed redundant but I kept it for readability for humans (in case someone don't know about `cast`). Feel free to remove it if you like :)

tests/io/vasp/test_inputs.py
DanielYang59: Yes it looks less readable, also I guess [type checking with `type(var) == type_of_var` feels a bit off](https://stackoverflow.com/a/152596/24021108)?

I don't think `kpt` would be a subclass of `int` ATM so it would be safe, but still not sure if we want to change this. I would let you decide :) Thanks for the comment.

Issue #3988, opened 2024-08-09
Breaking behavior: Removal of `pymatgen.io.vasp.sets.get_vasp_outcar` without warning


Andrew-S-Rosen:
### Python version

3.11

### Pymatgen version

pymatgen 2024.8.8

### Current behavior

```python
from pymatgen.io.vasp.sets import get_vasp_outcar
```

The above used to work on pymatgen < 2024.8.8 but does not exist any longer in pymatgen >= 2024.8.8. I don't believe there was a deprecation warning, so it may break other people's codes as-is.

### Expected Behavior

A deprecation warning before removal.

shyuep:
This has been fixed with the 2024.8.9 emergency release.

Andrew-S-Rosen:
Thank you!

janosh:
@shyuep please unrevert my #3865 PR. just add back the 
`get_vasprun_outcar` function that was accidentally dropped during merge conflict resolution which I might add was unnecessarily confusing because you moved everything to a `src` layout after i opened #3865. the reason it wasn't already merged before you [pointlessly moved all those files](https://github.com/materialsproject/pymatgen/issues/3874#issue-2344184349) (for which btw you never asked my or @mkhorton's approval either) is because [you asked](https://github.com/materialsproject/pymatgen/pull/3835#issuecomment-2118103809) that i do it after https://github.com/materialsproject/pymatgen/pull/3835 which was sensible so i did.

the split of several-thousand-line files is an obvious improvement. you're throwing away my time by undoing that work

in https://github.com/materialsproject/pymatgen/pull/3989 [you write](https://github.com/materialsproject/pymatgen/pull/3989#issue-2458280210)

> I did not agree to this.

what's that supposed to mean? there are hundreds of PRs that you don't agree to. if every one of those required your explicit approval, pymatgen development would grind to a standstill

mkhorton:
I think we should re-open #3865 and more comprehensively test.

It's a good idea to split these files up for maintainability. This can be done without backwards-incompatible changes; it seems like `get_vasprun_outcar` was an oversight.

However, I would go further than #3865. The `pymatgen.io.vasp.outputs` and `pymatgen.io.vasp.inputs` could also be split up into new folders without introducing backwards incompatibility, and just to improve ease-of-navigation. This would mean, e.g. `pymatgen.io.vasp.inputs.potcar` etc. Like the `/src` change, this sort of change does not need to introduce any breakage.

[...]

PR #3997, opened 2024-08-12
Re-organizing 

mkhorton:
## Summary

Re-opening changes initially introduced by @janosh in #3865 and later reverted by @shyuep (see [thread](https://github.com/materialsproject/pymatgen/issues/3988)).

**I will be taking responsibility for this PR.** This PR will require approval from both @janosh and @shyuep before merging.

Specifically,

- [ ] For each module changed, I will manually inspect `__dir__()` to make sure *all* imports remain unaffected and that there are no backwards-incompatible changes.
- [ ] Also split `pymatgen.io.vasp.inputs` and `pymatgen.io.vasp.outputs` into separate files.

This change is well-motivated to improve maintainability.

I hope we can get this PR merged without too much difficulty. Thank you again to @janosh for starting this PR, to @Andrew-S-Rosen for the bug report due to the errors caused, and to @shyuep for implementing a rapid fix.


src/pymatgen/io/vasp/sets/lobster.py
```diff
@@ -0,0 +1,107 @@
+# ruff: noqa: PGH003
```
DanielYang59, line 1: Thanks a lot for the wonderful work, I totally agree inputs/outputs VASP is too large to navigate through. 

Not sure if this is planned to be fixed or not, if not can you please add a TODO tag and pin me to fix these `PGH003` violations (as I cannot push commits directly) ? Also I believe it's a pretty helpful rule to have (ignore only specific rules instead everything). 

Thanks for the work again!

mkhorton:
The sklearn comparison is illustrative: even where there are large LOC, often there is only a single class per file, and they are quite readable. This is not the case for these specific files in pymatgen. I am not making axiomatic statements; I agree that a high number of LOC is not in itself an error.

Long-time pymatgen developers (we are in good company) might know exactly where in the file to go to make an edit, but before you can get there, one has to read and understand the structure of the code. It is nice to know what classes are available, for example, without needing to read the entire file first.

Point 3 I concur; perhaps it is better to split up the input sets according to functionality instead.

janosh:
Re 3, i think the most common use case is to have consistent settings across whatever mixture of workflows you run (relax, band gap, EOS, phonon, etc.). so after having run some number of relaxations, a user is unlikely to care about half a dozen other band gap workflows with inconsistent VASP settings to their relaxations. more likely, the only thing they want to know is if a band gap workflow from the same lineage (say MP) exists. hence splitting input sets by lineage makes sense imo.

i guess it doesn't have to be one or the other. if you look at [`atomate2/vasp/jobs`](https://github.com/materialsproject/atomate2/tree/409679b73085ce4c3326aad834bc449e89cf3fa7/src/atomate2/vasp/jobs), it has a mixture of splits by lineage (`mp.py`, `matpes.py`, `lobster.py`, `amset.py`) and by workflow type (`md.py`, `elastic.py`, `phonon.py`, ...)

shyuep:
If splitting by source is preferred, I am ok with that. Note that I don't really think atomate2's organizational structure should be the reference here. In any case, if the module split is a private implementation, then it really doesn't matter since we can always reorganize as needed without breaking downstream codes.

PR #4009, opened 2024-08-21
Drop Python 3.9 support

DanielYang59:
### Summary

- Drop Python 3.9 support, following NumPy https://github.com/numpy/numpy/issues/24932
- Remove a few outdated TODOs

### TODOs
- [x] Implement https://github.com/materialsproject/pymatgen/pull/3958#issuecomment-2259644593

### Future PR

- Looks like Python 3.13 rc1 is not supported by micromamba (perhaps I did it incorrectly), let's wait until [the stable release Oct 2024](https://peps.python.org/pep-0719/#schedule)?
- Fix ignore of `RUF017` rules

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,14 +28,14 @@ jobs:
         # maximize CI coverage of different platforms and python versions while minimizing the
         # total number of jobs. We run all pytest splits with the oldest supported python
         # version (currently 3.10) on windows (seems most likely to surface errors) and with
-        # newest version (currently 3.13) on ubuntu (to get complete coverage on unix).
+        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix).
         config:
           - os: windows-latest
             python: "3.10"
             resolution: highest
             extras: ci,optional
           - os: ubuntu-latest
-            python: "3.13.0rc1"
+            python: ">3.10"
             resolution: lowest-direct
             extras: ci,optional
           - os: macos-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
             resolution: highest
             extras: ci,optional
           - os: ubuntu-latest
-            python: "3.13"
+            python: "3.13.0rc1"
             resolution: lowest-direct
             extras: ci,optional
           - os: macos-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
             resolution: highest
             extras: ci,optional
           - os: ubuntu-latest
-            python: ">3.10"
+            python: "3.13"
             resolution: lowest-direct
             extras: ci,optional
           - os: macos-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,19 +27,19 @@ jobs:
       matrix:
         # maximize CI coverage of different platforms and python versions while minimizing the
         # total number of jobs. We run all pytest splits with the oldest supported python
-        # version (currently 3.9) on windows (seems most likely to surface errors) and with
-        # newest version (currently 3.12) on ubuntu (to get complete coverage on unix).
+        # version (currently 3.10) on windows (seems most likely to surface errors) and with
+        # newest version (currently 3.13) on ubuntu (to get complete coverage on unix).
         config:
           - os: windows-latest
-            python: "3.9"
+            python: "3.10"
             resolution: highest
             extras: ci,optional
           - os: ubuntu-latest
-            python: ">3.9"
+            python: ">3.10"
             resolution: lowest-direct
             extras: ci,optional
           - os: macos-latest
-            python: "3.10"
+            python: "3.11"
             resolution: lowest-direct
             extras: ci # test with only required dependencies installed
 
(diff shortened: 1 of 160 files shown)


.github/workflows/test.yml
```diff
@@ -27,19 +27,19 @@ jobs:
...
+            python: "3.10"
             resolution: highest
             extras: ci,optional
           - os: ubuntu-latest
-            python: ">3.9"
+            python: ">3.10"
             resolution: lowest-direct
             extras: ci,optional
           - os: macos-latest
-            python: "3.10"
```
DanielYang59, line 42: Need confirm: A bit unsure why we pinned 3.10 for MacOS, to test some version in between highest and lowest?

janosh: > perhaps to test some version in between highest and lowest?

that's right!

pyproject.toml
```diff
@@ -106,8 +105,7 @@ optional = [
     "h5py>=3.11.0",
     "jarvis-tools>=2020.7.14",
     "matgl>=1.1.1",
-    # TODO: track https://github.com/matplotlib/matplotlib/issues/28551
-    "matplotlib>=3.8,!=3.9.1",
+    "matplotlib>=3.8",
```
janosh, line 108: just to confirm, reading https://github.com/matplotlib/matplotlib/issues/28551 the entire faulty 3.9.1 release was yanked?

janosh (changes requested):
we should bump https://github.com/materialsproject/pymatgen/blob/3fee7d58fdc963b8fb8faf76ce92b68f7ccd0bf9/pyproject.toml#L175 to `py310` and apply resulting ruff fixes in this PR

dev_scripts/update_pt_data.py
```diff
@@ -218,7 +218,9 @@ def gen_iupac_ordering():
     ]  # At -> F
 
     order = sum((list(product(x, y)) for x, y in order), [])  # noqa: RUF017
-    iupac_ordering_dict = dict(zip([Element.from_row_and_group(row, group) for group, row in order], range(len(order))))
+    iupac_ordering_dict = dict(
+        zip([Element.from_row_and_group(row, group) for group, row in order], range(len(order)), strict=False)
```
janosh, line 222: ideally, most of these auto-inserted `zip(strict=False)` should be changed to `strict=True` for increased safety. a few cases will rely on `strict=False` for correct behavior though so it's a lot of work to identify which ones need to stay keep the prior behavior of `strict=False`. pinging @mkhorton in case he wants to weigh in

.github/workflows/test.yml
DanielYang59: Thanks for confirming

pyproject.toml
DanielYang59: Yes, see https://pypi.org/project/matplotlib/#history:
image

pyproject.toml
```diff
@@ -172,7 +170,7 @@ before-all = "ln -s /usr/lib64/libgfortran.so.5 /usr/lib64/libgfortran.so.3"
 repair-wheel-command = "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} --ignore-missing-dependencies"
 
 [tool.ruff]
-target-version = "py39"
+target-version = "py310"
```
DanielYang59, line 173: Thanks. Sorry I missed this one.

src/pymatgen/electronic_structure/plotter.py
```diff
@@ -371,7 +371,7 @@ def _make_ticks(self, ax: plt.Axes) -> plt.Axes:
     def _get_branch_steps(branches):
         """Find discontinuous branches."""
         steps = [0]
-        for b1, b2 in zip(branches[:-1], branches[1:], strict=False):
```
DanielYang59: I think the slicing in `branches[:-1]` is doing nothing here, but pin it in case I misunderstand anything.

To confirm:
```python
from itertools import pairwise


branches = ["a", "b", "c", "d"]

print("With zip:")
for b1, b2 in zip(branches[:-1], branches[1:], strict=False):
    print(f"{b1} + {b2}")

print("With itertool:")
for b1, b2 in pairwise(branches):
    print(f"{b1} + {b2}")
```

janosh: yes, `branches[:-1]` here was redundant as `zip` would have stopped before the last element anyway

DanielYang59: Thanks for confirming!

dev_scripts/update_pt_data.py
DanielYang59: Yes I wouldn't expect people to `zip` two unequal length sequences in most cases and not getting a warning when their lengths don't match. Might worth opening an issue to track this (this PR is already crazily huge).

DanielYang59:
Thanks @janosh @shyuep

QuantumChemist:
Hey 👋🏻 

I know it's just a minor thing, but I guess this [Requires Python 3.9+](https://python.org/downloads) should also be adjusted to [Requires Python 3.10+](https://python.org/downloads) in the README file then?

DanielYang59:
@QuantumChemist Thanks a ton for catching this! I missed this one.

Issue #1791, opened 2020-03-08
The list of vasp's INCAR tags supported by pymatgen currently.


hongyi-zhao:
Hi,

The complete list of vasp INCAR tags is here:
https://www.vasp.at/wiki/index.php/Category:INCAR

I want to know which of them supported by pymatgen currently?

Regards

mkhorton:
You can use any VASP INCAR tag via pymatgen using `user_incar_settings`.

hongyi-zhao:
I mean whether the pymatgen has already built-in all of these tags with the usage instructions when I use a appropriate python IDE, say, pycharm, to write the script.

hongyi-zhao:
The tag list is huge to some extent, so have the build-in usage instructions for them may be more useful for writing the job scripts with the help of python IDE's auto-completion functionality.

shyuep:
Pymatgen does not have instructions for VASP tags or any other software. Ultimately, VASP belongs to a completely different set of developers and they have to maintain their own documentation and help. It is certainly easy enough for us to implement something that obtains the help from the wiki....

hongyi-zhao:
Good idea.  Thanks a lot.

shyuep:
I implemented a basic version for now. It should be easy to make it nicer. See pymatgen.io.vasp.help.

hongyi-zhao:
Wonderful, I'll see it and test it.

shyuep:
```python
from pymatgen.io.vasp.help import VaspDoc
doc = VaspDoc()
doc.print_help("ISYM")
```

Pls review how to use Python classes. Sorry, but we cannot provide support for Python syntax.

hongyi-zhao:
Thanks.

PR #4015, opened 2024-08-25
Fix fixable `ruff` rules

DanielYang59:
### Summary

Fix fixable `ruff` rules

### TODOs
- [x]  Unit test of `io.aims` doesn't follow name convention, and has test files inside it.
- [x] Fix NPY201 https://github.com/materialsproject/pymatgen/blob/41e4c69911f2b724734e0917c436ee5a1de63091/pyproject.toml#L209
- [x] Fix PD011 (there might be a few false-positive for example in `io.abinit.pseudo` and io.cp2k keywords, https://github.com/astral-sh/ruff/issues/6432)
- [x] [**Need help**] Use of relative import in test cf24f12ccd5adf5c34a9c32899f87b2f790effe0

### Future PR
- Replace `http` URLs with `https`


tests/io/aims/conftest.py
```diff
@@ -1,19 +1,164 @@
...
+    monkeypatch.setenv("AIMS_SPECIES_DIR", f"{TEST_FILES_DIR}/io/aims/species_directory")
 
 
 @pytest.fixture(autouse=True)
 def _set_aims_species_dir_settings(monkeypatch: pytest.MonkeyPatch):
-    monkeypatch.setitem(SETTINGS, "AIMS_SPECIES_DIR", f"{MODULE_DIR}/species_directory")
+    monkeypatch.setitem(SETTINGS, "AIMS_SPECIES_DIR", f"{TEST_FILES_DIR}/io/aims/species_directory")
+
+
+def check_band(test_line: str, ref_line: str) -> bool:
```
DanielYang59, line 27: Need confirm: these helper functions for io.aims test is now relocated from `pymatgen.util.testing.aims`, is this intended? @shyuep

tests/io/aims/sets/test_bs_generator.py
```diff
@@ -2,14 +2,13 @@
 
 from __future__ import annotations
 
-from pathlib import Path
-
 from pymatgen.io.aims.sets.bs import BandStructureSetGenerator
-from pymatgen.util.testing.aims import Si, comp_system
+from pymatgen.util.testing import TEST_FILES_DIR
+
+from ..conftest import Si, comp_system  # noqa: TID252
```
DanielYang59, line 8: Not sure how to import without using relative import in this case...

src/pymatgen/io/vasp/help.py
```diff
@@ -55,7 +55,6 @@ def get_help(cls, tag: str, fmt: str = "text") -> str:
         tag = tag.upper()
         response = requests.get(
             f"https://www.vasp.at/wiki/index.php/{tag}",
-            verify=False,  # noqa: S501
```
DanielYang59, line 58: Need confirm, I can see this helper function was added with SSL certificate turned off (#1791), not sure if this is intended for any reason. The `print_help` method seems to work fine with `verify` enabled:
```python
from pymatgen.io.vasp.help import VaspDoc


doc = VaspDoc()
doc.print_help("ISYM")
```

src/pymatgen/io/vasp/help.py
```diff
@@ -74,7 +73,7 @@ def get_incar_tags(cls) -> list[str]:
             "https://www.vasp.at/wiki/index.php/Category:INCAR",
             "https://www.vasp.at/wiki/index.php?title=Category:INCAR&pagefrom=ML+FF+LCONF+DISCARD#mw-pages",
         ):
-            response = requests.get(page, verify=False, timeout=600)  # noqa: S501
```
DanielYang59, line 77: The `get_incar_tags` method doesn't seem to work with or without `verify`:
```
developer/pymatgen/src/pymatgen/io/vasp/help.py:77: GuessedAtParserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("html.parser"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.

The code that caused this warning is on line 77 of the file /Users/yang/developer/pymatgen/src/pymatgen/io/vasp/help.py. To get rid of this warning, pass the additional argument 'features="html.parser"' to the BeautifulSoup constructor.

  soup = BeautifulSoup(response.text)
```

src/pymatgen/io/abinit/abiobjects.py
```diff
@@ -339,7 +338,14 @@ class DefaultVariable:
 DEFAULT = DefaultVariable()
 
 
-class SpinMode(namedtuple("SpinMode", "mode nsppol nspinor nspden"), AbivarAble, MSONable):  # noqa: PYI024
+class _SpinMode(NamedTuple):
```
janosh, line 341: maybe call this `SpinModeTuple` or `SpinModeNT`

src/pymatgen/analysis/adsorption.py
```diff
@@ -300,7 +302,7 @@ def find_adsorption_sites(
             sites = [site + distance * np.asarray(self.mvec) for site in sites]
 
             ads_sites[key] = sites
-        ads_sites["all"] = sum(ads_sites.values(), [])  # noqa: RUF017
+        ads_sites["all"] = functools.reduce(operator.iadd, ads_sites.values(), [])
```
janosh, line 305: generally not so sure this ruff rule is actually an improvement. more verbose and less readable imo. maybe just ignore `RUF017` in general?

tests/io/aims/sets/test_static_restart_from_relax_generator.py
```diff
@@ -2,14 +2,13 @@
 
 from __future__ import annotations
 
-from pathlib import Path
-
 from pymatgen.io.aims.sets.core import StaticSetGenerator
-from pymatgen.util.testing.aims import O2, Si, comp_system
+from pymatgen.util.testing import TEST_FILES_DIR
+
+from ..conftest import O2, Si, comp_system  # noqa: TID252
```
janosh, line 8:
```suggestion
from tests.conftest import O2, Si, comp_system
```

janosh (changes requested):
thanks, good stuff as always!

will hold off on reviewing #4017 until this is merged

src/pymatgen/io/vasp/help.py
janosh: academic websites sometimes lack SSL certificates. perhaps VASP docs did in the past, but since they have SSL now, no reason not to enable

src/pymatgen/io/abinit/abiobjects.py
DanielYang59: ok!

tests/io/aims/sets/test_static_restart_from_relax_generator.py
DanielYang59: Can you double check this? This doesn't seem to work (we have to import from `pymatgen/tests/io/aims/conftest.py` not `pymatgen/tests/conftest.py`). Also `pymatgen.tests` isn't made a module at this moment (not sure if absolute import would work).

```python
____________________________________ ERROR collecting tests/io/aims/sets/test_bs_generator.py ____________________________________
ImportError while importing test module '/Users/yang/developer/pymatgen/tests/io/aims/sets/test_bs_generator.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/io/aims/sets/test_bs_generator.py:8: in <module>
    from tests.conftest import Si, comp_system  # noqa: TID252
E   ImportError: cannot import name 'Si' from 'tests.conftest' (/Users/yang/developer/pymatgen/tests/conftest.py)
```

src/pymatgen/analysis/adsorption.py
DanielYang59: I think this ruff is implemented to reduce time complexity ([`Checks for the use of sum() to flatten lists of lists, which has quadratic complexity.`](https://docs.astral.sh/ruff/rules/quadratic-list-summation/)), happy to revert this if you believe readability is more important than performance for these cases.

tests/io/aims/sets/test_static_restart_from_relax_generator.py
janosh: ah, then it should work with `from tests.io.aims.conftest import ...`

src/pymatgen/analysis/adsorption.py
janosh: good to know. didn't realize `sum` scales differently. i would say let's go with `[item for sublist in lists for item in sublist]` then since it doesn't require extra imports and list comprehension is often the most performant in python

tests/io/aims/sets/test_static_restart_from_relax_generator.py
DanielYang59: It doesn't seem to work on my machine either. I thought we could only:
1. Relative import, which is used right now but the `ruff` rule TID252 would suggest against this.
2. Absolute import, with `pymatgen.tests.io.aims.conftest`, but this would not work because `pymatgen.tests` are not a package in our case (no `__init__.py`).

Perhaps just ignore this rule for tests?
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:

janosh: > Perhaps just ignore this rule for tests?

sounds good to me. the absolute import from[ `tests` works in `pymatviz`](https://github.com/janosh/pymatviz/blob/1de350a146b53ae03f343ff9a78454ba7c976186/tests/test_bar.py#L12) but that's because it has an `__init__.py`

src/pymatgen/analysis/adsorption.py
DanielYang59: Wonderful idea. It looks like [list comprehension performs pretty well](https://realpython.com/python-flatten-list/#considering-performance-while-flattening-your-lists):

```
Time to flatten a 1000x1000 matrix (in milliseconds):

flatten_concatenation()..........1.95 ms
flatten_extend().................2.03 ms
flatten_reduce_iconcat().........2.68 ms
flatten_chain()..................4.60 ms
flatten_comprehension()..........7.79 ms  >> with list comprehension 
flatten_sum().................1113.22 ms
flatten_reduce_concat().......1117.15 ms
flatten_reduce_lambda().......1117.52 ms
flatten_reduce_add()..........1118.80 ms
```

tests/io/aims/sets/test_static_restart_from_relax_generator.py
DanielYang59: ok!

janosh (approved):
thanks a million @DanielYang59! 👍

DanielYang59:
No problem, thanks for reviewing!

PR #4017, opened 2024-08-26
Fix `S101`, replace all `assert` in code base (except for tests)

DanielYang59:
### Summary

- Fix `S101`, replace all `assert` in code base
- [x]  Use offending value in error message

**Please review all changes carefully to make sure the assert condition is correctly reversed** (there're just too many changes and I tried not to make mistakes).


janosh (changes requested):
thanks a lot for all this work @DanielYang59! 🙏 

will improve user experience a lot to get helpful error messages rather than a generic `AssertionError`. to make the messages more informative, i think it would be great to include the offending value/type in all `(Value|Type)Errors` (see 135f69ef7d for a fe examples)

DanielYang59:
> thanks a lot for all this work @DanielYang59! 🙏

No problem. Thanks for reviewing!

> to make the messages more informative, i think it would be great to include the offending value/type in all `(Value|Type)Errors` (see [135f69e](https://github.com/materialsproject/pymatgen/commit/135f69ef7db4b1318c0882fd46e08d31204f4b86) for a fe examples)

That's a very sensible suggestion, I would implement this later :)

DanielYang59:
Manually trace this link https://github.com/materialsproject/pymatgen/commit/135f69ef7db4b1318c0882fd46e08d31204f4b86 to the discussion on f-string, in case I want to get back to this thread.

janosh (approved):
excellent work! 🙏

src/pymatgen/analysis/diffraction/xrd.py
```diff
@@ -120,7 +120,8 @@ def __init__(self, wavelength="CuKa", symprec: float = 0, debye_waller_factors=N
             self.radiation = wavelength
             self.wavelength = WAVELENGTHS[wavelength]
         else:
-            raise TypeError(f"{type(wavelength)=} must be either float, int or str")
+            wavelength_type = type(wavelength).__name__
```
DanielYang59, line 123: Ah, thanks a lot for fixing all those (`type(var).__name__` for name directly instead of `<class 'xxx'>`). I totally forgot about them.

PR #4021, opened 2024-08-28
Replace HTTP URLs with HTTPS, avoid `from pytest import raises/mark`

DanielYang59:
### Summary

- Replace HTTP URLs with HTTPS
- Replace some expired URLs
- Avoid `from pytest import raises/mark` (a bit unsure about `MonkeyPatch` so just use `pytest.MonkeyPatch` if `pytest` is already imported)
- [ ] Regenerate documents


src/pymatgen/io/abinit/abiobjects.py
```diff
@@ -56,7 +56,7 @@ def lattice_from_abivars(cls=None, *args, **kwargs):
             raise ValueError(f"The sum of angdeg must be lower than 360, {ang_deg=}")
 
         # This code follows the implementation in ingeo.F90
-        # See also http://www.abinit.org/doc/helpfiles/for-v7.8/input_variables/varbas.html#angdeg
```
DanielYang59, line 59: I have to replace this with the latest doc instead of the original v7.8, because the original URL seems inaccessible.

DanielYang59:
@janosh  Please review this. Looks like we might need to regenerate docs, thanks!

janosh (approved):

janosh:
Currently on my phone. Could you apply the docs commit again?

src/pymatgen/io/exciting/inputs.py
```diff
@@ -197,7 +197,7 @@ def write_etree(self, celltype, cartesian=False, bandstr=False, symprec: float =
         """
         root = ET.Element("input")
         root.set(
-            "{http://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation",
```
DanielYang59, line 200: I was in my dreams sleeping (3am) and suddenly woke up and realized this might be a broken change. I would have a closer look when I really wake up.

DanielYang59: I have reverted this: https://github.com/materialsproject/pymatgen/blob/46d843c07b7dcfd1ab25ab41a01cfa60f5ce123d/src/pymatgen/io/exciting/inputs.py#L198-L202

I believe replacing http would be a breaking change because the namespace from the xml file would stay the same.

src/pymatgen/alchemy/materials.py
```diff
@@ -369,7 +369,7 @@ def to_snl(self, authors: list[str], **kwargs) -> StructureNL:
             history += [
                 {
                     "name": snl_metadata.pop("name", "pymatgen"),
-                    "url": snl_metadata.pop("url", "http://pypi.python.org/pypi/pymatgen"),
+                    "url": snl_metadata.pop("url", "https://pypi.python.org/pypi/pymatgen"),
```
DanielYang59, line 372: This looks like a breaking change too, the metadata (default value of `url` would have been changed):
https://github.com/materialsproject/pymatgen/blob/46d843c07b7dcfd1ab25ab41a01cfa60f5ce123d/src/pymatgen/alchemy/materials.py#L367-L375

janosh (approved):
ideally, we should have a link checker CI [similar to pymativz](https://github.com/janosh/pymatviz/blob/1de350a146b53ae03f343ff9a78454ba7c976186/.github/workflows/link-check.yml) to make sure all the new (and old links) are actually still alive. but i'm happy to leave that to a future PR. maybe something to open an issue for so it's not forgotton

DanielYang59:
> ideally, we should have a link checker CI [similar to pymativz](https://github.com/janosh/pymatviz/blob/1de350a146b53ae03f343ff9a78454ba7c976186/.github/workflows/link-check.yml) to make sure all the new (and old links) are actually still alive. but i'm happy to leave that to a future PR. maybe something to open an issue for so it's not forgotton

Wonderful. Added to my TODO list (which is not very long at this moment :) ).

janosh:
@DanielYang59 do you want to reapply the docs update commit before i merge?

DanielYang59:
@janosh certainly, got two warnings at the start, not sure if we should worry about that or not (it seems `favicon.ico` and `html_static_path = ["assets"]` don't exist?):
```
>>> invoke make-doc
rm: pymatgen.*.rst: No such file or directory
Running Sphinx v7.4.7
loading translations [en]... done
making output directory... done
WARNING: html_static_path entry 'assets' does not exist
WARNING: favicon file 'favicon.ico' does not exist
Converting `source_suffix = '.rst'` to `source_suffix = {'.rst': 'restructuredtext'}`.
building [mo]: targets for 0 po files that are out of date
writing output... 
building [html]: targets for 49 source files that are out of date
updating environment: [new config] 49 added, 0 changed, 0 removed
```

janosh:
> @janosh certainly, got two warnings at the start, not sure if we should worry about that or not (it seems favicon.ico and html_static_path = ["assets"] don't exist?):

safe to ignore for now. the docs need some more work at some point but that would be a bigger project

DanielYang59:
Great to know! Thanks.

PR #4022, opened 2024-08-29
Fix PT013: incorrect import of `pytest`

DanielYang59:
### Summary

- Fix [PT013: incorrect import of `pytest`](https://docs.astral.sh/ruff/rules/pytest-incorrect-pytest-import/)
> For consistency, pytest should be imported as import pytest and its members should be accessed in the form of pytest.xxx.yyy for consistency

Admittedly, `approx` is less verbose than `pytest.approx` (sad we cannot set an exception for `approx`), not sure if we want to fix this

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -216,7 +216,6 @@ ignore = [
     "PLR1702", # Too many nested blocks
     "PLR2004", # Magic-value-comparison TODO: fix these
     "PLW2901", # Outer for loop variable overwritten by inner assignment target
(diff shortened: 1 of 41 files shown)


DanielYang59:
I might close this one because `approx` is heavily used and `pytest and way too verbose`, [pytest doc even use `from pytest import approx`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.approx) so perhaps we should not say it's "incorrect import".

PR #4027, opened 2024-09-02
Recover commented out code in tests and mark with `pytest.mark.skip` instead

DanielYang59:
### Summary

- Recover commented out code in tests and mark with `pytest.mark.skip` instead (with `TODO: need someone to fix this` message for easy search), though currently [`ERA`](https://docs.astral.sh/ruff/rules/commented-out-code/) is ignored: https://github.com/materialsproject/pymatgen/blob/c06abf1a3c8d1507d9b128a91b51babbca650ab9/pyproject.toml#L192

### Rationale

Commented out code ages badly if not covered by any linter, perhaps we should mark them with `pytest.mark.skip` instead until someone come and fix them (hopefully).

### DIFF
--- a/tests/analysis/chemenv/coordination_environments/test_coordination_geometry_finder.py
+++ b/tests/analysis/chemenv/coordination_environments/test_coordination_geometry_finder.py
@@ -1,10 +1,17 @@
 from __future__ import annotations
 
+import json
+import os
+
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose
 from pytest import approx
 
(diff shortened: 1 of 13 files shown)


DanielYang59:
Looks like there isn't more commented out code in tests (searched with `#\s+def ` and `#\s+class `).

Also we only have 116 lines of code commented in tests (might include a few false positive), should be possible to enable `ERA` for tests at some point.

DanielYang59:
Thanks!

PR #4053, opened 2024-09-08
Avoid importing namespace package `pymatgen` directly 

DanielYang59:
### Summary

- Avoid importing namespace package `pymatgen` directly 
- [**Question**] Should we recommend the following alias (I thought `pmg` would be a shorthand of `pymatgen` instead of `pymatgen.core`): https://github.com/materialsproject/pymatgen/blob/15dc137c9325b58e7ce6b14bc14bd434e5f907f3/docs/index.md?plain=1#L181


janosh:
> [Question] Should we recommend the following alias (I thought pmg would be a shorthand of pymatgen instead of pymatgen.core):

no strong opinion but leaning towards not proliferating that usage

janosh (approved):

DanielYang59:
> > [Question] Should we recommend the following alias (I thought pmg would be a shorthand of pymatgen instead of pymatgen.core):
> 
> no strong opinion but leaning towards not proliferating that usage

Okay thanks for your reply, we might just `from pymatgen.core import xxx` (not the wild card `*`) then. And we might need to rebuild docs again (perhaps add rebuilding docs as an automate workflow as well?).

janosh:
> perhaps add rebuilding docs as an automate workflow as well?

yes, that should have been done a long time ago. in particular, there's no reason to keep the generated HTML for docs under version control. it's a major history pollutant

DanielYang59:
> > perhaps add rebuilding docs as an automate workflow as well?
> 
> yes, that should have been done a long time ago. in particular, there's no reason to keep the generated HTML for docs under version control. it's a major history pollutant

Okay I would have a look at this if you don't have much time ATM :)

docs/index.md
```diff
@@ -178,31 +178,32 @@ perform further structure manipulation or analyses.
 Here are some quick examples of the core capabilities and objects:
 
 ```python
-import pymatgen.core as pmg
+from pymatgen.core import Element, Composition, Lattice, Structure, Molecule
```
DanielYang59, line 181: Ah my bad, I manually typed in these import yesterday, forgot to sort them, should be:
```python
from pymatgen.core import Composition, Element, Lattice, Molecule, Structure
```

Would fix this in a following PR.

PR #3878, opened 2024-06-12
Add modification to aims input to match atomate2 magnetic order script

tpurcell90:
How it deals with spins is not through magmoms, but FHI-aims uses magmoms exclusively

## Summary

Modifying FHI-aims inputs to work with the magnetism workflows in atomate2

## Todos

If this is work in progress, what else needs to be done?

## Checklist

- [x] Google format doc strings added. Check with `ruff`.
- [x] Type annotations included. Check with `mypy`.
- [x] Tests added for new features/fixes.
- [x] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

PR #4054, opened 2024-09-08
Make AimsSpeciesFile a dataclass

tpurcell90:
As recommended by @janosh in #3878

## Summary

Major changes:

- Make AimsSpeciesFile a dataclass

## Checklist

### DIFF
--- a/src/pymatgen/io/aims/inputs.py
+++ b/src/pymatgen/io/aims/inputs.py
@@ -689,51 +689,24 @@ class AimsControlIn(MSONable):
         return cls(_parameters=decoded["parameters"])
 
 
+@dataclass
 class AimsSpeciesFile:
-    """An FHI-aims single species' defaults file."""
+    """An FHI-aims single species' defaults file.
 
-    def __init__(self, data: str, label: str | None = None) -> None:
-        """
-        Args:
-            data (str): A string of the complete species defaults file
-            label (str): A string representing the name of species
-        """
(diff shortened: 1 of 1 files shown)


src/pymatgen/io/aims/inputs.py
```diff
@@ -795,25 +768,24 @@ def from_element_and_basis_name(
...
     def as_dict(self) -> dict[str, Any]:
         """Dictionary representation of the species' defaults file."""
         return {"label": self.label, "data": self.data, "@module": type(self).__module__, "@class": type(self).__name__}
 
     @classmethod
-    def from_dict(cls, dct: dict[str, Any]) -> AimsSpeciesFile:
+    def from_dict(cls, dct: dict[str, Any]) -> Self:
         """Deserialization of the AimsSpeciesFile object"""
-        return AimsSpeciesFile(data=dct["data"], label=dct["label"])
+        return cls(**dct)
```
janosh, line 788: this method will now fail if `dct` contains additional keys besides `data` and `label` where it wouldn't before but could be a good safety mechanism unless that leeway was intentional?

janosh (approved):

src/pymatgen/io/aims/inputs.py
tpurcell90: I think this would be okay. I added the leeway there in case something changed in the future with the species.

PR #4057, opened 2024-09-09
Fix `apply_operation(fractional=True)`

kavanase:
## Summary
`Structure.apply_operation()` fails when using the `fractional=True` operation.
The issue is that the ``SymmOp`` is applied to the site frac coords (in the original lattice), but then these new frac coords (in the old basis) are used directly with the new (transformed) lattice, giving a 'doubled' transformation and messing with the output.

Fix added here, matching the implementation in [`doped`](https://github.com/SMTG-Bham/doped/blob/develop/doped/utils/symmetry.py#L211) (which also has a couple more options and a cleaner implementation).

MWE:
image
image

(Notebook and structure being used attached too as always 😉 , with more examples)

[Pymatgen_SymmOp_Fix_PR.zip](https://github.com/user-attachments/files/16934325/Pymatgen_SymmOp_Fix_PR.zip)

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -4297,8 +4297,9 @@ class Structure(IStructure, collections.abc.MutableSequence):
             def operate_site(site):
                 return PeriodicSite(
                     site.species,
-                    symm_op.operate(site.frac_coords),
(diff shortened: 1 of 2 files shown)


kavanase:
This fixed implementation is tested and shown to be working in the attached notebook as well

janosh (approved):
thanks, good stuff!

PR #4061, opened 2024-09-11
Fix branch directory check in `io.vasp.outputs.get_band_structure_from_vasp_multiple_branches`

### LINKED ISSUE #4060: get_band_structure_from_vasp_multiple_branches
### Python version

PYthon 3.12.5

### Pymatgen version

2024.8.9

### Operating system version

CentOS 7.6

### Current behavior

in pymatgen/io/vasp/output.py 

`get_band_structure_from_vasp_multiple_branches`

`if os.path.isfile(f"{dir_name}/branch_0"):`

### Expected Behavior

`get_band_structure_from_vasp_multiple_branches`

the line `if os.path.isfile(f"{dir_name}/branch_0"):`

check the `filename` for the `dirname`

So, it return `None` always.

I don't know why the newest version changed.

### Minimal example

```Python
from pymatgen.io.vasp.outputs import get_band_structure_from_vasp_multiple_branches

bs = get_band_structure_from_vasp_multiple_branches('./bs')

print(bs == None)
```

_No response_

DanielYang59:
### Summary

- Fix branch directory check in `io.vasp.outputs.get_band_structure_from_vasp_multiple_branches`, to fix #4060
- [ ] Add unit test

### DIFF
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -4262,29 +4262,30 @@ class VaspParseError(ParseError):
 
 
 def get_band_structure_from_vasp_multiple_branches(
-    dir_name: str,
+    dir_name: PathLike,
     efermi: float | None = None,
     projections: bool = False,
 ) -> BandStructureSymmLine | BandStructure | None:
     """Get band structure info from a VASP directory.
 
-    It takes into account that a run can be divided in several branches named
-    "branch_x". If the run has not been divided in branches the method will
-    turn to parsing vasprun.xml directly.
+    It takes into account that a run can be divided in several branches,
+    each inside a directory named "branch_x". If the run has not been
+    divided in branches the function will turn to parse vasprun.xml
+    directly from the selected directory.
 
     Args:
-        dir_name: Directory containing all bandstructure runs.
-        efermi: Efermi for bandstructure.
-        projections: True if you want to get the data on site projections if
-            any. Note that this is sometimes very large
+        dir_name (PathLike): Parent directory containing all bandstructure runs.
+        efermi (float): Fermi level for bandstructure.
+        projections (bool): True if you want to get the data on site
+            projections if any. Note that this is sometimes very large
 
     Returns:
-        A BandStructure Object.
+        A BandStructure/BandStructureSymmLine Object.
         None is there's a parsing error.
     """
-    # TODO: Add better error handling!!!
+    # TODO: Add better error handling
     if os.path.isfile(f"{dir_name}/branch_0"):
-        # Get all branch dir names
+        # Get all branch directory names
         branch_dir_names = [os.path.abspath(d) for d in glob(f"{dir_name}/branch_*") if os.path.isdir(d)]
 
         # Sort by the directory name (e.g, branch_10)
@@ -4299,7 +4300,7 @@ def get_band_structure_from_vasp_multiple_branches(
                 branches.append(run.get_band_structure(efermi=efermi))
             else:
                 # TODO: It might be better to throw an exception
-                warnings.warn(f"Skipping {dname}. Unable to find {xml_file}")
+                warnings.warn(f"Skipping {dname}. Unable to find {xml_file}", stacklevel=2)
 
         return get_reconstructed_band_structure(branches, efermi)
 
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -4284,7 +4284,7 @@ def get_band_structure_from_vasp_multiple_branches(
         None is there's a parsing error.
     """
     # TODO: Add better error handling
-    if os.path.isfile(f"{dir_name}/branch_0"):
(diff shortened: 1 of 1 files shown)


janosh (approved):
that was super fast! thanks a lot @DanielYang59, esp. for the new unit test! 👍

src/pymatgen/io/vasp/outputs.py
```diff
@@ -4262,54 +4262,53 @@ class VaspParseError(ParseError):
...
         sorted_branch_dir_names = sorted(branch_dir_names, key=lambda x: int(x.split("_")[-1]))
 
-        # Populate branches with Bandstructure instances
-        branches = []
-        for dname in sorted_branch_dir_names:
-            xml_file = f"{dname}/vasprun.xml"
-            if os.path.isfile(xml_file):
-                run = Vasprun(xml_file, parse_projected_eigen=projections)
-                branches.append(run.get_band_structure(efermi=efermi))
-            else:
```
DanielYang59, line 4300: I'm considering deprecating and removing this branch altogether because it's already not doing what this function is intended:
https://github.com/materialsproject/pymatgen/blob/6bb293896945106476d6676cb2b65550d791edb5/src/pymatgen/io/vasp/outputs.py#L4306-L4311

When this function falls into this branch, it might suggest either a wrong directory is selected or the directory is not structured as expected, user should really check the provided directory, instead of we still read a `vasprun.xml` directly from that directory?

What do you think? @janosh

janosh: probably best to keep as is to avoid a breaking change. never used that function myself so don't know the rationale behind returning

```py
Vasprun(xml_file, parse_projected_eigen=projections).get_band_structure(kpoints_filename=None, efermi=efermi)```

if there are no branch dirs

Comment (DanielYang59) [reply src/pymatgen/io/vasp/outputs.py:4300]: > probably best to keep as is to avoid a breaking change

Yes I totally agree with minimizing breaking change whenever possible, so I'm suggesting adding a deprecation warning to that particular branch instead of changing this behavior right away.

Also this function has several TODOs attached regarding error handling, so I assume change is needed:
https://github.com/materialsproject/pymatgen/blob/6bb293896945106476d6676cb2b65550d791edb5/src/pymatgen/io/vasp/outputs.py#L4285-L4286
https://github.com/materialsproject/pymatgen/blob/6bb293896945106476d6676cb2b65550d791edb5/src/pymatgen/io/vasp/outputs.py#L4301-L4302
https://github.com/materialsproject/pymatgen/blob/6bb293896945106476d6676cb2b65550d791edb5/src/pymatgen/io/vasp/outputs.py#L4307-L4309


The workflow of `get_band_structure_from_vasp_multiple_branches` is now like this:
```
--> check dir_name/branch_*/vasprun.xml  --> Yes --> continue to work on multiple branch bandstructure
                          |
                          |
                          No  --> try to read dir_name/vasprun.xml
                                   I think an exception should be raised here because it's already not
                                  "get band structure from multiple branches" from here
```

Comment (janosh) [reply src/pymatgen/io/vasp/outputs.py:4300]: thanks for clarifying. sorry, i missed the deprecating part in [your "deprecating and removing" message](https://github.com/materialsproject/pymatgen/pull/4061#discussion_r1756634794). i think with a one year deprecation period and no one complaining to keep the current behavior during that time, should be fine to remove it

File: tests/io/vasp/test_outputs.py [hunk 2]
```diff
@@ -1429,13 +1432,50 @@ def test_init(self):
...
         fpath = f"{TEST_DIR}/fixtures/static_silicon/OSZICAR"
         oszicar = Oszicar(fpath)
         assert oszicar.final_energy == approx(-10.645278)
-        assert set(oszicar.ionic_steps[-1]) == set({"F", "E0", "dE", "mag"})
+        assert set(oszicar.ionic_steps[-1]) == {"F", "E0", "dE", "mag"}
+
+
+class TestGetBandStructureFromVaspMultipleBranches:
+    def test_read_multi_branches(self):
+        """TODO: use real multi-branch bandstructure calculation."""
```
DanielYang59, line 1446: I would mark this as ready for review but I have to leave this TODO tag here because I have no experience with multi-branch bandstructure calculation. This may not be a legit real world calculation as the `vasprun.xml` is the same for all branches.

DanielYang59:
@shyuep Perhaps you could review this as well? 

I tried to add a unit test but I don't have any file for "VASP multi-branch bandstructure calculation" and had to create a dummy one instead, so I guess this is as far as I could get :)

tests/io/vasp/test_outputs.py
mkhorton: Hi @DanielYang59, do you want to resolve this comment? The PR looks good otherwise but I don't see it makes sense to include a test unless we are confident  in what is being tested.

mkhorton:
Thanks @DanielYang59, looks good. Set this to auto-merge

DanielYang59:
thanks!

PR #4063, opened 2024-09-12
(Not) Use `setup-uv` in CI tests workflow

### LINKED ISSUE #4044: Use `astral-sh/setup-uv` in CI
`uv` now has 1st-party GitHub Action support with built-in caching which could speedup our `test.yml` workflow. not sure if it clashes with our `mamba` setup

https://github.com/materialsproject/pymatgen/blob/1404220105d4dc1e421ac603b1cb83daf6b085c5/.github/workflows/test.yml#L64-L72

but if not, would be good to adopt this. should only require adding a new step:

```yml
      - name: Set up uv
        uses: astral-sh/setup-uv@v2
        with:
          enable-cache: true
```

DanielYang59:
### Summary

- [Use `setup-uv` in CI tests workflow](https://docs.astral.sh/uv/guides/integration/github/), to close #4044

## Performance benchmark

The following word "native" refers to package installed by native github action setups: `setup-Python/micromamba/uv`

**NOTE**: `setup-uv` installer might provide caching function, so the run time might be slower for the first run

### Dependency installation time

Time (in seconds)  is reported on the `Install pymatgen and dependencies` stage for "split 1", excluding conda and optional Ubuntu dependencies 

|                           | Ubuntu 22.04 | Windows 11 | MacOS |
|---------------------------|--------------|------------|-------|
| native mamba + pip        |        88      |      235      |   26    |
| native mamba + uv         |         85     |      150      |    28   |
| native mamba + native uv (first run)  |        93      |     254       |    24   |
| native mamba + native uv  (second run) |   97   |   141    |   23  |

**Conclusion**: not much difference between `setup-uv` and `pip install uv`, and [the M1 MacOS runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners) is just another level.

### Pytest running time

Time (in seconds) is reported for "split 1"

|                           | Ubuntu 22.04 | Windows 11 | MacOS |
|---------------------------|--------------|------------|-------|
| native mamba + pip        |       187       |       262     |    101   |
| native mamba + uv         |      181  |     249       |    114   |
| native mamba + native uv (first run)  |    183          |       258     |   100     |
| native mamba + native uv  (second run) |  186    |  249 |   100 |

**Conclusion**: Actual test running time is largely not influenced (as expected).

### Link to above workflow runs

|                           | Link |
|---------------------------|------|
| native mamba + pip     |   https://github.com/materialsproject/pymatgen/actions/runs/10824932738   |
| native mamba + uv         |   https://github.com/materialsproject/pymatgen/actions/runs/10824062208   |
| native mamba + native uv  (first run) |  https://github.com/materialsproject/pymatgen/actions/runs/10824811886    |
| native mamba + native uv  (second run) |   https://github.com/materialsproject/pymatgen/actions/runs/10824892797   |

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,10 +82,9 @@ jobs:
           # TODO1 (use uv over pip) uv install torch is flaky, track #3826
           # TODO2 (pin torch version): DGL library (matgl) doesn't support torch > 2.2.1,
           # see: https://discuss.dgl.ai/t/filenotfounderror-cannot-find-dgl-c-graphbolt-library/4302
-          # DEBUG: attach timestamp in ms
-          pip install torch==2.2.1 --verbose 2>&1 | while IFS= read -r line; do echo "$(date '+%Y-%m-%d %H:%M:%S.%3N') $line"; done
+          pip install torch==2.2.1
 
-          uv pip install --editable '.[${{ matrix.config.extras }}]' --resolution=${{ matrix.config.resolution }} --verbose 2>&1 | while IFS= read -r line; do echo "$(date '+%Y-%m-%d %H:%M:%S.%3N') $line"; done
+          uv pip install --editable '.[${{ matrix.config.extras }}]' --resolution=${{ matrix.config.resolution }}
 
       - name: Install optional Ubuntu dependencies
         if: matrix.config.os == 'ubuntu-latest'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,21 +68,12 @@ jobs:
         run: |
           micromamba create -n pmg python=${{ matrix.config.python }} --yes
 
-      # - name: Install uv
-      #   run: micromamba run -n pmg pip install uv
-
       - name: Install ubuntu-only conda dependencies
         if: matrix.config.os == 'ubuntu-latest'
         run: |
           micromamba install -n pmg -c conda-forge bader enumlib openff-toolkit packmol pygraphviz tblite --yes
 
-      # - name: Install the latest version of uv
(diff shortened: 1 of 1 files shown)


DanielYang59:
@janosh  Turns out there isn't much speed advantage for `setup-uv` over `pip install uv`, the first time running `setup-uv` is even slower. Perhaps we just keep things simple (also easy to maintain) and still use `pip install uv` for now.

janosh:
> Turns out there isn't much speed advantage for `setup-uv` over `pip install uv`, the first time running `setup-uv` is even slower.

good to know. i wasn't expecting it to be faster on a cold cache though. mainly hoping `setup-uv` would have good caching and be faster on subsequent runs

DanielYang59:
> > Turns out there isn't much speed advantage for `setup-uv` over `pip install uv`, the first time running `setup-uv` is even slower.
> 
>  i wasn't expecting it to be faster on a cold cache though. mainly hoping `setup-uv` would have good caching and be faster on subsequent runs

Yep, see the following tests, tested on second runs (`native mamba + native uv (second run)`) and didn't see obvious advantage

janosh:
thanks so much for your benchmarking! ❤️ that's super useful for the decision at hand _and_ later reference. i agree seems there's not much point in adopting `setup-uv`.

merging now. thanks for cleaning up the needless build-time deps install

DanielYang59:
> that's super useful for the decision at hand and later reference

No problem! That's what I was hoping too. Looks like `torch` is the huge bottleneck, hoping #3826 could have a solution soon :)

PR #4070, opened 2024-09-16
fix phase diagram break from 41e6d99

yang-ruoxi:
## Summary
This PR fixed the break of phase diagram plot caused by commit 41e6d99 which made changes not only to syntax but also plotting logic without careful review. This is a small fix but did require human hour to track down the issue, and caused downstream breakage on MP website and for pymatgen users who plot phase diagrams. 

Screenshot 2024-09-16 at 4 25 56 PM

### DIFF
--- a/src/pymatgen/analysis/phase_diagram.py
+++ b/src/pymatgen/analysis/phase_diagram.py
@@ -2972,7 +2972,7 @@ class PDPlotter:
                 for d in ["xref", "yref"]:
                     annotation.pop(d)  # Scatter3d cannot contain xref, yref
                     if self._dim == 3:
-                        annotation.update(x=x, y=y)
+                        annotation.update(x=y, y=x)
                         if entry.composition.is_element:
                             z = 0.9 * self._min_energy  # place label 10% above base
 


shyuep:
Sigh... why does this have to happen again?!

shyuep:
I have released an emergency version. Thanks for tracking this down.

janosh:
@yang-ruoxi thanks for finding and fixing! 👍 


> Sigh... why does this have to happen again?!

@shyuep you should have asked for a unit test to prevent future regressions.

shyuep:
Thanks for that helpful suggestion. 

This is a plotting code. I don't expect @yang-ruoxi to implement a unittest to check annotation positions. Also, I did ask that we avoid unnecessary cosmetic changes. Not like it did any good what I asked. Switching dict.update to the assignment syntax falls under that category and I am not even sure it is an improvement in code aesthetics even.

yang-ruoxi:
@shyuep thanks for the speedy release. 
@janosh while unit test may help, I don't think pymatgen users are expected, or able to, write tests for all possible scenarios of every class they consume. There should be scrutiny over multi-file changes that involve hundreds of lines of code.

janosh:
@yang-ruoxi i disagree. 100% coverage should absolutely be the goal even if we never get there. But a bug fix PR is an obvious place to inch towards it, since code breaking is a strong signal that it needs a test.

@shyuep shorter code is generally preferred

shyuep:
@janosh While unittests are always preferred, I will state again that I do not expect @yang-ruoxi to implement a unittest to check annotations - which is what this bug fix is about. Also, a proper response would have been to apologize for making the mistake (which I did not see anywhere in the responses) rather than just delivering these comebacks that suggest the onus is on others.

janosh:
> Also, a proper response would have been to apologize for making the mistake

let's not pretend you care about apologies. mistakes happen, esp. in a code base as large and with as much suboptimal code as pymatgen. i regret that it did. but the important thing is to have guard rails in place that allow you to code with confidence. that requires adding unit tests whenever something breaks which you could have asked me to write given i introduced the bug instead of merging immediately

>  I will state again that I do not expect @yang-ruoxi to implement a unittest to check annotations

no need to repeat yourself, i still disagree the 2nd time

shyuep:
@janosh The apology wasn't meant for me. And it does matter. Since you are not interested in giving it, we have nothing left to talk about. Go ahead and occupy yourself with other concerns.

mkhorton:
As a maintainer of this repo, I have to remind that we expect all discussions to be civil and polite.

For example, this is not an appropriate comment because I would consider this a personal attack:

> let's not pretend you care about apologies.

I would like to offer an apology to @yang-ruoxi for the discourse in this thread, and say thank you again for reporting the bug.

I take all comments in good faith because I know we all share the motivation to have code that is accurate, reliable and well-tested. I would like to make sure that we keep things friendly and welcoming while we pursue this goal. Thanks everyone.

janosh:
> For example, this is not an appropriate comment because I would consider this a personal attack:

@mkhorton in good faith, i'm too familiar with the characters involved to take @shyuep's request at face value:

>> Also, a proper response would have been to apologize for making the mistake

that was an obvious attempt at forcing submission rather than a genuine effort to steer towards a healthier community. asking for apologies should go hand in hand with acknowledging them when others make them unprompted and making some yourself. i've yet to see @shyuep do either

shyuep:
To avoid further off-topic whining about my intentions, I am locking this thread. I will not bother censoring any comments since they paint a clear picture.

PR #4071, opened 2024-09-17
Fix: fixed electronic step check with algo = exact and nelm = 1

yanghan234:
## Summary

Major changes:

This PR fixes the bug when parsing non-self-consistent calculation done with `ALGO=Exact` and `NELM=1`. In such senario, the electronic step of the calculation will never truly converge. However, this is what we expect in non-self-consistent calculations, as we only have one electronic step. In this PR, I hacked the `converged_electronic` method of the `Vasprun` class, which will return `True` in this senario.


yanghan234:
Tagging @utf for awareness.

shyuep:
Thanks,

PR #4072, opened 2024-09-17
fix chempot_diagram

yang-ruoxi:
## Summary
A strange behavior was caught for chemical potential diagram where the planes were crossed. 
Using the in-place [assignment expression](https://github.com/materialsproject/pymatgen/blob/150f816c55be80dd146cb20ebbf5f930633c8aeb/src/pymatgen/analysis/chempot_diagram.py#L649), the plot shows below: 

Screenshot 2024-09-17 at 2 38 07 PM

It it not obvious how it could cause the breakage. I spent sometime debugging but with no luck, it could be something much deeper. For now I suggest we revert it to the original form till we have a real solution. Would much appreciate a quick release due to breakage for MP...

### DIFF
--- a/src/pymatgen/analysis/chempot_diagram.py
+++ b/src/pymatgen/analysis/chempot_diagram.py
@@ -646,7 +646,7 @@ def simple_pca(data: np.ndarray, k: int = 2) -> tuple[np.ndarray, np.ndarray, np
     Returns:
         tuple: projected data, eigenvalues, eigenvectors
     """
-    data -= np.mean(data.T, axis=1)  # centering the data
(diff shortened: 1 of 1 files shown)


shyuep:
Can you try reverting the entire commit doing the in place assignment change and see if it solves the problem on your machine?

yang-ruoxi:
reverting this particular line solved the issue. Would you prefer reverting the entire commit for this PR?

shyuep:
OK, if it already fixes it. I will merge now. Thanks for the fix.

yang-ruoxi:
thank you!

PR #4073, opened 2024-09-18
Test source distribution and binary distribution (`.whl`) before release to PyPI

DanielYang59:
### Summary

- Test source distribution and binary distribution (`.whl`) before release to PyPI, currently the content of either source distribution or binary distribution is not check before pushing to PyPI, might lead to breakage if we miss anything


pyproject.toml
```diff
@@ -99,7 +99,7 @@ ci = ["pytest-cov>=4", "pytest-split>=0.8", "pytest>=8"]
 docs = ["invoke", "sphinx", "sphinx_markdown_builder", "sphinx_rtd_theme"]
 optional = [
     "ase>=3.23.0",
-    "beautifulsoup4",
+    "beautifulsoup4>=4.6.0",
```
DanielYang59, line 102: Now all dependencies in ci,optional have a lower bound, [but still get](https://github.com/materialsproject/pymatgen/pull/4073/checks):
```
warning: The direct dependency `pymatgen` is unpinned. Consider setting a lower bound when using `--resolution-strategy lowest` to avoid using outdated versions.
warning: The direct dependency `pymatgen[ci]` is unpinned. Consider setting a lower bound when using `--resolution-strategy lowest` to avoid using outdated versions.
warning: The direct dependency `pymatgen[optional]` is unpinned. Consider setting a lower bound when using `--resolution-strategy lowest` to avoid using outdated versions.
``

DanielYang59: Resolved

PR #4076, opened 2024-09-23
Install `BoltzTraP2` directly through `uv`

DanielYang59:
### Summary

- Install `BoltzTraP2` directly through `uv` as `BoltzTraP2` now support PEP-517 with build isolation
- [ ] ~~Look into why install failed for Windows~~


DanielYang59:
I tried but still cannot figure out why build for `BoltzTrap2` fails for Windows runners. But in any case, this wouldn't worsen our tests as it was installed for Ubuntu-only before.

PR #4078, opened 2024-09-24
JDFTx IO module

benrich37:
## Summary

This PR has been divided into the following 3 PR's

- 1:  (#4187) (Code change) Addition of `valences` property to ElementBase and Species
- 2:  (#4189) (Code addition) Addition of `inputs.py` module to `pymatgen.io.jdftx` and helper modules for representing and parsing JDFTx inputs
- 3:  (#4190) (Code addition) Addition of `outputs.py`module to `pymatgen.io.jdftx` and helper modules for representing and parsing JDFTx outputs

This PR can be ignored until those PR's are approved

## Todos

- Add methods which interface between input and output objects (namely creating input objects from output objects)
- Write broader inputs and outputs class to house JDFTXInfile and JDFTXOutfile along with additional inputs/outputs information


mkhorton:
Thanks @benrich37, I will review this

benrich37:
> Thanks @benrich37, I will review this

@mkhorton Sorry didn't see your comment, will pause on new commits until you're finished. Thank you!

src/pymatgen/io/jdftx/README.MD
```diff
@@ -0,0 +1,35 @@
+# JDFTx io
```
shyuep, line 1: I would suggest not putting this in a separate README, but rather in the actual `__init__.py` file. Speaking of which, you need to add a `__init__.py` for this module. There is no way this code would actually be properly imported.

src/pymatgen/io/jdftx/data.py
```diff
@@ -0,0 +1,80 @@
+"""Module containing reference data for JDFTx.
+
+This module contains reference data for JDFTx.
+"""
+
+from __future__ import annotations
+
+atom_valence_electrons = {
```
shyuep, line 8: Is this actually different from what you can get from the Element class?

src/pymatgen/io/jdftx/jeiters.py
```diff
@@ -0,0 +1,432 @@
+"""Module for JEiters class object.
```
shyuep, line 1: I am not sure what is the difference between this and the jeiter module? Plural vs singular?

shyuep (commented):
I have left some initial comments. Given this is a large PR, I will review it in steps.

src/pymatgen/io/jdftx/README.MD
benrich37: Should be fixed in my review_edits branch - let me know when you are done reviewing master in its current state and I will merge my new changes to master for this fork. Also thank you for taking the time to review this!

src/pymatgen/io/jdftx/data.py
benrich37: No - this dictionary has been removed. Functions needing a list of the element symbols enumerate through the Element class. Functions needing the number of valence electrons refer to the "get_atom_valence_electrons" method within data.py - the "valence" property raises ValueError for elements with mixed valence shells, so it cannot be used to sum up the number of valence electrons. (side note - can we add a "valences" property to Element to avoid this issue?)

src/pymatgen/io/jdftx/jeiters.py
benrich37: Yes - the original plan for JEiters was to be a list of JEiter with a few added properties and methods, but I had trouble figuring out how to get the inheritances right so now they are stored under slices instead. I've improved the docustring somewhat to make the purpose of these modules more clear.

src/pymatgen/io/jdftx/README.MD
benrich37: Sorry for confusion - wasn't sure if it would be cleaner to wait until you finished or to merge fixes as we went. Fixes to your comments are now merged to master (pinging for re-review just in case)

src/pymatgen/io/jdftx/data.py
mkhorton: You can directly propose changes to `Element` in this PR, this is ok as long as it is sensible and tested. This will be a better strategy than, say, a custom `get_atom_valence_electrons` function that lives alone inside only within the JDFTx code. The only caveat is not to break backwards compatibility when modifying core functionality.

src/pymatgen/io/jdftx/jeiters.py
mkhorton: The `JEiter` object looks like it might work well inheriting as a [`dataclass`](https://docs.python.org/3/library/dataclasses.html).

Can you explain the name `JEiter`, is it intended to be iterable?

I'd double check your `str` types too in case any of them might be better as `Literal`, if they can only take one of several values (rather than an arbitrary string).

For `JEiters`, I think the name is a bit confusing, perhaps this is also better kept inside the same file as `JEiter`. Note that `JEiters(List[JEiter])` should also work if you wanted to just add a few convenience methods to your list of `JEiter`.

mkhorton:
@benrich37 thanks for this PR and for your patience! Is the current branch up-to-date, and are there are specific areas of concern?

src/pymatgen/io/jdftx/data.py
benrich37: Awesome! I have added the "valences" property to ElementBase and Species class, as well as changed the "valence" property to call "valences" but to throw an error for mixed valence species as it did before. It is currently passing implicit tests through the JDFTx io testing, but I will add explicit testing in the appropriate place next. I have also modified some syntax to pre-existing code (but left its original form as comments) in the core.periodic_table.py file as pre-commit would not allow some of the code already there.

src/pymatgen/io/jdftx/jeiters.py
benrich37: The purpose of the JEiter class is just to hold the data pertaining to a single SCF iteration - I see now that the naming is confusing, I will refactor JEiter(s) to JElStep(s) for clarity. I've also added dataclass inheritance and all the pre-existing tests still pass - do you know if there's anything in particular I should test for in creating an inheritor of dataclass?

I also can migrate the JEiters (or now JElSteps) to JEiter.py. I also originally wrote the JEiters class to be defined as List[JEiter], but I believe I abandoned this approach either due to complications in how properties are inherited or in clarifying typing to one of the pre-commit programs. (the same fate also applied to JOutStructures and JDFTXOutfile)

benrich37:
> @benrich37 thanks for this PR and for your patience! Is the current branch up-to-date, and are there are specific areas of concern?

Hey! Thank you for your comments! Sorry I hadn't checked this PR in a few days. My main concern for this PR is its readability - I wrote the JDFTx output's data structure to be object oriented to help develop faster, but I can understand how the hierarchy of objects can make it difficult to understand what's going on. I feel confident with functionality itself though (aside from edge cases for JDFTx outputs that keep popping up)

computron:
From my side, can you explain the changes to periodic_table.py? Since it's a core class the changes should ideally be separated out from new functionality.

shyuep:
There is a failing test that I am pretty sure is related to the changes to periodic table.

benrich37:
> There is a failing test that I am pretty sure is related to the changes to periodic table.

Hi @shyuep ! Sorry yes - I thought it would be cleaner to have the Species valence(s) property defer to the Element valence(s) property, but I'm seeing now these need to be defined separately so that Species can incorporate oxidation state. Fixing right now.

Also @computron - that makes to me, I wasn't sure if it would be better to have internal redundancy to the periodic_table module in this PR or to update alongside this IO module. @mkhorton - would it be okay to go back to using the ad-hoc method in the io.jdftx.data module for the time being?

benrich37:
Quick update: some of the dataclasses are having issues calling attributes that used to work. Should be fixed soon

mkhorton:
>  I wasn't sure if it would be better to have internal redundancy to the periodic_table module in this PR or to update alongside this IO module.

Definitely let's try to minimize internal redundancy. That kind of thing can get out of control quickly and lead us towards a messy codebase.

I'll take another look at this soon, but if you want to arrange a quick call (~1hr) to review this PR I'm also open to that. I do not expect big issues but it might make it easier/quicker for us to get through it together.

benrich37:
@mkhorton  - all modules are ready to review!

src/pymatgen/core/periodic_table.py
```diff
@@ -204,7 +205,8 @@ def __getattr__(self, item: str) -> Any:
             if val is None or str(val).startswith("no data"):
                 warnings.warn(f"No data available for {item} for {self.symbol}")
                 val = None
-            elif isinstance(val, list | dict):
+            # elif isinstance(val, dict | list):
```
DanielYang59, line 208: I wish you don't mind me commenting :)

I'm not seeing any error from `pre-commit` for this line, perhaps you should double check your config?

Also do note their behaviour is not the same (`isinstance` would also consider childclass):
```python
class Foo(dict):
    def __init_subclass__(cls):
        return super().__init_subclass__()

user_dict = Foo(a=1)

print(isinstance(user_dict, dict))  # True
print(type(user_dict) in {dict, })  # False
```

src/pymatgen/core/periodic_table.py
```diff
@@ -523,7 +532,8 @@ def term_symbols(self) -> list[list[str]]:
         # Total ML = sum(ml1, ml2), Total MS = sum(ms1, ms2)
         TL = [sum(ml_ms[comb[e]][0] for e in range(v_e)) for comb in e_config_combs]
         TS = [sum(ml_ms[comb[e]][1] for e in range(v_e)) for comb in e_config_combs]
-        comb_counter = Counter(zip(TL, TS, strict=True))
+        # comb_counter: Counter = Counter(zip(TL, TS, strict=True))
```
DanielYang59, line 535: I'm not seeing any error here either: 

image

src/pymatgen/core/periodic_table.py
```diff
@@ -1639,9 +1656,18 @@ def get_el_sp(obj: int | SpeciesLike) -> Element | Species | DummySpecies:
             of properties that can be determined.
     """
     # If obj is already an Element or Species, return as is
-    if isinstance(obj, Element | Species | DummySpecies):
```
DanielYang59, line 1642: I don't really understand this change either?

DanielYang59, line 1642: I don't really understand this change?

src/pymatgen/core/periodic_table.py
benrich37: Hey Daniel! Thank you for your review! 

For my environment: I am running python 3.12 with a virtual environment set up using `uv` following the instructions on [the contributing page](https://pymatgen.org/contributing.html) (only with `uv create venv` replaced with `uv venv` as the syntax for uv has been updated)

For my issue: Running pre-commit with the line as `elif isinstance(val, list | dict)` raises the error within the `mypy` hook `error: Unsupported left operand type for | ("type[dict[Any, Any]]") [operator]`. I am seeing this is a problem on my end however, as I recently reformatted work computer to linux and this problem no longer occurs (previously I was developing on my personal mac laptop with an environment set up the same way). In both computers, the hooks for pre-commit was configured using the same pyproject.toml and pre-commit-config.yaml.

For the other issues: All the other changes to lines checking for object type was due to a similar or identical error raised mypy. I will check if these can also be fixed by just running pre-commit in linux

src/pymatgen/core/periodic_table.py
benrich37: Also fixed by running pre-commit on linux instead of mac, but for record-keeping the mypy errors on my mac for this line was `error: Unsupported left operand type for | ("type[dict[Any, Any]]")  [operator]` and `error: No overload variant of "zip" matches argument types "list[int]", "list[Union[float, Literal[0]]]", "bool"  [call-overload]`

src/pymatgen/core/periodic_table.py
benrich37: Also fixed by running on linux, but for record-keeping the mypy error on my mac for this was `Unsupported left operand type for | ("type[Element]")  [operator]`

benrich37:
@DanielYang59 previous weird edits I made to some of the existing code in periodic_table.py has been reverted - now I just need to figure out what's wrong with my laptop's execution of pre-commit! Thank you again for taking the time to review!

DanielYang59:
> Hey Daniel! Thank you for your review!

No problem, I'm not a maintainer, but I'm hoping my comments would be helpful.

> Running pre-commit with the line as elif isinstance(val, list | dict) raises the error within the mypy hook error: Unsupported left operand type for | ("type[dict[Any, Any]]") [operator]

This exactly suggests that the `pre-commit` is installed on python <= 3.9 interpreter.

I fully understand this Python module thing could be very confusing, here is how to resolve it.

---

### First of all, why this happened: 
As I don't have access to your env of course, my best guess is, you have multiple python interpreters install on your system (say one Python from the system package manager, for example `apt` for Ubuntu), and you installed `pre-commit` on your base interpreter. Then you create a virtual environment but didn't install `pre-commit`. If you run `pre-commit` at this point, it would come from your base Python interpreter, something like:

```bash
>>> which python3 
/usr/bin/python3  # the system installed base interpreter

# assume you haven't activated venv so far
>>> pip install pre-commit  

>>> source venv/bin/activate 

>>> which pre-commit
~/.local/bin/pre-commit  # pre-commit but from your base interpreter
```

### How to resolve this?
1. Install `pre-commit` on your venv interpreter as it would take higher priority (in general you might want to keep your base env clean)
2. Use explicit module run in the form of `path_to_python -m module_to_run`, something like `python3.12 -m pre-commit` to make sure pre-commit come from python 3.12 interpreter

src/pymatgen/core/periodic_table.py
DanielYang59: Yep that's again a python version issue, as `strict` is not available python <3.9, and the type overload doesn't exist.

.gitignore
```diff
@@ -27,6 +27,10 @@ setuptools*
 .*_cache
 # VS Code
 .vscode/*
+codecov*
+coverage.xml
+jdftx*.xml
```
DanielYang59, line 32: In general we might want to keep the project-level `.gitignore` clean  (admittedly it need some clean up) and only add global/general patterns as it would have project-level impact, `codecov/coverage` should fit into this but I don't think `jdftx*` or `actions-runner` do (I don't think many people try to trigger active locally).

Perhaps you could consider the following options:
- [Use a local "global" `.gitignore` file](https://stackoverflow.com/questions/7335420/global-git-ignore) for some very personal settings
  1. Write settings into a separate file somewhere in your local machine, say `~/.gitignore_global`
  2. Run `git config --global core.excludesFile '~/.gitignore_global'`
- Add a `.gitignore` file to a child directory if you really need to ignore some files for whatever reason, do note `.gitignore` could exist anywhere in the project (not just top level) and it would apply recursively to child directories

benrich37: This is very helpful - I will move those ignore lines soon to my global configuration. Also you were dead-on with the pre-commit problem! My laptop's default python3 is version 3.9 and I didn't realize the pre-commit I was using was installed there instead of the `uv` virtual environment. Installing the pip module onto the `uv` virtual environment and running `python -m pip install pre-commit` fixed everything! Thank you again for reviewing and helping me out on these issues! This has been very helpful for me.

DanielYang59: Yep co-existence of multiple python interpreters could be very confusing, great to know you sorted it out :)

---

Again I'm not a maintainer, I'm just sharing some personal experience here, and I would let you decide yourself:
- Personally I would try to avoid submitting a giant PR if possible, it would make it really hard to be reviewed (adding might be less an issue over changing) and also hard for your to work on it I believe (otherwise you could just finish a self-contained piece of it, instead of maneuvering the entire changeset all at once).

    Perhaps consider breaking this PR into atomic pieces, to start with say, you may want to break a PR by input files/output files (I don't have experience with JDFTx so I could be wrong here)
   
- Are all interfaces public (accessed by user)? If not, you might consider [adding an underscore for internal interfaces according to PEP 8](https://peps.python.org/pep-0008/#public-and-internal-interfaces) (e.g. `utils.py` -> `_utils.py`), a lot of methods seem to be "private" methods like `parse_ending_lines` -> `_parse_ending_lines`:
  > internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.
- `pymatgen` prefers [Google Style Python Docstrings](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html), looks like most part of your docstring adapts NumPy style

src/pymatgen/core/periodic_table.py
```diff
@@ -1639,6 +1654,7 @@ def get_el_sp(obj: int | SpeciesLike) -> Element | Species | DummySpecies:
             of properties that can be determined.
     """
     # If obj is already an Element or Species, return as is
+    # TODO: Why do we need to check "_is_named_isotope"?
```
DanielYang59, line 1657: Please try to avoid inserting a question as a comment for a commit as you're unlikely to get an answer this way.

For a general question, you could use the [discussions section](https://github.com/materialsproject/pymatgen/discussions), for something you suspect to be a bug, please [open an issue](https://github.com/materialsproject/pymatgen/issues)

Perhaps you could open an issue/discussion and discuss there for visibility?

---

I believe this reason is to guard against `DummySpecies` which doesn't have the `_is_named_isotope` attribute, @shyuep in case I'm wrong here. 

```python
from pymatgen.core import DummySpecies

d_sp = DummySpecies("X")
print(d_sp._is_named_isotope)  # AttributeError
```

If I were correct, perhaps this implementation should be tweaked to check type for readability (let's do this in a separate PR):

```diff
- if getattr(obj, "_is_named_isotope", None):
+ if isinstance(obj, DummySpecies):
```

Perhaps there could be exceptions? i.e. a `DummySpecies` with `_is_named_isotope` property?

DanielYang59: Thanks for writing up the discussion, also link to my previous comment https://github.com/materialsproject/pymatgen/pull/4078#discussion_r1841594048 in case you missed it

mkhorton:
> No problem, I'm not a maintainer, but I'm hoping my comments would be helpful.

They certainly are helpful! Really appreciate you stepping in here @DanielYang59 :)

src/pymatgen/core/periodic_table.py
mkhorton: @esoteric-ephemera do you know anything about the isotope handling here? Was it related to CIF parsing changes?

benrich37:
Note - I noticed from a merge conflict that some codespell exceptions were replaced by skipping certain files instead. To follow suit I'm replacing the exceptions I previously had in there with telling codespell to skip the out files in the test files for JDFTx IO

DanielYang59:
> Note - I noticed from a merge conflict that some codespell exceptions were replaced by skipping certain files instead. To follow suit I'm replacing the exceptions I previously had in there with telling codespell to skip the out files in the test files for JDFTx IO

Perhaps it makes sense to skip all **test files** (not tests) from `codespell` altogether? 

@benrich37 May I link to my previous comment again https://github.com/materialsproject/pymatgen/pull/4078#discussion_r1841594048? As far as I'm concerned:
- We might need to format docstring to Google Style 
- Currently all interfaces seem public, but I believe a portion of them should be private?

.gitignore
benrich37: Sorry for late reply to this! Not sure how I missed this - for the size of the PR, the main issue is that we need the IO module on pymatgen for the upcoming atomate2 paper. I like the idea of doing two separate pull requests for the input and output pieces of this module (and for the output piece, possibly breaking it down more by adding the "ground-level" io parsing objects first and then adding the objects that hold them (ie JElStep, followed by JElSteps, followed by JOutStructure, followed by JOutStructures, followed by JDFTXOutfileSlice, and then finally JDFTXOutfile)). As for how to do this, I'm imagining creating a new fork and slowly building up the content on the new fork from this fork by just copy-pasting. Before I start on this, do you know of a less manual approach? (ie a way to sync certain files between forks?) The JDFTx IO module on my fork is currently being used so it's important that the files shared on the new fork are always identical to the one on this fork.

For the public vs private interfaces, this I can definitely improve upon, there's a lot in there that the user is never supposed to interact with.

For the docustring style, updating to Google Style should not take too much time so I can get on that.

I can also update the directory names for the test files

DanielYang59: > Sorry for late reply to this! Not sure how I missed this 

No problem at all, I must have left too many comments :)

---

> for the size of the PR, the main issue is that we need the IO module on pymatgen for the upcoming atomate2 paper.

I'm not holding very strongly to break every PR into small pieces, and I believe this is less an issue for "adding" code over "changing" code. But I image working on a smaller changeset even for adding would be easier for you as well.


> As for how to do this, I'm imagining creating a new fork and slowly building up the content on the new fork from this fork by just copy-pasting. Before I start on this, do you know of a less manual approach? 

I'm not sure if I really understand your need here, but sounds like you don't really need a new [fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo) (copy of the entire repo) but a [new branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches) (branches are independent as well).

> (i.e. a way to sync certain files between forks?) The JDFTx IO module on my fork is currently being used by other students in my group so it's important that the files shared on the new fork are always identical to the one on this fork.

Again I don't really understand your need, because if you want two **identical** forks, why not working on the same fork (but different branches) directly?
- Or if you really need two identical copies, you might want to look at the [submodules](https://github.blog/open-source/git/working-with-submodules/), this is kind of like creating a soft link to another branch. 
- Or you could use git commands like [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) to "copy" a commit across branches/forks, or [`restore`](https://git-scm.com/docs/git-restore) to copy all changes to a certain file (`git restore --source=source-branch source_file.txt`)
- Or if you don't need **identical** copies but "updated" copies (the latest change/top on one fork/branch is pulled to another regularly), you might want to setup a [GitHub Action](https://docs.github.com/en/actions) to "pull/push" regularly

---

> For the public vs private interfaces, this I can definitely improve upon, there's a lot in there that the user is never supposed to interact with.

> For the docustring style, updating to Google Style should not take too much time so I can get on that.

Thanks a lot for the work, link to an [example Google Style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html) in case you're not familiar with this. Also I just updated the `CONTRIBUTION.md` to include this example.

benrich37: Oh right! I thought all pull requests from a fork had to come from the main branch - thank you again for this tip! It sounds like at the bare minimum then I should create a branch which only retains the edits to periodic_table.py, create a pull requests for those changes, and put this PR on pause until those changes get merged?

DanielYang59: > I thought all pull requests from a fork had to come from the main branch

Great you figured this out, yes you could freely select the source/target and repo/branch from the drop down menu when [creating a PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

image

---

> It sounds like at the bare minimum then I should create a branch which only retains the edits to periodic_table.py, create a pull requests for those changes, and put this PR on pause until those changes get merged?

For creating a branch, I don't think there's much restriction. For creating a PR, I believe you would need have differences (i.e. you could not create a PR from a source branch that is identical as the target branch). 

Also I don't think you need to "wait" or put any work on hold until another is merged, you could freely work on multiple branches/PRs in parallel (even if their changes overlap, but you need to resolve merge conflict in this case).

I think the most convenient way for you might be:
1. Create several feature branches from the **upstream master** (as you have put all your changes to your local master branch): `git checkout -b feature-branch upstream/master`
2. Migrate change for `periodic_table.py`(or anything you think should be packed as a separate PR) to this `feature-ptable`, and create a PR from this `feature-ptable` branch.
3. Repeat the same for JDFTx inputs/outputs, and in the end you get three independent feature branches.

To migrate all change in step 2, if you don't want to learn more `git` command for now ([`git restore`](https://git-scm.com/docs/git-restore)), I recommend just making a local copy of your current master branch which contents all change (otherwise you cannot find them when you `checkout` to another feature branch). And selectively copy/paste files to corresponding feature branch (say you copy and overwrite `periodic_table.py` to `feature-ptable` branch and so on).

Like the following:
```
upstream/master
       |

branch-ptable
       |
       o---o---o  <- Independent commits

branch-JDFTx-inputs
       |
       o---o---o  <- Independent commits 

branch-JDFTx-outputs
       |
       o---o---o  <- Independent commits 
```

benrich37: Thank you for walking me through this! I have created three new pull requests  (#4187, #4189, and #4190) from the "feature" branches as you've described, and set up some bash scripts on my local machine to run through `git restore` to keep them all in sync for the intended files. I will (unless anyone recommends against it) keep this PR open as I am hoping to add more methods in the input and output sections which reference one another, but I will hold off on adding them until the new PR's are approved for merging. (pinging @mkhorton @shyuep since this conversation accidentally got buried by new commits)

DanielYang59: Glad to know :)

> set up some bash scripts on my local machine to run through git restore to keep them all in sync for the intended files

Just curious why would you want to work on the `master` branch and "redistribute" them to feature branches? The target for creating a PR could be a non-master branch in case you didn't know this (your co-workers could merge their changes to your other branches as well). I guess working directly on those feature branches would be much convenient?

benrich37: I'm not too married to the idea of working on `master` and distributing respective changes to the feature branches - my thought right now is that the most important thing is that there exists at least one branch with all the required features present and working properly for my co-workers to use (they're doing a hefty amount debugging for this branch, but none are actually working directly on pymatgen - one is working on a fork of atomate2 and another on custodian). So for me it seems easiest to continue working on the branch of pymatgen that they're also using (where I can make sure new features will work for them and found bugs are properly squashed) and then distribute the changes to the feature branches where I can double check all the feature-specific test-checks are still passing.

DanielYang59: > my thought right now is that the most important thing is that there exists at least one branch with all the required features present and working properly for my co-workers to use

Thanks for letting me know, do work in a way that's easy for you. Feel free to ping me if there's anything I could help with.

benrich37:
Closing this PR until planned changes to integrate inputs and outputs submodules are fully developed

PR #4082, opened 2024-09-26
SLME Bug Fixes

kavanase:
This fixes some bugs in the `slme` code:
- Avoids overwriting the input data, which was leading to SLME results changing every time the function was (repeatedly) called.
- Removes some hard-coded settings for the plot and adds some labels


mkhorton:
Thanks @kavanase! Will merge. 

@knc6, notifying yourself too as a courtesy since I think you originally contributed this functionality, just in case the bug has affected you.

mkhorton:
Realised we should have asked for a test modification here too, but the fix seems clear enough to me.

PR #4093, opened 2024-10-03
Correctly compute rotational symmetry number for D*h point group

afonari:
Correct value is 2.

Example is H2. Also see See, for example, Table 10.1 and Appendix B of C. Cramer “Essentials of Computational Chemistry”, 2nd Ed.


shyuep:
Thanks.

PR #4097, opened 2024-10-08
Fix `selective_dyn` handling issue in `python>=3.12`

kavanase:
Hi,
I noticed I was getting some deprecation warnings like these, when parsing `vasprun.xml` files with `python=3.12`:
```
[DeprecationWarning("Testing an element's truth value will always return True in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead."), DeprecationWarning("Testing an element's truth value will always return True in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.")]
```

I traced this down to this line (https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/io/vasp/outputs.py#L1465).
With this small update, it fixes `selective_dyn` handling (as mentioned in the warning, this would always be set to `True` with future versions) and avoids these deprecation warnings.


shyuep:
Thanks.

PR #4101, opened 2024-10-10
bump `chgnet` to support numpy 2, skip failing `matgl` test for `torch` 2.4.1

DanielYang59:
### Summary

- Bump `chgnet` to support Numpy2

### Conflicting `torch` version

- `chgnet 0.4.0` requires `torch > 2.4.1` to be NumPy2 compatible https://github.com/CederGroupHub/chgnet/pull/202#discussion_r1756944575 
- `matgl` requires `torch <= 2.2.1` for `dgl` to work correctly https://github.com/materialsproject/pymatgen/pull/4101#issuecomment-2404278607

Should we skip two failing tests for `matgl` here as I assume this should be fixed from the `dgl` side? 
Input appreciated.


pyproject.toml
```diff
@@ -73,9 +73,7 @@ dependencies = [
     "tabulate>=0.9",
     "tqdm>=4.60",
     "uncertainties>=3.1.4",
-    # NumPy documentation suggests pinning the current major version as the C API is used
-    # https://numpy.org/devdocs/dev/depending_on_numpy.html#runtime-dependency-version-ranges
-    "numpy>=1.25.0,<3",
```
DanielYang59, line 78: I believe it should be [safe to remove the upper bound](https://numpy.org/devdocs/dev/depending_on_numpy.html#runtime-dependency-version-ranges) (especially considering numpy 3 wouldn't come anytime soon):
> Most libraries that rely on NumPy will not need to set an upper version bound: NumPy is careful to preserve backward-compatibility.

I think the original manual might have been misread here (ping @njzjz in case I'm wrong here):
> That said, if you are (a) a project that is guaranteed to release frequently, (b) use a large part of NumPy’s API surface, and (c) is worried that changes in NumPy may break your code, you can set an upper bound of <MAJOR.MINOR + N with N no less than 3, and MAJOR.MINOR being the current release of NumPy

If some project is very concerned (I assume not us) with numpy compatibility, they should pin `<MAJOR.MINOR + N` (not `<3`), which should be `<2.4.0` if current version is `2.1.0`.

DanielYang59:
The unit test failure:
>  E           FileNotFoundError: Cannot find DGL C++ graphbolt library at /home/runner/micromamba/envs/pmg/lib/python3.12/site-packages/dgl/graphbolt/libgraphbolt_pytorch_2.4.1.so

Seem to be related to DGL doesn't have that particular shared object for torch 2.4.1:
```
~/opt/miniconda3/envs/pymatgen/lib/python3.12/site-packages/dgl/graphbolt$ ls libgraphbolt_pytorch_2.*
libgraphbolt_pytorch_2.0.0.so  libgraphbolt_pytorch_2.1.0.so  libgraphbolt_pytorch_2.1.2.so  libgraphbolt_pytorch_2.2.1.so
libgraphbolt_pytorch_2.0.1.so  libgraphbolt_pytorch_2.1.1.so  libgraphbolt_pytorch_2.2.0.so
```

should be related to https://github.com/dmlc/dgl/issues/7247

pyproject.toml
njzjz: > If you use the NumPy C API (directly or via Cython), you can also pin **the current major version** to prevent ABI breakage.

image

DanielYang59: Cool! Thanks for the input. It's a bit unclear to me what "pin the current major version" refers to? I thought it refers to constrain the major numpy version to 2 (`2.0.0 <numpy<3`)?

But in any case, I don't think numpy 3 would be a concern anytime soon, what do you think?

njzjz: [NEP 53](https://numpy.org/neps/nep-0053-c-abi-evolution.html) has some explanations. In short, the ABI change in Numpy 2.0 is limited, and NumPy 2 ABI can be used with NumPy 1, but NumPy 1 ABI cannot be used with NumPy 2. However, there is no promise for the future NumPy version. For example, NumPy 3 ABI may or may not be able to be used with NumPy 2 - this is only known when NumPy 3 is developed.

DanielYang59: Yes I agree with that, but if I understand that manual correctly, a project should either:
- pin `<MAJOR.MINOR + N` (currently as `< 2.4`), or
- pin the current major version (`2 <= numpy < 3`)

Currently we are doing: `1.25 < numpy < 3`, if I understand correctly, "pin the current major version" refers to "setting the numpy major version as exactly 2" instead of "setting an upper bound for the major version"?

shyuep:
Since chgnet is already implemented in matgl, I would rather we unify support for everything that matgl supports using matgl.

DanielYang59:
Great to know, thanks! But I don't any experience with matgl or chgnet, perhaps need someone with more expertise on this to migrate the implementation

PR #4100, opened 2024-10-10
Support python3.13

DanielYang59:
### Summary

- Support python 3.13, to close #4284
- [ ] Remove all `pybtex` check after #4361
- [ ] Need to fix the EOS test failure on Ubuntu 3.13 (seemingly related to numerical precision but cannot recreate locally)
- [ ] Revert all platforms tested on Python 3.13 in `test.yml` test matrix (c9cc898301a7198488de660829a63ea62fafef3a)

#### Optional dependencies that miss Python 3.13 support:
- [ ] `openff-toolkit` (conda source)
- [ ] `chgnet`

PR #4102, opened 2024-10-10
pin python 3.12 in CI and `requires-python`

DanielYang59:
### Summary

- Pin Python 3.12 in CI for now, as it seems a lot of dependencies don't support Python 3.13 yet #4100


JaGeo:
The last set of tests never started. Thus, we are stuck here. Any thoughts, @DanielYang59 ?

DanielYang59:
> The last set of tests never started. Thus, we are stuck here. 

I don't know the exact reason, but my guess is the unstarted set is the cached set for our current config (python > 3.10 for Ubuntu). As the finished number of tests is exactly what we asked ` 42 successful, 1 skipped, and 10 expected checks `, the last 10 expected checks look like some kind of "ghost" from previous cache.

I didn't find any related reports in GitHub Community so I might open one later.

ab5424:
@DanielYang59 @JaGeo I think that the branch protection rules require a check under the name `test (ubuntu-latest, >3.10, lowest-direct, ci,optional)` to pass (hence the "Required" tag at the end of the line) but since that doesn't exist anymore so it can't report its status. I saw the same thing in one of my projects. The only workaround I know is to temporarily remove that rule from the repo setting.

DanielYang59:
@ab5424 Bravo! you are the savior.

> I think that the branch protection rules require a check under the name

I don't have write access to the repo, but sounds like the branch protection rules are dynamically generated based the test matrix from the master branch?

> The only workaround I know is to temporarily remove that rule from the repo setting.

Maintainers should be able to do a manual (force) merge (not using auto-merge)?

DanielYang59:
For some reason I'm still seeing "ghost" expected checks from other PR where the config should have been updated on master branch, perhaps the branch protection rule is hard coding to expect `ubuntu-latest, >3.10`? I cannot check myself as I don't have access

PR #4104, opened 2024-10-11
VASP Bugfixes: Volumetric data and XDATCAR parsing for monatomic structures

esoteric-ephemera:
Two bug fixes:
- For VASP 6.4.2 and up, all POSCAR-style strings include both the POTCAR symbol (e.g., `Mg_pv` rather than the elemental symbol `Mg`) and part of the POTCAR hash (`Si/79d9987ad87` rather than `Si`). While the hashing is already accounted for, the underscore was not. Fixed here, impacts CONTCAR and all volumetric data (CHGCAR, LOCPOT, AECCAR*)
- For any monatomic XDATCAR where the title of the POSCAR substring is the same as the single chemical species, the XDATCAR parsing functionality of pymatgen doesn't work. This PR fixes that behavior, adds a test for it, and cleans up the XDATCAR parsing a bit


esoteric-ephemera:
@shyuep and @mkhorton should be ready for review whenever you have time. I had to make some version changes to linting / testing in CI to get the tests to set up correctly and run (newer version of mypy throws a lot of errors that should be handled separately and the linux build failed because of py3.13)

shyuep:
Thanks. This looks good to me. Merged.

.github/workflows/lint.yml
```diff
@@ -18,11 +18,11 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.x'
+          python-version: '3.11'
```
DanielYang59, line 21: @esoteric-ephemera May I ask why do we pin the python version for lint workflow? I believe just pin `mypy` version should suffice?

I would fix these new `mypy` errors in #4110, and I might revert this pin?

esoteric-ephemera: Seemed to fix a lot of linting misfires but feel free to revert if I'm wrong. Not critical to this PR

DanielYang59: I guess the new linting issues is related to the new `mypy` 1.12 release (not the python version)

> Not critical to this PR

Absolutely, I would fix in #4110, thanks!

PR #4107, opened 2024-10-13
Properly test `ase` not installed error, use single `skip` mark for module level test skip

DanielYang59:
### Summary

- Properly test `ase` not installed error, the original test would not actually run because `ase` is installed in CI (ideally the test should be independent of whether that package is installed or not)
- Also fix not installed test for `seekpath` in `io.pwmat.inputs`
- Use module level skip when package not available


tests/io/test_ase.py
```diff
@@ -355,10 +338,14 @@ def test_msonable_atoms():
...
 def test_no_ase_err():
-    import pymatgen.io.ase
+    import pymatgen
 
     expected_msg = str(pymatgen.io.ase.NO_ASE_ERR)
-    with pytest.raises(PackageNotFoundError, match=expected_msg):
-        pymatgen.io.ase.MSONAtoms()
+
+    # Mock the import statement to raise ImportError when 'ase' is imported
+    with mock.patch("builtins.__import__", side_effect=PackageNotFoundError(expected_msg)):
```
DanielYang59, line 347: This patch should be improved to only raise error for import ase

tests/util/test_graph_hashing.py
```diff
@@ -37,14 +37,11 @@
 from __future__ import annotations
 
 import networkx as nx
-import pytest
 
+from pymatgen.util.graph_hashing import weisfeiler_lehman_graph_hash, weisfeiler_lehman_subgraph_hashes
 
-def test_graph_hash():
-    pytest.importorskip("emmet.core.graph_hashing")
```
DanielYang59, line 44: need confirm on this, [I'm not seeing `graph_hasing` from `emmet`](https://github.com/materialsproject/emmet/tree/main/emmet-core/emmet/core) but `pymatgen.util.graph_hashing`, perhaps these are relocated at some point?

https://github.com/materialsproject/pymatgen/blob/668f1aaffcaa0f3f56872b6eff93f27366acb356/src/pymatgen/util/graph_hashing.py#L76

https://github.com/materialsproject/pymatgen/blob/668f1aaffcaa0f3f56872b6eff93f27366acb356/src/pymatgen/util/graph_hashing.py#L151

tests/io/pwmat/test_inputs.py
```diff
@@ -129,10 +131,10 @@ def test_write_file(self):
         assert tmp_high_symmetry_points_str == high_symmetry_points.get_str()
 
 
-def test_err_msg_on_seekpath_not_installed(monkeypatch):
```
DanielYang59, line 132: Should I understand correctly, this doesn't really patch the import, but either:
- passes directly if `seekpath` is available, or
- capture error if not

PR #4113, opened 2024-10-16
Fix arg passing in `inverse` property of `SymmOp`

### LINKED ISSUE #4023: Bug in inverse of MagSymmOp
### Python version

Python 3.11.7

### Pymatgen version

2023.7.18

### Operating system version

_No response_

### Current behavior

When the `inverse` of `MagSymmOp` is called,  an error, `TypeError: MagSymmOp.__init__() missing 1 required positional argument: 'time_reversal'`, happens. This is probably because, in the definition in `class SymmOp`,
```py
@property
def inverse(self) -> Self:
    """Inverse of transformation."""
    inverse = np.linalg.inv(self.affine_matrix)
    return type(self)(inverse)
```
`type(self)` does not take a `time_reversal` argument.

### Expected Behavior

returns a `MagSymmOp` instance whose `affine_matrix` is inversed.

### Minimal example

_No response_

_No response_

DanielYang59:
### Summary

- Fix arg passing in `inverse` property of `SymmOp`, to fix #4023
- Replace some `np.array` with `np.asarray` for efficiency where we don't rely on creating a copy


src/pymatgen/core/operations.py
```diff
@@ -244,7 +244,9 @@ def translation_vector(self) -> np.ndarray:
     def inverse(self) -> Self:
         """Inverse of transformation."""
         inverse = np.linalg.inv(self.affine_matrix)
-        return type(self)(inverse)
+        if isinstance(self, MagSymmOp):
+            return MagSymmOp(inverse, self.time_reversal, self.tol)
+        return type(self)(inverse, self.tol)
```
DanielYang59, line 249: another subtle bug: the `tol` is lost

src/pymatgen/core/operations.py
```diff
@@ -243,12 +244,16 @@ def translation_vector(self) -> np.ndarray:
     @property
     def inverse(self) -> Self:
         """Inverse of transformation."""
-        inverse = np.linalg.inv(self.affine_matrix)
-        return type(self)(inverse)
+        new_instance = copy.deepcopy(self)
```
DanielYang59, line 247: strictly speaking we don't need to make a deep copy here as the only mutable attribute is overwritten by `np.linalg.inv(self.affine_matrix)` which would generate a new instance anyway, but I still decide to make deep copy to avoid potential breakage should any mutable attribute be added later.

PR #4122, opened 2024-10-18
Make `Incar` keys case insensitive, fix init `Incar` from dict val processing for str/float/int

### LINKED ISSUE #4119: VaspInput setter and Incar.check_params() are inconsistent
### Python version

Python 3.12.4

### Pymatgen version

2024.7.18

### Operating system version

_No response_

### Current behavior

When I set INCAR parametrs via `vasp_input['INCAR']['PARAM']=value`, they do not always pass `Incar.check_params()` checks.

Using the example below, I am getting:

```
...: BadIncarWarning: ENCUT: 550 is not a float
  vasp_input.incar.check_params()
...: BadIncarWarning: GGA: Cannot find Ps in the list of values
  vasp_input.incar.check_params()
```

### Expected Behavior

Setting the parameters passes the checks.

### Minimal example

```Python
from pymatgen.io.vasp.inputs import VaspInput
vasp_input = VaspInput.from_directory('.')
vasp_input['INCAR']['GGA'] = "MK"
vasp_input['INCAR']['GGA']
```

I used the following INCAR
```
ALGO = Normal
ENCUT = 550
ISMEAR = 0
NCORE = 16
SIGMA = 0.04
SYSTEM = C.hexagonal.191
```
Other requried inputs may be anything.

DanielYang59:
### Summary

- Fix `Incar.check_params` handling of case sensitivity, to fix #4119, also add unit tests
- [x] Issue warning if duplicated keys found when init `Incar`

### Make more operations (get/update/setdefault) in `Incar` case insensitive

- [x] Init `Incar` from dict (also the `from_dict` method) doesn't seem to use setter method, might be better off subclass UserDict
- [x] Update Incar class docstring to record the change in behaviour
- [x] Test other dict methods like update/setdefault/...

### Init `Incar` from dict doesn't call setter method, leading to unprocessed str/float/int value 

- [x] Setter method only trigger `proc_val` when val is string, leading to inconsistent behaviour when init from dict for int/float
- [x] Setter method not triggered on `update` method when subclassing builtin `dict`, leading to val not passing through `proc_val` when init from dict


DanielYang59:
ping the true expert @esoteric-ephemera in case you have any comments :) Thanks in advance!

src/pymatgen/io/vasp/inputs.py
```diff
@@ -734,7 +735,8 @@ def __init__(self, params: dict[str, Any] | None = None) -> None:
                     val.append(params["MAGMOM"][idx * 3 : (idx + 1) * 3])
                 params["MAGMOM"] = val
 
-            self.update(params)
```
DanielYang59, line 737: I have to drop the `update` method to use the setter method, I don't think this would break anything as the dict is just init. But still good to get more eyes on this change.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -744,7 +746,7 @@ def __setitem__(self, key: str, val: Any) -> None:
         """
         super().__setitem__(
             key.strip(),
-            type(self).proc_val(key.strip(), val.strip()) if isinstance(val, str) else val,
```
DanielYang59, line 747: I believe this is a subtle bug, where only `str` type would go through the `proc_val` preprocessor.

However there's another point that is unclear to me now, as the `proc_val` should be able to handle `list` keys as well (the `from_str` method): https://github.com/materialsproject/pymatgen/blob/32d9bebbebf24043bdf56683333053dcb1980d23/src/pymatgen/io/vasp/inputs.py#L867-L885

https://github.com/materialsproject/pymatgen/blob/32d9bebbebf24043bdf56683333053dcb1980d23/src/pymatgen/io/vasp/inputs.py#L847-L865

tests/io/vasp/test_inputs.py
```diff
@@ -638,6 +639,33 @@ def test_as_dict_and_from_dict(self):
         incar3 = Incar.from_dict(dct)
         assert incar3["MAGMOM"] == [Magmom([1, 2, 3])]
 
+    def test_from_file_and_from_dict(self):
```
DanielYang59, line 642: I didn't see much documentation, wondering should init directly from a key: value dict be allowed at all (in contrast to the MSONable dict)?

JaGeo:
I btw recently had issues with updating the incar. I would have expected that sigma and SIGMA would simply modify SIGMA but they did not. Instead, i had two different sigmas in the incar. I did not have time to investigate but it was a very recent pymatgen version

esoteric-ephemera:
@JaGeo did VASP parse them as different variables or just take the one defined furthest down the INCAR?

shyuep:
Yes, all tags are case insensitive. I don't think the Incar class actually checks. So if you set `sigma` and `SIGMA`, you end up with two sigma tags. The easiest solution is to automatically capitalize the key in setitem for Incar. I would use Camelcase for the values for Incar (even though it makes no difference where VASP is concerned). Usually VASP just looks at the first letter. E.g., ALGO = Accurate and ALGO=A means the same thing.

DanielYang59:
> did VASP parse them as different variables or just take the one defined furthest down the INCAR?

I believe VASP parses the first occurrence (can you confirm it's the bottom one?), but I could be wrong here. I tested with INCAR:
```
encut = 500
ENCUT = 400
``` 

And `grep ENCUT vasprun.xml` would return: `<i name="ENCUT">    500.00000000</i>`

DanielYang59:
> Why do this exactly?

> I don't think the Incar class actually checks. So if you set sigma and SIGMA, you end up with two sigma tags. The easiest solution is to automatically capitalize the key in setitem for Incar. I would use Camelcase for the values for Incar

I think I proposed that we cast Incar keys to upper case internally in order to prevent such duplicate and make internal checking easier.

Also the reason I proposed upper case over others (capitalize/lower) being:
- Most (if not all) our internal recordings (`incar_parameters.json`, and all kinds of recordings inside `vasp.inputs` class) use upper case, so we don't need to update them all
- VASP internally cast all keys to upper case when parsing (see the top comment in source code `drdatab.F`)
- VASP wiki also use upper case, so I assume it's more or less a convention?

DanielYang59:
Thanks for the helpful discussion and the quick fix @shyuep I would finish the remaining part of this PR

JaGeo:
@shyuep Thanks for the fast fix!

DanielYang59:
Just mention this in case you have any comments, it seems [the `ML_MODE` tag](https://www.vasp.at/wiki/index.php/ML_MODE) is an exception of the case insensitivity (which I never touch) added in #3625: https://github.com/materialsproject/pymatgen/blob/ed8542d8ebf18c0fa49e6d87386a52072bc31559/src/pymatgen/io/vasp/inputs.py#L932

Per the above VASP wiki:
> Warning: If any option other than the above is chosen or there is a spelling error (be careful to write everything in upper case or lower case letters) the code will exit with an error.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -709,10 +710,12 @@ class BadPoscarWarning(UserWarning):
     """Warning class for bad POSCAR entries."""
 
 
-class Incar(dict, MSONable):
+class Incar(UserDict, MSONable):
```
DanielYang59, line 713: Subclassing `UserDict` would make our life much easier, otherwise some methods like `update/setdefault` would not call the overridden setter method, and I believe that's [the recommended way to create a custom dict](https://docs.python.org/3/library/collections.html#collections.UserDict):
> [UserDict](https://docs.python.org/3/library/collections.html#collections.UserDict): wrapper around dictionary objects for easier dict subclassing

Similar to: https://github.com/materialsproject/pymatgen/blob/ed8542d8ebf18c0fa49e6d87386a52072bc31559/src/pymatgen/io/lobster/inputs.py#L56

Perhaps we should consider write a custom and reusable key case-insensitive dict at some point instead of writing this over and over again.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -709,43 +710,66 @@ class BadPoscarWarning(UserWarning):
...
+            and trailing white spaces.
+        - Cast keys to upper case.
         """
-        super().__setitem__(
-            key.strip().upper(),
-            type(self).proc_val(key.strip(), val.strip()) if isinstance(val, str) else val,
-        )
+        key = key.strip().upper()
+        # Cast float/int to str and feed into proc_val to clean up their types
+        val = self.proc_val(key, str(val)) if isinstance(val, str | float | int) else val
```
DanielYang59, line 759: This change:
```diff
- val = self.proc_val(key, str(val)) if isinstance(val, str) else val
+ val = self.proc_val(key, str(val)) if isinstance(val, str | float | int) else val
```

should have a healthy dose of stupidity as I believe casting int/float to str such that it would go through `proc_val` to sanitize the type again, otherwise they would bypass the `proc_val` check, leading to inconsistent behaviour init from dict and from str/file.

src/pymatgen/io/vasp/outputs.py
```diff
@@ -1718,11 +1719,11 @@ def __init__(
                     if tag == "generator":
                         self.generator = self._parse_params(elem)
```
DanielYang59, line 1720: Here is a well hidden bug, after we cast keys to upper case, we cannot use `Incar` to pack "generator" tags anymore, which is not INCAR tags:
```
{'program': 'vasp', 'version': '5.4.4.18Apr17-6-g9f103f2a35', 'subversion': '(build Apr 25 2017 00:43:00) complex            parallel', 'platform': 'LinuxIFC', 'date': '2017 09 29', 'time': '11:34:32'}
```

src/pymatgen/io/vasp/outputs.py
```diff
@@ -345,7 +345,7 @@ def __init__(
                 self.update_potcar_spec(parse_potcar_file)
                 self.update_charge_from_potcar(parse_potcar_file)
 
-        if self.incar.get("ALGO") not in {"CHI", "BSE"} and not self.converged and self.parameters.get("IBRION") != 0:
+        if self.incar.get("ALGO") not in {"Chi", "Bse"} and not self.converged and self.parameters.get("IBRION") != 0:
```
DanielYang59, line 348: Looks like before migrating to `UserDict`, the `proc_val` is not triggered through `update` method when init from a dict (instead of str), therefore these values remain as it when it should be capitalized.

shyuep:
Thanks.

DanielYang59:
No problem, thanks for the helpful discussion everyone

yantar92:
I think that one special key is SYSTEM - it is usually free text and capitalizing it can be awkward.

DanielYang59:
> I think that one special key is SYSTEM - it is usually free text and capitalizing it can be awkward.

That's a good point to me. On top of this, I think the `__eq__` for `Incar/Kpoint` should be overwritten to ignore the comment/`SYSTEM` for similar reason (docstring should be updated to reflect this behaviour of course). 

The use case might be comparing two INCARs to see if they would generate the same result, as comment doesn't really impact anything.

Free feel to push back on this if there's any reason that I'm missing, if not I might take on this in the upcoming days :)

PR #4123, opened 2024-10-18
fix: preserve site properties over cell transform

Lattay:
Make sure site properties are properly transferred from the original cell to the primitive or refined one.

## Summary

Major changes:

- fix 1: in `SpacegroupAnalyzer.get_refined_cell`, the flag `keep_site_properties` (used notably by `get_conventional_standard_structure`) would not behave as expected before because it used `number` returned by `spglib.refine_cell` to index into the `properties` (`properties` has one element per site, but `number` indexes into `_unique_species` which has one element per occupied irreducible position).
- fix 2: I noticed the same pattern in `SpacegroupAnalyzer.find_primitive`

You can verify that the new tests wouldn't pass before the change to `SpacegroupAnalyzer`.

## Checklist

- [-] (no relevant changes) Type annotations included. Check with `mypy`.
- [-] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))


shyuep:
Thanks

PR #4124, opened 2024-10-18
doc: make the test procedure explicit

Lattay:
The most common usual workflow with `pytest` is just `pip install . && pytest`, but this doesn't apply to pymatgen because of the test files.
Since I had to dig (just a bit) in the sources to get `pytest` to run I figured I could help the next contributor by making it clear how to point to the files using an environment variable.


DanielYang59:
I could be wrong here, but I don't think we need to do both, either we install pymatgen in editable mode to make the test files available, or we set `PMG_TEST_FILES_DIR` env var to point to the test file directory

shyuep:
You are supposed to install pymatgen in editable mode if you are a dev.

PR #4127, opened 2024-10-21
Fix gaussian input parser, and add test

sio-salt:
## Summary

Major changes:

- fix 1: In src/pymatgen/io/gaussian.py, within the from_str() method, the variable route_index is initially set to None and later assigned an integer representing the line number in the file. The code used `if route_index:` to check if route_index was set to a value other than None. However, when route_index is assigned 0, the condition if route_index: incorrectly evaluates to False. To fix this, I changed the condition to `if route_index is not None:`. This bug occurred when the Gaussian input file has no Link0 section and has the route section spanned multiple lines with only the first line starting with #, the parser did not read the file correctly.

- fix 2: Added the test function test_from_str_no_link0() for a Gaussian input file that previously caused an error in from_str(). This test ensures that the parser correctly handles input files without a Link0 section and with multi-line route sections.

- fix 3: Included example Gaussian input file for the .com extension used on Linux. This file, as described in Fix 1, has no Link0 section, has a multi-line route section, and only the first line of the route section starts with #.

## Checklist


shyuep:
Thanks

Issue #4028, opened 2024-09-02
CifParser taking forever without throwing an error


anyangml:
### Python version

python 3.11.5

### Pymatgen version

2024.7.18

### Current behavior

I have some cif files generated by diffusion models, and they are likely to be invalid. I am trying to use CifParser (default setting) to read those files, and save the structure if the structure turns out to be valid. For some of the files, it takes forever for CifParser to read the file without throwing an error. Below is an example 
```
# generated using pymatgen
data_V23Ni
_symmetry_space_group_name_H-M   'P 1'
_cell_length_a   1.49244463
_cell_length_b   3.79421616
_cell_length_c   30.42398643
_cell_angle_alpha   20.62129211
_cell_angle_beta   83.61568451
_cell_angle_gamma   62.99439240
_symmetry_Int_Tables_number   1
_chemical_formula_structural   V23Ni
_chemical_formula_sum   'V23 Ni1'
_cell_volume   0.00000000
_cell_formula_units_Z   1
loop_
 _symmetry_equiv_pos_site_id
 _symmetry_equiv_pos_as_xyz
  1  'x, y, z'
loop_
 _atom_site_type_symbol
 _atom_site_label
 _atom_site_symmetry_multiplicity
 _atom_site_fract_x
 _atom_site_fract_y
 _atom_site_fract_z
 _atom_site_occupancy
  V  V0  1  0.67231035  0.73907024  0.30680549  1
  V  V1  1  0.34705946  0.37159106  0.52686268  1
  V  V2  1  0.36653438  0.38854167  0.52153718  1
  V  V3  1  0.00921166  0.93152356  0.19919881  1
  V  V4  1  0.95677882  0.91787136  0.49960217  1
  Ni  Ni5  1  0.51781291  0.50244284  0.32125005  1
  V  V6  1  0.09629190  0.08547926  0.87447387  1
  V  V7  1  0.49141514  0.52637476  0.74801213  1
  V  V8  1  0.78377295  0.75774127  0.46150076  1
  V  V9  1  0.86805952  0.95672750  0.67616075  1
  V  V10  1  0.17673326  0.06560791  0.04854067  1
  V  V11  1  0.09662235  0.06218648  0.66348130  1
  V  V12  1  0.35039228  0.35385907  0.34141171  1
  V  V13  1  0.78324890  0.72533828  0.17591193  1
  V  V14  1  0.96514380  0.07838918  0.98198533  1
[...]

PR #4134, opened 2024-10-25
Clean up `neighbors.pyx` variable name

DanielYang59:
### Summary

- Minor clean up `neighbors.pyx` variable name, remnant from #4028, **I don't expect any functional changes**
- Use `unsigned int` (`npy_uint32`) over `int` for variables that we're certain would not be negative (counter, index when looping through "number of points/dimensions")

### DIFF
--- a/src/pymatgen/optimization/neighbors.pyx
+++ b/src/pymatgen/optimization/neighbors.pyx
@@ -78,9 +78,11 @@ def find_points_in_spheres(
             mask]
 
     cdef:
-        int i, j, k, l, m
-        double[3] maxr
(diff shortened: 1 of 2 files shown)


src/pymatgen/optimization/neighbors.pyx
```diff
@@ -78,15 +79,19 @@ def find_points_in_spheres(
             mask]
 
     cdef:
-        int i, j, k, l, m
-        double[3] maxr
```
DanielYang59, line 82: this `maxr` (and the corresponding `get_max_r` function) could be very easily mistaken for "maximum radius" in the near neighbor context (it's actually max repetition)

PR #4136, opened 2024-10-26
Read INCAR `SYSTEM` as is, `check_params` use `proc_val`

DanielYang59:
### Summary

- Manually update `incar_parameters.json`, to fix https://github.com/materialsproject/pymatgen/issues/4119#issuecomment-2431859839, need a more systematic script to update `incar_parameters.json` in the long run
- `check_params` use `proc_val`
- Remove some unnecessary casting in INCAR values as we're confident they are capitalized now
- Avoid casing INCAR  `SYSTEM` keyword, to fix https://github.com/materialsproject/pymatgen/pull/4122#issuecomment-2431216473


src/pymatgen/io/vasp/outputs.py
```diff
@@ -800,10 +800,10 @@ def run_type(self) -> str:
             "--",
             "None",
         }:
-            incar_tag = self.incar.get("METAGGA", "").strip().upper()
```
DanielYang59, line 803: I believe this would be covered by https://github.com/materialsproject/pymatgen/blob/6992aeef9847b03da81c80dabe27c84010e28abc/src/pymatgen/io/vasp/inputs.py#L1029

src/pymatgen/io/vasp/outputs.py
```diff
@@ -484,7 +484,7 @@ def _parse(
                             or elem.attrib["comment"]
                             == "INVERSE MACROSCOPIC DIELECTRIC TENSOR (including local field effects in RPA (Hartree))"
                         ):
-                            if self.incar.get("ALGO", "Normal").upper() == "BSE":
+                            if self.incar.get("ALGO", "Normal") == "Bse":
```
DanielYang59, line 487: After #4122 we should be confident that string values are now capitalized, these case casting should not be needed.

DanielYang59:
@yantar92 Hope this looks good to you

src/pymatgen/io/vasp/outputs.py
yantar92: If they are capitalized, why comparing with "Bse" (non-capitalized)

DanielYang59: Sorry I may not understand your question? `Bse` is capitalized?

yantar92: Sory, I got completely confused. Mostly because of apparent inconsistency between the code and VASP wiki. If you look at https://www.vasp.at/wiki/index.php/ALGO, you will see that BSE and many other values are all capitals:
> ALGO = Normal | VeryFast | Fast | Conjugate | All | Damped | Subrot | Eigenval | Exact | None | Nothing | CHI | G0W0 | GW0 | GW | scGW0 | scGW | G0W0R | GW0R | GWR | scGW0R | scGWR | ACFDT | RPA | ACFDTR | RPAR | BSE | TDHF

So, the code reads awkward now.

yantar92:
> a script is certainly needed but I don't have time to work on this at this moment.

https://github.com/materialsproject/pymatgen/issues/4119#issuecomment-2439617604-permalink

src/pymatgen/io/vasp/outputs.py
DanielYang59: > I got completely confused. Mostly because of apparent inconsistency between the code and VASP wiki

Yes I find it confusing too https://github.com/materialsproject/pymatgen/pull/4122#issuecomment-2421278991, and @esoteric-ephemera  said it's a convention to pymatgen https://github.com/materialsproject/pymatgen/pull/4122#issuecomment-2422769620

This is added likely since the start: 
image

As VASP itself is largely case insensitive, we just need to internally standardize for easy handling (no need to cast to upper/capitalize when comparing values and so on). I personally prefer to cast both value and tag to uppercase for the following reason:
- Get rid of the headache to memorize "key is uppercase", but "case is capitalized" and so on
- Most values (just like keys) are first-letter abbreviations of some sort, capitalizing them doesn't make much sense, though there're values that is indeed proper words like "Fast/Conjugate/All/..."

yantar92: > Yes I find it confusing too https://github.com/materialsproject/pymatgen/pull/4122#issuecomment-2421278991, and @esoteric-ephemera said it's a convention to pymatgen https://github.com/materialsproject/pymatgen/pull/4122#issuecomment-2422769620

I see, and I agree that abstracting things out into case-sensitive/insensitive class is the most robust way to go.

As for the convention, it annoyingly has exceptions:
```python
class Incar(UserDict, MSONable):
    """
    A case-insensitive dictionary to read/write INCAR files with additional helper functions.

    - Keys are stored in uppercase to allow case-insensitive access (set, get, del, update, setdefault).
    - String values are capitalized by default, except for keys specified
        in the `lower_str_keys` of the `proc_val` method.
    """
```
As you can see, capitalization is documented, except "keys specified in lower_str_keys". Currently it is a single key - `ML_MODE`.

However, do note that `Incar.check_params` does not expect such exceptions:
```python
            if allowed_values is not None:
                # Note: param_type could be a Union type, e.g. "str | bool"
                if "str" in param_type:
                    allowed_values = [item.capitalize() if isinstance(item, str) else item for item in allowed_values]
```
(aside: incar_parameters.json does not even list `ML_MODE`)

DanielYang59: > As you can see, capitalization is documented, except "keys specified in lower_str_keys". Currently it is a single key - ML_MODE.

Yep I updated the docstring to reflect this key/value case issue, but I'm not sure about the casing of `ML_MODE` either (I thought VASP is largely case insensitive ) https://github.com/materialsproject/pymatgen/pull/4122#issuecomment-2423472984

I never use that tag so cannot comment, it would be great if you (or someone else) could confirm this behaviour.

> However, do note that Incar.check_params does not expect such exceptions:

Thanks I would update it (one exception makes everything hard to handle)

yantar92: > I never use that tag so cannot comment, it would be great if you (or someone else) could confirm this behaviour.

I do not use this tag either, but clearly its value is case-sensitive - #3625 
I looked up https://www.vasp.at/wiki/index.php/ML_MODE and it says

>> Warning: If any option other than the above is chosen or there is a spelling error (be careful to write everything in upper case or lower case letters) the code will exit with an error.

I conclude that case-insensitivity is not universal.

Ironically, simply upcasing (rathen than capitalizing) everything would not require making an exception for ML_MODE.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -905,72 +965,15 @@ def from_str(cls, string: str) -> Self:
                     params[key] = cls.proc_val(key, val)
         return cls(params)
 
-    @staticmethod
-    def proc_val(key: str, val: str) -> list | bool | float | int | str:
+    @classmethod
+    def proc_val(cls, key: str, val: str) -> list | bool | float | int | str:
```
DanielYang59, line 969: Have to move these recordings (list_keys/bool_keys/...) as class variable for `check_params` to access them, also make `proc_val` class method. 

I don't think this would be breaking but still good for others to double check

src/pymatgen/io/vasp/outputs.py
DanielYang59: > I do not use this tag either, but clearly its value is case-sensitive - https://github.com/materialsproject/pymatgen/pull/3625

Yes I did see this PR and the VASP comment as well.

>  I conclude that case-insensitivity is not universal.
> Ironically, simply upcasing (rathen than capitalizing) everything would not require making an exception for ML_MODE.

I believe ML_MODE is the only exception for now. upcase everything (except for SYSTEM) would make everything easy, but I guess ML_MODE was not there when this part of code was written (12 years ago), and it's hard to change it now without breakage

yantar92: > t's hard to change it now without breakage

100% agree.

shyuep:
I am merging this. In the end, we are using the 80/20 rule. VASP keys are always all caps. VASP values are a mess of different things (int, floats, case-insensitive strings, case-sensitive strings, etc.). For now, we will just use capitalize since that works for most of the cases, e.g., ALGO=Fast,Normal, etc. I get that BSE should be full caps since it is an acronym but since it makes no difference to actual functionality, we are going to live with it.

PR #4141, opened 2024-10-27
VaspDoc.get_incar_tags: Use Mediawiki API

yantar92:
* src/pymatgen/io/vasp/help.py (VaspDoc.get_incar_tags): Use Mediawiki API instead of parsing the HTML source directly.  The old approach is not stable against changes in the tag list because of the way URLs are constructed.  pagefrom= parameters start from certain tag, which is not guaranteed to provide the complete tag list as the new tags are added before that tag given in pagefrom=.  At the moment of writing this commit, PRECFOCK tag is already missed using the old approach.

Following up:  https://github.com/materialsproject/pymatgen/issues/4119#issuecomment-2439617604-permalink

- [X] Install `pre-commit` hooks to auto-check types and linting before every commit

### DIFF
--- a/src/pymatgen/io/vasp/help.py
+++ b/src/pymatgen/io/vasp/help.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import re
 
 import requests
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks!

PR #4143, opened 2024-10-28
Improve element mismatch handling with POTCAR for `Poscar.from_file/str`

### LINKED ISSUE #4140: Poscar.from_file throws a cryptic error for VASP5-like POSCARs when POTCAR does not contain all the present elements
### Python version

Python 3.12.7

### Pymatgen version

Version: 2024.10.22

### Operating system version

_No response_

### Current behavior

Consider the following scenario:
1. We have a POSCAR file in VASP5 format with, say, C and Na sites
2. We have an obsolete POTCAR file only specifying C potentials
3. We try to read the POSCAR using Poscar.from_file

When I do try it, I get an error like:
```
  File "/home/yantar92/.venv/lib/python3.12/site-packages/pymatgen/io/vasp/inputs.py", line 282, in from_file
    return cls.from_str(file.read(), names, read_velocities=read_velocities)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yantar92/.venv/lib/python3.12/site-packages/pymatgen/io/vasp/inputs.py", line 478, in from_str
    struct = Structure(
             ^^^^^^^^^^
  File "/home/yantar92/.venv/lib/python3.12/site-packages/pymatgen/core/structure.py", line 3956, in __init__
    super().__init__(
  File "/home/yantar92/.venv/lib/python3.12/site-packages/pymatgen/core/structure.py", line 1051, in __init__
    raise StructureError(f"{len(species)=} != {len(coords)=}")
pymatgen.core.structure.StructureError: len(species)=1 != len(coords)=721
```

Such error is extremely hard to understand.

I note that the error is not exactly a bug - docstring of from_file method does say that POTCAR symbols take priority by default. However, (1) it is very hard to understand what is going on from the error thrown; (2) the existing behaviour does not make much sense for VASP5 POSCARs where the symbol names are explicitly specified.

### Expected Behavior

Ideally, I'd like Poscar.from_file fall back to using POSCAR symbols if POTCAR does not specify them all and throw a warning.

### Minimal example

```Python
from pymatgen.io.vasp.inputs import Poscar, Potcar
Poscar.from_file('POSCAR')
```

[POSCAR.tar.gz](https://github.com/user-attachments/files/17531141/POSCAR.tar.gz)

I cannot share the POTCAR file because of license terms, but here is the spec:
```python
[PotcarSingle(symbol='C', functional='PBE', TITEL='PAW_PBE C 08Apr2002', VRHFIN='C: s2p2', n_valence_elec=4)]
```

DanielYang59:
### Summary

- The glob logic of POTCAR matching might need some attention, as files other than `POTCAR`  might take higher priority than `POTCAR`
- Suppress a lot of expected warning in VASP.inputs tests, to avoid warning pollution

### For element mismatch in POTCAR and POSCAR, to fix #4140:
- If is able to overwrite, a warning would be issued (previously would silently overwrite)
- If unable to overwrite as reported in #4140 (fewer elements in POTCAR than POSCAR), a more semantic value error would be raised


DanielYang59:
@yantar92 This situation seems a bit trickier than I thought (I don't have much experience trying to handling an incompatible POTCAR). I'm still finding a cleaner way to handle this.

`check_for_potcar` doesn't really check the content (elements) of a POTCAR, but more likely: 
1. check existence (by calling `Potcar.from_file()`); 
2. then generate a `default_names`: https://github.com/materialsproject/pymatgen/blob/4b10dd3fb37d8012d46dee58678ac3a695845104/src/pymatgen/io/vasp/inputs.py#L278-L280

---

And it turns out its the `default_names` generated from POTCAR that caused the cryptic error: https://github.com/materialsproject/pymatgen/blob/4b10dd3fb37d8012d46dee58678ac3a695845104/src/pymatgen/io/vasp/inputs.py#L418-L428


In your specific case, the `n_atoms` is `[1, 720]` (corresponding to your POSCAR) but the `default_names` is `["C", ]` therefore there would be an `IndexError` raised but caught (I believe this is sort of intended, see the comment in the above code block)

This block also generates `atomic_symbols` as `["C", ]`, and propagates to the `Structure` init and leads to the final exception we saw (and the length of `atomic_symbols` and `coords` mismatch).

https://github.com/materialsproject/pymatgen/blob/4b10dd3fb37d8012d46dee58678ac3a695845104/src/pymatgen/io/vasp/inputs.py#L485-L492

---

At this stage I believe not suppress the `IndexError` but re-raise as a semantic `ValueError` is the best solution (also confirmed this would not break any test for `vasp.inputs/outputs`), unless there's any reason not to do so?

src/pymatgen/io/vasp/inputs.py
```diff
@@ -346,12 +346,23 @@ def from_str(
...
+        # ...
+        # atomic_symbols is ["H", "H", "O"]
+        # symbols is ["H", "O"]
         atomic_symbols: list[str] = []
 
         try:
             n_atoms: list[int] = [int(i) for i in lines[5].split()]
             ipos: int = 6
 
+            symbols: list[str] = []
```
DanielYang59, line 364: To prevent `symbols` from being unbound

src/pymatgen/io/vasp/inputs.py
```diff
@@ -424,8 +435,12 @@ def from_str(
...
                     atomic_symbols.extend([default_names[idx]] * n_atom)
                 vasp5_symbols = True
-            except IndexError:
-                pass
+            except IndexError as exc:
+                # TODO: need discussion on exception handling
+                warnings.warn(
+                    f"Elements in POTCAR {default_names} don't match POSCAR {symbols}", BadPoscarWarning, stacklevel=2
+                )
+                raise ValueError(f"Elements in POTCAR {default_names} don't match POSCAR {symbols}") from exc
```
DanielYang59, line 443: Is there any use case the need this `IndexError` being suppressed instead of interrupted?

src/pymatgen/io/vasp/inputs.py
```diff
@@ -345,15 +346,25 @@ def from_str(
...
         atomic_symbols: list[str] = []
+        vasp5or6_symbols: bool = False
+        symbols: list[str] | None = None
 
         try:
             n_atoms: list[int] = [int(i) for i in lines[5].split()]
             ipos: int = 6
 
         except ValueError:
-            vasp5_symbols = True
```
DanielYang59, line 356: I guess this tag `vasp5_symbols` (added 2012) is used to differentiate VASP4 and VASP5/6? Instead of VASP 5 vs VASP 6?

image

I have absolutely zero knowledge about VASP 4.x (seemingly before 2008), and didn't even manage to find a VASP 4 POSCAR.

Anyone with experience on VASP 4.x please free feel to comment :)

yantar92:
> At this stage I believe not suppress the IndexError but re-raise as a semantic ValueError is the best solution

Well. Reading again through the code of Poscar.from_str, it looks like its default behaviour is very permissive - it even assigns arbitrary elements in some scenarios:

>         ... This will ensure at least a
>         unique element is assigned to each site and any analysis that does not
>         require specific elemental properties should work fine.

Throwing a warning as you did is 100% helpful.
Whether to throw a ValueError or not, I am not sure. Maybe you can ask the maintainers?

DanielYang59:
Thanks for the comment.

>  Whether to throw a ValueError or not, I am not sure. Maybe you can ask the maintainers?

I didn't fully understand the code when I left this comment, but I'm now pretty sure re-raising when not alter the original behaviour, only makes the error more understandable.

---

I rewrote the code logic to the following:

```python
if default_names is not None and symbols is not None:
    if len(symbols) > len(default_names):
        raise ValueError(f"{default_names=} (likely from POTCAR) has fewer elements than POSCAR {symbols}")

    # Overwrite only when elements are different
    if default_names[: len(symbols)] != symbols:
        atomic_symbols = []
        for idx, n_atom in enumerate(n_atoms):
            atomic_symbols.extend([default_names[idx]] * n_atom)

        warnings.warn(f"Elements in POSCAR would be overwritten by {default_names=}", stacklevel=2)
```

Basically there're three branches:
- If the POTCAR (`default_names`) doesn't have enough elements, in the original code it could cause an `IndexError` which is suppressed. But this would leave a partially finished `atomic_symbols` which is then passed to `Structure` init and eventually leading to the cryptic error message you observed.
- If the POTCAR has enough elements:
    - Elements would be overwritten if they're different, now a warning is inserted.
    - In the original implementation elements would be rewritten even if they're the same, this effectively does nothing but would now by skipped for efficiency.

---

For some reason `enumlib` test would fail meaning there must have been conditions not covered by VASP test, I would have a look.

src/pymatgen/io/vasp/inputs.py
```diff
@@ -435,6 +435,8 @@ def from_str(
         # This is in line with VASP's parsing order that the POTCAR
         # specified is the default used.
         if default_names is not None and symbols is not None:
+            vasp5or6_symbols = True
```
DanielYang59, line 438: I don't think this tag is doing anything as `symbols is not None` effectively filter VASP4 cases, but for some reason enumlib test might fail, so there must have been some cases that are missed

PR #4150, opened 2024-11-01
Added new Flag for `AutoOxiStateDecorationTransformation`

jmmshn:
# New flag for `AutoOxiStateDecorationTransformation`

The situation we have in mind is `AutoOxiStateDecorationTransformation` -> `OrderDisorderedStructureTransformation` pipeline.

The `OrderDisorderedStructureTransformation` expects oxidation numbers to help rank the structure guesses.  However, if you have metallic systems, the sequences of transformation breaks.  In those cases it's not super clear how the ordered structures should be ranked anyways unless you start invoking more sqs-type logic.  For the simple case I think we can allow the users to come up with some quick and dirty guess for ordered structures so they can start computing stuff before they might move on to more fine-grained structure selection.

Added a new `zeros_on_fail` flag.
If True and and  the `BVAnalyzer` fails to come up with a guess for the oxidation guess, we will set the all the oxidation states to zero.

### DIFF
--- a/src/pymatgen/transformations/standard_transformations.py
+++ b/src/pymatgen/transformations/standard_transformations.py
@@ -107,6 +107,7 @@ class AutoOxiStateDecorationTransformation(AbstractTransformation):
         max_radius=4,
         max_permutations=100000,
         distance_scale_factor=1.015,
+        zeros_on_fail=False,
     ):
         """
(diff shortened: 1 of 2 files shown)


shyuep:
Merged. Thanks.

PR #4154, opened 2024-11-05
Fix `borg.hive.SimpleVaspToComputedEntryDrone.assimilate` ValueError when core file missing

### LINKED ISSUE #4132: assimilate method in SimpleVaspToComputedEntryDrone is not raising ValueError
### Python version

Python 3.12.3

### Pymatgen version

2024.10.22

### Operating system version

Ubuntu 22.04 LTS

### Current behavior

assimilate method in apps.borg.hive.SimpleVaspToComputedEntryDrone is not raising the ValueError when checking if all the necessary files are present ("INCAR", "POTCAR", "CONTCAR", "OSZICAR", "POSCAR"). Instead, the method returns None, i think it happens because the raise keyword is placed inside the try statement.

### Expected Behavior

assimilate method should correctly detect the absence of required VASP files.

### Minimal example

```Python
from pymatgen.apps.borg.hive import SimpleVaspToComputedEntryDrone
drone = SimpleVaspToComputedEntryDrone()
entry = drone.assimilate(".")
print(entry)
```

_No response_

DanielYang59:
### Summary

- Fix `borg.hive.SimpleVaspToComputedEntryDrone.assimilate` ValueError when core file missing, to fix #4132
- Add unit test
- Improve type annotation

### DIFF
--- a/src/pymatgen/apps/borg/hive.py
+++ b/src/pymatgen/apps/borg/hive.py
@@ -89,7 +89,7 @@ class VaspToComputedEntryDrone(AbstractDrone):
         self,
         inc_structure: bool = False,
         parameters: list[str] | None = None,
-        data: list | None = None,
+        data: list[str] | None = None,
     ) -> None:
         """
         Args:
@@ -233,10 +233,10 @@ class SimpleVaspToComputedEntryDrone(VaspToComputedEntryDrone):
             filenames = {"INCAR", "POTCAR", "CONTCAR", "OSZICAR", "POSCAR", "DYNMAT"}
             if "relax1" in files and "relax2" in files:
                 for filename in ("INCAR", "POTCAR", "POSCAR"):
-                    search_str = f"{path}/relax1", f"{filename}*"
+                    search_str = f"{path}/relax1/{filename}*"
                     files_to_parse[filename] = glob(search_str)[0]
                 for filename in ("CONTCAR", "OSZICAR"):
-                    search_str = f"{path}/relax2", f"{filename}*"
+                    search_str = f"{path}/relax2/{filename}*"
                     files_to_parse[filename] = glob(search_str)[-1]
             else:
                 for filename in filenames:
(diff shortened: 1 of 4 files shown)


src/pymatgen/apps/borg/hive.py
```diff
@@ -228,7 +234,10 @@ def assimilate(self, path):
             else:
                 for filename in filenames:
                     files = sorted(glob(os.path.join(path, f"{filename}*")))
-                    if len(files) == 1 or filename in ("INCAR", "POTCAR") or (len(files) == 1 and filename == "DYNMAT"):
```
DanielYang59, line 231: The first condition (`len(files) == 1`) would short-circuit the last condition: https://github.com/materialsproject/pymatgen/blob/bd9fba9ec62437b5b62fbd0b2c2c723216cc5a2c/src/pymatgen/apps/borg/hive.py#L231

Perhaps it should be the following?
```python
if len(files) == 1 or filename in {"INCAR", "POTCAR", "DYNMAT"}:
```

src/pymatgen/apps/borg/hive.py
```diff
@@ -162,9 +168,6 @@ def get_valid_paths(self, path):
             return [parent]
         return []
 
-    def __str__(self):
-        return " VaspToComputedEntryDrone"
```
DanielYang59, line 166: Looks like here is an extra leading space?

src/pymatgen/apps/borg/hive.py
```diff
@@ -270,14 +279,17 @@ def assimilate(self, path):
                 return ComputedStructureEntry(structure, energy, parameters=param, data=data)
             return ComputedEntry(structure.composition, energy, parameters=param, data=data)
 
+        except ValueError as exc:
+            if "not all necessary files are present" in str(exc):
+                raise
+
+            # TODO: other value error should be suppressed?
```
DanielYang59, line 286: TODO tag

src/pymatgen/apps/borg/hive.py
```diff
@@ -220,15 +233,18 @@ def assimilate(self, path):
             filenames = {"INCAR", "POTCAR", "CONTCAR", "OSZICAR", "POSCAR", "DYNMAT"}
             if "relax1" in files and "relax2" in files:
                 for filename in ("INCAR", "POTCAR", "POSCAR"):
-                    search_str = f"{path}/relax1", f"{filename}*"
```
DanielYang59, line 223: This is likely a bug, `glob.glob` expect `str` instead of `tuple[str, str]`

PR #4155, opened 2024-11-05
Add informative error message to `SymmetryUndeterminedError`

kavanase:
After pulling my hair out for a little while, trying to figure out why I was getting `SymmetryUndeterminedError`s when trying to initialise `SpacegroupAnalyzer` with a particular structure (which currently don't come with any message), I realised that `spglib.get_error_message()` can be called to get an error message for why no result was obtained. 
This is a small update to include the `spglib.get_error_message()` message in `SymmetryUndeterminedError`, so the user has some idea why it failed.

In action:
image

image


mkhorton:
Appreciate the improvement, thanks @kavanase !

mkhorton (approved):

mkhorton:
Sorry for the delay in merging, auto-merge was broken and I hadn't realised. Should be fixed now.

PR #4159, opened 2024-11-09
Avoid using full equality (`==`) to compare float, avoid `assert_array_equal` compare float array

### LINKED ISSUE #4158: tests use equality to compare floating point numbers
Floating point numbers (angles) are compared in transformation tests at

https://github.com/materialsproject/pymatgen/blob/bd9fba9ec62437b5b62fbd0b2c2c723216cc5a2c/tests/transformations/test_advanced_transformations.py#L840

Using equality to compare floating point numbers inevitably fails
e.g. https://buildd.debian.org/status/fetch.php?pkg=pymatgen&arch=amd64&ver=2024.10.29%2Bdfsg1-1&stamp=1731061592&raw=0
```
        # make sure that the orthorhombic supercell is different from the cubic cell
        assert not np.array_equal(
            supercell_generator_cubic.transformation_matrix,
            supercell_generator_orthorhombic.transformation_matrix,
        )
        assert transformed_orthorhombic.lattice.abc != transformed_cubic.lattice.abc
        # only angels are expected to be the same because of force_90_degrees = True
>       assert transformed_cubic.lattice.angles == transformed_orthorhombic.lattice.angles
E       assert (89.999970160...0000000086139) == (89.999970160...0000000086142)
E         
E         At index 2 diff: 90.00000000086139 != 90.00000000086142
E         Use -v to get more diff

/<<PKGBUILDDIR>>/.pybuild/test_python3.12/tests/transformations/test_advanced_transformations.py:840: AssertionError
```

Floating point comparison needs to be used instead
```
assert_allclose(transformed_cubic.lattice.angles, lattice.angles)
```

DanielYang59:
### Summary

- Avoid using full equality `==` to compare float, to fix #4158
- Avoid [`assert_array_equal` on int array](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html):
    > Assert fails with numerical imprecision with floats:
    > Use [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) or one of the nulp (number of floating point values) functions for these cases instead:
- Tweak `_proj` implementation, ~3x speedup
- (Partially) Replace sequence of float comparison with `==` (list/tuple/dict ...): https://github.com/materialsproject/pymatgen/blob/bd9fba9ec62437b5b62fbd0b2c2c723216cc5a2c/tests/core/test_bonds.py#L56
- Other type/comment tweaks


src/pymatgen/transformations/advanced_transformations.py
```diff
@@ -1872,11 +1877,11 @@ def is_one_to_many(self) -> bool:
         return True
 
 
-def _proj(b, a):
+def _proj(b: NDArray, a: NDArray) -> NDArray:
     """Get vector projection (np.ndarray) of vector b (np.ndarray)
     onto vector a (np.ndarray).
     """
-    return (b.T @ (a / np.linalg.norm(a))) * (a / np.linalg.norm(a))
+    return (np.dot(b, a) / np.dot(a, a)) * a
```
DanielYang59, line 1884: This new implementation is slightly more readable and gives ~3x speedup:
```
Original Implementation Time: 601.05 ms
New Implementation Time: 182.84 ms
```


With script:
```python
import numpy as np
from numpy.typing import NDArray
from time import perf_counter_ns


def _proj_original(b: NDArray, a: NDArray) -> NDArray:
    return (b.T @ (a / np.linalg.norm(a))) * (a / np.linalg.norm(a))

def _proj_new(b: NDArray, a: NDArray) -> NDArray:
    return (np.dot(b, a) / np.dot(a, a)) * a

def verify_projection():
    a = np.random.rand(1000)
    b = np.random.rand(1000)
    proj1 = _proj_original(b, a)
    proj2 = _proj_new(b, a)
    assert np.allclose(proj1, proj2)

def benchmark_projections(n_iter=100000):
    a = np.random.rand(1000)
    b = np.random.rand(1000)

    # Measure original implementation
    start_time = perf_counter_ns()
    for _ in range(n_iter):
        _proj_original(b, a)
    time_original = perf_counter_ns() - start_time

    # Measure new implementation
    start_time = perf_counter_ns()
    for _ in range(n_iter):
        _proj_new(b, a)
    time_new = perf_counter_ns() - start_time

    print(f"Original Implementation Time: {time_original / 1e6:.2f} ms")
    print(f"New Implementation Time: {time_new / 1e6:.2f} ms")

verify_projection()

print("\nBenchmarking both implementations...")
benchmark_projections()

```

tests/analysis/xas/test_spectrum.py
```diff
@@ -38,22 +38,22 @@ def setUp(self):
         self.site2_xanes = XAS.from_dict(site2_xanes_dict)
 
     def test_e0(self):
-        assert approx(self.k_xanes.e0) == 7728.565
```
DanielYang59, line 41: I believe `a == pytest.approx(ref_val)` is the [recommended usage](https://docs.pytest.org/en/stable/reference/reference.html) plus it's slightly more readable. However do note `approx` is not symmetric:
> a == pytest.approx(b, rel=1e-6, abs=1e-12): True if the relative tolerance is met w.r.t. b or if the absolute tolerance is met. Because the relative tolerance is only calculated w.r.t. b, this test is asymmetric and you can think of b as the reference value. In the special case that you explicitly specify an absolute tolerance but not a relative tolerance, only the absolute tolerance is considered.

tests/analysis/chemenv/coordination_environments/test_coordination_geometries.py
```diff
@@ -89,7 +89,7 @@ def test_coordination_geometry(self):
         assert cg_oct.IUCr_symbol_str == "[6o]"
 
         cg_oct.permutations_safe_override = True
-        assert cg_oct.number_of_permutations == 720.0
```
DanielYang59, line 92: this should be int?

JaGeo: Yes.

DanielYang59: 😆 Thanks for chiming in, I was just going through all changes again

tests/io/lammps/test_data.py
```diff
@@ -562,18 +562,18 @@ def test_init(self):
...
         topo_override = Topology(sites=mol, ff_label="ff_map", charges=outer_charge, velocities=outer_velo)
         assert topo_override.type_by_sites == ["D"] * 10
-        np.testing.assert_array_equal(topo_override.charges, outer_charge)
-        np.testing.assert_array_equal(topo_override.velocities, outer_velo)
+        assert_array_equal(topo_override.charges, outer_charge)
+        assert_array_equal(topo_override.velocities, outer_velo)
         # test using a list of sites instead of SiteCollection
         topo_from_list = Topology(sites=mol.sites)
         assert topo_from_list.type_by_sites == topo.type_by_sites
-        np.testing.assert_array_equal(topo_from_list.charges, topo.charges)
```
DanielYang59, line 575: `assert_array_equal` usage across `io.lammps.test_data` is very likely incorrect, looks like float array

PR #4160, opened 2024-11-09
Clean up `io.vasp.outputs.Outcar` docstring

DanielYang59:
### Clean up `io.vasp.outputs.Outcar` docstring/comment

- Greatly enhance type annotation for parsers in `Outcar` with more specific types, clarify docstring: `Renders accessible from self.data` (the internal `self.data` dictionary) vs `Renders accessible from attributes` (instance attributes)
- Other docstring/comment cleanups

### Others
- Change `getattr` default value to `False` instead of `None` when used for condition check


src/pymatgen/io/vasp/outputs.py
```diff
@@ -2323,7 +2326,7 @@ def read_table_pattern(
         attribute_name: str | None = None,
         last_one_only: bool = True,
         first_one_only: bool = False,
-    ) -> list:
+    ) -> list:  # TODO: clarify table-like data type
```
DanielYang59, line 2329: Tag: important: clarify type of table-like data

src/pymatgen/util/io_utils.py
```diff
@@ -48,36 +52,45 @@ def clean_lines(
             yield clean_string
 
 
-def micro_pyawk(filename, search, results=None, debug=None, postdebug=None):
```
DanielYang59, line 51: > [!CAUTION]
> I didn't see the `debug/postdebug` being used across the entire code base, it's very likely some debug args, schedule them for removal after one-year grace period.

image

src/pymatgen/io/vasp/outputs.py
```diff
@@ -2605,7 +2703,7 @@ def read_nmr_efg_tensor(self) -> list[NDArray]:
         """Parses the NMR Electric Field Gradient Raw Tensors.
 
         Returns:
-            A list of Electric Field Gradient Tensors in the order of Atoms from OUTCAR.
+            list[NDArray]: Electric Field Gradient Tensors in the order of atoms.
```
DanielYang59, line 2706: double check type

src/pymatgen/io/vasp/outputs.py
```diff
@@ -2046,7 +2071,8 @@ def __init__(self, filename: PathLike) -> None:
                     e_wo_entrp = float(match[1])
                 if e0 is None and (match := e0_pattern.search(clean)):
                     e0 = float(match[1])
-            if all([nelect, total_mag is not None, efermi is not None, run_stats]):
+
+            if nelect is not None and total_mag is not None and efermi is not None and run_stats:
```
DanielYang59, line 2075: Tag: use truthy to check `run_stats` (dict[str, float | None]) might yield unexpected results as None in dict is also truthy

src/pymatgen/io/vasp/outputs.py
DanielYang59: Cannot firm its type as I never do this calculation and no test is available, but should be correct as per the function implementation of `make_symmetric_matrix_from_upper_tri`

src/pymatgen/util/num.py
```diff
@@ -22,14 +30,21 @@ def round_to_sigfigs(num, sig_figs):
     return round(num, prec)
 
 
-def make_symmetric_matrix_from_upper_tri(val):
-    """Given a symmetric matrix in upper triangular matrix form as flat array indexes as:
-    [A_xx,A_yy,A_zz,A_xy,A_xz,A_yz]
+def make_symmetric_matrix_from_upper_tri(val: ArrayLike) -> NDArray:
```
DanielYang59, line 33: Looks like this helpful function is designed to work for 3x3 matrix alone (rank-2 tensor), and wouldn't scale well as the `idx` is hard-coded, also it doesn't work along the diagonals, I would add a check to reject non 3x3 arrays

src/pymatgen/util/num.py
```diff
@@ -22,17 +30,30 @@ def round_to_sigfigs(num, sig_figs):
...
+         [A_xy, A_yy, A_yz],
+         [A_xz, A_yz, A_zz]]
+
+    Args:
+        val (ArrayLike): Flattened upper triangular elements.
+
+    Returns:
+        NDArray: The symmetric matrix.
     """
-    idx = [0, 3, 4, 1, 5, 2]
```
DanielYang59, line 31: I believe this helper function is designed to work for 3x3 matrix alone, as such I might just make things simple and reject non 3x3 arrays with `ValueError`.

The new implementation gives a ~3 speedup:
```
Original implementation time: 0.378586 seconds
Simplified implementation time: 0.118705 seconds
```


Test script (from GPT):
```python
import numpy as np
from timeit import timeit

def original_make_symmetric_matrix_from_upper_tri(val):
    """Original implementation."""
    idx = [0, 3, 4, 1, 5, 2]
    val = np.array(val)[idx]
    mask = ~np.tri(3, k=-1, dtype=bool)
    out = np.zeros((3, 3), dtype=val.dtype)
    out[mask] = val

    out.T[mask] = val
    return out

def simplified_make_symmetric_matrix_from_upper_tri(val):
    """Simplified implementation."""
    array = np.asarray(val)
    if array.shape != (6,):
        raise ValueError(f"Expected val of length 6, got {array.shape}")

    return np.array([
        [array[0], array[3], array[4]],
        [array[3], array[1], array[5]],
        [array[4], array[5], array[2]],
    ])

# Test input
val = [1, 2, 3, 4, 5, 6]

# Measure execution time
original_time = timeit(
    "original_make_symmetric_matrix_from_upper_tri(val)",
    globals=globals(),
    number=100000,
)
simplified_time = timeit(
    "simplified_make_symmetric_matrix_from_upper_tri(val)",
    globals=globals(),
    number=100000,
)

# Display results
print(f"Original implementation time: {original_time:.6f} seconds")
print(f"Simplified implementation time: {simplified_time:.6f} seconds")
```

src/pymatgen/io/vasp/outputs.py
DanielYang59: I would clarify the docstring instead, as the full type of a table is super long (`list[list[Any] | dict[str, Any]]`) and I don't want to add a type alias as it's used only once.

Also the type could vary if last_one_only or first_one_only is True, which would require writing 3 overloads

src/pymatgen/io/vasp/outputs.py
DanielYang59: I was wrong, test is available as `test_nmr_efg`

src/pymatgen/io/vasp/outputs.py
```diff
@@ -2768,23 +2894,23 @@ def read_neb(
         if self.data.get("tangent_force"):
             self.data["tangent_force"] = float(self.data["tangent_force"][0][1])
 
-    def read_igpar(self) -> None:
```
DanielYang59, line 2771: I didn't find a test for this parser method

PR #4163, opened 2024-11-10
Clarify `util.string.formula_double_format` return type, fix using `int` to round to nearest int

DanielYang59:
### Summary

- [ ] Should `formula_double_format` return string instead of float?
- Clarify `util.string.formula_double_format` return type as `Literal[""] | float`
- Add types for `util.string`
- **Bug fix**: `int` would not round to nearest int (but truncate decimal part): https://github.com/materialsproject/pymatgen/blob/bd9fba9ec62437b5b62fbd0b2c2c723216cc5a2c/src/pymatgen/util/string.py#L136-L137
https://github.com/materialsproject/pymatgen/blob/bd9fba9ec62437b5b62fbd0b2c2c723216cc5a2c/src/pymatgen/util/string.py#L129
- **Two minor behavior changes** for `formula_double_format`:
    1. Use `isclose` to round float very close to `1`:
  ```diff
  - if ignore_ones and afloat == 1:
  + if ignore_ones and math.isclose(afloat, 1, abs_tol=tol):  
  ```
  2. `tol` now includes the boundary (unit test added):
  ```diff
  - if abs(afloat - int(afloat)) < tol:
  + if math.isclose(afloat, round(afloat), abs_tol=tol):
  ```
  
  ```python
  print(math.isclose(1, 1+1E-8, abs_tol=1E-8))  # True
  ```

### DIFF
--- a/src/pymatgen/util/string.py
+++ b/src/pymatgen/util/string.py
@@ -138,16 +138,16 @@ def formula_double_format(
     ignore_ones: bool = True,
     tol: float = 1e-8,
 ) -> float | str:
-    """Make pretty formulas by formatting the amounts.
+    """Format a float for pretty formulas.
     E.g. "Li1.0 Fe1.0 P1.0 O4.0" -> "LiFePO4".
 
     Args:
-        afloat (float): a float.
+        afloat (float): The float to be formatted.
         ignore_ones (bool): if true, floats of 1.0 are ignored.
-        tol (float): Absolute tolerance to round to nearest int. i.e. 2.0000000001 -> 2.
+        tol (float): Absolute tolerance to round to nearest int. i.e. (2 + 1E-9) -> 2.
 
     Returns:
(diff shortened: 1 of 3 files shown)


src/pymatgen/util/string.py
```diff
@@ -97,57 +107,65 @@ def to_unicode_string(self):
...
     Returns:
-        A string representation of the float for formulas.
+        str | float: A string representation of the float for formulas.
     """
-    if ignore_ones and afloat == 1:
+    if ignore_ones and math.isclose(afloat, 1, abs_tol=tol):
         return ""
-    if abs(afloat - int(afloat)) < tol:
+    if abs(afloat - int(afloat)) <= tol:
         return int(afloat)
```
DanielYang59, line 155: Here is a bug, `int()` would not round to "nearest" integer, it would just truncates the decimal part:
```python
print(int(2.9))  # 2
```

PR #4167, opened 2024-11-12
Fix: BSPlotterProjected._Orbitals_SumOrbitals() method

ShengLin1001:
**Fix**: Ensure BSPlotterProjected._Orbitals_SumOrbitals() checks if sum_morbs is not None, as a NoneType object cannot be iterated.

### Major changes:

- Fixed the BSPlotterProjected._Orbitals_SumOrbitals() method to handle cases where sum_morbs is None.

## Description
When running the following code, I encountered an error in the get_projected_plots_dots_patom_pmorb() method, which uses _Orbitals_SumOrbitals(). The error occurred because sum_morbs was None, making it non-iterable.

This fix addresses the issue by ensuring that sum_morbs is checked to confirm it is not None before any iteration, especially when plotting a single orbital. The order of statements has been adjusted accordingly.

### Example Code to Reproduce:

```python
from pymatgen.electronic_structure.plotter import DosPlotter, BSPlotter, BSPlotterProjected, BSDOSPlotter
from pymatgen.io.vasp import BSVasprun, Vasprun

v = BSVasprun("./bs/vasprun.xml", parse_projected_eigen=True)
bs = v.get_band_structure(kpoints_filename="./bs/KPOINTS", line_mode=True)
plotter = BSPlotterProjected(bs)
plotter.get_projected_plots_dots_patom_pmorb({"Au": ["dxy"]}, {"Au": [1]}, ylim=[-5, 5])
```

This change ensures more robust handling of cases where sum_morbs is None to prevent runtime errors.


ShengLin1001:
And I also have a question, why we need run this

```python
elif len(dictio[elt][0]) > 1:
```
for example code, the len("dxy") = 3, what's meaning of this?

shyuep:
Thanks. But the electronic_structure module has been moved to pymatgen-core. Please copy the changes there and resubmit a PR. Thanks!

PR #4169, opened 2024-11-13
CP2K fixes

janosh:
this PR fixes several bugs encountered trying to use the `cp2k` modules in `pymatgen` which don't seem to have seen active development over the last 4 years (so some errors from upstream breaking changes expected)

## Key Changes

1. **Default Verbosity**: change `Keyword` + `Section` `verbose` default to `False` to stop printing explanatory comments for sections and flags into _every_ CP2K input file (don't explain stuff that can just be looked up in the docs)

2. **YAML Migration**: Switched from `ruamel.yaml` to `pyyaml` to resolve `AttributeError` in `setup_cp2k_data` function.

3. **Basis Set & Potential Handling**:
   - Added support for string-based potential specifications (used to raise `AttributeError` no `.filename`)
   - Added fallback to global basis sets and potentials in `get_basis_and_potential`
   - Added `cp2k_data_dir` keyword parameter in `get_basis_and_potential` that takes precedence over `PMG_CP2K_DATA_DIR` from `pmgrc.yaml`

## Bug Fixes

- Fixed `KeyError` in `calculation_type` when `self.data.get("dft")` returns `None`
- Fixed trailing whitespace in `Section._get_str`


shyuep:
Merged. Thanks. One option is to break this out as an add-on for separate management.

PR #4175, opened 2024-11-15
Cleanup `codespell` ignore patterns

DanielYang59:
## Cleanup `codespell` ignore patterns

Currently a lot of global ignore patterns are very specific to a certain module (or even one line), there's no good reason to ignore a pattern globally for such cases. Admittedly it will be a balance between convenience (no one want to insert a `# codespell:ignore` tag every line) and correctness (some patterns are very specific to few lines across the code base, but are real typos elsewhere, like `lamda/coul`, do note).

Changes:
  - Keep the global pattern when it's very common: Elements like `Nd/Te`, or very common for a commonly used module, `TITEL` for VASP for example, or our good friend `atomate`
  - Rename some internal/test variables **when it's not breaking** (usually increase readability)
  - Ignore all JSON files as comments are not allowed in JSON, we might safely assume everything is intended in JSON
  - [ ] Is it possible to add file-specific ignore tags (for example ignore `ontop` for `adsorption.py`, or `coul` for `LAMMPS`)? Currently [`codespell` seems to provide block-level ignore tag](https://github.com/codespell-project/codespell/issues/3381) but doesn't take a pattern


tests/apps/battery/test_conversion_battery.py
```diff
@@ -63,13 +63,13 @@ def setUp(self):
...
                     ["LiCoO2"],
-                    ["Li6CoO4", "CoO"],
+                    ["Li6CoO4", "CoO"],  # codespell:ignore CoO
                     ["Li6CoO4", "Co"],
                 ],
                 "entries_discharge": [
                     ["Li(CoO2)2"],
                     ["LiCoO2"],
-                    ["Li6CoO4", "CoO"],
+                    ["Li6CoO4", "CoO"],  # codespell:ignore pres
```
DanielYang59, line 72: pattern typo

DanielYang59:
Thanks! Currently `codespell` doesn't seem to support block/file level support with pattern, I would update once they do.

PR #4176, opened 2024-11-15
Change XAS behavior under negative frequencies

esoteric-ephemera:
The XAS spectrum class used to not throw a `ValueError` when more than 5% of the user-input intensities are non-positive (cverified this was still the behavior as of v2020.4.29).  More, the warning that is thrown is misleading, as it suggests that the majority of intensities are non-positive.

This change in behavior has led to issues with MP's API users, see [here for an example](https://matsci.org/t/avoiding-value-error-double-check-the-intensities-most-of-them-are-nonpositive/58748/2)

Changed this to throw a warning only when more than 5% of the intensities are ***negative***. Zero intensity shouldn't be penalized since this is physical (there's no absorption in that range, e.g. if a transition is forbidden)

Also added an option to zero out negative intensities (false by default)

### DIFF
--- a/src/pymatgen/analysis/xas/spectrum.py
+++ b/src/pymatgen/analysis/xas/spectrum.py
@@ -10,11 +10,13 @@ import numpy as np
 from scipy.interpolate import interp1d
 
 from pymatgen.analysis.structure_matcher import StructureMatcher
+from pymatgen.core import Element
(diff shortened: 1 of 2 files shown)


mkhorton:
Thanks @esoteric-ephemera, appreciate the context with the matsci.org thread too

PR #4177, opened 2024-11-15
Modify FHI-aims inputs to not use species string

tpurcell90:
FHI-aims does not read spin from the species and that causes issues with magnetic calculations

## Summary

Major changes:

- fix 1: Don't use species_string for FHI-aims

## Checklist


ansobolev:
@tpurcell90 This seems to work for the defect workflow, thanks; however, as aims accepts anything in the `species` line (it's just a label), I would rather use something like `site.label` for site description in `geometry.in`. This will allow more precise accounting for species defaults in heterogeneous calculations (something like adsorbate on the surface). But that's for sure open for discussion and not the blocking issue.

tpurcell90:
ready for review

shyuep:
Thanks.

tpurcell90:
would it be possible to make a release with this soon? We need this fix for atomate2

PR #4180, opened 2024-11-17
[Breaking/Fix] Skip isotopes when iterating through `core.Element`

DanielYang59:
### Summary

- Skip isotopes when iterating through `core.Element`, to fix #4178, rationale:
   - I might expect looping through `Element` to behave similarly to looping through the periodic table, i.e. isotopes should not sneak in here, do you like this idea @esoteric-ephemera? 
- Implementation based on https://docs.python.org/3/library/enum.html:
> EnumType is responsible for setting the correct __repr__(), __str__(), __format__(), and __reduce__() methods on the final enum, as well as creating the enum members, properly handling duplicates, providing iteration over the enum class, etc.
- Add `property` `Element.named_isotopes` for all named_isotopes so far.
- Clean up `core.periodic_table` docstring (remove non-existent `PeriodicTable` class).

This **would be a breaking change** (also might be called a fix), so comments are hugely appreciated.

---

```python
from pymatgen.core import Element

for idx, ele in enumerate(Element):
    print(idx, ele)
```

### Before
```
0 H
1 H  # Deuterium
2 H  # Tritium
3 He
4 Li
5 Be
6 B
7 C
......
```

### Now (nothing special, just skip isotopes of hydrogen)
```
0 H
1 He
2 Li
3 Be
4 B
5 C
......
```


shyuep:
I agree isotopes shouldn't appear as part of default behavior.

esoteric-ephemera:
That was an oversight on my part, thanks for the fix! Also agree that enumerating over elements shouldn't list isotopes. Maybe adding a method to `Element` to show people which isotopes are available would be useful tho?

DanielYang59:
> That was an oversight on my part, thanks for the fix!

No problem at all, together we improve test coverage :)

> Maybe adding a method to Element to show people which isotopes are available would be useful tho?

Fair point, is current implementation looking good to you (also I don't have much experience overwriting an `Enum` so not sure if there's any pitfall like builtin `dict/list/...`, do review with a pinch of salt)?
```python
Element.named_isotopes # ---> (Element.D, Element.T)
```

DanielYang59:
@janosh I guess you're very experienced with `Enum` (I'm not), perhaps you could help me double-check the implementation (esp. around overwriting)?

PR #4181, opened 2024-11-17
Fix incorrect comparison logic and update tests

### LINKED ISSUE #4166: [BUG] Incorrect usage of `np.all() > tol` for element-wise value comparison
The following method: https://github.com/materialsproject/pymatgen/blob/bd9fba9ec62437b5b62fbd0b2c2c723216cc5a2c/src/pymatgen/io/lobster/outputs.py#L1694-L1737

Used a pattern like `np.array().all() > tolerance` with the intention to perform element-wise value comparison, however [this seems to be misused as `all()`](https://numpy.org/doc/stable/reference/generated/numpy.all.html#numpy.all):  
> "Test whether all array elements along a given axis evaluate to True."

```python
import numpy as np


arr = np.array([1, 2, 3])

print(arr.all())  # True

print(np.array([False, 1, 2]).all())  # False
```

Should be something like:
```diff
- if abs(band2 - 1.0).all() > limit_deviation:
# However this logic might need double check as it only evaluates to True 
# when ALL deviations are greater than threshold
+ if np.all(np.abs(band2 - 1) > limit_deviation):
# Perhaps it should be the following?
+ if np.any(np.abs(band2 - 1) > limit_deviation):

# The same to the other 
- elif band2.all() > limit_deviation:
+ elif np.all(band2 > limit_deviation)
# The following seems more sensible?
+ elif np.any(band2 > limit_deviation)

```

I might go ahead and change this if I didn't misunderstand anything? @JaGeo

naik-aakash:
## Changes
 
Based on the suggestions form @QuantumChemist ( https://github.com/materialsproject/pymatgen/issues/4166#issuecomment-2480508677) and @DanielYang59 (https://github.com/materialsproject/pymatgen/issues/4166#issuecomment-2480925279) , the code has been fixed to correctly evaluate whether the matrix values are above threshold and tests have been updated accordingly

# Todo

- [x] Add 1 or 2 more test cases

### DIFF
--- a/src/pymatgen/io/lobster/outputs.py
+++ b/src/pymatgen/io/lobster/outputs.py
@@ -1716,13 +1716,26 @@ class Bandoverlaps(MSONable):
         for spin in (Spin.up, Spin.down) if spin_polarized else (Spin.up,):
             num_occ_bands = number_occ_bands_spin_up if spin is Spin.up else number_occ_bands_spin_down
 
-            for array in self.band_overlaps_dict[spin]["matrices"]:
-                sub_array = np.asarray(array)[:num_occ_bands, :num_occ_bands]
+            for overlap_matrix in self.band_overlaps_dict[spin]["matrices"]:
+                sub_array = self.get_sub_array(num_occ_bands=num_occ_bands, overlap_matrix=overlap_matrix)  # type: ignore[arg-type]
 
                 if not np.allclose(sub_array, np.identity(num_occ_bands), atol=limit_deviation, rtol=0):
                     return False
 
         return True
+
+    @staticmethod
+    def get_sub_array(num_occ_bands: int, overlap_matrix: np.ndarray) -> np.ndarray:
+        """Helper method to get the subarray of the overlap matrix for the occupied bands
+
+        Args:
+            num_occ_bands (int): Number of occupied bands
+            overlap_matrix (np.ndarray): Overlap matrix
+
+        Returns:
+            np.ndarray: Subarray of the overlap matrix for the occupied bands
+        """
+        return np.asarray(overlap_matrix)[:num_occ_bands, :num_occ_bands]
 
     @property
     @deprecated(message="Use `band_overlaps_dict` instead.", category=DeprecationWarning)
--- a/src/pymatgen/io/lobster/outputs.py
+++ b/src/pymatgen/io/lobster/outputs.py
@@ -1710,29 +1710,17 @@ class Bandoverlaps(MSONable):
         Returns:
             bool: True if the quality of the projection is good.
         """
-        for matrix in self.band_overlaps_dict[Spin.up]["matrices"]:
-            for iband1, band1 in enumerate(matrix):
-                for iband2, band2 in enumerate(band1):
-                    if iband1 < number_occ_bands_spin_up and iband2 < number_occ_bands_spin_up:
-                        if iband1 == iband2:
-                            if abs(band2 - 1.0).all() > limit_deviation:
-                                return False
-                        elif band2.all() > limit_deviation:
-                            return False
+        if spin_polarized and number_occ_bands_spin_down is None:
+            raise ValueError("number_occ_bands_spin_down has to be specified")
+        
+        for spin in (Spin.up, Spin.down) if spin_polarized else (Spin.up,):
+            num_occ_bands = number_occ_bands_spin_up if spin == Spin.up else number_occ_bands_spin_down
 
-        if spin_polarized:
-            for matrix in self.band_overlaps_dict[Spin.down]["matrices"]:
-                for iband1, band1 in enumerate(matrix):
-                    for iband2, band2 in enumerate(band1):
-                        if number_occ_bands_spin_down is None:
-                            raise ValueError("number_occ_bands_spin_down has to be specified")
+            for array in self.band_overlaps_dict[spin]["matrices"]:
+                sub_array = np.asarray(array)[:num_occ_bands, :num_occ_bands]
 
-                        if iband1 < number_occ_bands_spin_down and iband2 < number_occ_bands_spin_down:
-                            if iband1 == iband2:
-                                if abs(band2 - 1.0).all() > limit_deviation:
-                                    return False
-                            elif band2.all() > limit_deviation:
-                                return False
+                if not np.allclose(sub_array, np.identity(num_occ_bands), atol=limit_deviation, rtol=0):
+                    return False
 
         return True
 
--- a/src/pymatgen/io/lobster/outputs.py
+++ b/src/pymatgen/io/lobster/outputs.py
@@ -1712,7 +1712,7 @@ class Bandoverlaps(MSONable):
         """
         if spin_polarized and number_occ_bands_spin_down is None:
             raise ValueError("number_occ_bands_spin_down has to be specified")
-        
+
         for spin in (Spin.up, Spin.down) if spin_polarized else (Spin.up,):
             num_occ_bands = number_occ_bands_spin_up if spin == Spin.up else number_occ_bands_spin_down
 
--- a/src/pymatgen/io/lobster/outputs.py
+++ b/src/pymatgen/io/lobster/outputs.py
@@ -1717,25 +1717,12 @@ class Bandoverlaps(MSONable):
             num_occ_bands = number_occ_bands_spin_up if spin is Spin.up else number_occ_bands_spin_down
 
             for overlap_matrix in self.band_overlaps_dict[spin]["matrices"]:
-                sub_array = self.get_sub_array(num_occ_bands=num_occ_bands, overlap_matrix=overlap_matrix)  # type: ignore[arg-type]
+                sub_array = np.asarray(overlap_matrix)[:num_occ_bands, :num_occ_bands]
 
                 if not np.allclose(sub_array, np.identity(num_occ_bands), atol=limit_deviation, rtol=0):
                     return False
 
         return True
-
-    @staticmethod
-    def get_sub_array(num_occ_bands: int, overlap_matrix: np.ndarray) -> np.ndarray:
-        """Helper method to get the subarray of the overlap matrix for the occupied bands
-
-        Args:
-            num_occ_bands (int): Number of occupied bands
-            overlap_matrix (np.ndarray): Overlap matrix
-
-        Returns:
-            np.ndarray: Subarray of the overlap matrix for the occupied bands
-        """
-        return np.asarray(overlap_matrix)[:num_occ_bands, :num_occ_bands]
 
     @property
     @deprecated(message="Use `band_overlaps_dict` instead.", category=DeprecationWarning)
--- a/src/pymatgen/io/lobster/outputs.py
+++ b/src/pymatgen/io/lobster/outputs.py
@@ -1714,7 +1714,7 @@ class Bandoverlaps(MSONable):
             raise ValueError("number_occ_bands_spin_down has to be specified")
 
         for spin in (Spin.up, Spin.down) if spin_polarized else (Spin.up,):
-            num_occ_bands = number_occ_bands_spin_up if spin == Spin.up else number_occ_bands_spin_down
+            num_occ_bands = number_occ_bands_spin_up if spin is Spin.up else number_occ_bands_spin_down
 
             for array in self.band_overlaps_dict[spin]["matrices"]:
                 sub_array = np.asarray(array)[:num_occ_bands, :num_occ_bands]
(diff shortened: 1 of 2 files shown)


JaGeo:
@naik-aakash could you add a few more tests here next week (not today) and make sure it is working as intended?

tests/io/lobster/test_outputs.py
```diff
@@ -1603,7 +1603,7 @@ def test_has_good_quality(self):
         assert not self.band_overlaps2_new.has_good_quality_check_occupied_bands(
             number_occ_bands_spin_up=10, limit_deviation=0.0000001
         )
-        assert not self.band_overlaps2.has_good_quality_check_occupied_bands(
+        assert self.band_overlaps2.has_good_quality_check_occupied_bands(
```
JaGeo, line 1606: So, the result will change from false to true?

naik-aakash: Yes, as the logic was not correct before, the results will change.

naik-aakash:
> @naik-aakash could you add a few more tests here next week (not today) and make sure it is working as intended?

Sure can do this 😄

src/pymatgen/io/lobster/outputs.py
```diff
@@ -1710,29 +1710,17 @@ def has_good_quality_check_occupied_bands(
...
-                            if iband1 == iband2:
-                                if abs(band2 - 1.0).all() > limit_deviation:
-                                    return False
-                            elif band2.all() > limit_deviation:
-                                return False
+        if spin_polarized and number_occ_bands_spin_down is None:
+            raise ValueError("number_occ_bands_spin_down has to be specified")
+
+        for spin in (Spin.up, Spin.down) if spin_polarized else (Spin.up,):
+            num_occ_bands = number_occ_bands_spin_up if spin == Spin.up else number_occ_bands_spin_down
```
DanielYang59, line 1717: Thanks a ton @naik-aakash I guess it would be slightly better to use `is` to check identify here (though should yield the same results)?  i.e. `if spin is Spin.up`

DanielYang59:
Forgive me, I'm going slightly off-topic again, perhaps also replace the tolerance (`limit_deviation`) with more readable scientific notation? https://github.com/materialsproject/pymatgen/blob/c72c9e93ca5922f8fbd1d0e0bb385cfc1ae26420/tests/io/lobster/test_outputs.py#L1545-L1604

naik-aakash:
> Forgive me, I'm going slightly off-topic again, perhaps also replace the tolerance (`limit_deviation`) with more readable scientific notation?
> 
> https://github.com/materialsproject/pymatgen/blob/c72c9e93ca5922f8fbd1d0e0bb385cfc1ae26420/tests/io/lobster/test_outputs.py#L1545-L1604
> 
> Also perhaps separate the tests of `has_good_quality_maxDeviation` and `has_good_quality_check_occupied_bands` into two separate test methods? (I guess they're independent right?)

How about renaming `limit_deviation` to `deviation_tol` ? Does this make it more readable scientific notation ?

And regarding your second request, yes they are independent and thus have been seperated now in the tests.

src/pymatgen/io/lobster/outputs.py
naik-aakash: Made the change

naik-aakash:
> `has_good_quality_check_occupied_bands`

Ahh okay got what you meant. Was not clear to me it referred to numeric values and not vairable naming and Yess it is public, thus will lead to a breaking change.

src/pymatgen/io/lobster/outputs.py
```diff
@@ -1710,29 +1710,17 @@ def has_good_quality_check_occupied_bands(
...
-                            raise ValueError("number_occ_bands_spin_down has to be specified")
-
-                        if iband1 < number_occ_bands_spin_down and iband2 < number_occ_bands_spin_down:
-                            if iband1 == iband2:
-                                if abs(band2 - 1.0).all() > limit_deviation:
-                                    return False
-                            elif band2.all() > limit_deviation:
-                                return False
+        if spin_polarized and number_occ_bands_spin_down is None:
+            raise ValueError("number_occ_bands_spin_down has to be specified")
```
DanielYang59, line 1714: We might be slightly missing test coverage here, something like:
```python
        with pytest.raises(ValueError, match="number_occ_bands_spin_down has to be specified"):
            self.band_overlaps1.has_good_quality_check_occupied_bands(
            number_occ_bands_spin_up=4,
            spin_polarized=True,
        )
```

naik-aakash: Yes, I will update the tests as listed in the todo today

DanielYang59: No rush at all, I had a general idea to add and would share with you a bit later today.

DanielYang59: I'm thinking about something like the following:

```python
    def test_has_good_quality_check_occupied_bands_patched(self):
        """Test with patched data."""
        actual_deviation = 0
        limit_deviation = 0.1

        number_occ_bands_spin_up = 9
        number_occ_bands_spin_down = 5

        band_overlaps = copy.deepcopy(self.band_overlaps1)

        for spin in (Spin.up, Spin.down):
            for i_arr, array in enumerate(band_overlaps.band_overlaps_dict[spin]["matrices"]):
                number_occ_bands = number_occ_bands_spin_up if spin is Spin.up else number_occ_bands_spin_down

                shape = array.shape
                assert np.all(np.array(shape) >= number_occ_bands)
                assert shape[0] == shape[1]  # square array

                # Generate a noisy background array
                patch_array = np.random.uniform(0, 10, shape)

                # Patch the top-left sub-array (the part that would be checked)
                patch_array[:number_occ_bands, :number_occ_bands] = np.identity(number_occ_bands) + np.random.uniform(-actual_deviation, actual_deviation, (number_occ_bands, number_occ_bands))

                band_overlaps.band_overlaps_dict[spin]["matrices"][i_arr] = patch_array

        result = band_overlaps.has_good_quality_check_occupied_bands(
            number_occ_bands_spin_up=number_occ_bands_spin_up,
            number_occ_bands_spin_down=number_occ_bands_spin_down,
            spin_polarized=True,
            limit_deviation=limit_deviation,
        )

        assert result == (actual_deviation <= limit_deviation)
```

This is a very rough scratch so more work need to be done. But the basic idea is to patch the data and call the checker method on the generated data (I understand generating a "just right" data in real world must be super hard). However I believe those hard coded parameters would need some extra work (this is the heavy lifting part, thank you). Also do note `pytest.mark.parametrize` cannot be used with `TestCase` (unfortunately TestBandoverlaps is inheriting it), so it would be a bit tricky to parametrize the tests (like generate several combinations of good/bad deviations).

naik-aakash: Added this exeptions test and also added a test to check if extracted sub_array is desired

tests/io/lobster/test_outputs.py
naik-aakash: Hi @JaGeo , refactored the code a bit and also added a test to ensure we extract and check correct data

src/pymatgen/io/lobster/outputs.py
QuantumChemist: Hey @DanielYang59 ,
can you explain the idea with the noise patch a bit more?
I see no need to add an extra backgorund noise patch as there are already unit tests to check good and bad data quality (return true and false).

naik-aakash: At this point, I do not think this change is really necessary, so I will leave it out. The main issue was with comparison logic in the code (which has been fixed thanks to @DanielYang59 and @QuantumChemist's suggestions), and current test cases already check for good and bad deviations, although not in concise form.

DanielYang59: > I am not able to understand the need for the patched test code snippet your shared. Can you please elaborate a bit. Do you want to say we should basically test a list of deviations and codense the code in test ?

Perhaps "patch" is not a proper word in this case, I believe we need to test the method on:
- Generated data array that is fully under our control (array with known data distribution), something like https://github.com/materialsproject/pymatgen/pull/4181#discussion_r1846118944
- Real world data (existing tests)

Because I believe it would be super hard (or even not possible) to fully test the method on real-world data (say it's hard to get band overlaps with certain deviation).

QuantumChemist: I still don#T see the benefit. Could you explain what kind of test case one would cover with this?

naik-aakash: Okay, got the point. Thanks @DanielYang59 , will accommodate such test as well.

DanielYang59: > I still don't see the benefit. 

I might be overdoing tests. Basically our current `has_good_quality_check_occupied_bands` method is to check whether all arrays in `band_overlaps_dict` is `allclose` to a identity array. 

With real world data, it's hard to know the data distributions (we don't really know the deviation of that array, we just run the method and get `True/False`, and at this point we're checking the consistency of the method, i.e. "the method return True before, and it should return True now"). 

With generated data, we know exactly how much the array deviates from identity, and we know exactly what the method should return (`True/False`) **before** even running the method, so we're checking the correctness here.

QuantumChemist: Oh yeah, that makes a lot of sense indeed and is a better unit test practice as well. Thank you!

naik-aakash: Hi @DanielYang59 , thanks for more explanations it surely helped to understand better. I have now added another test with patched data code snippet you shared in your suggestion. 

I have also tweaked your code snippet to also test for some deviations values /spin up, spin down band number variations.

DanielYang59: No problem, glad you found it useful. Thanks a lot for all these work :)

src/pymatgen/io/lobster/outputs.py
```diff
@@ -1710,32 +1710,33 @@ def has_good_quality_check_occupied_bands(
...
+            for overlap_matrix in self.band_overlaps_dict[spin]["matrices"]:
+                sub_array = self.get_sub_array(num_occ_bands=num_occ_bands, overlap_matrix=overlap_matrix)  # type: ignore[arg-type]
+
+                if not np.allclose(sub_array, np.identity(num_occ_bands), atol=limit_deviation, rtol=0):
+                    return False
 
         return True
 
+    @staticmethod
+    def get_sub_array(num_occ_bands: int, overlap_matrix: np.ndarray) -> np.ndarray:
```
DanielYang59, line 1728: Would it be overkill to make this a separate method?

tests/io/lobster/test_outputs.py
```diff
@@ -1614,14 +1616,104 @@ def test_has_good_quality(self):
...
             number_occ_bands_spin_up=2, limit_deviation=0.1
         )
         assert self.band_overlaps2.has_good_quality_check_occupied_bands(number_occ_bands_spin_up=1, limit_deviation=1)
         assert self.band_overlaps2_new.has_good_quality_check_occupied_bands(
             number_occ_bands_spin_up=1, limit_deviation=2
         )
 
+    def test_has_good_quality_check_occupied_bands_patched(self):
+        """Test with patched data."""
+        import copy
```
DanielYang59, line 1629: Perhaps we don't need to lazy load `copy`?

src/pymatgen/io/lobster/outputs.py
naik-aakash: Thank you for taking time to explain me with example cases. Will remember these for tests I will write in future 😄

tests/io/lobster/test_outputs.py
naik-aakash: Done

DanielYang59: Thanks

src/pymatgen/io/lobster/outputs.py
DanielYang59: It's purely personal taste, but I'm feeling having a method just for `return np.asarray(overlap_matrix)[:num_occ_bands, :num_occ_bands]` might be too overkill? I.e. we could just do this inline?

naik-aakash: oki, I will get rid of this method and the associated test then

DanielYang59: Thanks for the change, basically adding a test here would be testing whether slicing a `np.array` works correctly or not, I guess it's not really necessary

shyuep:
Thanks.

PR #4183, opened 2024-11-18
Compositions from weight str

CompRhys:
Often in alloy literature we get composition strings given in weight rather than molar quantities. This PR extends upon the `from_weight_dict` method to give a more general `from_weights` method that will take strings as well as CompositionLike


src/pymatgen/core/composition.py
```diff
@@ -639,22 +640,52 @@ def from_dict(cls, dct: dict) -> Self:
         return cls(dct)
 
     @classmethod
```
shyuep, line 642: I think adding a few init examples for this method would be useful. Similar to the one at the init method of Composition.

shyuep:
Thanks. Merged.

PR #4187, opened 2024-11-21
Addition of "valences" property to ElementBase and Species for mixed-valence ions

benrich37:
## Summary

Note: This code change is **no longer required for #4078** and the `valences` property does **not** return `(idx, ne)` for every valence subshell but only **every partially filled valence subshell** (unless all subshells are filled, which causes only the last subshell to be returned). I don't see a use for this criteria, so I will fix this to behave as intended. I am also okay with letting this PR be closed and abandoned since it is no longer needed for #4078. 

Major changes:

- feature 1: For both ElementBase and Species class, the code within the "valence" property was copy-pasted into a new "valences" property, with the error raised for ambiguous valence (more than one valence subshell) removed 
(`if len(valence) > 1:
            raise ValueError(f"{self} has ambiguous valence")`)
and return signature changed from `tuple[int | np.nan, int]` to `list[tuple[int | np.nan, int]]` (`return valence` instead of `return valence[0]`)
- fix 1: To remove redundancy, the "valence" property for `ElementBase` and `Species` was re-written to raise the ambiguous valence ValueError if `len(self.valences) > 1`, and otherwise return `self.valences[0]`

## TODO

- Loosen return criteria for `valences` to return all valence subshells while maintaining selectivity of `valence`

### DIFF
--- a/src/pymatgen/core/periodic_table.py
+++ b/src/pymatgen/core/periodic_table.py
@@ -29,6 +29,7 @@ if TYPE_CHECKING:
 
     from pymatgen.util.typing import SpeciesLike
 
+
 # Load element data from JSON file
 with open(Path(__file__).absolute().parent / "periodic_table.json", encoding="utf-8") as ptable_json:
     _pt_data = json.load(ptable_json)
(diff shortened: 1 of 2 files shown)


benrich37:
closing this until I make those fixes

PR #4189, opened 2024-11-21
Inputs - JDFTx IO Module

benrich37:
## Summary

Major changes:

- feature 1: `inputs.py` module containing `JDFTXInfile` class to give pythonic representation to the inputs for a JDFTx calculation
- feature 2: "helper" module `generic_tags.py` holding "Tag" objects (`AbstractTag` and inheritors of `AbstractTag`) to represent the different data structures JDFTx expects for inputs
- feature 3: "helper" module `jdftxinfile_master_format.py` to create proper "Tag" objects for input tags
- feature 4: "helper" module `jdftxinfile_ref_options.py` to hold lists of acceptable input strings for different input tags (ie list of acceptable XC functionals for the tag "elec-ex-corr")


mkhorton:
Thanks for your patience with this one, long PRs adding a big new feature are always more difficult to review. I think this nicely self-contained and is in good shape to merge in. We can save further improvements for follow-up PRs that we can review/merge more quickly, and we can take the same approach with #4190 too.

One thing that wasn't obvious to me, why the `/tmp` directory in the test files? I couldn't see where it was used in the diff.

benrich37:
> Thanks for your patience with this one, long PRs adding a big new feature are always more difficult to review. I think this nicely self-contained and is in good shape to merge in. We can save further improvements for follow-up PRs that we can review/merge more quickly, and we can take the same approach with #4190 too.
> 
> One thing that wasn't obvious to me, why the `/tmp` directory in the test files? I couldn't see where it was used in the diff.

Awesome thank you! Sorry for the new push - had a minor bug from inconsistent tracking of units. And the `tmp` directory is used to hold files that are created and deleted during the tests (ie testing that an input file can be written and then read correctly). I thought it would make the tests run quicker to have the directory exist permanently instead of created when needed but I can change that if the alternative is better

mkhorton:
> I thought it would make the tests run quicker to have the directory exist permanently instead of created when needed but I can change that if the alternative is better.

It'd be better to remove it :) Creating a directory doesn't take much time, and might add some confusion to others to see it there. Tests should do "set up" and "tear down" as needed.

> So edits for minor bugs I should keep within https://github.com/materialsproject/pymatgen/pull/4078 and leave https://github.com/materialsproject/pymatgen/pull/4189 and https://github.com/materialsproject/pymatgen/pull/4190 untouched, right?

As long as you merge in changes from the main branch (or rebase) #4078 once #4189 and #4190 are merged, or you can create new PRs. Either way is fine, I will try and be responsive!

benrich37:
@mkhorton Sorry for the delay! "tmp" directory now only exists during the execution of the required tests by defining it as a module-scope fixture defined in [`tests.io.jdftx.shared_test_utils`](https://github.com/benrich37/pymatgen/blob/feature-jdftx-inputs/tests/io/jdftx/shared_test_utils.py#L38-L42). I tried using the `tmp_path` class variable in the `PymatgenTest` class but I discovered that test methods within test classes inheriting `unittest.TestCase` cannot use the `pytest.mark.parametrize` decorator so the fixture-approach ended up working out better.

shyuep:
Is htis ready to be merged?

PR #4192, opened 2024-11-23
Clean up `util.num.make_symmetric_matrix_from_upper_tri` and `util.io_util.micro_pyawk`, fix new `ruff` errors

DanielYang59:
### Simplify/clarify `util.num.make_symmetric_matrix_from_upper_tri` and add tests

> [!CAUTION]
> `make_symmetric_matrix_from_upper_tri` now raise `ValueError` for non 3x3 array

I believe `utils.num.make_symmetric_matrix_from_upper_tri` is designed to [work on 3x3 matrix alone](https://github.com/materialsproject/pymatgen/pull/4192#discussion_r1855063190) (the new index is hard-coded): https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/util/num.py#L31-L32

As such I might greatly simplify the implementation f2c325299aa0162f0ee3f0a50507ab95991363bb which [gives ~3x speedup](https://github.com/materialsproject/pymatgen/pull/4192#discussion_r1855063190) and much better readability.

### Clean up `util.io_util.micro_pyawk`

> [!WARNING]
> The `debug/postdebug` is likely debug args, no usage across the entire code bass, schedule them for removal after one-year grace period in case someone really need them.

### Others

- Remove: https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/util/io_utils.py#L98-L99

    I believe this is doing nothing (set umask to 0, but reset it back the original umask immediately)
- Fix new `ruff` errors and bump `pre-commit` accordingly


src/pymatgen/util/num.py
```diff
@@ -22,17 +30,30 @@ def round_to_sigfigs(num, sig_figs):
...
+         [A_xy, A_yy, A_yz],
+         [A_xz, A_yz, A_zz]]
+
+    Args:
+        val (ArrayLike): Flattened upper triangular elements.
+
+    Returns:
+        NDArray: The symmetric matrix.
     """
-    idx = [0, 3, 4, 1, 5, 2]
```
DanielYang59, line 31: I believe this helper function is designed to work for 3x3 matrix alone, as such I might just make things simple and reject non 3x3 arrays with `ValueError`.

The new implementation gives a ~3 speedup:
```
Original implementation time: 0.378586 seconds
Simplified implementation time: 0.118705 seconds
```


Test script (from GPT):
```python
import numpy as np
from timeit import timeit

def original_make_symmetric_matrix_from_upper_tri(val):
    """Original implementation."""
    idx = [0, 3, 4, 1, 5, 2]
    val = np.array(val)[idx]
    mask = ~np.tri(3, k=-1, dtype=bool)
    out = np.zeros((3, 3), dtype=val.dtype)
    out[mask] = val

    out.T[mask] = val
    return out

def simplified_make_symmetric_matrix_from_upper_tri(val):
    """Simplified implementation."""
    array = np.asarray(val)
    if array.shape != (6,):
        raise ValueError(f"Expected val of length 6, got {array.shape}")

    return np.array([
        [array[0], array[3], array[4]],
        [array[3], array[1], array[5]],
        [array[4], array[5], array[2]],
    ])

# Test input
val = [1, 2, 3, 4, 5, 6]

# Measure execution time
original_time = timeit(
    "original_make_symmetric_matrix_from_upper_tri(val)",
    globals=globals(),
    number=100000,
)
simplified_time = timeit(
    "simplified_make_symmetric_matrix_from_upper_tri(val)",
    globals=globals(),
    number=100000,
)

# Display results
print(f"Original implementation time: {original_time:.6f} seconds")
print(f"Simplified implementation time: {simplified_time:.6f} seconds")
```

src/pymatgen/symmetry/groups.py
```diff
@@ -18,7 +18,6 @@
 from monty.design_patterns import cached_class
 from monty.serialization import loadfn
 
-from pymatgen.core.operations import SymmOp
```
DanielYang59: looks like `pre-commit` hasn't been updated to recognize the `TC004` rule and would reposition this import

DanielYang59: bump ruff hook in pre-commit seems to resolve this 7825975594f95c9d9c73efa9a5c4a58e0e43419e

PR #4195, opened 2024-11-25
Read NBANDS in `io.vasp.outputs.Outcar` init

### LINKED ISSUE #3843: Support parsing of NBANDS from a VASP OUTCAR file
### Feature Requested

In VASP, the NBANDS parameter describes the number of bands. It can be modified on-the-fly by VASP, potentially even overriding the setting chosen by the user. This is done based on the number of cores requested for the calculation. Therefore, knowing the NBANDS VASP actually uses is quite useful. This information is available in the OUTCAR but is not currently parsed. We should add support for this as an attribute of `pymatgen.io.vasp.outputs.Outcar` (see below).

### Proposed Solution

In Custodian, we have a block of code that looks like:

```python
nbands = None
with open(os.path.join(directory, "OUTCAR")) as file:
    for line in file:
        # Have to take the last NBANDS line since sometimes VASP
        # updates it automatically even if the user specifies it.
        # The last one is marked by NBANDS= (no space).
        if "NBANDS=" in line:
            try:
                d = line.split("=")
                nbands = int(d[-1].strip())
                break
            except (IndexError, ValueError):
                pass
```

We can basically port this over to Pymatgen's `pymatgen.io.vasp.outputs.Outcar` class as a `.nbands` attribute (and then remove the code block from Custodian).

### Relevant Information

_No response_

DanielYang59:
### Summary

- Read NBANDS in `io.vasp.outputs.Outcar` init, to close #3843
- Add overwriting tests and test file (355 KB), I didn't find any reusable file (`zgrep 'I found NBANDS' OUTCAR*` returns nothing)

#### VASP version stability
- 5.2.11 (unit test)
- 5.4.4 (unit test)
- 6.3.0 (unit test)
- 6.3.2 (unit test, overridden)

#### Reason for using current pattern instead of `NBANDS = (\d+)`

As mentioned in https://github.com/materialsproject/pymatgen/issues/3843#issue-2308586992, the NBANDS could be modified to be "number of cores compatible". [More specifically](https://www.vasp.at/wiki/index.php/NBANDS):
> When executed on multiple CPUs, VASP automatically increases the number of bands, so that NBANDS is divisible by the number of CPU cores. If [NCORE](https://www.vasp.at/wiki/index.php/NCORE) > 1, NBANDS is increased until it is divisible by the number of cores in a group ([NCORE](https://www.vasp.at/wiki/index.php/NCORE)). If [KPAR](https://www.vasp.at/wiki/index.php/KPAR) > 1, NBANDS is increased until it is divisible by the number of cores in a group.

For example if I set NBANDS to 30 on a 64-core job, it would be modified to 32 with the following message:
```
>>> grep NBANDS OUTCAR_set_30_but_changed_to_32 
   NBANDS = 30
|     I found NBANDS = 30. Now, NBANDS = 32.                                  |
   k-points           NKPTS =    126   k-points in BZ     NKDIM =    126   number of bands    NBANDS=     32
```


tests/io/vasp/test_outputs.py
```diff
@@ -1327,6 +1327,18 @@ def test_onsite_density_matrix(self):
...
+
+        # Test VASP 5.4.4
+        assert Outcar(f"{VASP_OUT_DIR}/OUTCAR.LOPTICS.vasp544").data["nbands"] == 128
+
+        # Test VASP 6.3.0
+        assert Outcar(f"{VASP_OUT_DIR}/OUTCAR_vasp_6.3.gz").data["nbands"] == 64
+
     def test_nplwvs(self):
         outcar = Outcar(f"{VASP_OUT_DIR}/OUTCAR.gz")
         assert outcar.data["nplwv"] == [[34560]]
```
DanielYang59, line 1344: ### Question

Why would the "total number of plane waves (NPLWV)" tag take a nested list format?  
https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/io/vasp/outputs.py#L2139-L2147

https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/tests/io/vasp/test_outputs.py#L1330-L1332

DanielYang59:
@Andrew-S-Rosen Is currently implementation looking good to you? :)

Andrew-S-Rosen:
Due to time commitments, I can just say that I trust you. 👍 Happy to report if something comes up.

DanielYang59:
Thanks for saying that, I tested on multiple versions (included in unit test) and didn't find any issue for now, do let me know if you notice any :)

PR #4198, opened 2024-11-25
Fix array comparison in `core.Structure.merge_sites`, also allow `int` property to be merged instead of `float` alone, `mode` only allow full name

### LINKED ISSUE #4197: Structure.merge_sites fails when selective_dynamics is property present
### Python version

Python 3.12.7

### Pymatgen version

Version: 2024.10.22

### Operating system version

_No response_

### Current behavior

Error is thrown when I try to run merge_sites.

  File "/home/yantar92/.venv/lib/python3.12/site-packages/pymatgen/core/structure.py", line 4604, in merge_sites
    if props[key] is not None and self[i].properties[key] != props[key]:
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

### Expected Behavior

merge_sites works without errors :)

### Minimal example

```Python
import pymatgen.core as pmg
str1 = pmg.Structure.from_file('POSCAR1')
str2 = pmg.Structure.from_file('POSCAR2')
merge = str1.copy()
for site in str2:
    merge.append(site.species, site.frac_coords, properties=site.properties)
merge.merge_sites(mode='average')
```

[POSCAR2.gz](https://github.com/user-attachments/files/17900367/POSCAR2.gz)
[POSCAR1.gz](https://github.com/user-attachments/files/17900368/POSCAR1.gz)

DanielYang59:
### Summary

- Fix array comparison in `core.Structure.merge_sites`, to fix  #4197
- [x] Test for property reset warning is missing
- [x] Add test for `is_np_dict_equal ` for custom `__eq__`
- Add check for illegal `mode` initials (not starting with `a/s/d`, currently don't check full name yet)

---

> [!WARNING]  
> Minor behaviour change (fix) in `merge_sites` property handling 

Only allowing `float` to be averaged may not be a good idea, we should include `int` as well, otherwise `properties={"prop1": 100}` would lead to the site property being reset to `None` while `properties={"prop1": 100.0}` doesn't, which is super confusing
https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/tests/core/test_structure.py#L1677-L1679

Also the docstring is clarified to explain this behaviour.

---

> [!WARNING]  
> `mode` only allow "full name" instead of initials after an one-year deprecation period

I believe it's beneficial to allow only full name (`"sum", "delete", "average"`) instead of checking first letter only: 
- more readable: `mode="sum"` instead of `mode="s"`
- consistent with currently type annotation: https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/core/structure.py#L4719


src/pymatgen/core/structure.py
```diff
@@ -4746,7 +4746,7 @@ def merge_sites(self, tol: float = 0.01, mode: Literal["sum", "delete", "average
                 offset = self[i].frac_coords - coords
                 coords += ((offset - np.round(offset)) / (n + 2)).astype(coords.dtype)
                 for key in props:
-                    if props[key] is not None and self[i].properties[key] != props[key]:
```
DanielYang59, line 4749: Using `array_equal` may not be a good idea either as the property could be (sequence of) floats, the test failure should be related to this: https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/tests/core/test_structure.py#L1667-L1681

yantar92: Property can actually be anything, including value supplied by user.

src/pymatgen/core/structure.py
```diff
@@ -4717,43 +4717,53 @@ def scale_lattice(self, volume: float) -> Self:
...
+                - "average" means that the site is deleted but the properties are averaged.
 
         Returns:
-            Structure: self with merged sites.
+            Structure: Structure with merged sites.
         """
-        dist_mat = self.distance_matrix
+        # TODO: change the code the allow full name after 2025-12-01
+        # TODO2: add a test for mode value, currently it only checks if first letter is "s/a"
+        if mode.lower() not in {"sum", "delete", "average"} and mode.lower()[0] in {"s", "d", "a"}:
```
DanielYang59, line 4735: Want to hear more options on this, I guess it's beneficial to allow only full name (`"sum", "delete", "average"`) instead of checking first letter only to make use of the IDE auto-complete feature

src/pymatgen/core/structure.py
```diff
@@ -4717,43 +4717,56 @@ def scale_lattice(self, volume: float) -> Self:
...
             for n, i in enumerate(inds[1:]):
                 sp = self[i].species
                 if mode.lower()[0] == "s":
                     species += sp
                 offset = self[i].frac_coords - coords
                 coords += ((offset - np.round(offset)) / (n + 2)).astype(coords.dtype)
                 for key in props:
-                    if props[key] is not None and self[i].properties[key] != props[key]:
+                    if props[key] is not None and not np.array_equal(self[i].properties[key], props[key]):
                         if mode.lower()[0] == "a" and isinstance(props[key], float):
```
DanielYang59, line 4762: I guess only allowing `float` to be averaged may not be a good idea, we should include `int` as well, otherwise `properties={"prop1": 100}` would lead to the site property being reset to `None` while `properties={"prop1": 100.0}`, which is super confusing
https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/tests/core/test_structure.py#L1677-L1679

Also the docstring would be clarified to explain this behaviour.

PR #4199, opened 2024-11-26
Fix `py.typed` usage for namespace packages

### LINKED ISSUE #3413: Is the current `py.typed` setup ideal?
### Problem

I am using `mypy` on a package with Pymatgen as a dependency and am getting back the following even though there is a `py.typed` marker in the base pymatgen directory. Any insights?

```
schemas/_aliases/emmet.py:11: error: Skipping analyzing "pymatgen.core.composition": module is installed, but missing library stubs or py.typed marker  [import]
schemas/_aliases/emmet.py:12: error: Skipping analyzing "pymatgen.core.lattice": module is installed, but missing library stubs or py.typed marker  [import]
schemas/_aliases/emmet.py:13: error: Skipping analyzing "pymatgen.core.periodic_table": module is installed, but missing library stubs or py.typed marker  [import]
schemas/_aliases/emmet.py:14: error: Skipping analyzing "pymatgen.core.structure": module is installed, but missing library stubs or py.typed marker  [import]
schemas/_aliases/emmet.py:15: error: Skipping analyzing "pymatgen.entries.computed_entries": module is installed, but missing library stubs or py.typed marker  [import]
schemas/_aliases/emmet.py:16: error: Skipping analyzing "pymatgen.io.vasp.inputs": module is installed, but missing library stubs or py.typed marker  [import]
```

### Proposed Solution

Confirm `py.typed` marker carries over to the full package.

### Alternatives

_No response_

DanielYang59:
- [ ] [Fix `py.typed` marker file placement](https://typing.readthedocs.io/en/latest/spec/distributing.html#packaging-typed-libraries) to close #3413:
    > For namespace packages (see [PEP 420](https://peps.python.org/pep-0420/)), the py.typed file should be in the submodules of the namespace, to avoid conflicts and for clarity.
- [ ] Add a test for `py.typed` files, verify the original reported issue is gone (`Skipping analyzing "pymatgen.io.vasp.inputs": module is installed, but missing library stubs or py.typed marker  [import]`)
- [ ] Double check config of `analysis / ext / io`, are they meant to be namespace packages as well?

---

Looks like some sub-modules (`analysis / ext / io`)  are implicit namespace packages as well (no `__init__.py`)? i.e. `pymatgen` is namespace package, so do `pymatgen.ext/pymatgen.analysis/pymatgen.io`. Not sure if this is intended, AFAIK sharing the `pymatgen` namespace is sufficient to support `pymatgen-analysis-defects/analysis/diffusion`, `pymatgen-io-validation `?

Also quote [Anthony Sottile](https://www.youtube.com/watch?v=2Xvb79hOUdM):
> imo a namespace package is just always a bad idea -- you're better to use a plugin system -- here's an example of how to set one up: [https://youtu.be/fY3Y_xPKWNA](https://www.youtube.com/watch?v=fY3Y_xPKWNA)

```
>>> import pymatgen
>>> pymatgen
<module 'pymatgen' (namespace) from ['/home/yang/developer/pymatgen/src/pymatgen']>

>>> import pymatgen.io
>>> pymatgen.io
<module 'pymatgen.io' (namespace) from ['/home/yang/developer/pymatgen/src/pymatgen/io']>
>>> import pymatgen.ext
>>> pymatgen.ext
<module 'pymatgen.ext' (namespace) from ['/home/yang/developer/pymatgen/src/pymatgen/ext']>
>>> import pymatgen.analysis
>>> pymatgen.analysis
<module 'pymatgen.analysis' (namespace) from ['/home/yang/developer/pymatgen/src/pymatgen/analysis']>
```

https://github.com/materialsproject/pymatgen/tree/master/src/pymatgen/ext
https://github.com/materialsproject/pymatgen/tree/master/src/pymatgen/io
https://github.com/materialsproject/pymatgen/tree/master/src/pymatgen/analysis

### DIFF
--- a/tests/test_pkg.py
+++ b/tests/test_pkg.py
@@ -1,35 +1,107 @@
+"""Check the content of source code and test build source distribution.
+
+Binary distribution (wheel) is checked in test workflow.
+"""
+
 from __future__ import annotations
 
-import os
+import subprocess
+import tarfile
 from glob import glob
+from pathlib import Path
+from typing import TYPE_CHECKING
 
-import pytest
+from monty.tempfile import ScratchDir
 
-SRC_TXT_PATH = "src/pymatgen.egg-info/SOURCES.txt"
+if TYPE_CHECKING:
+    from pymatgen.util.typing import PathLike
+
+PROJECT_ROOT = Path(__file__).parent.parent
+NAMESPACE_PKGS = {"analysis", "ext", "io"}  # TODO: double check
 
 
-@pytest.mark.skipif(
-    not os.path.isfile(SRC_TXT_PATH),
-    reason=f"{SRC_TXT_PATH=} not found. Run `pip install .` to create",
-)
-def test_egg_sources_txt_is_complete():
-    """Check that all source and data files in pymatgen/ are listed in pymatgen.egg-info/SOURCES.txt."""
+def test_source_code():
+    """Check the source code in the working directory."""
+    src_txt_path = PROJECT_ROOT / "src/pymatgen.egg-info/SOURCES.txt"
 
-    with open(SRC_TXT_PATH, encoding="utf-8") as file:
-        sources = file.read()
+    _check_src_txt_is_complete(PROJECT_ROOT, src_txt_path)
 
-    # check that all files listed in SOURCES.txt exist
+    # Check existence of `py.typed` file in each sub-package
+    _check_py_typed_files(PROJECT_ROOT / "src/pymatgen")
+
+
+def test_build_source_distribution():
+    """Test build the source distribution (sdist), also check `py.typed` files."""
+    with ScratchDir("."):
+        # Build the source distribution
+        subprocess.run(["python", "-m", "pip", "install", "--upgrade", "build"], check=True)
+        subprocess.run(["python", "-m", "build", "--sdist", PROJECT_ROOT, "--outdir", ".", "-C--quiet"], check=True)
+
+        # Decompress sdist
+        sdist_file = next(Path(".").glob("*.tar.gz"))
+        sdist_dir = Path(sdist_file.name.removesuffix(".tar.gz"))
+        with tarfile.open(sdist_file, "r:gz") as tar:
+            # TODO: remove attr check after only 3.12+
+            if hasattr(tarfile, "data_filter"):
+                tar.extractall("", filter="data")
+            else:
+                tar.extractall("")  # noqa: S202
+
+        # Check existence of `py.typed` file in each sub-package
+        _check_py_typed_files(sdist_dir / "src/pymatgen")
+
+
+def _check_src_txt_is_complete(project_root: PathLike, src_txt_path: PathLike) -> None:
+    """Check that all source code and data files are listed in given SOURCES.txt.
+
+    Args:
+        project_root (PathLike): Path to the directory containing "src/pymatgen/".
+        src_txt_path (PathLike): Path to the "SOURCES.txt" file.
+    """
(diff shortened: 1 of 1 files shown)


tests/test_pkg.py
```diff
@@ -1,35 +1,107 @@
...
+
+    # Check that all files listed in "SOURCES.txt" exist
     for src_file in sources.splitlines():
-        assert os.path.isfile(src_file), f"{src_file!r} does not exist!"
-
-    # check that all files in pymatgen/ are listed in SOURCES.txt
-    for ext in ("py", "json*", "yaml", "csv"):
-        for filepath in glob(f"pymatgen/**/*.{ext}", recursive=True):
-            unix_path = filepath.replace("\\", "/")
-            if unix_path.endswith("dao.py"):
```
DanielYang59, line 29: I believe `dao.py` is already included in sdist and bdist (not just this branch, but from [PyPI](https://pypi.org/project/pymatgen/#files))

tests/test_pkg.py
```diff
@@ -1,35 +1,107 @@
...
+from monty.tempfile import ScratchDir
+
+if TYPE_CHECKING:
+    from pymatgen.util.typing import PathLike
+
+PROJECT_ROOT = Path(__file__).parent.parent
+NAMESPACE_PKGS = {"analysis", "ext", "io"}  # TODO: double check
+
+
+def test_source_code():
```
DanielYang59, line 23: I'm not sure if that makes sense at all to compare source code/distribution with `SOURCES.txt` (would it be possible that the contents of sdist is defined by `SOURCES.txt`?), need to [double check the definition](https://setuptools.pypa.io/en/stable/deprecated/commands.html#egg-info-create-egg-metadata-and-set-build-tags):
> This command performs two operations: it updates a project’s .egg-info metadata directory (used by the bdist_egg, develop, and test commands), and it allows you to temporarily change a project’s version string, to support “daily builds” or “snapshot” releases. It is run automatically by the sdist, bdist_egg, develop, and test commands in order to update the project’s metadata, but you can also specify it explicitly in order to temporarily change the project’s version string while executing other commands. (It also generates the .egg-info/SOURCES.txt manifest file, which is used when you are building source distributions.)

In any case, I believe trying to test building sdist/bdist is necessary to make at least they could be built properly (bdist build is tested in test workflow)

tests/test_pkg.py
```diff
@@ -1,35 +1,107 @@
...
+                tar.extractall("", filter="data")
+            else:
+                tar.extractall("")  # noqa: S202
 
-SRC_TXT_PATH = "src/pymatgen.egg-info/SOURCES.txt"
+        # Check existence of `py.typed` file in each sub-package
+        _check_py_typed_files(sdist_dir / "src/pymatgen")
 
 
-@pytest.mark.skipif(
```
DanielYang59, line 11: I have a feeling that making `test_egg_sources_txt_is_complete` optional is not a good idea, we should setup the test such that it would always run in unit test, otherwise we may or may not be checking the contents at all:
https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/tests/test_pkg.py#L11-L15

I would update `test_pkg` to cover bdist/sdist build

DanielYang59:
I would close it for now as looks like this involves decision on types of some sub-modules (`analysis / ext / io`) from maintainers, free feel to take over this PR.

DanielYang59:
Still I would prefer closing this and let maintainers take over, esp. around the migration from namespace system to entry points.

PR #4203, opened 2024-11-28
Remove redundant `total_ordering` decorator usage

DanielYang59:
### Summary

- Remove redundant `total_ordering` decorator usage, as `total_ordering` is already applied on superclasses and child classes did not redefine any rich comparison operations.
- Clean up comment of  equality definition of `core.IStructure`
- Minor import clean up (avoid `from io import StringIO` when it's used only twice as `io` is already imported)
- Minor docstring clean up

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -1097,21 +1097,22 @@ class IStructure(SiteCollection, MSONable):
 
     def __eq__(self, other: object) -> bool:
         """Define equality by comparing all three attributes: lattice, sites, properties."""
-        if not (hasattr(other, "lattice") and hasattr(other, "sites") and hasattr(other, "properties")):
+        needed_attrs = ("lattice", "sites", "properties")
+        if not all(hasattr(other, attr) for attr in needed_attrs):
             return NotImplemented
+
+        # TODO (DanielYang59): fix below type
+        other = cast(Structure, other)  # make mypy happy
 
         if other is self:
             return True
 
-        if isinstance(other, collections.abc.Sized) and len(self) != len(other):
+        if len(self) != len(other):
             return False
 
         if self.lattice != other.lattice:
             return False
         if self.properties != other.properties:
-            return False
-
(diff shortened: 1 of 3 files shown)


src/pymatgen/core/structure.py
```diff
@@ -1098,9 +1098,9 @@ def __init__(
         self._properties = properties or {}
 
     def __eq__(self, other: object) -> bool:
+        """Define equality by comparing all three attributes: lattice, sites, properties."""
         needed_attrs = ("lattice", "sites", "properties")
 
-        # Return NotImplemented as in https://docs.python.org/3/library/functools.html#functools.total_ordering
```
DanielYang59, line 1103: I guess `total_ordering` was applied on `IStructure` at some point but I believe this comment should be removed now:

https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/core/structure.py#L1006-L1007

Also current implementation of `IStructure` doesn't fulfill the [prerequisite of applying `total_ordering`](https://docs.python.org/3/library/functools.html#functools.total_ordering):
> The class must define one of __lt__(), __le__(), __gt__(), or __ge__(). In addition, the class should supply an __eq__() method.

src/pymatgen/core/structure.py
```diff
@@ -1098,22 +1099,24 @@ def __init__(
...
         if not all(hasattr(other, attr) for attr in needed_attrs):
             return NotImplemented
 
-        # TODO (DanielYang59): fix below type
-        other = cast(Structure, other)  # make mypy happy
```
DanielYang59, line 1108: I believe checking attribute (duck-typing-like) would better reflect the goal of the implementation, i.e. we don't need `other` to be a `Structure` type, we just need it to have certain attributes (`lattice`, `sites` and `properties`)

DanielYang59: I would revert this for now as I'm not 100% sure what is the best way to check for `in` and `len` support

PR #4205, opened 2024-11-29
Fix `assert_str_content_equal`, add tests for testing utils

### LINKED ISSUE #4202: [Dev/Test] Migrate `PymatgenTest` from `unittest.TestCase` to `pytest`
Current [`pytest` framework isn't fully compatible with `unittest.TestCase`](https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses), we should consider migrate to `pytest` framework:
> The following pytest features do not work, and probably never will due to different design philosophies:

>    [Fixtures](https://docs.pytest.org/en/stable/reference/fixtures.html#fixture) (except for autouse fixtures, see [below](https://docs.pytest.org/en/stable/how-to/unittest.html#mixing-fixtures));
>    [Parametrization](https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize);
>    [Custom hooks](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#writing-plugins);

---

For example `pytest.mark.parametrize` doesn't work: 
- https://github.com/pytest-dev/pytest/issues/5382

```python
from unittest import TestCase

import pytest


class TestPytest(TestCase):
    @pytest.mark.parametrize("foo", ["hello", "world"])
    def test_demo(self, foo):
        assert isinstance(foo, str)
```

Gives:
```
    def _callTestMethod(self, method):
>       if method() is not None:
E       TypeError: TestPytest.test_demo() missing 1 required positional argument: 'foo'
```

While the following works just fine:
```diff
- class TestPytest(TestCase):
+ class TestPytest:
```

DanielYang59:
### Summary

- Fix `assert_str_content_equal` not raising `AssertionError` 
- Add unit tests for `util.testing` as it's intended for public usage (and [quite some external code use `PymatgenTest`](https://github.com/search?q=%22%28PymatgenTest%29%3A%22&type=code))
- More human-readable error messages for assert methods in `PymatgenTest`
- [**NEED CONFIRM**] Simplify single-file module `testing/__init__.py` to `testing.py`

---

> [!WARNING]  
> [**NEED CONFIRM**] Simplify single-file module `testing/__init__.py` to `testing.py` (**Is this intended for some purpose?**), I don't think this would be breaking (though the definition of `MODULE_DIR` would change from `util/testing` to `util`, but no code import this AFAIK as the `util/testing` has nothing else, I believe it's intended for `STRUCTURES_DIR` and I have updated this)

---

### Fix `assert_str_content_equal` not raising `AssertionError` 
Current implementation **would not raise `AssertionError` when two strings are not equal** (it only returns a bool): https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/util/testing/__init__.py#L66-L69

And [current usage of it](https://github.com/search?q=repo%3Amaterialsproject%2Fpymatgen%20self.assert_str_content_equal&type=code) take the pattern of `self.assert_str_content_equal` without `assert`.

Several issues found:
- #4206 
- #4207

### DIFF
--- a/src/pymatgen/util/testing.py
+++ b/src/pymatgen/util/testing.py
@@ -122,8 +122,13 @@ class PymatgenTest(TestCase):
         Returns:
             Structure
         """
-        struct = cls.TEST_STRUCTURES.get(name) or loadfn(f"{STRUCTURES_DIR}/{name}.json")
+        try:
+            struct = cls.TEST_STRUCTURES.get(name) or loadfn(f"{STRUCTURES_DIR}/{name}.json")
+        except FileNotFoundError as exc:
+            raise FileNotFoundError(f"structure for {name} doesn't exist") from exc
+
         cls.TEST_STRUCTURES[name] = struct
+
         return struct.copy()
 
     def serialize_with_pickle(
--- a/src/pymatgen/util/testing.py
+++ b/src/pymatgen/util/testing.py
@@ -30,7 +30,7 @@ if TYPE_CHECKING:
 
 MODULE_DIR: Path = Path(__file__).absolute().parent
 
-STRUCTURES_DIR: Path = MODULE_DIR / ".." / "structures"
+STRUCTURES_DIR: Path = MODULE_DIR / "structures"
 
 TEST_FILES_DIR: Path = Path(SETTINGS.get("PMG_TEST_FILES_DIR", f"{ROOT}/../tests/files"))
 VASP_IN_DIR: str = f"{TEST_FILES_DIR}/io/vasp/inputs"
--- a/src/pymatgen/util/testing.py
+++ b/src/pymatgen/util/testing.py
@@ -65,7 +65,8 @@ class PymatgenTest(TestCase):
 
     @staticmethod
     def assert_msonable(obj: Any, test_is_subclass: bool = True) -> str:
-        """Test if an object is MSONable and verify the contract is fulfilled.
+        """Test if an object is MSONable and verify the contract is fulfilled,
+        and return the serialized object.
 
         By default, the method tests whether obj is an instance of MSONable.
         This check can be deactivated by setting `test_is_subclass` to False.
@@ -150,8 +151,7 @@ class PymatgenTest(TestCase):
                 with the original object using the `__eq__` method.
 
         Returns:
-            list: Nested list with the objects deserialized with the specified
-            protocols.
+            list[Any]: Objects deserialized with the specified protocols.
         """
         # Build a list even when we receive a single object.
         got_single_object = False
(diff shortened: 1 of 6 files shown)


src/pymatgen/util/testing.py
```diff
@@ -18,42 +19,105 @@
...
-MODULE_DIR = Path(__file__).absolute().parent
-STRUCTURES_DIR = MODULE_DIR / ".." / "structures"
-TEST_FILES_DIR = Path(SETTINGS.get("PMG_TEST_FILES_DIR", f"{ROOT}/../tests/files"))
-VASP_IN_DIR = f"{TEST_FILES_DIR}/io/vasp/inputs"
-VASP_OUT_DIR = f"{TEST_FILES_DIR}/io/vasp/outputs"
-# fake POTCARs have original header information, meaning properties like number of electrons,
+    from pymatgen.core import Structure
+    from pymatgen.util.typing import PathLike
+
+_MODULE_DIR: Path = Path(__file__).absolute().parent
```
DanielYang59, line 31: **Need confirm**: make `MODULE_DIR` private, as the [`util.testing` directory contains nothing else other than`__init__.py`](https://github.com/materialsproject/pymatgen/tree/master/src/pymatgen/util/testing), I guess it's used to define `STRUCTURES_DIR` only:

https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/util/testing/__init__.py#L27-L28

src/pymatgen/util/testing.py
```diff
@@ -18,42 +19,105 @@
...
+            raise AssertionError(
+                "Strings are not equal (whitespaces ignored):\n"
+                f"{' Actual '.center(50, '=')}\n"
+                f"{actual}\n"
+                f"{' Expected '.center(50, '=')}\n"
+                f"{expected}\n"
+            )
+
     @classmethod
     def get_structure(cls, name: str) -> Structure:
```
DanielYang59, line 115: TODO: `get_structure` with Path doesn't seem to work

src/pymatgen/util/testing.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # Check if the object can be accurately reconstructed from its dict representation
+        if obj.as_dict() != type(obj).from_dict(obj.as_dict()).as_dict():
+            raise ValueError(f"{obj_name} object could not be reconstructed accurately from its dict representation.")
+
+        # Verify that the deserialized object's class is a subclass of the original object's class
+        json_str = json.dumps(obj.as_dict(), cls=MontyEncoder)
+        round_trip = json.loads(json_str, cls=MontyDecoder)
+        if not issubclass(type(round_trip), type(obj)):
+            raise TypeError(f"The reconstructed {round_trip.__class__.__name__} object is not a subclass of {obj_name}")
```
DanielYang59, line 96: This error message is more readable.

Before: `<class 'util.test_testing.TestAssertMSONable.test_not_round_trip.<locals>.NotAKpoints'> != <class 'pymatgen.io.vasp.inputs.Kpoints'>`

Now: `The reconstructed NotAKpoints object is not a subclass of Kpoints`

PR #4209, opened 2024-11-30
Schedule `PymatgenTest` for migration from `unittest` to `pytest` 

DanielYang59:
### Summary
- Avoid catching all warnings, add a message filter, and clean up [assert no warning tests](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests) with warnings filter
- [ ] Should be reviewed/merged after #4212

### Schedule `PymatgenTest` for migration after a grace period

- [x] We cannot migrate `PymatgenTest` directly as it's public and [quite some external code use `PymatgenTest`](https://github.com/search?q=%22%28PymatgenTest%29%3A%22&type=code), added a grace period **until 2026-01-01** and let downstream code finish their migration first (**get rid of some unittest-specific functions**), as [`pytest` is not compatible with some `unittest` functions](https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses)
- [ ] **OR, do we want to deprecated `PymatgenTest` with a more general `MatSciTest` (or some other name) as it's intended to be used by other packages?**
- [x] Issue migration warning, and test methods specific to `TestCase` such that we know potential side effect, also provide a migration guide (not really our job, but at least provide links to some resources), [`ruff` does have rules](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt) (for example [PT009](https://docs.astral.sh/ruff/rules/pytest-unittest-assertion/)) but we cannot trust this as not every project enables `ruff`

---

### Super brief migration guide
- Setup/teardown methods (`setUp`, `setUpClass`, `tearDown`, `tearDownClass`):
    1. [Recommended by `pytest`: Use `fixtures`](https://docs.pytest.org/en/stable/reference/fixtures.html#fixture)
    OR
    2. Use `pytest`'s [xUnit-style setup/teardown](https://docs.pytest.org/en/stable/how-to/xunit_setup.html): `[setup/teardown]_[class/method/function]`.
- Assertion methods (`assertTrue`, `assertFalse`, `assertEqual`, etc.): Replace with direct Python `assert` statements.


src/pymatgen/util/testing/__init__.py
```diff
@@ -41,6 +42,17 @@ class PymatgenTest(TestCase):
     # dict of lazily-loaded test structures (initialized to None)
     TEST_STRUCTURES: ClassVar[dict[str | Path, Structure | None]] = dict.fromkeys(STRUCTURES_DIR.glob("*"))
 
+    @classmethod
+    def setUpClass(cls):
```
DanielYang59, line 46: Need to re-think this, as it would emit a warning no matter unittest-specific functions are used or now, which would be very annoying

src/pymatgen/analysis/elasticity/elastic.py
```diff
@@ -479,8 +479,7 @@ def from_pseudoinverse(cls, strains, stresses) -> Self:
             "questionable results from vasp data, use with caution."
         )
         stresses = np.array([Stress(stress).voigt for stress in stresses])
-        with warnings.catch_warnings():
```
DanielYang59, line 482: Looks like a bare `with warnings.catch_warnings():` (without any filter or such) is not doing anything

src/pymatgen/command_line/mcsqs_caller.py
```diff
@@ -191,9 +190,7 @@ def _parse_sqs_path(path) -> Sqs:
         process = Popen(["str2cif"], stdin=input_file, stdout=output_file, cwd=path)
         process.communicate()
 
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
```
DanielYang59, line 195: I'm not see any warning from this, would remove the filter

tests/command_line/test_mcsqs_caller.py
```diff
@@ -103,5 +103,6 @@ def test_mcsqs_caller_runtime_error(self):
         struct.replace_species({"Ti": {"Ti": 0.5, "Zr": 0.5}, "Zr": {"Ti": 0.5, "Zr": 0.5}})
         struct.replace_species({"Pb": {"Ti": 0.2, "Pb": 0.8}})
         struct.replace_species({"O": {"F": 0.8, "O": 0.2}})
-        with pytest.raises(RuntimeError, match="mcsqs exited before timeout reached"):
```
DanielYang59, line 106: This test is failing as a different `RuntimeError` is raised:
```
>       with pytest.raises(RuntimeError, match="mcsqs exited before timeout reached"):
E       AssertionError: Regex pattern did not match.
E        Regex: 'mcsqs exited before timeout reached'
E        Input: 'mcsqs did not generate output files, is search_time sufficient or are number of instances too high?'
```

But looks like both exceptions are very similar (all pointing to timeout) as I would just change the error message here: https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/command_line/mcsqs_caller.py#L136-L160

tests/command_line/test_gulp_caller.py
```diff
@@ -132,15 +131,6 @@ def test_structure_lines_no_frac_coords(self):
         assert "cell" not in inp_str
         assert "cart" in inp_str
 
-    @unittest.skip("Not Implemented yet")
-    def test_specie_potential(self):
-        pass
-
-    @unittest.expectedFailure
-    def test_library_line_explicit_path(self):
```
DanielYang59, line 140: I would remove this test as it is hard-coded to a specific path, even with `expectedFailure` it doesn't really seem to be checking anything

DanielYang59:
I would reopen this after #4212 is reviewed/merged as I opened too many PRs for me to track

PR #4212, opened 2024-12-01
Deprecate `PymatgenTest`, migrate tests to `pytest` from `unittest`

### LINKED ISSUE #4202: [Dev/Test] Migrate `PymatgenTest` from `unittest.TestCase` to `pytest`

DanielYang59:
### Summary
- Migrate `pymatgen` tests to `pytest`, to close #4202
- [x] `conftest.py` doesn't seem to be used at all
- [x] Add tests for legacy `PymatgenTest` to avoid accidental breakage

> [!WARNING]
> - [**NEED CONFIRM**] Deprecate `PymatgenTest` with `MatSciTest`, set **deadline at 2026-01-01**

---

#### `setUp` -> `setup_method` even when not inheriting from `TestCase`

Test classes that doesn't subclass `TestCase` and define `setUp` method would work in `pytest`, but [it's not a native support](https://docs.pytest.org/en/stable/deprecations.html), so I would still migrate them here:
> One thing that might catch users by surprise is that plain setup and teardown methods are not pytest native, they are in fact part of the nose support.
> Native pytest support uses setup_method and teardown_method (see [Method and function level setup/teardown](https://docs.pytest.org/en/stable/how-to/xunit_setup.html#xunit-method-setup)), so the above should be changed to:


tests/io/test_lmto.py
```diff
@@ -23,20 +23,20 @@
 MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
 
 
-class TestCtrl(PymatgenTest):
-    def setUp(self):
+class TestCtrl(MatSciTest):
+    def setup_method(self):
         os.chdir(TEST_DIR)
```
DanielYang59, line 28: Can chdir be replaced by tempdir?

src/pymatgen/util/testing/__init__.py
```diff
@@ -35,8 +35,15 @@
 FAKE_POTCAR_DIR = f"{VASP_IN_DIR}/fake_potcars"
 
 
-class PymatgenTest(TestCase):
-    """Extends unittest.TestCase with several assert methods for array and str comparison."""
+class MatSciTest:
```
DanielYang59, line 38: I originally didn't plan to deprecate `PymatgenTest` in this PR, but it turns out we have to do this here, otherwise tests that subclassing `PymatgenTest` would still rely on the `TestCase` and `pytest` methods like `setup_method/class` would not be compatible, meaning we have to do another migrate after `PymatgenTest` dropped `TestCase` inheritance. 

So I guess the easiest way is not to drop inheriting `TestCase` by `PymatgenTest`, but replacing `PymatgenTest` with another one that doesn't subclassing `TestCase`.

I drafted a name here but free feel to comment if you have better ideas (batch rename would be super easy with an IDE so don't worry).

shyuep:
There seems to be conflicts that need to be adjusted.

DanielYang59:
Fixed, thanks for reviewing. Looks like COD might be temporarily down thus the failure

DanielYang59:
@shyuep Can you maybe review this? Merge conflicts solved. Thanks.

DanielYang59:
Thanks

PR #4218, opened 2024-12-06
Explicit UTF-8 encoding for VASP input files with `zopen`, and `open` for other text files

### LINKED ISSUE #4214: For band structure, how to correctly display the symbol of the \Gamma point
### Python version

python--3.12

### Pymatgen version

2024.11.13

### Operating system version

Win-11

### Current behavior

When I was drawing the band structure, I found some problems that the gamma points were not displayed correctly.

### Expected Behavior

How should I solve this problem?

### Minimal example

_No response_

_No response_

DanielYang59:
### Summary
- Explicit UTF-8 encoding when reading/writing VASP input files with `zopen`, to fix #4214
- Explicit UTF-8 encoding for standard `open`
- `pytest` ignore `UserWarning` only by default instead of all warnings
- [x] Add test for non-ASCII char in Kpoints comment

### Rationale 

[PEP 597 – Add optional EncodingWarning](https://peps.python.org/pep-0597/):
> [Using the default encoding is a common mistake](https://peps.python.org/pep-0597/#using-the-default-encoding-is-a-common-mistake)

> Developers using macOS or Linux may forget that the default encoding is not always UTF-8.

> For example, using long_description = open("README.md").read() in setup.py is a common mistake. Many Windows users cannot install such packages if there is at least one non-ASCII character (e.g. emoji, author names, copyright symbols, and the like) in their UTF-8-encoded README.md file.

> Of the 4000 most downloaded packages from PyPI, 489 use non-ASCII characters in their README, and 82 fail to install from source on non-UTF-8 locales due to not specifying an encoding for a non-ASCII file. [[1]](https://peps.python.org/pep-0597/#id10)

> Another example is logging.basicConfig(filename="log.txt"). Some users might expect it to use UTF-8 by default, but the locale encoding is actually what is used. [[2]](https://peps.python.org/pep-0597/#id11)

Meanwhile [PEP 686 – Make UTF-8 mode default](https://peps.python.org/pep-0686/) should resolve this altogether but it wouldn't be in place until Python 3.15

Ruff rule: [unspecified-encoding (PLW1514)](https://docs.astral.sh/ruff/rules/unspecified-encoding/#unspecified-encoding-plw1514)


QuantumChemist:
As you have started this already, let me know when I shall step in (if needed).

DanielYang59:
Thanks for saying that I could handle this as there aren't too many to fix across the entire code base (~300) and unsafe fixes from `ruff` is available, should be able to fix them pretty fast (though need to double check the `ruff` fix as it's in preview)


```bash
>>> ruff check . --select PLW1514 --preview

Found 282 errors.
No fixes available (282 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

QuantumChemist:
Yes, I have a bit of time to spare for a review 😎
Ping me when you are done ☺️

dev_scripts/update_pt_data.py
```diff
@@ -123,7 +123,7 @@ def parse_radii():
             print(el)
     with open("periodic_table2.yaml", mode="w") as file:
         yaml.dump(data, file)
-    with open("../pymatgen/core/periodic_table.json", mode="w") as file:
+    with open(f"{PKG_DIR}/core/periodic_table.json", mode="w") as file:
         json.dump(data, file)
```
DanielYang59, line 127: I would not change `open` for `json.dump` at this moment, as [`json.dump` would escape non-ASCII chars by default](https://docs.python.org/3/library/json.html#json.dump):
> If ensure_ascii is true (the default), the output is guaranteed to have all incoming non-ASCII characters escaped. If ensure_ascii is false, these characters will be output as-is.

tests/phonon/test_thermal_displacements.py
```diff
@@ -335,8 +334,7 @@ def test_visualization_directionality_criterion(self):
         self.thermal.visualize_directionality_quality_criterion(
             filename=f"{self.tmp_path}/U.vesta", other=self.thermal, which_structure=0
         )
-        with open(f"{self.tmp_path}/U.vesta") as file:
-            file.seek(0)  # set position to start of file
```
DanielYang59, line 339: Looks like ref point reset is not necessary here as there is no prior operation

DanielYang59:
@QuantumChemist I went through the changes myself and didn't see much issue (not trying to bias you), please help me review those changes if you have time, really appreciate that.

I would separate the remaining change to `zopen` to a separate PR as this one is already huge, and I would need some investigation on the behaviour of `zopen` as it wraps around multiple `gzip/bz2/lzma` to make sure encoding arg works for them as well.

Thanks again!

QuantumChemist:
@DanielYang59 , sorry this weekend was big exhaustion time and today I also had a day off but now I have the time and energy for reviewing x)

dev_scripts/update_pt_data.py
QuantumChemist: then let's put the encoding explicitly

src/pymatgen/core/__init__.py
```diff
@@ -32,11 +32,11 @@
...
-SETTINGS_FILE = os.path.join(os.path.expanduser("~"), ".config", ".pmgrc.yaml")
-OLD_SETTINGS_FILE = os.path.join(os.path.expanduser("~"), ".pmgrc.yaml")
-MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
-PKG_DIR = os.path.dirname(MODULE_DIR)
-ROOT = os.path.dirname(PKG_DIR)
+SETTINGS_FILE: str = os.path.join(os.path.expanduser("~"), ".config", ".pmgrc.yaml")
+OLD_SETTINGS_FILE: str = os.path.join(os.path.expanduser("~"), ".pmgrc.yaml")
+MODULE_DIR: str = os.path.dirname(os.path.abspath(__file__))
+PKG_DIR: str = os.path.dirname(MODULE_DIR)
+ROOT: str = os.path.dirname(PKG_DIR)
```
QuantumChemist, line 39: would there be a benefit to make them `Path` objects?

QuantumChemist (commented):
I think there is not a lot to review here xD
I'm just wondering how we could make sure that no "open ... " line was missed out 🤔 

I can also test this implementation on Wednesday (will not be able to test it tomorrow) with a `Γ` in a KPOINTS file 👀

dev_scripts/update_pt_data.py
DanielYang59: Oh yes of course, sorry I missed this one, thanks for catching this!

DanielYang59:
> sorry this weekend was big exhaustion time and today I also had a day off but now I have the time and energy for reviewing x)

That's totally alright, no one would expect you to work on off times :) Thanks a lot!

DanielYang59:
> Thanks for noticing this, I originally used the [ruff PLW1514 rule](https://docs.astral.sh/ruff/rules/unspecified-encoding/) to identify them, and indeed it seems to miss some (understandably, as that rule is still in preview mode, I would report the missing ones to them).

Looks like I was wrong here, upon a closer look it appears to me all `open` within code (some might be in comment) now has `encoding` in text mode

> https://github.com/materialsproject/pymatgen/pull/4218#discussion_r1876792894

I believe this code shown is at the time of comment and would not be updated automatically, i.e. the `open` in `update_pt_data.py` already has encoding now

tests/io/vasp/test_inputs.py
```diff
@@ -1293,6 +1293,11 @@ def test_automatic_monkhorst_vs_gamma_style_selection(self):
                     kpoints = Kpoints.automatic_density_by_lengths(struct, [len_density] * 3)
                     assert kpoints.style == Kpoints.supported_modes.Gamma
 
+    def test_non_ascii_comment(self):
+        """Non-ASCII comment like 'Γ' might not be encoded correctly in Windows."""
+        kpoints = Kpoints.from_file(f"{VASP_IN_DIR}/KPOINTS_band")
+        assert kpoints.labels[0] == "Γ"
```
DanielYang59, line 1299: This should be able to verify the encoding, in my Windows machine (system language English but install Chinese input, so the default encoding becomes `gb2312`).

If I revert the explicit encoding, I would get exact error as that issue:
```
    def test_non_ascii_comment(self):
        """Non-ASCII comment like 'Γ' might not be encoded correctly in Windows."""
        kpoints = Kpoints.from_file(f"{VASP_IN_DIR}/KPOINTS_band")
>       assert kpoints.labels[0] == "Γ", kpoints.labels[0]
E       AssertionError: 螕
```

QuantumChemist:
> > sorry this weekend was big exhaustion time and today I also had a day off but now I have the time and energy for reviewing x)
> 
> That's totally alright, no one would expect you to work on weekends/holidays :) Thanks a lot!
> 
I know that nobody expects to work on weekends, I just thought I would have the time and energy but was not the case against my expectation. 
>
> > I can also test this implementation on Wednesday (will not be able to test it tomorrow) with a Γ in a KPOINTS file 👀
> 
> In fact I tested on my local machine, and of course additional test is always welcome (I should add a test in CI, in TODO list now). However a pitfall may be: your system could be using a default encoding where `Γ` is decoded correctly (apparently the default Chinese "GB" encoding is not one of them), so please make sure you could recreate the incorrect decoding error before testing the fix (otherwise you would get the correct result with and without this fix).
> 
> > I'm just wondering how we could make sure that no "open ... " line was missed out
> 
> Thanks for noticing this, I originally used the [ruff PLW1514 rule](https://docs.astral.sh/ruff/rules/unspecified-encoding/) to identify them, ~and indeed it seems to miss some (understandably, as that rule is still in preview mode, I would report the missing ones to them).~
> 
> **Update:** looks like I was wrong here, upon a closer look I believe all `open` within code (some might be in comment [6a90d2d](https://github.com/materialsproject/pymatgen/commit/6a90d2d9fd858ddbec19c7e4330a462f6b584ff3)) now has `encoding` in text mode.
> 
> I believe this code shown with [#4218 (comment)](https://github.com/materialsproject/pymatgen/pull/4218#discussion_r1873344121) is at the time of comment and would not be updated automatically, i.e. the `open` in `update_pt_data.py` already has encoding.
> 
> Not really useful anymore

An `EncodingWarning` in the pyproject.toml sounds like the most solid option to me.

tests/io/vasp/test_inputs.py
DanielYang59: No I didn't remove it, the code diff shown by a comment would not be updated (i.e. I updated the err msg in 2608e8a2eb107e510b2545e10edf4bafd7a1fae1 after this comment so you wouldn't see that part)

QuantumChemist: Ah oki good then 😃 GitHub is also bugging for me today. Nice then! 👍🏻

src/pymatgen/core/__init__.py
QuantumChemist: so? what is your decision? I'm also fine with `str`, I just try to keep in mind to mae things as flexible and/or adjustable as possible. 😃

DanielYang59: Sorry for the lack of response I was busy fix the CI failures yesterday. I don't hold very strongly to prefer one over another at this moment (the decision by Janosh) and I might prefer to keep it as is for now until both have comparable performance maybe? At at that time we could globally migrate to `pathlib` handling?

DanielYang59:
> I know that nobody expects to work on weekends, I just thought I would have the time and energy but was not the case against my expectation.

I guess there is always a gap between the ideal image of self and and real us :) You have already done so much BTW

> An EncodingWarning in the pyproject.toml sounds like the most solid option to me.

Thanks for the comment, with the [unspecified-encoding (PLW1514)](https://docs.astral.sh/ruff/rules/unspecified-encoding/#unspecified-encoding-plw1514) rule enable `open` should have already been covered now, only thing left is `zopen` AFAIK so I might migrate this guard there #4222 otherwise test would fail from here .

src/pymatgen/core/__init__.py
QuantumChemist: sounds good to me!

QuantumChemist:
I wasn't aware that `EncodingWarning` is only available from Python 3.13 on, though, so maybe your current setup is more than good enough already.

src/pymatgen/core/__init__.py
DanielYang59: Thanks for the confirm :)

DanielYang59:
> I wasn't aware that EncodingWarning is only available from Python 3.13 on, though, so maybe your current setup is more than good enough already.

I would see if I could get the warning filter to work in #4222 so let's discuss there maybe, apparently the filter I added in  bbd53bfa8a008ea5c2e96d3f2096186a60081ea9 is not doing what I expected as no error is thrown after I intentionally removed some encoding in 2ccf30c72e7e16c95f3c8c0cd2aa216ab097c1eb

The `EncodingWarning` `zopen` issued is the following custom one but there must have been something wrong with [my warning filter syntax](https://docs.python.org/3/library/warnings.html#describing-warning-filters):
https://github.com/materialsvirtuallab/monty/blob/26acf0b2900b5074143ed64ccd1bdea6ba9f6705/src/monty/io.py#L25

DanielYang59:
@QuantumChemist Sorry there was a major typo, the EncodingWarning was added in Python 3.10 not Python 3.13

The reason for the custom `EncodingWarning` in `monty.io.zopen`: `monty` only bumped min Python to Python 3.10 two days ago and at the time `zopen` change was made Python 3.9 was still supported, so the custom warning could be dropped now :)

QuantumChemist:
> @QuantumChemist Sorry there was a major typo, [the EncodingWarning was added in Python 3.10](https://peps.python.org/pep-0597/) not Python 3.13
> 
> The reason for the custom `EncodingWarning` in `monty.io.zopen`: `monty` only bumped min Python to Python 3.10 two days ago in [materialsvirtuallab/monty#709](https://github.com/materialsvirtuallab/monty/pull/709) and at the time `zopen` change was made Python 3.9 was still supported, so [the custom warning could be dropped now](https://github.com/materialsvirtuallab/monty/pull/732/commits/d54bde4c6fa03839e7c2716fa46d164d10ae0474) :)

I see it was a typo xD

> I would see if I could get the warning filter to work in https://github.com/materialsproject/pymatgen/pull/4222 so let's discuss there maybe

yeah, I will have a closer look in the other PR too :)

shyuep:
Is this ready to be merged?

DanielYang59:
> Is this ready to be merged?

Yes as far as I'm aware, thank you!

QuantumChemist:
> Is this ready to be merged?

yes, ready to be merged.

mkhorton:
Thanks @DanielYang59 and thanks for the review @QuantumChemist, I've set this to auto-merge

PR #4219, opened 2024-12-07
`zopen`: explicit binary/text `mode` , and explicit `encoding` as UTF-8

DanielYang59:
### Summary

- Explicit binary/text `mode` for `zopen`, I have opened a PR from `monty`'s side to prevent this https://github.com/materialsvirtuallab/monty/pull/730
- [x] Explicitly use UTF-8 `encoding` for `zopen`

Detailed reasoning and example are given in https://github.com/materialsvirtuallab/monty/pull/730#issue-2724465284. In a word, not passing `b/t` in `mode` would cause different and unexpected behaviour when opening a text file (`file.txt`) versus a compressed file (`file.gz`), for the former we would get `TextIO` which gives string but `BinaryIO` for the latter which gives bytes.

---

There're few cases where this wouldn't cause problem, for example [`json.load` accepts both binary and text IO](https://docs.python.org/3/library/json.html#json.load):
>  json.load(fp, *, cls=None, object_hook=None, parse_float=None, parse_int=None, parse_constant=None, object_pairs_hook=None, **kw)

> Deserialize fp (a .read()-supporting [text file](https://docs.python.org/3/glossary.html#term-text-file) or [binary file](https://docs.python.org/3/glossary.html#term-binary-file) containing a JSON document) to a Python object using this [conversion table](https://docs.python.org/3/library/json.html#json-to-py-table).

For such cases, I tested with the [pdag3_complete_dos.json.gz](https://github.com/materialsproject/pymatgen/blob/master/tests/files/electronic_structure/dos/pdag3_complete_dos.json.gz) file, "text mode" seems slightly faster:
```
Binary mode: Loaded data (first 100 chars): {'@module': 'pymatgen.electronic_structure.dos', '@class': 'CompleteDos', 'efermi': -0.28973031, 'st...
Time taken (binary mode): 0.3674 seconds
Text mode: Loaded data (first 100 chars): {'@module': 'pymatgen.electronic_structure.dos', '@class': 'CompleteDos', 'efermi': -0.28973031, 'st...
Time taken (text mode): 0.3450 seconds
```


src/pymatgen/io/cif.py
```diff
@@ -1761,5 +1761,5 @@ def write_file(
         mode: Literal["w", "a", "wt", "at"] = "w",
```
DanielYang59, line 1761: Bad allowed `mode`

tests/io/qchem/test_utils.py
```diff
@@ -36,6 +36,7 @@ def test_process_parsed_hess(self):
             binary = file.read()
             data_132 = [struct.unpack("d", binary[ii * 8 : (ii + 1) * 8])[0] for ii in range(len(binary) // 8)]
 
+        # TODO: why not utf-8 encoding?
         with zopen(f"{TEST_DIR}/parse_hess/HESS", mode="rt", encoding="ISO-8859-1") as file:
```
DanielYang59, line 40: Tag

PR #4221, opened 2024-12-11
Revert "`zopen`: explicit binary/text `mode` , and explicit `encoding` as UTF-8"

shyuep:
Reverts materialsproject/pymatgen#4219. Tests fail because of this. @DanielYang59 Pls fix and resubmit PR.


DanielYang59:
Sure thing would look into this

DanielYang59:
@shyuep I would look into those test failure in current master branch as they're seemingly related to my PRs

DanielYang59:
AFAIK that is not related to #4219 but recent `monty` releases `numpy<2` pin and `numpy` bumped to 2, while `netcdf4` doesn't support NP2, in any case I would be responsible for this and would try to fix ASAP

PR #4223, opened 2024-12-11
Fix CI failure 

DanielYang59:
Fix CI failure

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ dependencies = [
     "uncertainties>=3.1.4",
     # NumPy documentation suggests pinning the current major version as the C API is used
     # https://numpy.org/devdocs/dev/depending_on_numpy.html#runtime-dependency-version-ranges
-    "numpy>=1.25.0,<3",
+    "numpy>=1.25.0,<2",
 ]
 version = "2024.11.13"
 


pyproject.toml
```diff
@@ -56,7 +56,7 @@ classifiers = [
 dependencies = [
     "joblib>=1",
     "matplotlib>=3.8",
-    "monty>=2024.10.21",
+    "monty>=2024.10.21,<2024.12.10",  # DEBUG: latest monty breaks abinit
```
DanielYang59, line 59: I must have broken `Namespace` from `monty` somewhere, need further investigation

src/pymatgen/io/abinit/pseudos.py
```diff
@@ -601,7 +601,7 @@ def _dict_from_lines(lines, key_nums, sep=None) -> dict:
     if len(lines) != len(key_nums):
         raise ValueError(f"{lines = }\n{key_nums = }")
 
-    kwargs = Namespace()
```
DanielYang59, line 604: I must have broken `Namespace` dict from `monty` somewhere, also need to double check coverage:

https://github.com/materialsproject/pymatgen/blob/b28b4c24d92830327f2aaf199804b726d680437c/src/pymatgen/io/abinit/pseudos.py#L604-L627

src/pymatgen/io/abinit/pseudos.py
```diff
@@ -601,7 +601,9 @@ def _dict_from_lines(lines, key_nums, sep=None) -> dict:
...
```
DanielYang59, line 604: The original implementation uses the `Namespace` dict, which according to the original docstring (before my clean up in https://github.com/materialsvirtuallab/monty/pull/722):
> A dictionary that does not permit to redefine its keys.

The issue is two-fold:
1. My fault goes first, looks like the override on `update` is not complete to handle `Iterator` correctly, and surely I would fix this ASAP:
```python
from monty.collections import Namespace

namespace_dict = Namespace()

namespace_dict.update(zip(["a", "b"], [1, 2]))
print(namespace_dict)

namespace_dict.update({"a":1, "b":2})
print(namespace_dict)
``` 

After fixing the issue, however the usage of `Namespace` in `oncvpsp_header` method of `io.abinit.pseudo.NcAbinitHeader` seems incorrect with `pop`:
https://github.com/materialsproject/pymatgen/blob/b28b4c24d92830327f2aaf199804b726d680437c/src/pymatgen/io/abinit/pseudos.py#L783

DanielYang59:
Thanks!

PR #4226, opened 2024-12-16
Fix file existence check in ChargemolAnalysis to verify directory instead.

lllangWV:
atomic_densities dir for chargemol is a dir not a file. 

I am working with chargemol in a high throughput context with Atomate2. When trying to perform chargemol calculations I could not get past `if not os.path.isfile(atomic_densities_path):`. I believe this is because if you provide a directory, which is what `atomic_densities_path` should be, this will always trigger preventing the operation. Unless I am wrong with the functioning of `os.path.isfile`.

## Summary

Major changes:

- feature 1: Changed isfile to isdir
- fix 1: Changed isfile to isdir


mkhorton:
Thanks for flagging this @lllangWV 

Tagging @mhsiron @Andrew-S-Rosen for this one, I haven't used the Chargemol functionality myself. Not clear how this worked previously if it is an error?

Andrew-S-Rosen:
Almost certainly this is the correct patch, although I must admit that I don't think I've used this function in a long time (likely before any refactoring).

DanielYang59:
I guess this is an slight oversight from (which is very likely done by batch regular expression replacement or similar):
- https://github.com/materialsproject/pymatgen/pull/3690

shyuep:
Thanks.

PR #4228, opened 2024-12-17
Replace some deprecated API

DanielYang59:
### Summary

Replace some deprecated API:
- `spglib` dict interface
- [x] Revert deprecation warning as error


mkhorton:
Thanks @DanielYang59!

PR #4229, opened 2024-12-17
AseAtomsAdaptor: Support arbitrary selective dynamics constraints

yantar92:
## Summary

Major changes:

- Implement full selective_dynamics support when converting to/from ASE atoms.

## Todos

- I need directions to add tests.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


jmmshn:
Hi @yantar92, 

> I need directions to add tests.

You can use this example here:
https://github.com/materialsproject/pymatgen/blob/da607e86f9ce8aec942067c6c1a4fda6e04915dd/tests/io/test_ase.py#L186-L203

Also can you detail how this is different than the existing `set_constraint` method?  It looks like some ase classes being used there changed, does this change require an update to the ASE version?

yantar92:
> Also can you detail how this is different than the existing `set_constraint` method? It looks like some ase classes being used there changed, does this change require an update to the ASE version?

I replaced `FixAtoms` ASE constraint with `FixCartesian`. `FixCartesian` allows fixing individual coordinates while `FixAtoms` fixes everything.

`FixCartesian` thus provides everything needed to implement "selective dynamics" from VASP. It is a strict superset of `FixAtoms`.

`FixCartesian` is available in ASE since forever (introduced in e2e1799d5, 16 years ago). There should no bo issues with compatibility.

yantar92:
I added new tests now.

shyuep:
Thanks.

PR #4233, opened 2024-12-23
Fix ion formula check in `ion_or_solid_comp_object` of `analysis.pourbaix_diagram`

### LINKED ISSUE #4231: Mistake in regex string of ion_or_solid_comp_object leads to example string formula "Na[+]" failing
### Python version

Python 3.12.3

### Pymatgen version

 2024.11.13

### Operating system version

Windows 10

### Current behavior

The following code leads to an error (ValueError: + is an invalid formula!)
```
from pymatgen.analysis.pourbaix_diagram import ion_or_solid_comp_object
fails = ion_or_solid_comp_object('Na[+]')
```
Even though the definition of ion_or_solid_comp_object states that the argument string should be formatted as 'Na[+]'
```
def ion_or_solid_comp_object(formula):
    """Get an Ion or Composition object given a formula.

    Args:
        formula: String formula. Eg. of ion: NaOH(aq), Na[+];
            Eg. of solid: Fe2O3(s), Fe(s), Na2O

    Returns:
        Composition/Ion object
    """
    if re.match(r"\[([^\[\]]+)\]|\(aq\)", formula):
        comp_obj = Ion.from_formula(formula)
    elif re.search(r"\(s\)", formula):
        comp_obj = Composition(formula[:-3])
    else:
        comp_obj = Composition(formula)
    return comp_obj
```
In it's current form, the string needs to be '[+]Na' in order for it to work:
```
from pymatgen.analysis.pourbaix_diagram import ion_or_solid_comp_object
works = ion_or_solid_comp_object('[+]Na')
```

### Expected Behavior

If 'Na[+]' is the desired string format, the definition of ion_or_solid_comp_object should be modified

L398 of pourbaix_diagram.py is currently:
```
if re.match(r"\[([^\[\]]+)\]|\(aq\)", formula):
```
https://github.com/materialsproject/pymatgen/blob/da607e86f9ce8aec942067c6c1a4fda6e04915dd/src/pymatgen/analysis/pourbaix_diagram.py#L398C18-L398C19

I suggest replacement with 
```
if re.match(r".*\[([^\[\]]+)\]|\(aq\)", formula):
```

I am not super familiar with git workflows and pull requests so I have sent this as an issue. Please advise if you think this is something that should be fixed and if you wish me to do a pull request :)

### Minimal example

```Python
from pymatgen.analysis.pourbaix_diagram import ion_or_solid_comp_object

works = ion_or_solid_comp_object('[+]Na')
fails = ion_or_solid_comp_object('Na[+]')
```

_No response_

DanielYang59:
### Summary

- Fix ion formula in `ion_or_solid_comp_object` of `analysis.pourbaix_diagram`, to fix #4231
- [x] Add tests
- [x] Add type annotations


src/pymatgen/analysis/pourbaix_diagram.py
```diff
@@ -62,12 +66,15 @@
 
 logger = logging.getLogger(__name__)
 
-PREFAC = 0.0591
+PREFAC: float = 0.0591
```
DanielYang59, line 69: I guess it would be better to use to original $\frac{RT}{nF}$ instead of rounded one?

src/pymatgen/analysis/pourbaix_diagram.py
```diff
@@ -377,34 +395,30 @@ def from_dict(cls, dct: dict) -> Self:
...
-    if re.match(r"\[([^\[\]]+)\]|\(aq\)", formula):
-        comp_obj = Ion.from_formula(formula)
-    elif re.search(r"\(s\)", formula):
-        comp_obj = Composition(formula[:-3])
-    else:
-        comp_obj = Composition(formula)
-    return comp_obj
+    # Formula for ion
+    try:
+        return Ion.from_formula(formula)
```
DanielYang59, line 415: TODO: one issue with current implementation, `Na2O` would be treated as `Ion` but from the docstring it should be `Composition`

src/pymatgen/analysis/pourbaix_diagram.py
```diff
@@ -736,12 +768,12 @@ def get_pourbaix_domains(pourbaix_entries, limits=None):
...
         Args:
-            pourbaix_entries ([PourbaixEntry]): Pourbaix entries
+            pourbaix_entries (list[PourbaixEntry]): Pourbaix entries
                 with which to construct stable Pourbaix domains
-            limits ([[float]]): limits in which to do the pourbaix
+            limits (list[list[float]]): limits in which to do the pourbaix
                 analysis
 
-        Returns:
+        Returns:  # TODO: incorrect return type doc
```
DanielYang59, line 776: TODO tag

src/pymatgen/analysis/pourbaix_diagram.py
```diff
@@ -377,34 +395,30 @@ def from_dict(cls, dct: dict) -> Self:
...
-    return comp_obj
+    # Formula for ion
+    if formula.endswith("(aq)") or re.search(r"\[.*\]$", formula) or "-" in formula or "+" in formula:
+        return Ion.from_formula(formula)
 
+    # Formula for solid
+    if formula.endswith("(s)"):
+        return Composition(formula[:-3])
 
-ELEMENTS_HO = {Element("H"), Element("O")}
```
DanielYang59, line 407: This is unlikely breaking, was a public var, but should be a class var

src/pymatgen/analysis/pourbaix_diagram.py
```diff
@@ -1075,22 +1131,3 @@ def domain_vertices(self, entry):
             list of vertices
         """
         return self._pbx._stable_domain_vertices[entry]
-
-
-def generate_entry_label(entry):
```
DanielYang59, line 1080: This should be a helper private function for `PourbaixPlotter` (as per its docstring), also as far as I'm aware, there is [no usage outside this class](PourbaixPlotter), so I might make it a private staticmethod instead

shyuep:
Thanks.

src/pymatgen/analysis/pourbaix_diagram.py
tschaume: This change broke crystaltoolkit. See materialsproject/crystaltoolkit#446 for the fix. I will publish a release candidate

DanielYang59: Hi thanks for letting me know and really sorry to know that. 

Admittedly this is a bit unexpected to me as I don't expect the `analysis.pourbaix_diagram` module to provide a  definition of a set of element hydrogen and oxygen (which seems very specific to certain task IMO).

mkhorton:
@DanielYang59 Removing `ELEMENTS_HO` in https://github.com/DanielYang59/pymatgen/commit/9551b5a14227be7ed295c0f2d020709f4f16bb66 was a breaking change, and needed to be mentioned in https://github.com/materialsproject/pymatgen/blob/master/docs/compatibility.md

This kind of change is not recommended since it can unnecessarily break downstream codes without a benefit in terms of bug fix or new functionality.

Thank you for fixing the regex!

DanielYang59:
@mkhorton thanks for letting me know and sorry for the breakage, which I was already aware of in https://github.com/materialsproject/pymatgen/pull/4233#discussion_r1929392743

> Removing ELEMENTS_HO in https://github.com/DanielYang59/pymatgen/commit/9551b5a14227be7ed295c0f2d020709f4f16bb66 was a breaking change

Yep I was aware it could potentially be breaking (thought it would unlikely be used from outside, and not covered by test as a public interface) and therefore mentioned this in https://github.com/materialsproject/pymatgen/pull/4233#discussion_r1896587008, looks like I should be more explicit about this potential

> needed to be mentioned in https://github.com/materialsproject/pymatgen/blob/master/docs/compatibility.md

Thanks for the reminder, looks like this doc has been slipping my mind for previous breaking/deprecation changes. Would certain keep this in mind.

> This kind of change is not recommended since it can unnecessarily break downstream codes without a benefit in terms of bug fix or new functionality.

Well received! I thought it was a misplaced class variable as it's quite weird for `analysis.pourbaix_diagram` to provide a definition of a set of two elements

PR #4234, opened 2024-12-24
Drop GULP libs from code base

DanielYang59:
Drop GULP libs from code base and refactored `gulp_caller` to take potential file as arg, rationale: 
- `pymatgen` just provides an interface to other packages, but we don't need to carry the dependencies/libs for them?


shyuep:
Thanks. This is an old file because we used to actually test the gulp functionality. This is now more or less defunct.

DanielYang59:
No problem at all, yep we dropped the GULP binary but I forgot about these libs. 

And happy new year of course :)

Issue #1600, opened 2019-09-23
Spin down projection does not work


nicodemosvarnava:
**Describe the bug**
When I try to get the projection on elements and orbitals for my bands
the spin down case does not work.
i.e.
Mn_bands = bands.get_projections_on_elements_and_orbitals({"Mn": ["s", "p","d"]})
print(Mn_bands[Spin.down][n_b][n_k]["Mn"]["s"])
I get the error:
KeyError: <Spin.down: -1>
When I replace Spin.down with Spin.up I don't get errors
**To Reproduce**
Mn_bands = bands.get_projections_on_elements_and_orbitals({"Mn": ["s", "p","d"]})
print(Mn_bands[Spin.up][n_b][n_k]["Mn"]["s"])

**Desktop (please complete the following information):**
 - OS: Mac
 - Version [e.g. 10.14.6]

shyuep:
This implies you do not have a spin down DOS. Is your run spin polarized?

nicodemosvarnava:
Shyuep,

The calculation is spin polarized but with spin-orbit coupling 
so the spin up/down channels are not well defined anymore.


Now

mkhorton:
We would need the relevant file to be able to investigate this.

What do you get if you print `print(Mn_bands.keys())`?

shyuep:
Hmmm... yes, that is paradoxical. 

The most helpful thing would be to generalize the DOS and BS objects to support SOC calculations as well as non-SOC>

Issue #1540, opened 2019-07-25
Improper pdos for the result from SOC calculation (VASP)


Nokimann:
**Describe the bug**
There are four spin numbers in vasprun.xml when we calculate DOS with SOC. They represent 'total, mx, my, and mz' for each OrbitalType. But DOS Object only considers Up and Down as Total and mx (no more than two spin numbers). And, I think Up and Down are not proper namings because they don't come from spin polarization.

**Desktop (please complete the following information):**
CentOS 7
pymatgen / 2019.7.2 / pypi_0 / pypi

**Additional context**

utf:
I'd add this issue is also present with band structure projections. You first have orbital projections and then mx, my, mz spin projections. The spin projection information is not stored for SOC calculations.

Nokimann:
@utf
Did you mean there are sub-class or sub-functions under the object of the orbital projection? Or, I have to construct mx, my, mz spin projections manually?

PR #4239, opened 2024-12-29
Fix DOS parsing for SOC calculations

kavanase:
## Summary

Currently, `pymatgen` does not correctly parse the density of states (DOS) from VASP SOC calculations. This issue was alluded to in https://github.com/materialsproject/pymatgen/issues/1540 & https://github.com/materialsproject/pymatgen/issues/1600.
The `tdos`, `idos`, `complete_dos`, and `pdos` attributes of `Vasprun` give arrays for both `Spin.up` and `Spin.down`, but only `Spin.up` should be present for SOC calculations. Looking at the code, this is found to be the result of `pymatgen`'s `Vasprun` parser (in `_parse_dos()`) assuming that spin arrays other than `"spin 1"` are `Spin.down`, which is true for non-SOC calculations but in SOC calculations "spins" 2-4 are the x/y/z spin projections of the SOC DOS (see https://www.vasp.at/wiki/index.php/LSORBIT):

https://github.com/materialsproject/pymatgen/blob/aab23ff5764ff88057f643f7eca40a5bf81f5aac/src/pymatgen/io/vasp/outputs.py#L1621-L1626

This is fixed here by first checking if it is a SOC calculation (by checking the number of `"spin X"` arrays), and only parsing the first, `Spin.up`, array if it is a SOC calculation, for both `tdos`/`idos` and `pdos`s. This also has the benefit of making `Vasprun` parsing about 10% faster for SOC calculations from my testing. 
I also added a test for this fixed parsing (note there was no other SOC `vasprun.xml` parsing test in the `pymatgen` tests prior to this).

https://github.com/materialsproject/pymatgen/blob/f67bf28f011506545b31141e48e9445600cf551a/src/pymatgen/io/vasp/outputs.py#L1634-L1642

## Checklist


kavanase:
Just to show it now in action, only `Spin.up` arrays present as desired:
image

shyuep:
Many thanks for fixing this.

PR #4240, opened 2024-12-29
Minor update to make `FermiDos` more robust

kavanase:
This is a minor fix to the `get_doping` method in `FermiDos` to make it a little more robust, so that it can handle rare cases where there is only either 0 or 1 energy increments between the VBM and CBM indices (e.g. Dirac cone systems or toy test cases). This is done by starting the CB integral (for electron concentrations) from either the mid-gap index or, if mid-gap index is the VBM, from VBM+1 (= CBM index), and vice versa for the VB integral for hole concentrations.


shyuep:
Thanks. Can you add a unittest for the functionality pls?

kavanase:
Done! Didn't add one at first as it _should_ be a relatively rare case where this actually matters (e.g. Dirac cone systems or toy test cases), but added now!
Some tests seem to be failing for unrelated reasons

mkhorton:
Are you sure they're unrelated? I see e.g.

```
    def test_get_gap(self):
>       assert self.dos.get_gap() == approx(2.0589, abs=1e-4)
E       assert np.float64(2.3163) == 2.0589 � 1.0e-04
E         
E         comparison failed
E         Obtained: 2.3163
E         Expected: 2.0589 � 1.0e-04
```

Thanks for the PR!

mkhorton:
Tagging @mhsiron @Andrew-S-Rosen for this one, I haven't used the Chargemol functionality myself. Not clear how this worked previously if it is an error?

kavanase:
@mkhorton ah yes sorry! I just looked at one failing test and it seemed to be something different.

I've fixed that one now, and it's a nice validation as it now gives a `Dos.get_gap()` value which better matches the `eigenvalue_band_properties` result (which is typically the most reliable).

esoteric-ephemera:
Hi @kavanase and @shyuep  - was this change intended to only impact `FermiDos`? The changes are made to the parent classes and are now changing parsed bandgaps in the Materials Project (noticed this from failing CI in `emmet`) which **don't** use `FermiDos`

Maybe this scheme is more accurate and I need to investigate if that's the case - either way, if the intention was not to impact other classes besides `FermiDos`, I'd ilke to get this fixed quickly

kavanase:
Hi @esoteric-ephemera,
The change was intended to affect all `Dos` objects, to be more accurate for gap determination. For the tests in `pymatgen` and other test cases in [`doped`](https://doped.readthedocs.io/en/latest/), the updated code seemed to be more accurate, but yes from the plots you attach the legacy scheme is def better...

Some of the additional commits on top of the base ones may have adversely affected things. I can investigate and do some more testing; would you be able to link the failing CI and/or share the code you're using to generate those DOS plots?
Apologies for this!
Thanks!

kavanase:
I will try quickly implementing the interpolation-approach mentioned, if we agree that's the best option then?

kavanase:
Implementing this new interpolated scheme gives basically the same results as the legacy scheme here, because the DOS is going from ~0.1 to 0, and the default `tol` is 1e-4, so interpolating across this gives a point basically right beside the 0-point

image
image

kavanase:
@esoteric-ephemera I believe the issues you noted before should all now be sorted with the latest versions of `pymatgen` (from https://github.com/materialsproject/pymatgen/pull/4288, https://github.com/materialsproject/pymatgen/pull/4299), so the `pymatgen` version pin in `emmet-core` (https://github.com/materialsproject/emmet/commit/afc7495858c140da61d79bc82079100b545015a0) can be removed?

esoteric-ephemera:
Hey @kavanase, thanks for the fixes! Yes we [just released](https://github.com/materialsproject/emmet/pull/1206) the version pin on `pymatgen` in `emmet-core`. Since we're planning a release of new data, I think we'll wait to update how electronic structure is parsed, but from what I can tell, the differences introduced by this change are pretty small and sensible (e.g., in materials with zero gap, the CBM and VBM are correctly identical now)

kavanase:
Ok cool!

PR #4253, opened 2025-01-14
ASE <---> pymatgen trajectory conversion

### LINKED ISSUE #4252: Add method `Trajectory.to_ase_trajectory()`
### Feature Requested

`class Trajectory` has a [`from_file` method](https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/core/trajectory.py#L552) that supports ASE `.traj` files. arguably, if from ASE works, would be great if to ASE also works

### Proposed Solution

add method `Trajectory.to_ase_trajectory()`

### Relevant Information

pinging @esoteric-ephemera and @Andrew-S-Rosen in case you'd like to weigh in

esoteric-ephemera:
Adds feature to convert between ASE and pymatgen trajectories via `from_ase` and `to_ase` methods on `Trajectory`.  Allows for carrying energies, forces, stresses, etc. between ASE and pymatgen trajectories, rather than just structural info as previously implemented.

Merging would close #4252. From the discussion there: this PR ensures that any calculator attr's that are set by `AseAtomsAdaptor` are not lost in the trajectory conversion processs.

To-do:
~- Tests~


esoteric-ephemera:
@shyuep and @mkhorton  should be ready for review whenever you get a chance

src/pymatgen/core/trajectory.py
```diff
@@ -734,3 +732,160 @@ def _get_site_props(self, frames: ValidIndex) -> SitePropsType | None:
                 return [self.site_properties[idx] for idx in frames]
             raise ValueError("Unexpected frames type.")
         raise ValueError("Unexpected site_properties type.")
+
+    def to_ase_trajectory(self, **kwargs) -> AseTrajectory:
```
shyuep, line 736: I suggest we just call it to_ase. Given this class is already called Trajectory, I don't think it is a stretch to imagine to_ase meaning the ASE format for trajectory.

src/pymatgen/core/trajectory.py
```diff
@@ -734,3 +732,160 @@ def _get_site_props(self, frames: ValidIndex) -> SitePropsType | None:
...
+
+        Returns:
+            ASE Trajectory
+        """
+        if NO_ASE_ERR is None:
+            return self.pmg_to_ase_trajectory(self, **kwargs)
+        raise ImportError("ASE is required to write .traj files. pip install ase")
+
+    @classmethod
+    def ase_to_pmg_trajectory(
```
shyuep, line 751: This should be just called from_ase.

src/pymatgen/core/trajectory.py
```diff
@@ -734,3 +732,160 @@ def _get_site_props(self, frames: ValidIndex) -> SitePropsType | None:
...
+        if is_pbc:
+            return cls.from_structures(structures, constant_lattice=constant_lattice, frame_properties=frame_properties)
+        return cls.from_molecules(
+            structures,
+            constant_lattice=constant_lattice,
+            frame_properties=frame_properties,
+        )
+
+    @staticmethod
+    def pmg_to_ase_trajectory(
```
shyuep, line 830: I don't think we need a separate method from to_ase?

shyuep (commented):
Thanks. see comments.

src/pymatgen/core/trajectory.py
esoteric-ephemera: Thanks @shyuep - the intent was to have two methods that don't depend on the `Trajectory` class being instantiated to use as general converters. Similar design as `AseAtomsAdaptor`(`get_atoms` and `get_structure` are both staticmethods)

I'm OK changing this if you'd rather just have the `to_ase` method

shyuep: I don't think we need a method that is independent of Trajectory.

shyuep:
Looks good. Thanks. Merged.

PR #4255, opened 2025-01-16
Make run_type in mixing scheme consistent with entries

peikai:
The run_type for entries had been changed from 'R2SCAN' to 'r2SCAN', whereas the `MaterialsProjectDFTMixingScheme()` is still expecting to find 'R2SCAN' entries.

As a result, all r2SCAN entries would be ignored when carrying out GGA(+U)/R2SCAN mixing scheme corrections, thereby affecting the construction of GGA(+U)/R2SCAN mixing phase diagram [^1].

Error message would be:

```
pymatgen\entries\mixing_scheme.py:176: UserWarning: Invalid run_type='r2SCAN' for entry mp-1018134-r2SCAN. Must be one of ['GGA', 'GGA+U', 'R2SCAN']. This entry will be ignored.
```

[^1]: https://docs.materialsproject.org/methodology/materials-methodology/thermodynamic-stability/phase-diagrams-pds#gga-ggau-r2scan

### DIFF
--- a/src/pymatgen/entries/mixing_scheme.py
+++ b/src/pymatgen/entries/mixing_scheme.py
@@ -40,7 +40,7 @@ class MaterialsProjectDFTMixingScheme(Compatibility):
     may lead to unexpected results.
 
     This is the scheme used by the Materials Project to generate Phase Diagrams containing
-    a mixture of GGA(+U) and R2SCAN calculations. However in principle it can be used to
+    a mixture of GGA(+U) and r2SCAN calculations. However in principle it can be used to
     mix energies from any two functionals.
     """
 
@@ -48,7 +48,7 @@ class MaterialsProjectDFTMixingScheme(Compatibility):
         self,
         structure_matcher: StructureMatcher | None = None,
         run_type_1: str = "GGA(+U)",
-        run_type_2: str = "R2SCAN",
+        run_type_2: str = "r2SCAN",
         compat_1: (Compatibility | None) = MaterialsProject2020Compatibility(),  # noqa: B008
         compat_2: Compatibility | None = None,
         fuzzy_matching: bool = True,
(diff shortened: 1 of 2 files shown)


mkhorton:
Good catch, thanks @peikai!

If you could, can you change the check to ensure it is not case sensitive? This is probably the most robust solution, in case some people are still using entries serialised with the former `run_type`.

peikai:
@mkhorton, Sure, I updated the calculation parameters for archive test entries, changing 'R2SCAN' to 'r2SCAN' in parameters['run_type']. Then, added an unittest to check run types in mixing scheme are consistent to calculation parameters. 

In addition, comments that use R2SCAN as a run type are also revised, according to a principle that the run type inherent from calculation parameters of VASP should use 'r2SCAN' to keep consistent with the `parameter` dictionary in entry object. Whereas when talking about the thermo type, 'R2SCAN' is retained to keep consistency since it has been a frequent glossary. 

Also, I would like to mention that there are two run_type terms in an entry object, located in the parameter and the data dictionary respectively. For an R2SCAN entry, the run_type in parameter dict is 'r2SCAN', whereas the run_type in data dict is 'R2SCAN', which still easy to make confusion.

esoteric-ephemera:
Hey @peikai, we recently changed the `run_type` values used by the Materials Project in the `emmet` package to account for how DFT functionals are written in literature. So MP's data now uses `r2SCAN` as the `run_type` rather than `R2SCAN`. Similarly, other data now has `PBEsol` as the `run_type`

There's a mismatch between what the pymatgen corrections/phase diagram class expects and what MP data returns. In the long run, might be better to implement @mkhorton's suggestion since there shouldn't be functionals with names that differ only in capitalization. Happy to do so

peikai:
Hey @esoteric-ephemera, I agree with the improvement on glossary consistency to calculation. Also, the run_type in data dictionary might need update, for all entries, or link them as one directly. The commits above had incorporated an unittest for run_type consistency during mixing scheme correcting.

peikai:
@mkhorton Oh, I misunderstood the word case-sensitive. 

> change the check to ensure it is not case sensitive

But I still like my fixes that conform to the glossary norm, which are actually based on clear principles. So, I closed this PR. It's welcome if @esoteric-ephemera can have a relay.

peikai:
No problem! Thanks for the instructions.

PR #4256, opened 2025-01-17
Fix `Composition` for mixed species & element compositions

kavanase:
`Composition` objects initialised as e.g. `Composition({"Cd2+": 1, "Cd": 1, "Te2-": 2})` do not behave as expected, giving e.g. `Composition.formula = "Cd3 Te2"` instead of `Cd2 Te2`. This is due to `Composition.__getitem__()` matching all species and elements/non-species of a given input string; so that `Composition["Cd"]` here would return `2` (as desired), but than `Composition.items()` (which is use for `get_el_amt_dict`, `formula` and more) gives `("Cd2+", 1), ("Cd", 2), ("Te", 2)` which is incorrect.
In the end, it's an easy fix to just define the `items()` method to avoid the `Composition.__getitem__()` (which is only needed when the user does `Composition[...]`).

Test added, which failed with previous code and passes now.

Context: Took a while to find what was going on here. Was originally flagged by this `ValueError` with `ComputedStructureEntry`: https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/entries/computed_entries.py#L588

With original `pymatgen` code:
image

image

With fixed code:
image

image

Flagged by tests in [ShakeNBreak](https://shakenbreak.readthedocs.io/), because defect generation in `pymatgen-analysis-defects` mixes species with just element symbols; e.g. converting to just element symbol with `get_element` here: https://github.com/materialsproject/pymatgen-analysis-defects/blob/main/pymatgen/analysis/defects/core.py#L566-L576

### DIFF
--- a/src/pymatgen/core/composition.py
+++ b/src/pymatgen/core/composition.py
@@ -203,6 +203,7 @@ class Composition(collections.abc.Hashable, collections.abc.Mapping, MSONable, S
         return iter(self._data)
 
     def items(self) -> ItemsView[Species | Element | DummySpecies, float]:
+        """Returns Dict.items() for the Composition dict."""
         return self._data.items()
 
     def __contains__(self, key) -> bool:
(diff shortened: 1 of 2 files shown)


shyuep:
Great thanks.

PR #4260, opened 2025-01-23
Draft: Use external Package for FHI-aims IO interface

tpurcell90:
## Summary

Major changes:

- moved FHI-aims specific content into a separate package (pyfhiaims) for ease of maintenance (stop multiple libraries from implementing the same basic interface)

## Todos

- Workout how to best test the aims io interface/dependency tracker
- check with community this does not break their setups

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
If the idea is to move FHI aims to a new external package, I think it should be done completely. Then there is no need for the unit tests to be done here. It can be done in the new package. We can also remove all the FHI related test files.

tpurcell90:
Okay that is reasonable, I will move the tests there.

mkhorton:
To clarify, @tpurcell90 does `pyfhiaims` implement a `pymatgen.io` namespace package? It's possible to implement I/O separately, e.g. see [`pymatgen-io-espresso`](https://github.com/Griffin-Group/pymatgen-io-espresso) for a nice example of this.

This allows independent maintenance while still retaining the `pymatgen.io` imports for easier backwards compatibility.

I would be concerned if we remove the I/O wholesale that it might cause problems for others who depend on it.

tpurcell90:
I can take a look at that. And implement it

tpurcell90:
@mkhorton I added it as an add-on. Should I delete the files here to make it easier to maintain?

shyuep:
Thanks!

PR #4268, opened 2025-01-27
Add `Structure.get_symmetry_dataset` convenience method

janosh:
`moyopy` is a successor to `spglib` that's faster and written in Rust. in my usage so far for analyzing MLFF-relaxed WBM structures for https://matbench-discovery.materialsproject.org/tasks/geo-opt, it seems about 4x faster than `spglib` and gives identical results in nearly all 257k cases. a few exceptions were reported in https://github.com/CompRhys/aviary/pull/96 and resolution tracked in https://github.com/spglib/moyo/issues/54

this PR adds a convenience method for directly accessing `moyopy` symmetry analysis results on pymatgen `Structures`. it also defines a new optional dependency set `symmetry` in `pyproject.toml`

```
symmetry = ["moyopy>=0.3", "spglib>=2.5"]
```

pinging @lan496

related ASE issue: https://gitlab.com/ase/ase/-/issues/1612

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -23,8 +23,6 @@ from fnmatch import fnmatch
 from typing import TYPE_CHECKING, Literal, cast, get_args, overload
 
 import numpy as np
-from duecredit import due
-from duecredit.doi import Doi
 from monty.dev import deprecated
 from monty.io import zopen
 from monty.json import MSONable
@@ -45,6 +43,7 @@ from pymatgen.core.units import Length, Mass
 from pymatgen.electronic_structure.core import Magmom
 from pymatgen.symmetry.maggroups import MagneticSpaceGroup
 from pymatgen.util.coord import all_distances, get_angle, lattice_points_in_supercell
+from pymatgen.util.due import Doi, due
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable, Iterator, Sequence
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -48,6 +48,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Iterable, Iterator, Sequence
     from typing import Any, ClassVar, SupportsIndex, TypeAlias
 
+    import moyopy
     import pandas as pd
     from ase import Atoms
     from ase.calculators.calculator import Calculator
@@ -3337,6 +3338,29 @@ class IStructure(SiteCollection, MSONable):
             Structure: with the requested cell type.
         """
         return self.to_cell("conventional", **kwargs)
+
+    def get_moyo_dataset(self, **kwargs) -> moyopy.MoyoDataset:
+        """Get a MoyoDataset object from the structure.
+
+        Args:
+            **kwargs: Additional arguments passed to MoyoDataset constructor.
+
+        Returns:
+            moyopy.MoyoDataset: Object containing moyopy symmetry analysis for structure.
+
+        Raises:
+            ImportError: If moyopy is not installed.
+        """
+        try:
+            import moyopy
+        except ImportError:
+            raise ImportError("moyopy is not installed. Run pip install moyopy.")
+
+        import moyopy.interface
+
+        # Convert structure to MoyoDataset format
+        moyo_cell = moyopy.interface.MoyoAdapter.from_structure(self)
+        return moyopy.MoyoDataset(cell=moyo_cell, **kwargs)
 
 
 class IMolecule(SiteCollection, MSONable):
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -20,7 +20,7 @@ import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from fnmatch import fnmatch
-from typing import TYPE_CHECKING, Literal, cast, get_args
+from typing import TYPE_CHECKING, Literal, cast, get_args, overload
 
 import numpy as np
 from monty.dev import deprecated
@@ -50,6 +50,7 @@ if TYPE_CHECKING:
 
     import moyopy
     import pandas as pd
+    import spglib
     from ase import Atoms
     from ase.calculators.calculator import Calculator
     from ase.io.trajectory import Trajectory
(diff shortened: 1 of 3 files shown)


janosh:
it occurs to me i forgot to add a `@due.dcite` decorator to this method. @lan496 do you have something you'd like users of this method to cite? maybe just a Zenodo release or a `citation.cff` file for now? here's a similar example:

https://github.com/materialsproject/pymatgen/blob/296a0ec856b22da847b2ccd0e55e146e071cb8fd/src/pymatgen/analysis/prototypes.py#L31-L35

src/pymatgen/core/structure.py
```diff
@@ -48,6 +48,7 @@
     from collections.abc import Callable, Iterable, Iterator, Sequence
     from typing import Any, ClassVar, SupportsIndex, TypeAlias
 
+    import moyopy
```
shyuep, line 51: This makes moyopy an actual dependency. Not an optional one. This needs to be in a try catch.

src/pymatgen/core/structure.py
```diff
@@ -3338,6 +3339,29 @@ def to_conventional(self, **kwargs) -> Structure:
         """
         return self.to_cell("conventional", **kwargs)
 
+    def get_moyo_dataset(self, **kwargs) -> moyopy.MoyoDataset:
```
shyuep, line 3342: From the method name, it is unclear it is a symmetry related function. I would name it get_moyo_symmetry_dataset. Alternatively, why not just put the same functionality in get_symmetry_dataset and have a new kwarg called `use_moyo=False`.

shyuep (commented):
Thanks. Please see comments.

src/pymatgen/core/structure.py
janosh: `get_moyo_symmetry_dataset` sounds good.

> Alternatively, why not just put the same functionality in get_symmetry_dataset and have a new kwarg called use_moyo=False.

i'm reluctant to do that as you could no longer have a single return type and so you would give up on auto-complete suggestions in your IDE:

Screenshot 2025-01-27 at 12 13 12

src/pymatgen/core/structure.py
janosh: it's inside `if TYPE_CHECKING:`, so only imported at type-checking time

src/pymatgen/core/structure.py
janosh: besides the better auto-complete UX, directly returning the `Moyo/SpglibDataset` also has the benefit of reducing maintenance burden. when new keys and symmetry results are added to `moyopy` in the future, they don't have to be translated from `MoyoDataset` to `dict` format

mkhorton: If we have one method name for both backends, we would need to make sure the return value has some kind of useful type (eg some minimal dataclass) and not just a bare dict, otherwise I agree the maintenance burden would be too high when the return values inevitably become inconsistent between the different backends.

mkhorton: This could include converting the spglib dict into the `MoyoDataset` format if it seems appropriate, but we really should just have one type returned.

mkhorton: Ok, we take on the maintenance burden. We should still have a good idea of exactly what is being returned here _especially_ if there are multiple backends.

As a proposal, how about we have `get_symmetry_dataset -> moyopy.MoyoDataset | spglib.SpglibDataset` as suggested, but we  use [`typing.overload`](https://docs.python.org/3/library/typing.html#typing.overload) to ensure everything is typed nicely?

Then we get the benefit of the general method name, the multiple backends, and the typing being correct.

lan496:
> do you have something you'd like users of this method to cite? maybe just a Zenodo release or a citation.cff file for now?

@janosh For now, I think the citation for spglib is enough.
```
@article{spglib,
  author = {Atsushi Togo, Kohei Shinohara and Isao Tanaka},
  title = {Spglib: a software library for crystal symmetry search},
  journal = {Sci. Technol. Adv. Mater., Meth.},
  volume = {4},
  number = {1},
  pages = {2384822--2384836},
  year = {2024},
  doi = {10.1080/27660400.2024.2384822},
  url = {https://doi.org/10.1080/27660400.2024.2384822},
}
```

janosh:
the remaining question after [8974a61](https://github.com/materialsproject/pymatgen/pull/4268/commits/8974a61ed1929e8bed619414721d5ec13f9b67e5) what the consistent default `symprec` and `angle_tolerance` across `moyopy` and `spglib` should be. currently each code uses different defaults which can result in different symmetries returned by each backend. i believe the current `pymatgen` default `symprec=0.1` is far too loose. maybe the `spglib` default of `1e-5` is a bit too strict. which is why advocate we adopt the `moyopy`'s default of `1e-4`. curious to hear your thoughts @lan496. also, the `angle_tolerance` in `moyopy` is in radians whereas `spglib` uses degrees, so we need to take of unit conversion and also pick a consistent default

mkhorton:
Also curious to hear a perspective on that, concur the 0.1 default is probably far too loose. I think maybe fine for purposes of loosely grouping structures, for rough relaxations etc. We should probably add a note to the docstring to give a hint to the user what symprecs might be appropriate for what use cases.

shyuep:
I don't have a strong opinion on symprec, but I will give some context. The general idea of all structure / symmetry comparisons in pymatgen is this: if I take two structures that have the "same" (either in structure matcher or in symmetry) and relax them in DFT, would I end up in the same place? In that context, 0.1A is good enough. It is highly unlikely that two structures that differ by 0.1A would relax to two different final structures or symmetries. 

For spglib, I have found that sometimes there are some paradoxical behavior. A structure would fail with either too low or too high symmprec, but is ok in some intermediate range of symprec. 

In the end, it really depends on your philosophy. Basically, if I take a structure and just moves one of the atoms by 0.11A, do you want that to break symmetry?

shyuep:
Sorry, I accidentally merged. Feel free to submit a separate PR to change the default symprec.

janosh:
> Feel free to submit a separate PR to change the default symprec.

happy to but need some consensus on what the default should be. unless i hear otherwise, it'll be `1e-4`. @lan496 what would you recommend for `angle_tolerance`?

shyuep:
@janosh I propose 0.01A. I think 1e-4 is far too strict. 0.01 is already an order of magnitude smaller than the shortest chemical bond.

shyuep:
I can just make this change myself.

shyuep:
One thing to note, there are severe inconsistencies between moyopy and spglib. E.g., spglib returns the international symbol (which any sane symmetry analysis code would do). moyopy just returns the number. This makes it even more imperative that the default return type needs to be a consistent dict / named tuple. Otherwise the end user is going to be confused.

janosh:
You can get both from moyopy. The reason I think 0.1 is a bad default symprec is you shouldn't make the assumption for the user that they want the symmetry of the structure that their input might relax to. By default, you should assume they want the symmetry for the actual structure they pass.

shyuep:
I implemented some basic changes. Feel free to make further mods and submit a PR. 

I don't necessarily disagree that we should return the symmetry of the structure that they pass. But that is a slippery argument. Symmetry is always based on tolerances. One can argue that even if an atom moves by 1e-100 A from the correct position, that already breaks symmetry in the strictest sense of the word. Surely we can all agree that a 1e-100 tolerance is too strict and will cause too many structures to be rendered P1.

The logical determination of symmetry tolerances should be based on **real-world perturbations with an actual impact on properties**. If a 0.01A shift has negligible impact on properties and is in fact 10 times smaller than the shortest chemical bond, it is good enough to say that the structures are symmetrically equivalent. We do not need a stricter cutoff based on some mathematical argument.

lan496:
From an implementation perspective, a symmetry finder (spglib and moyopy) struggles with too small or big symprec.
For too small symprec (say ~1e-10), the numerical accuracy for floating-point number may result in inconsistency: for example, detected symmetry operations with the large symprec may not form a group. 
For too large symprec (~0.1), I sometimes observed weird behaviors as well. I think it comes from the inconsistency in the detected translation subgroup and symmetry operations.

Recently, I've performed a benchmark of these symmetry finders with several symprec choices to Materials Project's entries ([link](https://github.com/spglib/moyo/blob/main/bench/mp/analysis.ipynb)). Let me check how symprec affects detected space-group types.

shyuep:
You cannot determine the necessary symprec simply from just running on all MP structures. If symprec = 0.1 gives cubic and symprec = 0.00001 gives orthorhombic, which is correct?

At room temperature (300K), the typical atomic vibration amplitude is around 0.05 - 0.1 A. We can agree that thermal vibrations shouldn't break symmetry. So a tolerance of 0.01A is fine. (This is already the default for the Structure.get_space_group method).

lan496:
I do not intend to suggest determining a default symprec from statistics.

The following figures show how detected crystal systems are changed by symprec. Low-symmetry systems (triclinic and monoclinic) are especially affected by symprec choice as following an intuition. 

image

image

Interactive HTML for these figures:
[sankey.zip](https://github.com/user-attachments/files/18601356/sankey.zip)

PR #4270, opened 2025-01-27
Add `seed: int = 0` parameter to `Structure.perturb()` method

janosh:
- Update tests to verify reproducible random perturbations with same seed
- Add test cases for different seed values and min_distance parameter


shyuep:
I am confused why this is really needed. The definition of a random perturbation means it is non-deterministic. The only determinism is an inequality based on the amplitude of the perturbation.

janosh:
needed for downstream packages who use the `perturb` method in testing

PR #4274, opened 2025-01-28
Move `occ_tol` to init in `OrderDisorderedStructureTransformation`

Tinaatucsd:
## Summary

Major changes:
- fix 1:  Fixed incompatibility between `OrderDisorderedStructureTransformation` and `StandardTransmuter`:
- Previously, `occ_tol` was a method parameter in `apply_transformation`, making it unconfigurable when used with `StandardTransmuter` (which only passes transformation and optionally `extend_collection` parameters when calling `append_transformation`)
- Moved `occ_tol` to class initialization to allow proper configuration when used in transformation chains

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/src/pymatgen/transformations/standard_transformations.py
+++ b/src/pymatgen/transformations/standard_transformations.py
@@ -454,7 +454,7 @@ class OrderDisorderedStructureTransformation(AbstractTransformation):
     ALGO_BEST_FIRST = 2
     ALGO_RANDOM = -1
 
-    def __init__(self, algo=ALGO_FAST, symmetrized_structures=False, no_oxi_states=False):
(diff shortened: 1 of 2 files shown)


jmmshn:
Thanks @Tinaatucsd, it looks like `occ_tol` should have been placed as an attribute of the class since only the first two parameters are called by the `alchemy` part of the code. 

https://github.com/materialsproject/pymatgen/blob/296a0ec856b22da847b2ccd0e55e146e071cb8fd/src/pymatgen/alchemy/materials.py#L123C1-L126C1

shyuep:
Thanks.

PR #4276, opened 2025-01-29
Fix timeout in `command_line.enumlib_caller.EnumlibAdaptor`

### LINKED ISSUE #4185: Timeout in MagneticStructureEnumerator (and EnumlibAdaptor) does not work
### Python version

3.11.5

### Pymatgen version

2024.11.13

### Operating system version

_No response_

### Current behavior

`enum = MagneticStructureEnumerator(structure)` should stop after 5 minutes by default,
but it does not. Passing a different timeout does not change anything. I'm now making sure that the timeout arg is passed to the EnumlibAdaptor (see bug in #4184).

I did some tests but I could not figure out why the enum.x process does not get killed.
The main logic from https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/src/pymatgen/command_line/enumlib_caller.py#L281 is here:

```
    def _run_multienum(self):
        with subprocess.Popen([enum_cmd], stdout=subprocess.PIPE, stdin=subprocess.PIPE, close_fds=True) as process:
            if self.timeout:
                timed_out = False
                timer = Timer(self.timeout*60, lambda p: p.kill(), [process])

                try:
                    timer.start()
                    output = process.communicate()[0].decode("utf-8")
                finally:
                    if not timer.is_alive():
                        timed_out = True
                    timer.cancel()

                if timed_out:
                    raise TimeoutError("Enumeration took too long")

            else:
                output = process.communicate()[0].decode("utf-8")
```

It seems the the process stops communicating. I tried to use the timeout argument
of the `process.communicate()` method but it does not work neither.

One thing that worked correctly is the same code above with 
enum_cmd="/usr/bin/sleep 6000". This process gets killed as it should.

I have the same behavior both in my laptop and on perlmutter.

Any thoughts about what's causing this?

### Expected Behavior

`enum = MagneticStructureEnumerator(structure)` should stop after 5 minutes by default.

### Minimal example

_No response_

_No response_

DanielYang59:
## Summary

- Fix timeout in `command_line.enumlib_caller.EnumlibAdaptor`, to fix #4185
- [x] Fix unit test, current unit test doesn't seem to work correctly (`Fortran runtime error: Cannot open file 'struct_enum.in': No such file or directory`)


tests/command_line/test_enumlib_caller.py
```diff
@@ -119,11 +118,11 @@ def test_partial_disorder(self):
...
-    @unittest.skip("Fails seemingly at random.")
     def test_timeout(self):
         struct = Structure.from_file(filename=f"{TEST_FILES_DIR}/cif/garnet.cif")
         SpacegroupAnalyzer(struct, 0.1)
         struct["Al3+"] = {"Al3+": 0.5, "Ga3+": 0.5}
-        adaptor = EnumlibAdaptor(struct, 1, 1, enum_precision_parameter=0.01, timeout=0.0000000000001)
+
+        adaptor = EnumlibAdaptor(struct, enum_precision_parameter=0.01, timeout=0.01)
         with pytest.raises(TimeoutError, match="Enumeration took too long"):
-            adaptor._run_multienum()
```
DanielYang59, line 129: The original test command is incorrect, a bare `_run_multienum` doesn't generate the necessary input files, `_gen_input_file` is needed, so perhaps just use `run` instead.

After fixing the run command, the run time on my local Apple M4 machine is ~ 1s, and apparently the timeout is not really working (as reported in the original issue)

DanielYang59:
Hi @fraricci sorry for the late response, can you help me test if this patch works? I don't have a "long enough" enumlib job to test with (the unit test job last only 0.04 s and I'm not sure if it's really validating the behaviour), thanks a lot!

tests/command_line/test_enumlib_caller.py
fraricci: Hey @DanielYang59, I just saw your fix here.  Sure I can test it on my structure if still needed.

fraricci:
Hey @DanielYang59, I just saw your fix here. Sure I can test it on my structure if still needed.

DanielYang59:
Sure, please have a try and let me know whether the fix works for you

fraricci:
I tested and still it does not work. 
I tried to insert some print to understand and found that it does not enter into the `try` condition: 
a print inside that before the output line is not executed

Also, a version that uses `sleep 60` reveals that this logic does catch the timeout but 
it still waits the end of the command to return the raised error
```
ENUM_CMD = ["/usr/bin/sleep","30"]
timeout = 6
with subprocess.Popen(ENUM_CMD, stdout=subprocess.PIPE, stdin=subprocess.PIPE, close_fds=True) as process:
    print(timeout)
    try:
        print("running")
        output = process.communicate(timeout=timeout)[0].decode("utf-8")
        print("running ended")
        
    except subprocess.TimeoutExpired as exc:
        print("timeout")
        raise TimeoutError("Enumeration took too long")
```
I think we need to kill the process inside the except.

DanielYang59:
Hi thanks for testing. I cannot really say for sure because I haven't really used `enumlib` (maybe it doesn't respond to SIGTERM or something?).

In this case is it possible to share the structure and code for me to recreate this?

fraricci:
Yeah, I think enum.x is not responding as other commands do.
Here is a structure which enumlib should take some time for computing all the configurations.
[POSCAR_mp-1218641.txt](https://github.com/user-attachments/files/18770183/POSCAR_mp-1218641.txt)

fraricci:
I see, calling the EnumlibAdapter on that structure does not work.
Here my code:
```
from pymatgen.analysis.magnetism.analyzer import MagneticStructureEnumerator

struct = Structure.from_file("./POSCAR_mp-1218641.txt")
enum_mag = MagneticStructureEnumerator(st,transformation_kwargs={"timeout":1})
```
Note: you need the other fix to get the timout arg passed correctly to the EnumlibAdapter.

Let me know how that goes.

DanielYang59:
Great! Really appreciate the code snnipet!

I just had a look but I don't think that code snnipet is enough to recreate that issue. In my case, the total runtime of the entire script (at around 150 seconds) is indeed longer than the `timeout` (note the `timeout` is in **minutes** so it's 60 seconds). However the run time of `_run_multienum` is around 0.01 seconds so we're far from hitting that timeout.

---

I also did a quick line profiling of the script:
- `__init__` `of MagneticStructureEnumerator` points out that `_generate_ordered_structures` is the bottlenect (100% of total run time): [profile_init_MagneticStructureEnumerator.log](https://github.com/user-attachments/files/18811817/profile_init_MagneticStructureEnumerator.log)
- `duplicate_checker.matches_ordering(check_structure)` (350 hits) takes up 88% of the run time on `_generate_ordered_structures`: [profile_generate_ordered_structures.log](https://github.com/user-attachments/files/18811820/profile_generate_ordered_structures.log)

I didn't go further than this as it's already unrelated to `timeout` on the `enumlib_caller` side, just mention this in case you want to optimize this part.

fraricci:
Thanks for pointing out that the bottleneck might be not the enumeration but the "duplicate_checker" step.
Here a custom structure that can be used directly by EnumlibAdaptor
[struct_for_testing_enumlib.json](https://github.com/user-attachments/files/18825409/struct_for_testing_enumlib.json)
and here the code I used for testing.
```
from pymatgen.command_line.enumlib_caller import EnumlibAdaptor
import time

struct = loadfn("./struct_for_testing_enumlib.json")

adaptor = EnumlibAdaptor(struct_for_enumlib,
                         max_cell_size=10, #increasing this while take longer and hit the timeout
                         timeout=0.1)

# Expect TimeoutError
adaptor.run()
```

This should take enough time and reveals that the current implementation is actually working, but still missing the process.terminate()/kill() in the exception (as mentioned above).
Another process that is taking some time is the step "getting_structures" in the enumlibAdaptor.run() which follows the enumeration, and that fooled me in the beginning in finding the culprit.

Let me know what you find and we can close this I think.

DanielYang59:
@fraricci Can I have your permission to include the `[struct_for_testing_enumlib.json](https://github.com/user-attachments/files/18825409/struct_for_testing_enumlib.json)` file into the test suite? Previous test doesn't seem to do what it should, we need to update it.

fraricci:
Good news, glad it helped out testing correctly this issue. 
I had already tested myself adding `process.terminate()` in the exception and it works fine.
Sure, go ahead and include that structure into the test suit.

If you can also change the error message as I mentioned above, that could be great.

Thanks!

DanielYang59:
> I had already tested myself adding process.terminate() in the exception and it works fine.

Great to know. You could directly checkout my branch (no need to manually apply the change)

> If you can also change the error message as I mentioned above, that could be great.

Happy to change, but I'm not seeing anything related to this?

fraricci:
It was just easier to add one line than go through updating branch ;-)

I mentioned this in a message above:

> Could you change the message into:
> f"Enumeration took more than the provided timeout {self.timeout} minutes"

DanielYang59:
Done! Super weird, I still couldn't find that message

fraricci:
It's a review message and I see that it is "pending". Maybe that's why you can't see it.

Anyway, thanks. It's good to go.

DanielYang59:
> It's a review message and I see that it is "pending". Maybe that's why you can't see it.

That makes sense, you have to commit the review message otherwise people cannot see it: https://github.com/orgs/community/discussions/10369

> Anyway, thanks

No problem at all. Can never be done without your help :)

shyuep:
Thanks all for the great bug fix!

DanielYang59:
No problem at all!

PR #4262, opened 2025-01-24
WIP: add script to generate lobster basis information

JaGeo:
## Summary

This is a script to read in POTCARs and generate the required basis set information files for all LOBSTER modules. 

As we might want to generate these files for newer versions of POTCARs, I wanted to make them available. 

(@Luccerboi for information)

PR #4278, opened 2025-02-01
[Breaking] Fix valence electron configuration parsing for `PotcarSingle.electron_configuration`

### LINKED ISSUE #4269: Possibly incorrect electron_configuration returned by vasp.inputs.PotcarSingle
### Python version

3.10.16

### Pymatgen version

2025.1.24

### Operating system version

_No response_

### Current behavior

It seems pymatgen returns incorrect valence configuration while parsing VASP POTCAR. Consider following code, where the POTCAR file corresponds to Hg (VASP-5.4, GGA, PBE, PAW, default):
```py
from pymatgen.io.vasp.inputs import PotcarSingle

with open("POTCAR", "r", encoding="utf-8") as file:
    data = file.read()

print(PotcarSingle(data).electron_configuration)
```

It prints:
```
[(5, 'd', 10), (4, 'f', 14)]
```

### Expected Behavior

Previously, I tested `pymatgen==2023.8.10`, it used to return:
```
[(6, 's', 2), (5, 'd', 10)]
```

The second line (valence) in POTCAR file is:
```
12.0000000000000 
```

Also, there is this line:
```
   POMASS =  XXX.XXX; ZVAL   =   12.000    mass and valenz
```

So, I suppose the previous behavior (i.e, `[(6, 's', 2), (5, 'd', 10)]`, 12 valence electrons) was correct.

Atomic configuration block:
```
   Atomic configuration
   15 entries
     n  l   j            E        occ.
     1  0  0.50         XXX      2.0000
...
     4  3  3.50         XXX     14.0000
     5  0  0.50         XXX      2.0000
     5  1  1.50         XXX      6.0000
     5  2  2.50         XXX     10.0000
     6  0  0.50         XXX      2.0000
     6  1  1.50         XXX      0.0000
```

### Minimal example

```Python
We can create a temporary directory and create virtualenv for testing:

mkdir test
cd test
python -m venv .venv
source .venv/bin/activate
pip install pymatgen

Then please use POTCAR of Hg and run the above python code.

To test older version of pymatgen, do the same instead install:

pip install pymatgen==2023.8.10 numpy==1.*
```

Sorry, but you need access to VASP POTCAR files, which cannot be shared due to copyright issues.

DanielYang59:
## Summary

- Fix valence electron configuration parsing in `PotcarSingle.electron_configuration`, to fix #4269
- Likely a follow up PR of #3944
- [x] Update unit tests
- [x] Test other versions of PSP?


DanielYang59:
I have a feeling that we might need to rewrite the `electron_configuration` property to parse the `Atomic configuration` section of POTCAR instead of generating from Madelung energy ordering rule, as electron configuration from POTCAR may or may not agree with periodic table.

For example for the `PAW_PBE Fe 06Sep2000` POTCAR, the valence electron configuration is "3d7 4s1" instead of "3d6 4s2": https://github.com/materialsproject/pymatgen/blob/5b997f78c9ae785e7a18a157c697d996f7dba746/tests/io/vasp/test_inputs.py#L1386

```
   Atomic configuration
    9 entries
     n  l   j            E        occ.
     1  0  0.50        xxx   2.0000
     2  0  0.50        xxx   2.0000
     2  1  1.50        xxx   6.0000
     3  0  0.50        xxx   2.0000
     3  1  1.50        xxx   6.0000
     3  2  2.50        xxx   7.0000
     4  0  0.50        xxx   1.0000
     4  1  1.50        xxx   0.0000
     4  3  2.50        xxx   0.0000
```

Can you please give me some advice on this? @rkingsbury Thanks in advance

JaGeo:
@DanielYang59 
I agree.
Maybe, take a look at https://github.com/materialsproject/pymatgen/pull/4262
It partially parses this info already

DanielYang59:
Hi @JaGeo good day and thanks for the comment

> Maybe, take a look at https://github.com/materialsproject/pymatgen/pull/4262 It partially parses this info already

Beautiful! Looks like the `dev_scripts.get_lobster_basis_from_potcars.get_valence` function could replace `electron_configuration` property? In this case do you want me to take over the valence electron property reader part of `PotcarSingle` (perhaps copy and refactor your `get_valence` code) and then you work on top of this? Or do you want to work on both?

> the code is poorly written but

I guess no code is perfectly written without multiple refactor? :)

JaGeo:
@DanielYang59 yes, it would be great if you take over and make this code a method for  the potcar. Then, I can adapt the developer code.

DanielYang59:
>  it would be great if you take over and make this code a method for the potcar

Great! I guess we could either:
- Simply change (fix) the implement of `electron_configuration` property (this PR would be breaking then)
- Or go the safer way, deprecate current `electron_configuration` with a corrected property (something like `val_elec_config` or something, naming could be discussed later of course)

Which do you suggest? I might personally prefer the former now that its already giving the wrong results so I would consider this a fix?

---

> I also defines a cutoff for an occupied orbital which might help for non-integer cases

I'm not really aware of non-integer occupancies from POTCARs, can you perhaps point me to some such POTCARs such that I could put more thinking on them?

JaGeo:
@DanielYang59 i also think it should be the former as it currently is implemented wrong in my opinion.

Li, Be, maybe? Some have 0.01 occupancy, I think. If you have some potcars available, you can use the dev script to spot them.

DanielYang59:
> i also think it should be the former as it currently is implemented wrong in my opinion.

Cool I would just replace for now but should be easy to switch anytime 

> Li, Be, maybe? Some have 0.01 occupancy, I think. If you have some potcars available, you can use the dev script to spot them.

Great I would have a look, I thought it's H.* (H.25, H.35, ...)

src/pymatgen/io/vasp/inputs.py
```diff
@@ -2127,24 +2127,52 @@ def __repr__(self) -> str:
...
+            n, ang_moment, _j, _E, occ = int(parts[0]), int(parts[1]), float(parts[2]), float(parts[3]), float(parts[4])
+
+            all_config.append((n, l_map[ang_moment], occ))
+
+        # Get valence electron configuration (defined by ZVAL)
+        valence_config: list[tuple[int, str, float]] = []
+        total_electrons = 0.0
+
+        for n, subshell, occ in reversed(all_config):
+            if occ >= 0.01:  # TODO: hard-coded occupancy cutoff
```
DanielYang59, line 2168: TODO tag: hard-coded cutoff

tests/io/vasp/test_inputs.py
```diff
@@ -1377,13 +1377,101 @@ def test_nelectrons(self):
...
+            PotcarSingle.from_file(f"{FAKE_POTCAR_DIR}/POT_GGA_PAW_PBE_54/POTCAR.H.25.gz").electron_configuration,
+            [
+                (1, "s", 0.25),
+            ],
+        )
+
+        # Test occupancy cut-off (Be: 2s1.99 2p0.01)
+        assert_config_equal(
+            PotcarSingle.from_file(f"{FAKE_POTCAR_DIR}/POT_GGA_PAW_PBE_54/POTCAR.Be.gz").get_electron_configuration(
+                occu_cutoff=0.1
```
DanielYang59, line 1468: I think we might need to rethink about this cutoff or note a warning in docstring.

For the Be electron configuration, it's:
```
     1  0  0.50        xxx   2.0000
     2  0  0.50        xxx   1.9900
     2  1  0.50        xxx   0.0100
     3  2  1.50        xxx   0.0000
```

If we choose cutoff as > 0.01, the 1s orbital would also be counted as valence (as 1.99 < ZVAL)

JaGeo: You could round to the next integer using this cutoff to get the energy config and if not possible raise an error

JaGeo: But happy to hear other opinions

src/pymatgen/core/periodic_table.py
```diff
@@ -452,33 +453,48 @@ def icsd_oxidation_states(self) -> tuple[int, ...]:
...
-        data = [parse_orbital(s) for s in e_str.split(".")]
-        if data[0][0] == "[":
-            sym = data[0].replace("[", "").replace("]", "")
+        # Split e_str (e.g. for Fe "[Ar].3d6.4s2" into ["[Ar]", "3d6", "4s2"])
+        data: list = [parse_orbital(s) for s in e_str.split(".")]
+
+        # Fully expand core-electron configuration (replace noble gas notation string)
+        if isinstance(data[0], str):
+            sym: str = data[0].replace("[", "").replace("]", "")
             data = list(Element(sym).full_electronic_structure) + data[1:]
```
rkingsbury, line 494: It looks to me like this is the only substantive change to `core.periodic_table.py`, which I support. Can you please make sure to mirror these changes in `Ion.electronic_structure` as well? Unfortunately the method doesn't inherit; it had to be re-implemented.

rkingsbury:
@DanielYang59 thanks for flagging. I'm not qualified to comment on the best way to connect this to `POTCAR` output; my changes in #3944 were guided only by trying to follow the canonical Madelung's rule. But in general it seems parsing information specific to the calculation is always better than assuming a standard configuration.

I made one comment above to please mirror changes here in `Species`, because it actually does not inherit from `Element`.

src/pymatgen/core/periodic_table.py
DanielYang59: Thanks for looking into this! Sure I would port changes to `Species.electronic_structure`. 

This change was made to make `mypy` happy otherwise it would report: `tuple` don't have `replace` method: https://github.com/materialsproject/pymatgen/blob/5b997f78c9ae785e7a18a157c697d996f7dba746/src/pymatgen/core/periodic_table.py#L477-L478

tests/io/vasp/test_inputs.py
DanielYang59: > You could round to the next integer using this cutoff to get the energy config and if not possible raise an error

If sounds pretty similar to [above "tolerance" method](https://github.com/materialsproject/pymatgen/pull/4278#discussion_r1938440804) if I understand correctly? I.e. when current number of electrons is "close enough" to ZVAL (closeness defined by tolerance), stop search further. 

In this case, if we apply `tolerance = 0.1`, 2p orbital would be ignored, first counted orbital would be 2s (1.99 electrons), and as it's close enough to ZVAL=2, 1s wouldn't be counted.

Am I understanding this correctly?

JaGeo: Yes, exactly. 

The Lobster basis function search is slightly different. As I just need the orbitals that might be relevant for a projection from the plane waves. Here, we look for the exact valence electrons (and only  integer numbers might be allowed).

@shyuep any opinions here as this would be a breaking change?

DanielYang59: Great thanks for confirming. I would implement the change and everyone please feel free to comment (certainly there would be edge cases that I'm not aware of), also ping @esoteric-ephemera maybe

shyuep:
Thanks all for the fix.

Issue #4247, opened 2025-01-07
port MPSurfaceSet from atomate(1) to pymatgen


rkingsbury:
### Feature Requested

The `atomate` repository contains an input set for surface calculations - [`MPSurfaceSet`](https://github.com/hackingmaterials/atomate/blob/2e541f297165254ac054b50e8132adaa282a01e0/atomate/vasp/workflows/base/adsorption.py#L370) - whose parameters differ slightly from `MVLSlabSet` and appears to have been used in the adsorption workflows. The `InputSet` was intended to be moved to pymatgen, but it has not yet happened.

### Proposed Solution

Add `MPSurfaceSet` to `pymatgen.io.vasp.sets.py`. Update the API if needed for compatibility with recent changes to `VaspInputSet`

rkingsbury:
@abhardwaj73 from my group is using this in his research and has agreed to tackle this

shyuep:
Please post this issue on the atomate Issues?

shyuep:
Ok thanks.,

PR #4287, opened 2025-02-11
Updated vasp input sets with MPSurfaceSet from atomate1

abhardwaj73:
## Summary

Major changes:

- feature 1: Addition of MPSurfaceSet mentioned [here](https://github.com/hackingmaterials/atomate/blob/2e541f297165254ac054b50e8132adaa282a01e0/atomate/vasp/workflows/base/adsorption.py#L370)

## Todos

- feature 2: Addition of tests for this new vasp input set

Mentioned in this issue #4247 @rkingsbury


shyuep:
I would rather not have two surface sets with only slgihtly different parameters and confusing names. The changes to the inputs are small enough that you can go ahead and modify MVLSlabSet directly and atomate should use MVLSlabSet.

rkingsbury:
@abhardwaj73 it looks like the adsorption workflows in `atomate2` use a modified `MPRelaxSet` ([link](https://github.com/materialsproject/atomate2/blob/4f4b607a10a90852d76cca3aa45e073c1bcfaf83/src/atomate2/vasp/jobs/adsorption.py#L147)). After this PR is done, let's compare the revised `MVLSlabSet` parameters with the modified set used there, and update the `atomate2` import if appropriate.

src/pymatgen/io/vasp/sets.py
```diff
@@ -2499,6 +2527,51 @@ def as_dict(self, verbosity: int = 2) -> dict[str, Any]:
...
+            "POTIM": 1.0,
+            "LDAU": ldau,
+            "EDIFF": 1e-5,
+            "ISYM": 0,
+        }
+        incar.update(incar_config)
+        incar.update(self.user_incar_settings)
+        return incar
+
+
```
rkingsbury, line 2574: Thanks @abhardwaj73 . Since we were asked only to update `MVLSlabSet` to match the `MPSurfaceSet` parameters, this class can now be removed.

src/pymatgen/io/vasp/sets.py
```diff
@@ -2435,14 +2435,42 @@ class MVLSlabSet(VaspInputSet):
     @property
     def incar_updates(self) -> dict[str, Any]:
         """Updates to the INCAR config for this calculation type."""
+
+        """Determine LDAU based on slab chemistry without adsorbates"""
+
```
rkingsbury, line 2440: For historical provenance purposes, please add some comment lines either here or at the top of the `MVLSlabSet` class explaining that the parameters were updated in March 2025 to match the `MPSurfaceSet` in atomate1.

rkingsbury (changes requested):
Looks good to me @abhardwaj73 , thanks. See a few small comments.

Also, please update the tests [here](https://github.com/materialsproject/pymatgen/blob/7cdd28852cb3672b458073ed6c179ebb713f672f/tests/io/vasp/test_sets.py#L1434) if needed so they pass with the new parameters.

PR #4288, opened 2025-02-11
`Dos.get_cbm_vbm` updates

kavanase:
Following discussion with @esoteric-ephemera after the updates to this function in https://github.com/materialsproject/pymatgen/pull/4240; the best approach for determining the VBM/CBM eigenvalues from a DOS object is to interpolate between the DOS points above/below the chosen threshold `tol`, and take the eigenvalues at the points where the (interpolated) DOS value equals `tol`.

This gives band gaps matching the expected values for the (current) `emmet-core` tests, and also brings this behaviour in line with the `dos.get_interpolated_gap` function (in practice, results in reduced partially-redundant code here).

Examples (used by @esoteric-ephemera in https://github.com/materialsproject/pymatgen/pull/4240 discussion):
Green lines show determined VBM/CBM positions
image

image


kavanase:
Just to note, the linting test failure is unrelated to this PR, seems to be something due to `Xdatcar` parsing? (Possibly due to a `mypy` version update)?

kavanase:
Tests are all passing btw

PR #4296, opened 2025-02-19
Make dict representation of `SymmetrizedStructure` MSONable

### LINKED ISSUE #3018: SymmetrizedStructure.to(json) raises TypeError
```py
from pymatgen.analysis.structure_analyzer import SpacegroupAnalyzer

structure = Structure(
    lattice=Lattice.cubic(3),
    species=("Fe", "Fe"),
    coords=((0, 0, 0), (0.5, 0.5, 0.5)),
)
symmetrized_structure = SpacegroupAnalyzer(structure).get_symmetrized_structure()
symmetrized_structure.to("dummy.json")
```

raises

```py
TypeError: Object of type SymmOp is not JSON serializable
```

DanielYang59:
### Summary

- Make dict representation of `SymmetrizedStructure` MSONable, to fix #3018
- [x] Unit test


src/pymatgen/symmetry/structure.py
```diff
@@ -148,3 +153,11 @@ def from_dict(cls, dct: dict) -> Self:
...
             wyckoff_letters=dct["wyckoff_letters"],
         )
+
+    def to(self, filename: PathLike = "", fmt: FileFormats = "", **kwargs) -> str:
+        """Use `MontyEncoder` as default JSON encoder."""
+        filename, fmt = str(filename), cast(FileFormats, fmt.lower())
+        if fmt == "json":
+            kwargs.setdefault("cls", MontyEncoder)
+
+        return super().to(filename, fmt, **kwargs)
```
DanielYang59, line 163: This cannot pass until #4295 is merged (otherwise `kwarg` wouldn't be passed to `json.dumps`)

src/pymatgen/symmetry/structure.py
```diff
@@ -44,6 +47,7 @@ def __init__(
         """
         self.spacegroup = spacegroup
         uniq, inverse = np.unique(equivalent_positions, return_inverse=True)
+        inverse = np.atleast_1d(inverse)  # Ensures `inverse` is 1D
```
DanielYang59, line 50: I believe this is needed, otherwise `numpy` would throw an exception when `inverse` is scalar:

```
        equivalent_indices: list[list[int]] = [[] for _ in range(len(uniq))]
        equivalent_sites: list[list[PeriodicSite]] = [[] for _ in range(len(uniq))]
        wyckoff_symbols: list[list[str]] = [[] for _ in range(len(uniq))]
>       for idx, inv_ in enumerate(inverse):
E       TypeError: iteration over a 0-d array
```

mkhorton:
Thanks @DanielYang59, and thanks for flagging it as breaking. Could you clarify what is actually breaking here? As far as I can see it's just the extra keys in the dict representation?

If breaking, could you also make sure the PR includes an update to the [compatibility page](https://github.com/materialsproject/pymatgen/blob/master/pyproject.toml) to make sure people are notified. Thanks!

DanielYang59:
Hi @mkhorton thanks for reviewing!

> Could you clarify what is actually breaking here? As far as I can see it's just the extra keys in the dict representation?

Yes AFAIK the breaking change would be the change of dict representation (which could also be considered a fix, but I assume it's better safe than sorry).

PR #4299, opened 2025-02-21
Update inequality in `get_linear_interpolated_value`

kavanase:
This is a small but important fix for the `get_linear_interpolated_value` function in `pymatgen.util.coord`, used in various places within the codebase. 

This function works by finding the first index in `x_values` that is greater than / equal to `x`, and then uses this index and the previous (`idx-1`) with `x_values` and `y_values` to interpolate:
https://github.com/materialsproject/pymatgen/blob/v2025.2.18/src/pymatgen/util/coord.py#L136-L147

The issue is that if the _first_ index in `x_values` is equal to `x`, this causes the ValueError to be thrown (as `indices[0]==0`); where it thinks that the input `x` to interpolate is outside the interpolable range. Changing the inequality to `>` rather than `>=` resolves this issue.

MWE:
```
from pymatgen.util.coord import get_linear_interpolated_value

get_linear_interpolated_value([0.0001, 0.0002], [1,2], 0.0001)
```
Current `pymatgen` `master`:
image
Where we can see "`x=0.0001 is out of range of provided x_values (0.0001, 0.0002)`" does not make sense.

This branch:
image

And confirming the error is still correctly thrown with an `x` value that is actually outside the range:
image

### DIFF
--- a/src/pymatgen/analysis/local_env.py
+++ b/src/pymatgen/analysis/local_env.py
@@ -3745,23 +3745,23 @@ class EconNN(NearNeighbors):
 
         if self.use_fictive_radius:
             # calculate fictive ionic radii
-            firs = [_get_fictive_ionic_radius(site, neighbor) for neighbor in neighbors]
+            fict_ionic_radii = [_get_fictive_ionic_radius(site, neighbor) for neighbor in neighbors]
         else:
             # just use the bond distance
-            firs = [neighbor.nn_distance for neighbor in neighbors]
+            fict_ionic_radii = [neighbor.nn_distance for neighbor in neighbors]
 
         # calculate mean fictive ionic radius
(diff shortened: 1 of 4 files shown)


kavanase:
Also there was a pre-commit `codespell` failure due to a variable name in `local_env` which was fixed in https://github.com/materialsproject/pymatgen/pull/4299/commits/8fdea1637b7c6ae933102d769ded02d1ca23163e

kavanase:
This also adds some very small updates to make the behaviour of `DOS.get_interpolated_gap` and `Dos.get_interpolated_gap` fully consistent

mkhorton (approved):
Thanks for catching and fixing this @kavanase! Appreciate the code readability improvements too.

PR #4303, opened 2025-02-25
Fix mcl kpoints

dgaines2:
Fixed errors in two of the k-points for the MCL reciprocal lattice (according to Table 16 in Setyawan-Curtarolo 2010)

M2 and D1 aren't included in the recommended k-point path, but third-party software that plots k-point paths using pymatgen labelled M2 in the path instead of M1 due to it being the "same" k-point.


mkhorton:
Thanks for catching this @dgaines2! A fairly subtle bug, I hope there are not more present. Perhaps a good excuse to go back over these to be sure.  Out of interest, what was the third-party software you noticed the bug in?

Including a screenshot from [the paper](https://arxiv.org/pdf/1004.2974) for documentary purposes:

image

mkhorton (approved):

dgaines2:
Thanks for reviewing this! The third-party software is [sumo](https://smtg-bham.github.io/sumo/)

shyuep:
Thanks.

PR #4315, opened 2025-03-06
patch to allow pyzeo to work

daniel-sintef:
Hi everyone,

Pyzeo is a more maintained version of the cython but has different imports. This small patch allows users to swap out zeo++ with pyzeo.


shyuep:
Great thanks. Can you add / enable some unittests to check that this is working? Thanks.

mkhorton:
You would also need to add it as an [optional dependency](https://github.com/materialsproject/pymatgen/blob/master/pyproject.toml), thanks for the edit!

daniel-sintef:
Note that even with the patch one would still need to manually install zeo++, i've added some logic to the test.yaml in the workflow for this, but can't really test it on my end. The tests pass if zeo++ is installed. Historically testing the zeo++ plugin was just skipped for this reason, but maybe it can be automated and included?

daniel-sintef:
@shyuep @mkhorton a gentle reminder to take a look whenever you have time. -Daniel

shyuep:
Thanks!

daniel-sintef:
Thanks everyone!

PR #4321, opened 2025-03-11
Fix `from_ase_atoms` for `Molecule`

### LINKED ISSUE #4320: `Molecule.from_ase_atoms()` does not pass kwargs appropriately
### Python version

3.12

### Pymatgen version

2025.3.10

### Operating system version

_No response_

### Current behavior

The `Molecule.from_ase_atoms` method does not accept dedicated keyword arguments for the `Molecule` class. I demonstrate this below:

```python
from ase.build import molecule
from pymatgen.core import Molecule

atoms = molecule("CH3")
Molecule.from_ase_atoms(atoms, charge_spin_check=False)
```

The above returns:
```
TypeError: Structure.__init__() got an unexpected keyword argument 'charge_spin_check'
```

However, the following works as expected:

```python
from ase.build import molecule

atoms = molecule("CH3")
AseAtomsAdaptor.get_molecule(atoms, charge_spin_check=False)
```

This is not a hard patch. I personally can't fix it at the moment though. PRs welcome!

DanielYang59:
### Summary

- Fix `from_ase_atoms` for `Molecule`, to close #4320
- [ ]  Add tests


src/pymatgen/core/structure.py
```diff
@@ -995,19 +995,6 @@ def to_ase_atoms(self, **kwargs) -> Atoms:
 
         return AseAtomsAdaptor.get_atoms(self, **kwargs)
 
-    def from_ase_atoms(self, **kwargs) -> Structure:
```
DanielYang59, line 998: TODO: need to make sure this is needed by any other concrete subclass

DanielYang59, line 998: **Need careful check on this**: AFAIK there is no other concrete class other than `IMolecule/IStructure` of the abstract base class`SiteCollection` that would reply on inheriting `from_ase_atoms`?

Andrew-S-Rosen: I don't believe there are. Even if there were, then presumably we would currently be missing out on such instances anyway since the pre-existing methods were on `Structure` and `Molecule`, right?

Andrew-S-Rosen: Is there a problem here in that it will produce an `IStructure` rather than a `Structure`?

tests/core/test_structure.py
```diff
@@ -2429,6 +2445,14 @@ def test_to_file_alias(self):
         assert self.mol.to(out_path) == self.mol.to_file(out_path)
         assert os.path.isfile(out_path)
 
+    def test_to_from_ase_atoms(self):
+        pytest.importorskip("ase")
+
+        atoms = self.mol.to_ase_atoms()
+        assert isinstance(atoms, Atoms)
+
+        assert type(IMolecule.from_ase_atoms(atoms)) is Molecule
```
DanielYang59, line 2454: Not sure if we want to change this behaviour or not(I found current behaviour a bit confusing), currently `from_ase_atoms` would always give the mutable version (`Molecule/Structure`)

src/pymatgen/core/structure.py
```diff
@@ -4133,6 +4133,19 @@ def from_file(cls, filename: PathLike) -> Self | None:
...
+
+        Args:
+            kwargs: Passed to AseAtomsAdaptor.get_molecule.
+
+        Returns:
+            Molecule
+        """
+        from pymatgen.io.ase import AseAtomsAdaptor
+
+        return AseAtomsAdaptor.get_molecule(self, **kwargs)
```
DanielYang59, line 4147: A very subtle behaviour change (could potentially be breaking), currently `SiteCollection.from_ase_atoms` calls `get_structure`:
https://github.com/materialsproject/pymatgen/blob/55d0934792c707447b75fe971b4b3bb299cf0e27/src/pymatgen/core/structure.py#L998-L1009

 This changes the behaviour for IMolecule to call `get_molecule` instead (looks like a wrapper on top of `get_structure` with additional charge and spin multiplicity):

https://github.com/materialsproject/pymatgen/blob/55d0934792c707447b75fe971b4b3bb299cf0e27/src/pymatgen/io/ase.py#L379-L406

tests/core/test_structure.py
Andrew-S-Rosen: I agree that this is currently not the expected behavior.

DanielYang59: Thanks for confirming this, I have refactored to return the corresponding class instead. There is a another issue though and I would add a new comment

src/pymatgen/core/structure.py
```diff
@@ -995,18 +995,20 @@ def to_ase_atoms(self, **kwargs) -> Atoms:
 
         return AseAtomsAdaptor.get_atoms(self, **kwargs)
 
-    def from_ase_atoms(self, **kwargs) -> Structure:
-        """Convert ase.Atoms to pymatgen Structure.
+    @classmethod
+    def from_ase_atoms(cls, atoms: Atoms, **kwargs) -> Self:
+        """Convert ase.Atoms to pymatgen (IStructure)/(I)Molecule.
```
Andrew-S-Rosen, line 1000: Minor comment: (IStructure) --> (I)Structure

Andrew-S-Rosen:
This looks good to me. I will give this through the end of the week for others to chime in, and if there's nothing I will do a final review and merge over the weekend.

src/pymatgen/core/structure.py
DanielYang59: Great catch, thanks!

tests/core/test_structure.py
```diff
@@ -2687,9 +2714,17 @@ def test_relax_gfnxtb(self):
...
         assert isinstance(atoms, Atoms)
         assert len(atoms) == len(self.mol)
+
         assert AseAtomsAdaptor.get_molecule(atoms) == self.mol
+
+        assert type(Molecule.from_ase_atoms(atoms)) is Molecule
+
+        # Make sure kwargs are correctly passed
+        atoms = molecule("CH3")
+        Molecule.from_ase_atoms(atoms, charge_spin_check=False)
```
Andrew-S-Rosen, line 2730: To make sure that this continues to work as expected, I would suggest adding one more test:

```python
atoms = molecule("CH3")
with pytest.raises(ValueError): # or whatever error it is
        Molecule.from_ase_atoms(atoms)
```

This will ensure that your test is actually doing what we intend it to do even if there are changes elsewhere. Happy to merge after this.

DanielYang59: I don't really understand but it seems the following would not throw an error? 
```python
from ase.build import molecule
from pymatgen.core import Molecule

Molecule.from_ase_atoms(molecule("CH3"))
```

Andrew-S-Rosen: Can you try the following instead?

```python
atoms = molecule("CH3")
with pytest.raises(ValueError, match="not possible"):
        Molecule.from_ase_atoms(atoms, spin_multiplicity=1)
```

DanielYang59: > Can you try the following instead?

Sure thing.

---

> That way nobody deletes the test thinking it's useless.

I added a comment saying this is to test `kwargs` passing (also there's another test to check the return type) so it's unlikely people would think this is useless, but sure happy to add as there isn't much additional cost.

docs/compatibility.md
```diff
@@ -6,72 +6,35 @@ nav_order: 3
 
 # Compatibility
 
-Pymatgen is a tool used for academic research and is actively developed by
-a large community of people. As such, releases are frequent, and new features
-and capabilities are constantly being added.
+Pymatgen is a tool used for academic research and is actively developed by a large community of people. As such, releases are frequent, and new features and capabilities are constantly being added.
 
-However, `pymatgen` is also used as a library by other tools, and as such breaking
```
DanielYang59, line 13: I assume hard wrap in markdown is not necessary, removed them here and let text editors handle line wraps automatically. Correct me if I'm wrong.

Andrew-S-Rosen:
Thanks, @DanielYang59! Merging.

Andrew-S-Rosen:
Ah, looks like I don't have permission to merge until the CI passes...

DanielYang59:
> Ah, looks like I don't have permission to merge until the CI passes...

No rush at all! I'm working on fixing these new linting errors in #4327

shyuep:
Merged

DanielYang59:
Thanks!

PR #4323, opened 2025-03-13
Tweak POSCAR / XDATCAR to accommodate malformed files

esoteric-ephemera:
Related to [this matsci.org issue](https://matsci.org/t/xdatcar-error-when-reading-xdatcar-could-not-convert-string-to-float/61831): sometimes the XDATCAR can be malformed because fortran uses fixed format floats when printing. In those cases, there's no space between coordinates:
```
Direct configuration= 2
-0.63265286-0.11227753 -0.15402785
-0.12414874 -0.01213420 -0.28106824
...
```
In some cases, this issue is reparable (a negative sign separates coordinates). This PR implements a fix when it is reparable and adds a few Trajectory-like convenience features to `Xdatcar`


shyuep:
Thanks.

PR #4324, opened 2025-03-14
GibbsComputedStructureEntry update to handle float temperature values

slee-lab:
## Summary

Issue: when temperature %100==0 but is type float, you get KeyError because G_ELEMS has keys of integer values.

Major changes:

- feature 1: updated to handle float temperature values 
- fix 1: G_ELEMS[str(self.temp)] is changed to G_ELEMS[str(int(self.temp))]

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
Thanks,.

PR #4327, opened 2025-03-15
Fix new `ruff` error in linting pipeline

DanielYang59:
## Summary

- Fix new `ruff` error in linting pipeline
- Avoid configuring the root logger


src/pymatgen/cli/get_environment.py
```diff
@@ -35,10 +34,6 @@ def main() -> int:
     print("Chemical Environment package (ChemEnv)")
     print(chemenv_config.package_options_description())
 
-    logging.basicConfig(
```
DanielYang59, line 38: I believe pymatgen as a lib should not config the root logger?

tests/io/vasp/test_sets.py
```diff
@@ -2138,11 +2138,11 @@ def test_incar(self):
 
     def test_kpoints(self):
         kpoints1 = self.lobsterset1.kpoints
-        assert kpoints1.comment.split()[6], 6138
+        assert kpoints1.comment.split()[5] == "6138"
```
DanielYang59, line 2141: @JaGeo perhaps you could have a look at this?

Looks like this test was not correctly working from the beginning 573642dffcacf156de38ed7aafa82182d079af2a, using pattern like:
```python
self.assertTrue(kpoints1.comment.split(" ")[6], 6138)
```

But for [`assertTrue(expr, msg=None)`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue) it doesn't compare the first arg with the second, it checks whether the first expression is true and the second arg is the message if not

pyproject.toml
```diff
@@ -220,6 +220,7 @@ ignore = [
     "S603",    # Check source for use of "subprocess" call
     "S607",    # Start process with relative path
     "SIM105",  # Use contextlib.suppress() instead of try-except-pass
+    "SIM905",  # Split static strings
```
DanielYang59, line 223: I would ignore [SIM905](https://docs.astral.sh/ruff/rules/split-static-string/) for now as in our case it could avoid unnecessary multi-line lists 

Would go back to this if performance overhead has been a concern at some point

tests/ext/test_matproj.py
```diff
@@ -530,6 +530,10 @@ def test_api_key_is_none(self):
 )
 class TestMPResterNewBasic(PymatgenTest):
```
DanielYang59, line 531: TODO: not sure if this is solvable

tests/core/test_surface.py
```diff
@@ -630,8 +630,8 @@ def test_bonds_broken(self):
         # expect 2 and 6 bonds broken so we check for this.
         # Number of broken bonds are floats due to primitive
         # flag check and subsequent transformation of slabs.
-        assert slabs[0].energy, 2.0
-        assert slabs[1].energy, 6.0
+        assert slabs[0].energy == approx(8.0)
+        assert slabs[1].energy == approx(24.0)
```
DanielYang59, line 634: @fyalcin it has been a while and I'm not sure if you could still remember this...But this test was not working from the start for the following reason as in https://github.com/materialsproject/pymatgen/pull/4327/commits/b8d2bc997252b32d6f2ec617310465964a1a0a14#r1996705049

I have to change the value here for now

tests/ext/test_matproj.py
DanielYang59: Looks like some legacy MPRester test, would remove them in a separate PR

PR #4330, opened 2025-03-17
Update type-hinting for `pymatgen.io.qchem.outputs`

Andrew-S-Rosen:
## Summary

I updated several I/O parsers in `pymatgen.io.qchem.outputs` such that they are properly type-hinted with `str | Path` arguments rather than just `str`.

### DIFF
--- a/src/pymatgen/io/qchem/outputs.py
+++ b/src/pymatgen/io/qchem/outputs.py
@@ -34,7 +34,7 @@ except ImportError:
 
 if TYPE_CHECKING:
     from typing import Any
-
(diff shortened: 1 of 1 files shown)


DanielYang59:
I guess we have a type alias `PathLike` which might be preferred to annotate such "pathlike" things: https://github.com/materialsproject/pymatgen/blob/168999dc7acac4f4dda6296b5cef1ca512ed33a6/src/pymatgen/util/typing.py#L28

```python
from pymatgen.util.typing import PathLike
```

PR #4334, opened 2025-03-25
Updated Potentials Class in FEFF io to consider radius

CharlesCardot:
Changed the Potentials class to consider the same radius that is used in
the Atoms class. This is necessary to avoid a situation where the radius is small enough to only have a subset of the unique elements in a structure, but all the elements have potentials defined, which causes FEFF to fail when run.

## Summary

Major changes:

- fix 1: **Previous behavior**: When creating a FEFFDictset using the feff.io tools, the potentials class defined a potential for every unique element in a structure, while the atoms class defined an atom coordinate for every atom in a radius around the absorbing atom. If the radius was defined to be small, only a subset of the unique atoms in the structure would be included in the atom class. **New behavior**: Updated the potentials class to only create potentials for atoms inside the radius specified when creating a FEFFDictset. Without this, a too small radius for a structure with unique elements outside of that radius would cause FEFF to fail, given that there was a Potential defined for an element that did not exist in the Atoms flag.

## Todos

None, the work is complete

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
Thanks very much. Can you add some tests to test_feff.py for the updated functionality?

CharlesCardot:
I've added a test, and I've changed the behavior of the Potentials class to be able to handle situations where the radius is not defined (in which case it will revert to the old behavior of defining a potential for every atom in the structure).

shyuep:
Thanks!

shyuep:
Thanks. Merged

Issue #4336, opened 2025-03-28
[Proposal] Better periodic table dev recording (`periodic_table.yaml`)- One property per file?


DanielYang59:
We might want to improve our development `periodic_table.yaml` recording:
- It's simply too large and monolithic as a single file, and as such maintaining it (and track git history) is quite hard.
- It doesn't carry enough metadata, for example we don't know the "source/reference" of each data entry, also some information/metadata is missing, for example the temperature at which "Electrical resistivity" is measured.

As such I want to propose that we split the recording to one property per file (perhaps core and static properties in one file), for the following benefits:
- Easier to maintain, we could add/modify certain property each time relatively easily, and also the git history would be much easier to read
- It's more similar to how certain data is presented in handbooks, for example the [CRC Handbook of Chemistry and Physics](https://en.wikipedia.org/wiki/CRC_Handbook_of_Chemistry_and_Physics 95th) have "Electrical Resistivity of Pure Metals" table which list the data for each element but at difference temperatures (and therefore we need more detailed metadata)

---

This is inspired by the following issues:
- https://github.com/materialsproject/pymatgen/issues/4312
- https://github.com/materialsproject/pymatgen/issues/4313

shyuep:
Maintenance is not done on the json, but on the yaml (which is a lot easier to read and modify). That file is in dev scripts. 
I am not opposed to splitting up by element but I suspect 100 file open calls will be more expensive than loading one big file. If you can prove to me that loading multiple element data files is just as fast as loading one big file, I am willing to consider it.

The way the periodic_table.json should be thought of is like the "min" versions of javascript for webpages, as opposed to the non-min version that is used in development. The reason why you use the min version, which is difficult to read and edit, is that it loads faster.

Element is something that must be loaded by default in pymatgen. I don't think there is a situation where elemental data is not loaded.

shyuep:
![Image](https://github.com/user-attachments/assets/dcb0ec51-7f80-4bc9-9e0f-08ad2b460da7)

This is what the YAML looks like in the PyCharm IDE. Navigation to a different element is trivial based on the file structure and updating a value is super easy. I added a comment in the periodic_table.py to make sure people understand how the data is updated.

DanielYang59:
> I am not opposed to splitting up by element but I suspect 100 file open calls will be more expensive than loading one big file. If you can prove to me that loading multiple element data files is just as fast as loading one big file, I am willing to consider it.

[...]

Issue #4313, opened 2025-03-04
More sources/references in periodic table


CakeIsYummy21:
### Feature Requested

More sources in the periodic table code or perhaps a site with all the sources, so you can put this python library and its sources as valid sources in your paper/essay. An idea would be like that: https://reference.wolfram.com/language/note/ElementDataSourceInformation.html

### Proposed Solution

More sources

Thank you : )

PR #4338, opened 2025-03-28
[DEV] `ptable_table.json` generate with unified interface

DanielYang59:
## Summary
- Clean up development resources for `ptable_table.json` generation, #4336 
- Allow sources/reference, #4313
- Remove seemingly unused `bvparm2011` (1d9ab4d276aaf8e46fca3c2925928f6fd12f5537) and `periodic_table.db`(benchmark from c80afb5d483ca3cc43ce0c155e7184eb8eb9e7b6)

### JSON/YAML generation directly from source with unified parser interface

```
Source recordings (different parsers for csv/yaml/...)
↓
Each generate one or more Property (property (hold unit/reference) -> element -> data)
↓
YAML for development and JSON for production
```

- Assume all values for the same property have the same unit 
- Assume single reference for single property for now

---

### Follow up PRs
- [ ] JSON generator directly update `core.periodic_table.json`
- [ ] Have unit in a separate field, and need to update unit handle mechanism in `core.Element`
- [ ] Lazily load optional/extra properties
- [ ] Some property are missing in current JSON (present in YAML): "Ground level", "NMR Quadrupole Moment"
- [ ] "Electronic structure" in [current JSON](https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/core/periodic_table.json) and [YAML](https://github.com/materialsproject/pymatgen/blob/master/dev_scripts/periodic_table.yaml) seems quite different (new YAML for this PR uses the JSON to be consistent with production code)
- [ ] Resolve known differences in `_compare_json.py`

---

### Compare new JSON with current 

Save a copy of current (master branch) JSON/YAML for quick lookup (uploading YAML not allowed): [master_json_yaml.zip](https://github.com/user-attachments/files/19547439/master_json_yaml.zip)

`pymatgen/dev_scripts/periodic_table_resources/_compare_json.py`


shyuep:
I want to point out that ChatGPT can make this extremely easy. I passed ChatGPT the yaml and prompted it to "separate out the units of each property for each element from the value, keeping the dictionary schema consistent". Below is the result. 

```
Below is a Python script that reads your existing YAML file, scans every property for values that appear to include a numeric (or “no data”) part followed by units, and then rewrites those entries as dictionaries with separate “value” and “units” keys. The overall schema (i.e. the element keys and the structure for sub‐dictionaries, lists, etc.) remains unchanged.

You can add this file (for example, name it transform_periodic_table.py) to your project and run it. It will load periodic_table.yaml and produce a new file called periodic_table_transformed.yaml with the modified structure.

Below is the complete code:

import yaml
import re

def separate_units(val):
    """
    If val is a string of the form "<number> <unit>" or "no data <unit>",
    separate out the numeric part (or "no data") and the unit.
    Otherwise, return the original value.
    """
    if isinstance(val, str):
        # Match a number (integer or float) or the string "no data", followed by optional whitespace and the unit.
        m = re.match(r"^(no data|\d+(?:\.\d+)?)(?:\s+(.*))?$", val)
        if m:
            num_part = m.group(1)
            unit_part = m.group(2) or ""
            # Convert numeric part to float if possible and if not "no data"
            if num_part != "no data":
                try:
                    num_val = float(num_part)
                except Exception:
                    num_val = num_part
            else:
                num_val = num_part
            return {"value": num_val, "units": unit_part.strip()}
    return val

def transform(data):
    """
    Recursively process the YAML data.
    If a value is a string, attempt to separate out a measurement and its unit.
    If the value is a dict or list, recursively process it.
    """
    if isinstance(data, dict):
        new_data = {}
        for k, v in data.items():
            if isinstance(v, str):
                new_data[k] = separate_units(v)
            elif isinstance(v, (dict, list)):
                new_data[k] = transform(v)
            else:
                new_data[k] = v
        return new_data
    elif isinstance(data, list):
        return [transform(item) for item in data]
    else:
        return data

# Load the original periodic table YAML file
with open("periodic_table.yaml", "r", encoding="utf-8") as f:
    data = yaml.safe_load(f)

# Transform the data to separate units from the values
transformed_data = transform(data)

# Write out the transformed data to a new YAML file
with open("periodic_table_transformed.yaml", "w", encoding="utf-8") as f:
    yaml.dump(transformed_data, f, sort_keys=False)

print("Transformation complete. New file saved as 'periodic_table_transformed.yaml'.")

How It Works
	•	separate_units function:
This function uses a regular expression to detect a pattern where the value starts with either a number (integer or float) or the phrase "no data", followed by optional whitespace and then the unit. If the pattern is found, it returns a dictionary with the keys "value" and "units". If not, it simply returns the original value.
	•	transform function:
This function recursively walks through the entire YAML data structure. For each string value it finds, it applies the separate_units function. If the value is already a dictionary or a list, it recurses into that structure.
	•	I/O:
The script reads your original periodic_table.yaml and writes the transformed version to periodic_table_transformed.yaml.

Once you run this script, your properties (for example, "Boiling point: 3573 K") will be converted into a nested structure like:

Boiling point:
  value: 3573.0
  units: K

This keeps the overall dictionary schema (each element remains a top-level key with its properties) consistent.

You can now run this script in your PyCharm project, and the transformed file will be available for further use.
```

DanielYang59:
Beautiful, thanks a ton for chiming in! That would be really helpful. I would work on this likely next week :)

dev_scripts/periodic_table_resources/oxidation_states.yaml
```diff
@@ -0,0 +1,101 @@
+Oxidation states:
+  unit: null
+  reference: https://en.wikipedia.org/wiki/Template:List_of_oxidation_states_of_the_elements
```
DanielYang59, line 3: I would assume all values have the same unit for now, we could convert to the same unit **before** saving the data recording anyway (avoid having a separate unit for each element makes things more readable).

For the `reference` I could image cases where we have to combine data from different sources (ideally we might want single source for a single property for consistency though), for example the "hardness" for metals/non-metal might come from difference sources

dev_scripts/periodic_table_resources/ionic_radii.csv
```diff
@@ -0,0 +1,98 @@
...
+Cr,hs,,,,,94,,,,,,
+Mn,ls,,,,,81,72,67,47,39.5,60,
+Mn,hs,,,,,97,78.5,,,,,
+Fe,ls,,,,,75,69,72.5,,39,,
+Fe,hs,,,,,92,78.5,,,,,
+Co,ls,,,,,79,68.5,,,,,
+Co,hs,,,,,88.5,75,67,,,,
+Ni,ls,,,,,83,70,62,,,,
+Ni,hs,,,,,,74,,,,,
+Cu,,,,,91,87,68,,,,,
```
DanielYang59, line 31: Need to double check Cu ionic radii for +3 (68 pm), the original CSV gives data as `68 ls`, but current JSON parser treat it as base radii:
https://github.com/materialsproject/pymatgen/blob/95fb605417bca6beae1362dd7ffb725bee3a8190/src/pymatgen/core/periodic_table.json#L3688-L3692

image

dev_scripts/generate_periodic_table_json.py
```diff
@@ -0,0 +1,351 @@
...
+        if "Electron affinity (kJ/mol)" in table.columns and table["Name"].astype(str).str.contains("Hydrogen").any()
+    )
+    ea_df = ea_df.drop(columns=["References", "Electron affinity (kJ/mol)", "Element"])
+
+    # Drop superheavy elements
+    max_z: int = max(Element(element).Z for element in Element.__members__)
+    ea_df = ea_df[pd.to_numeric(ea_df["Z"], errors="coerce") <= max_z]
+
+    # # Drop heavy isotopes (by detecting duplicate Z)
+    # ea_df = ea_df.drop_duplicates(subset="Z", keep="first")
```
DanielYang59, line 265: Not sure if we want to include electron affinities for heavy isotopes or not, our current parser seems to drop them: https://github.com/materialsproject/pymatgen/blob/95fb605417bca6beae1362dd7ffb725bee3a8190/dev_scripts/update_pt_data.py#L263-L265

But its manually added for D in f463ac147b74731dd6e804f9ec902f38627d278f

Guess we just keep them

DanielYang59:
Quite weird, looks like the following elements are present in current (master branch) JSON, but not YAML (thought this JSON was generated from the YAML?):

```
{'Mc', 'Bh', 'Nh', 'Fl', 'Lv', 'Sg', 'Mt', 'Db', 'Ts', 'Og', 'Ds', 'Cn', 'Rf', 'Rg', 'Hs'}
```

shyuep:
@DanielYang59 This looks like a lot of effort. Honestly, I think these features are barely going to be used. I suggest you simplify. Have a small json that contains only Z, atomic mass, electronegativity and ionic radii (which is always needed) and another that contains all other information and loaded with lru_cache. Most of the time, we only need to load the former.

DanielYang59:
> This looks like a lot of effort. Honestly, I think these features are barely going to be used. I suggest you simplify. Have a small json that contains only Z, atomic mass, electronegativity and ionic radii (which is always needed) and another that contains all other information and loaded with lru_cache. 

Sure thing happy to do that ;) I think we're getting very close, just need a final clean up of the "elemental data" YAML source file.

As i noticed our current JSON is quite different from our current YAML:
- YAML contains these element but JSON doesn't: `{'Mc', 'Bh', 'Nh', 'Fl', 'Lv', 'Sg', 'Mt', 'Db', 'Ts', 'Og', 'Ds', 'Cn', 'Rf', 'Rg', 'Hs'}`
- JSON contains these properties but YAML doesn't: "Ground level", "NMR Quadrupole Moment"


After this clean up, they should be consistent

dev_scripts/periodic_table_resources/elemental_properties.yaml
```diff
@@ -720,105 +1549,262 @@ Coefficient of linear thermal expansion:
     Zr: 5.7
 Common oxidation states:
```
DanielYang59, line 1550: reformat

tests/core/test_periodic_table.py
```diff
@@ -367,7 +367,7 @@ def test_radii(self):
         el = Element.Pd
         assert el.atomic_radius == approx(1.40)
         assert el.atomic_radius_calculated == approx(1.69)
-        assert el.van_der_waals_radius == approx(2.10)
```
DanielYang59, line 370: I haven't looked into the reason, but our current Van der waals radius seems incorrect:

https://github.com/materialsproject/pymatgen/blob/4c7892f5c9dcc51a1389b3ad2ada77632989a13e/dev_scripts/radii.csv?plain=1#L1
https://github.com/materialsproject/pymatgen/blob/4c7892f5c9dcc51a1389b3ad2ada77632989a13e/dev_scripts/radii.csv?plain=1#L47

dev_scripts/periodic_table_resources/_periodic_table.yaml
```diff
@@ -0,0 +1,4231 @@
...
+    As: 887, Se: 958, Br: 332, Kr: 119.93, Rb: 961, Sr: 1655, Y: 3609, Zr: 4682, Nb: 5017,
+    Mo: 4912, Tc: 4538, Ru: 4423, Rh: 3968, Pd: 3236, Ag: 2435, Cd: 1040, In: 2345,
+    Sn: 2875, Sb: 1860, Te: 1261, I: 457.4, Xe: 165.1, Cs: 944, Ba: 2143, La: 3743,
+    Ce: 3633, Pr: 3563, Nd: 3373, Pm: 3273, Sm: 2076, Eu: 1800, Gd: 3523, Tb: 3503,
+    Dy: 2840, Ho: 2993, Er: 3141, Tm: 2223, Yb: 1469, Lu: 3675, Hf: 4876, Ta: 5731,
+    W: 5828, Re: 5869, Os: 5285, Ir: 4701, Pt: 4098, Au: 3129, Hg: 629.88, Tl: 1746,
+    Pb: 2022, Bi: 1837, Po: 1235, Rn: 211.3, Ra: 2010, Ac: 3573, Th: 5093, U: 4200,
+    Np: 4273, Pu: 3503, Am: 2880, Cm: 3383}
+Brinell hardness:
+  unit: sup^2 MN m
```
DanielYang59, line 1099: Avoid format string as md

src/pymatgen/core/periodic_table.py
```diff
@@ -226,30 +231,33 @@ def __getattr__(self, item: str) -> Any:
                     tokens = no_bracket.replace("about", "").strip().split(" ", 1)
                     if len(tokens) == 2:
                         try:
-                            if "10<sup>" in tokens[1]:
```
DanielYang59, line 229: Double check: make sure this only appears for `{"electrical_resistivity", "coefficient_of_linear_thermal_expansion"}`

src/pymatgen/core/periodic_table.py
```diff
@@ -211,45 +213,53 @@ def __getattr__(self, item: str) -> Any:
...
             if val is None or str(val).startswith("no data"):
                 warnings.warn(f"No data available for {item} for {self.symbol}", stacklevel=2)
                 val = None
+
             elif isinstance(val, list | dict):
                 pass
+
             else:
                 try:
                     val = float(val)
```
DanielYang59, line 228: Looks like the value unit handling mechanism has to be refactored in this PR (though I wanted to keep this PR atomic but looks like I don't have a choice), currently unit is separated from the value, so `float(val)` would not throw an exception.

But I'm hoping after this refactor this part would be easier to read :)

DanielYang59: I think this PR is getting overly large (plus dramatically changing `core`), so I would just change JSON generator to write val + unit as a single string just like before, to focus this PR on JSON generation, and separate val-unit separation to another PR.

DanielYang59:
@shyuep  this PR is ready from my side, as it might be changing the core functionalities of pymatgen, I would like to push forward as slowly and carefully as possible, so I would separate the workflow into multiple steps:
- This PR would only generate a duplicate JSON saved in `dev_scripts` without really changing our production recording
- Any change/fix in production JSON recording would be submitted as a separate PR for visibility 
- More detailed TODOs see the description of this PR in `Follow up PRs`

shyuep:
This is merged since it only affects the dev directory. Thanks.

DanielYang59:
No problem at all, thanks for reviewing! 

Yes I was trying to separate JSON generator and rewrite of JSON loader/parser for ptable.py into separate PRs.

PR #4344, opened 2025-04-01
[Breaking] Update "electron affinities" in `periodic_table.json`

DanielYang59:
## Summary

Looks like some data entries from Wikipedia changed quite dramatically (some are just flucations), not sure if we want to update them?

https://github.com/materialsproject/pymatgen/blob/4c7892f5c9dcc51a1389b3ad2ada77632989a13e/dev_scripts/update_pt_data.py#L243-L246

### DIFF
--- a/src/pymatgen/core/periodic_table.json
+++ b/src/pymatgen/core/periodic_table.json
@@ -857,7 +857,7 @@
             14447.678,
             15028.907
         ],
-        "Electron affinity": 0.80482
+        "Electron affinity": 0.8044863
     },
     "At": {
         "Atomic mass": 210,
@@ -5602,7 +5602,7 @@
         "Ionization energies": [
             13.598434599702
         ],
-        "Electron affinity": 0.754598
+        "Electron affinity": 0.75419519
     },
     "He": {
         "Atomic mass": 4.002602,
@@ -6688,7 +6688,7 @@
             86438.9,
             88113.3
         ],
-        "Electron affinity": 1.5643615
+        "Electron affinity": 1.56405712
     },
     "K": {
         "Atomic mass": 39.0983,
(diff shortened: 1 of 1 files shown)


shyuep:
This one we need to be careful about. I am unclear where the old values come from. I believe it is the official Pauling definition. It might be old but it is what people have been using all this while. This can potentially cause havoc since MP uses the electronegativity in generating the formula.

DanielYang59:
Yes can never be too careful when changing such core data recordings, looks like to me it might came from the same source (wikipedia) as used in the PR (not sure if wikipedia updated its recording at some point?):
https://github.com/materialsproject/pymatgen/blob/55798119194076fd28cb28b539423779905c6ff9/dev_scripts/update_pt_data.py#L242-L247

shyuep:
Maybe a quick check would be to query for all MP formulas, rerun the Compositiion parsing and reduced formula nad compare the results? If there are no changes to formulas, we can potentially consider it. But again, this is something often used. E.g., matminer uses it as a feature in some instances.

DanielYang59:
Just to make sure we're on the same page, this PR tried to update the "electron affinity" not the "electronegativity" (`X`):

https://github.com/materialsproject/pymatgen/blob/55798119194076fd28cb28b539423779905c6ff9/src/pymatgen/core/periodic_table.py#L358-L361

https://github.com/materialsproject/pymatgen/blob/55798119194076fd28cb28b539423779905c6ff9/src/pymatgen/core/periodic_table.py#L295-L307

And indeed `Composition` uses `X` to sort the formula: 
https://github.com/materialsproject/pymatgen/blob/55798119194076fd28cb28b539423779905c6ff9/src/pymatgen/core/composition.py#L365-L373

And AFAIK there's no places across pymatgen accessing this property: 
image

PR #4360, opened 2025-04-15
Speed up `Vasprun` parsing some more

kavanase:
Me again.
From further profiling and playing around, I found I could speed up `_parse_vasp_array` (one of the main bottlenecks when using `parse_dos = True` (default), `parse_eigen = True` (default) and/or `parse_projected_eigen = True` (False by default)), using `numpy`'s parse from string function.
e.g. parsing a SOC defect supercell vasprun via [`doped`](https://doped.readthedocs.io/en/latest/) with these updates (with `parse_projected_eigen=True` to get eigenvalues/magnetisation) decreases parsing time from ~8.5s to ~4.8s.

All changes here should be covered by tests already in the codebase.

### DIFF
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -138,8 +138,8 @@ def _parse_vasp_array(elem) -> list[list[float]] | NDArray[float]:
         return [[i == "T" for i in v.text.split()] for v in elem]
 
     try:
-        # numerical data, try parse with numpy fromstring for efficiency:
-        return np.loadtxt([e.text for e in elem])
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks. but I don't think we need to use string concat? I believe np.loadtxt would be able to handle the text without concat.

shyuep:
Example:
```
In [1]: import numpy as np

In [3]: np.loadtxt(["1 2", "3 4"])
Out[3]:
array([[1., 2.],
       [3., 4.]])
```

shyuep:
Even better, no reshaping needed.

kavanase:
Ah yes! Good points. Done ⬆️

PR #4368, opened 2025-04-19
[WIPBetter input set generator and sets for io.lammps

vir-k01:
## Summary

Major changes:

- feature 1: ...
- fix 1: ...

## Todos

- feature 2: ...
- fix 2:

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


src/pymatgen/core/periodic_table.py
```diff
@@ -776,6 +777,19 @@ def is_rare_earth(self) -> bool:
         """
         return self.is_lanthanoid or self.is_actinoid or self.symbol in {"Sc", "Y"}
 
+    @property
+    @deprecated(
```
DanielYang59, line 781: I guess reverting changes to `core.periodic_table` and `io.exciting` is not intended?

esoteric-ephemera: Just behind the master branch, synced up now

src/pymatgen/io/lammps/generators.py
```diff
@@ -10,23 +10,448 @@
...
 from pymatgen.io.lammps.sets import LammpsInputSet
-
-__author__ = "Ryan Kingsbury, Guillaume Brunin (Matgenix)"
-__copyright__ = "Copyright 2021, The Materials Project"
-__version__ = "0.2"
+from pymatgen.util.typing import PathLike
 
 MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
 TEMPLATE_DIR = f"{MODULE_DIR}/templates"
+_BASE_LAMMPS_SETTINGS = {
```
gbrunin, line 31: To address the different default settings for solids or molecules, maybe this could be a nested dictionary like 
```
_COMMON_SETTINGS = {}  # Add any default that would be common to any type of data. 
_BASE_LAMMPS_SETTINGS = {
    "structure": _COMMON_SETTINGS | {"atom_style": "atomic", "boundary": ("p", "p", "p")},
    "molecule": _COMMON_SETTINGS | {"atom_style": "full", "boundary": ("f", "f", "f")},
    "general": _COMMON_SETTINGS,
}
```
that would then be used according to the type of the object passed to `get_input_set`.

src/pymatgen/io/lammps/generators.py
```diff
@@ -10,23 +10,448 @@
...
+        write_data = {"forcefield.lammps": FF_string}
+        if additional_data:
+            write_data.update({"extra.data": additional_data})
+            settings_dict.update({"extra_data_flag": "include"})
+
+        # Replace all variables
+        input_str = Template(self.inputfile.get_str()).safe_substitute(**settings_dict)  # type: ignore[union-attr]
+        lines = input_str.split("\n")
+        # Filter out the lines where the substitution resulted in a line starting with '#'
+        filtered_input_str = "\n".join([line for line in lines if not line.lstrip().startswith("#")])
```
gbrunin, line 449: This may remove some useful comments/stage names that have been written by the user or in the template and that should stay there.
I would suggest to replace missing flags by "###" and fitler the lines starting with "###" (assuming that comments start with a single #).

src/pymatgen/io/lammps/generators.py
```diff
@@ -10,23 +10,448 @@
...
+        input_file = LammpsInputFile.from_str(filtered_input_str, keep_stages=self.keep_stages)
+
+        return LammpsInputSet(
+            inputfile=input_file, data=data, calc_type=self.calc_type, additional_data=write_data, **kwargs
```
gbrunin, line 453: It would be nice to pass the argument `template_file=...` so that a trace of the initial template is kept somehow. This could probably be achieved in different ways though, not sure what would be best.

src/pymatgen/io/lammps/generators.py
```diff
@@ -10,23 +10,448 @@
...
+            present_settings = self.settings.dict
+            for k, v in updates.items():
+                present_settings.update({k: v})
+            self.settings = LammpsSettings(validate_params=self.validate_params, **present_settings)
+        else:
+            self.settings.update(updates)
+
+    def get_input_set(
+        self,
+        data: Structure | LammpsData | CombinedData,
```
gbrunin, line 303: This would have to accept Molecule as well.

src/pymatgen/io/lammps/generators.py
```diff
@@ -10,23 +10,448 @@
...
+            self.inputfile = LammpsInputFile.from_file(os.path.join(TEMPLATE_DIR, file), keep_stages=self.keep_stages)
+
+        if isinstance(self.settings, dict):
+            settings = self.settings.copy()
+            if self.include_defaults:
+                base_settings = _BASE_LAMMPS_SETTINGS.copy()
+                for key, value in base_settings.items():
+                    if key not in settings:
+                        settings[key] = value
+            self.settings = LammpsSettings(validate_params=self.validate_params, **settings)
```
gbrunin, line 276: Just a suggestion, but these few lines could be replaced by
```
settings = self.settings.copy()
if self.include_defaults:
    settings = base_settings | settings
self.settings = LammpsSettings(validate_params=self.validate_params, **settings)
```

src/pymatgen/io/lammps/generators.py
```diff
@@ -10,23 +10,448 @@
...
+        if isinstance(data, str):
+            try:
+                data = LammpsData.from_str(data, atom_style=atom_style)
+            except ValueError:
+                raise ValueError(f"Data file {data} not recognized. Please check the format.")
+
+        species = ""
+        if isinstance(data, Structure):
+            species = " ".join({s.symbol for s in data.species})
+            data = LammpsData.from_structure(data, atom_style=atom_style)
```
gbrunin, line 348: For Molecules, I think the `LammpsData.from_molecule` classmethod should be implemented.

gbrunin (commented):
Hi Vir,
Thanks for opening this PR, it looks very good and much better than what we had already done. I have left a few comments - those are only suggestions, I don't have a strong opinion on any of them.

I think indeed including support for molecules would be beneficial. I have added a few comments/suggestions going in that direction, let me know what you think.

The force field description is definitely a challenge with so many different ways to specify it. I'm not sure checking/validating what the user specifies is absolutely necessary at this stage - after all the code is also not checking the validity of other user-specified settings. I would leave it to the user to know how to specify the force field they want to use. But if others have ideas on how to implement that efficiently, I'm definitely not against it either :-)

Looking forward to having this integrated in pymatgen!

Best,
Guillaume

src/pymatgen/io/lammps/generators.py
gpetretto: Just as a small comment, I wanted to point out that in principle the `Lattice` has a `pbc` attribute:
https://github.com/materialsproject/pymatgen/blob/39308ec3b5d9bffdaa19ca63a62af2536311e390/src/pymatgen/core/lattice.py#L48
so in principle even mixed use cases could be handled directly with a `Structure` object.
Maybe it would be easier to have the `boundary` and `atom_style` settings be part of the `data` as produced by `LammpsData.from_structure`(`from_molecule`) instead? Would it make sense?

vir-k01: Thank you so much for the comments! I've made some of the changes, and let me try what @gpetretto mentioned for the molecules

davidwaroquiers:
@vir-k01 any news about this PR ?
Can we provide some help or would you like us to take this on if needed on your side and possible on our side ?

vir-k01:
@davidwaroquiers apologies for slacking off on this. After a bit of experimentation, I've added a `LammpsData.from_molecule` method and added different defaults as per what @gbrunin suggested. Tbh although this supports using molecules, a better way would be for the user to first make the `Topology` and `ForceField` objects as they've been currently implemented and use the `LammpsData.from_ff_and_topologies` method instead. This would preserve information about bonding. I suppose this shouldn't be an issue though for force fields that do not rely on such info.

esoteric-ephemera:
@mkhorton and @shyuep when you get a chance, would you be able to review this PR? We'd need a new pymatgen release to get @vir-k01's LAMMPS workflows tested on the atomate2 side. Thanks!

@vir-k01's addressed the earlier review comments from community members

esoteric-ephemera:
@mkhorton and @shyuep bumping this - the PR should be ready to go!

shyuep:

PR #4372, opened 2025-04-21
Reapply update to ptable vdw radii CSV source and JSON with CRC handbook

### LINKED ISSUE #4370: [DEV] Update CSV source for vdw/metallic radii data of `core.periodic_table.json` with CRC Handbook
#4345 only corrected inconsistency when data exists in BOTH the CSV and JSON, however I noticed some more issues are not corrected **when the data only exist in JSON but not the CSV** (meaning the data source is likely unknown or at least not from the CSV...)

Note the parent hash of #4338 is 4a3d06780cf32fdd8ad38f480ad588a82860e0e6 in case someone want to ensure it's not the new JSON generator the culprit.

DanielYang59:
## Summary

- Update ptable vdw/metallic radii CSV source, to fix #4370
- [x] Revert #4345 and apply changes to CSV


DanielYang59:
@rkingsbury I believe this PR has reverted #4345 and updated the [radii.csv](https://github.com/materialsproject/pymatgen/blob/master/dev_scripts/periodic_table_resources/radii.csv) accordingly. Would appreciate it if you could help me double check as I currently don't have access to a CRC handbook (you could/should work on the CSV instead of the JSON :D ).

Currently the JSON generator is still work in progress and it would not overwrite the `core/periodic_table.json` directly, but rather generate a temporary duplicate at `dev_scripts/periodic_table_resources/_periodic_table.json`. As there's still something inconsistency between the JSON and it's parser in `core/periodic_table.py` to be addressed around how value/unit is handled and I'm working on this.

---

There's another [temporary JSON compare script](https://github.com/materialsproject/pymatgen/blob/master/dev_scripts/periodic_table_resources/_compare_json.py) at `dev_scripts` in case you want to compare two JSONs.

rkingsbury:
Thanks @DanielYang59 !  Please connect with me via email and I can share the CRC data with you for double checking.

Note that I don't recall modifying the metallic radii in my previous PR; only the VdW ones. I'm not exactly sure about the meaning of "metallic radii".

DanielYang59:
> Note that I don't recall modifying the metallic radii in my previous PR;

Ah okay I didn't noticed that thanks for remaining me!

I just tracked down the git history (JSON is simply doesn't go well with version control). The "radii.csv" was added first in bec79b2803c3fe5939819535ef76375867257550 (already with "Metallic radii" data), and #1192 & #1241 directly added "metallic radii" to the JSON seemingly from a different source.

It's quite weird because the CSV data seems to [come from Wikipedia](https://en.wikipedia.org/wiki/Atomic_radii_of_the_elements_(data_page)#Metallic_radius) and #1192 seems to refering to Wikipedia but actually adding a different value

For example the metallic radius of K is 2.27 in CSV and wikipedia, but #1192 added a value of 2.381

I would revert changes to metallic radii to be consistent with CSV until we could verify the source from #1192

rkingsbury:
> > I can share the CRC data with you for double checking.
> 
> Thanks for sharing the vdw radii page. I just double checked all vdw radii, and noticed only one difference (could also be the 105th ed CRC handbook after the 91st ed):
> 
> * vdw radius for Pa is 2.43 in the handbook, 2.1 after [Update vdW radii to CRC handbook values #2122](https://github.com/materialsproject/pymatgen/pull/2122)

Great! Thank you for double checking. I'd go with the latest CRC value (2.43). A better value may have become available since the edition that I used in the previous PR.

DanielYang59:
> I'd go with the latest CRC value (2.43). 

Yes it has been applied in 939ea85ee71f901dd0f7020d8f4251a2710347d4

Issue #4373, opened 2025-04-21
[DEV] Add source for "Ground level" and "NMR Quadrupole Moment" to ptable JSON


DanielYang59:
Repost from #4338, it looks like ["Ground level"](https://github.com/materialsproject/pymatgen/blob/ac95b80fc914d237bc31b0e3cd74c1d95f85e135/dev_scripts/update_pt_data.py#L258-L279) (also from `ionization_energies_nist.html`) and ["NMR Quadrupole Moment"](https://github.com/materialsproject/pymatgen/pull/1099) are directly added to JSON without keeping a local copy of the source

DanielYang59:
Close as this is just for my own reference and I'm working on it

DanielYang59:
link #4374

PR #4374, opened 2025-04-21
Separate "Ground level" and "NMR Quadrupole Moment" from `elemental_properties.yaml` for ptable JSON

DanielYang59:
#4373

- [x] Add "NMR Quadrupole Moment" as a separate YAML source
- [x]  "ground level" seems to be generated from the same NIST source as ionization energies


DanielYang59:
close, just do this in #4376

PR #4391, opened 2025-05-06
add custom from_dict method for proper initialization of attributes of IcohpCollection

naik-aakash:


naik-aakash:
Dear @shyuep ,

I wanted to know what the implications would be if one does not use the `unitized` decorator for the Vasprun.final_energy property and method [here](https://github.com/naik-aakash/pymatgen/blob/fd13e102aa263bd1cb44f9d091fc9bc0399107cb/src/pymatgen/io/vasp/outputs.py#L698) 

I am asking this because I ran into serialization issues when working on the app in crystal-toolkit (See https://github.com/materialsproject/crystaltoolkit/pull/459). 

The issue to be precise orginates from plotly not being able to serialize FloatWithUnit class objects.

JaGeo:
Hi @shyuep , @mkhorton , @esoteric-ephemera ,

does anyone of you have more insight on this issue? We ran into this when preparing our LOBSTER data for the CrystalToolkit and corresponding visualization.

esoteric-ephemera:
Not sure I understand: how is `Vasprun` being serialized by plotly? There's no real deserialization of `Vasprun`: while it's `MSONable`, the roundtrip `from_dict` can't deserialize a `Vasprun` from its `as_dict()`

You can also use the `as_dict` representation where all `unitized` objects are base numeric types:
```
Vasprun(...).as_dict()['output']['final_energy']
```

naik-aakash:
> Not sure I understand: how is `Vasprun` being serialized by plotly? There's no real deserialization of `Vasprun`: while it's `MSONable`, the roundtrip `from_dict` can't deserialize a `Vasprun` from its `as_dict()`
> 
> You can also use the `as_dict` representation where all `unitized` objects are base numeric types:
> 
> ```
> Vasprun(...).as_dict()['output']['final_energy']
> ```

Hi  @esoteric-ephemera,  yes `Vasprun` cannot be deserialized, and for the app that i am working in crystaltoolkit, deserialization is not needed as I need it only once during initialization. 

Maybe I got it wrong before, I think serialization happens via dash, not plotly, I suppose here https://github.com/materialsproject/crystaltoolkit/blob/79e6a9981f0af0a97109e2f7e9343aefc0d75a1c/crystal_toolkit/core/mpcomponent.py#L270,

The Error I get exactly is this `TypeError: Object of type FloatWithUnit  is not JSON serializable` when I try to initialize the app.

esoteric-ephemera:
My guess is that something is not going right on the crystaltoolkit side since I'm able to do
```
json.dumps(Vasprun(...).as_dict())
```
without issue. Maybe just pass the `as_dict` of `Vasprun` as an object to the dash store for now?

naik-aakash:
> My guess is that something is not going right on the crystaltoolkit side since I'm able to do
> 
> ```
> json.dumps(Vasprun(...).as_dict())
> ```
> 
> without issue. Maybe just pass the `as_dict` of `Vasprun` as an object to the dash store for now?

Thanks @esoteric-ephemera , this can work, maybe I will adapt a bit of stuff in our LobsterPy API too then it should work fine without needing any changes.

PR #4392, opened 2025-05-06
Brute force order matcher speedup

kavanase:
This is a small PR to add a `break_on_tol` option to `BruteForceOrderMatcher`, which allows one to massively speed up the molecule matching in cases where one just cares _if_ the molecules match within a given RMSD (rather than screening over _all_ possible permutations, to get the lowest possible RMSD). 

The speedup factor can be anywhere from 0% to several orders of magnitude, depending on the number of possible permutations and when a matching permutation is found.

Also includes a small typo fix that for an error message that reference the original class name (in https://github.com/materialsproject/pymatgen/pull/1938) which was later renamed. 

I've also added some tests for this.


shyuep:
Thanks,

PR #4400, opened 2025-05-09
Use `np.nan` instead of 0 for no uncertainty with `ufloat`, to avoid unnecessary warnings

### LINKED ISSUE #4386: `UFloat` Warnings with `ComputedEntry.energy`
### Python version

Python 3.10-3.12

### Pymatgen version

2025.5.2

### Operating system version

_No response_

### Current behavior

From running our `doped` tests with the latest `pymatgen` release, we get many of these warnings (which breaks tests):
```
Using UFloat objects with std_dev==0 may give unexpected results.
```

From setting the warning level to error, the traceback shows:
```
File ~/Packages/doped/doped/core.py:1428, in DefectEntry.sc_entry_energy(self)
   1425     return self._sc_entry_energy
   1427 self._sc_entry_hash = hash(self.sc_entry)
-> 1428 self._sc_entry_energy = self.sc_entry.energy
   1430 return self._sc_entry_energy

File ~/Packages/pymatgen/src/pymatgen/entries/computed_entries.py:350, in ComputedEntry.energy(self)
    347 @property
    348 def energy(self) -> float:
    349     """The *corrected* energy of the entry."""
--> 350     return self.uncorrected_energy + self.correction

File ~/Packages/pymatgen/src/pymatgen/entries/computed_entries.py:367, in ComputedEntry.correction(self)
    362 """
    363 Returns:
    364     float: the total energy correction / adjustment applied to the entry in eV.
    365 """
    366 # adds to ufloat(0.0, 0.0) to ensure that no corrections still result in ufloat object
--> 367 corr = ufloat(0.0, 0.0) + sum(ufloat(ea.value, ea.uncertainty) for ea in self.energy_adjustments)
    368 return corr.nominal_value

File ~/miniconda3/lib/python3.10/site-packages/uncertainties/core.py:1024, in ufloat(nominal_value, std_dev, tag)
    992 """
    993 Create an uncertainties Variable
    994 
   (...)
   1021    and `tag` which match the input values.
   1022 """
   1023 if std_dev == 0:
-> 1024     warn("Using UFloat objects with std_dev==0 may give unexpected results.")
   1025 return Variable(nominal_value, std_dev, tag=tag)

UserWarning: Using UFloat objects with std_dev==0 may give unexpected results.
```

I think this is due to a recent update in `uncertainties`, which now includes this warning, which occurs when there are no energy adjustments with uncertainties in `ComputedEntry.energy_adjustments`. Some discussion here: https://github.com/lmfit/lmfit-py/issues/999

### Expected Behavior

Not totally sure, but it seems like that warning is there to stay, so might make sense to catch/handle it here somehow?
Only use `ufloat` when there is an energy adjustment with non-zero uncertainty, otherwise just normal floats for correction values?

### Minimal example

```Python

ce = ComputedEntry(composition="H", energy=1)
ce.energy  # shows warning
```

_No response_

kavanase:
Closes #4386

### DIFF
--- a/src/pymatgen/analysis/reaction_calculator.py
+++ b/src/pymatgen/analysis/reaction_calculator.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, overload
 import numpy as np
 from monty.fractions import gcd_float
 from monty.json import MontyDecoder, MSONable
-from uncertainties import ufloat
+from uncertainties import UFloat, ufloat
 
 from pymatgen.core.composition import Composition
 from pymatgen.entries.computed_entries import ComputedEntry
@@ -100,7 +100,7 @@ class BalancedReaction(MSONable):
     __repr__ = __str__
 
     @overload
-    def calculate_energy(self, energies: dict[Composition, ufloat]) -> ufloat:
+    def calculate_energy(self, energies: dict[Composition, ufloat]) -> UFloat:
         pass
 
     @overload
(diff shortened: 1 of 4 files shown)


shyuep:
Thanks. I agree this is probably a more accurate representation. 0 implies total certainty. I am tagging @rkingsbury since I believe he is the person who wrote most of this. I will wait for his comments before I merge.

shyuep:
Note that there are some failing tests due to the change in uncertainty definition.

kavanase:
Ah I missed the reaction calculator usage. Should be fixed now

rkingsbury:
Thanks @kavanase ; yes I think this fix should be fine. In fact, it may have been our original intent to default to `np.nan` when there was no uncertainty but we switched back to using zero so that corrections could be added to. I'd be concerned about unintended behavior in `test_compatibility.py` , test_correction_calculator.py`, and `test_mixing_scheme.py`, but they all appear OK with this change.

rkingsbury:
This was originally implemented by @awvio (with me supervising)

Issue #1929, opened 2020-08-14
SpacegroupAnalyzer.get_conventional_standard_structure returns a wrong structure


lan496:
**To Reproduce**
If I understand correctly, a structure and a symmetry-refined one have the same minimum distance between sites.
So, three structures `structure`, `conventional_refined`, and `spglib_refined` in the following script should have the same minimum distance.

```python
import numpy as np
from pymatgen.core import Structure
from pymatgen.analysis.structure_analyzer import SpacegroupAnalyzer


def min_dist(structure):
    all_dists = structure.distance_matrix[np.triu_indices(structure.num_sites, 1)]
    return np.min(all_dists)


if __name__ == '__main__':
    # spacegroup: C2/m
    matrix = [[3.945, 0.0, 3.945], [0.0, 3.945, 3.945], [0.0, 0.0, 11.834999999999999]]
    species = ['O'] * 7
    frac_coords = np.array([
        [0., 0.5, 0.66666667],
        [0.5, 0., 0.],
        [0.5, 0., 0.33333333],
        [0.5, 0., 0.66666667],
        [0.5, 0.5, -0.33333333],
        [0.5, 0.5, 0.],
        [0.5, 0.5, 0.33333333],
    ])
    structure = Structure(matrix, species, frac_coords)
    print("original:", min_dist(structure))

    sga = SpacegroupAnalyzer(structure)

    conventional_refined = sga.get_conventional_standard_structure()
    print("conventional_standard:", min_dist(conventional_refined))

    spglib_refined = sga.get_refined_structure()
    print("refine_cell:", min_dist(spglib_refined))

    conventional_refined.to(fmt='cif', filename='conventional_standard.cif', symprec=1e-2)
    spglib_refined.to(fmt='cif', filename='refined.cif', symprec=1e-2)
```

But I got the following output. The structure created by `SpacegroupAnalyzer.get_conventional_standard_structure` has a different distance from that of the original structure.
```
original: 2.789536223885567
conventional_standard: 1.610539489774544
refine_cell: 2.7895361959902023
```

**Screenshots**
The structure created by `SpacegroupAnalyzer.get_conventional_standard_structure` is as follows:
![conventional](https://user-images.githubusercontent.com/7272505/90248981-a55b0600-de74-11ea-9275-23a08e396cb1.png)

The structure created by `SpacegroupAnalyzer.get_refined_structure` is as follows:
[...]

PR #4404, opened 2025-05-13
Monoclinic Symmetry Handling Fix

kavanase:
This issue is related to https://github.com/materialsproject/pymatgen/issues/1929 (for which a patch solution was added that fixed one case of this occurrence, but not in general).

When handling monoclinic symmetry within `SpacegroupAnalyzer`, the standardisation attempts to reorder the lattice vectors depending on whether the alpha angle is >90 degrees or <90 degrees. If it is exactly 90 degrees, it defaults to the original lattice, however there was an issue with this implementation where it assumed that the beta and gamma angles were also 90 degrees (and so set the lattice matrix as `[[a, 0, 0], [0, b, 0], [0, 0, c]]`), which may not (and in most cases should not) be the case.

This was causing weird behaviour for me where the volume of the primitive structures being returned by `SpacegroupAnalyzer` was different to `Structure.find_primitive()` (by a non-integer factor). I've added a test for this case, and also confirmed that the general implementation here works for the failure case noted in https://github.com/materialsproject/pymatgen/issues/1929

### DIFF
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -771,95 +771,72 @@ class SpacegroupAnalyzer:
                     latt2 = Lattice([m[tp2[0]], m[tp2[1]], m[2]])
                     lengths = latt2.lengths
                     angles = latt2.angles
-                    if angles[0] > 90:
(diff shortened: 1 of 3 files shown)


shyuep:
Thanks!

PR #4412, opened 2025-05-19
[WIP] Update VASP outputs.py for more robust piezo and elastic tensor extraction from OUTCAR

knoori:
**Purpose:**
This PR updates the VASP `outputs.py` to allow for the extraction from OUTCAR of piezoelectric properties from PEAD calculations (LCALCEPS). It also updates the code to allow for the correct extraction of the electronic (ion-clamped) piezoelectric stress tensor for both LEPSILON and LCALCEPS calculations. The full elastic tensor is now also extracted in the case of a finite difference (FD) calculation that allows for variable cell deformation. 

**Modifications:**
The `Outcar` class is modified to include a finite difference `self.fd` check if IBRION is 5 or 6. We also check for RPA (LRPA) in OUTCAR and whether a variable cell relaxation is being performed (ISIF is 3), and set `self.lrpa` and `self.varcell` accordingly.

We add a check for LCALCEPS in OUTCAR and set `self.lcalceps` accordingly. Similar to the workflow for LEPSILON, we call `self.read_epsilon()` if `self.lcalceps` is True, and call `self.read_lepsilon_ionic()` if `self.fd` is True. Note that we can re-use the `read_epsilon()` and `read_lepsilon_ionic()` functions for the LCALCEPS workflow.

We update `read_lepsilon()` to check if `self.lrpa` is False (it should be for piezo calculations, cf. [VASP wiki](https://www.vasp.at/wiki/index.php/LEPSILON)) and modify the search string to correctly extract the electronic piezo tensor (verified to work at least for VASP v6.4.3).

If we have a FD calculation (`self.fd = True`) and ISIF is 3  (`self.varcell = True`), we can extract the full elastic tensor using `self.read_elastic_tensor()`. This function has been updated by modifying the search string to ensure the full 6x6 tensor is correctly read (verified for VASP v6.4.3). 

In `as_dict` we add checks for `self.lcalceps`, `self.lcalceps` AND `self.fd`, and `self.fd` AND `self.varcell`. We append the corresponding electronic piezo, ionic piezo, and elastic tensors in a manner consistent with the existing code. 

**NB** I’m not sure if the changes are substantial enough to warrant new tests, but I can write them if needed.

### DIFF
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -41,7 +41,6 @@ from pymatgen.io.vasp.inputs import Incar, Kpoints, Poscar, Potcar
 from pymatgen.io.wannier90 import Unk
 from pymatgen.util.io_utils import clean_lines, micro_pyawk
 from pymatgen.util.num import make_symmetric_matrix_from_upper_tri
-from pymatgen.util.typing import Kpoint, Tuple3Floats, Vector3D
 
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


shyuep:

Issue #3940, opened 2024-07-19
Calling PhaseDiagram.as_dict() breaks the link between different Entries


CompRhys:
### Feature Requested

When saving the `PhaseDiagram` it could reduce the memory footprint if to keep track of the ids of entries rather than just save multiple copies of entries in the separate lists. 

### Proposed Solution

Refactor the PhaseDiagram class to use `@property` to return the different lists of entries that are kept track of based on lists of ids as opposed to making references that result in duplication when serializing.

### Relevant Information

This isn't a priority and would only be worth merging if the differences in speed are negligible. Posting this issue because it may be a good first issue for someone to tackle if I can't get round to it in the near term.

CompRhys:
closed by #4414

PR #4414, opened 2025-05-20
Phase Diagram Serialization Improvement

fadams-umd:
## Summary

Addresses #3940 by modifying the serialization of `pymatgen.analysis.phase_diagram.PhaseDiagram` to (1) remove redundant copies of the list of all the entries and (2) serialize the subset `qhull_entries` as a list of indices from which the list of entries can be recreated upon deserialization.

### DIFF
--- a/src/pymatgen/analysis/phase_diagram.py
+++ b/src/pymatgen/analysis/phase_diagram.py
@@ -392,7 +392,6 @@ class PhaseDiagram(MSONable):
         return {
             "@module": type(self).__module__,
             "@class": type(self).__name__,
-            "all_entries": [e.as_dict() for e in self.all_entries],
             "elements": [e.as_dict() for e in self.elements],
             "computed_data": self.computed_data,
         }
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @fadams-umd, this seems much more efficient.

For backwards compatibility, for those who have already serialized a phase diagram with the older method, could you add an appropriate `from_dict()` method that can handle both formats?

felix-adams:
Certainly! I just added a check for a dictionary key only present in the old format, and adjusted deserialization accordingly.

mkhorton:
We're getting a test failure on `analysis.test_phase_diagram.TestPhaseDiagram.test_as_from_dict` that's blocking the merge. It looks like the test output just needs to be updated to reflect the changes made in this PR.

felix-adams:
I was inadvertently modifying the dictionary argument in the `from_dict` method, which was throwing off the tests. Creating a copy solved the issue.

CompRhys:
@fadams-umd thanks for working on this! I originally noted this issue when working with patched phase diagrams which inflate this issue even more due to shared copies between patches. The only reasonable way to store them currently is as pickles which is an awful way to persist long term data.

mkhorton:
Apologies for the delay in merging this @felix-adams!

shyuep:
Thanks!

PR #4419, opened 2025-05-27
Fix Molecule.get_boxed_structure when reorder=False

gpetretto:
## Summary

In `Molecule.get_boxed_structure`, when `reorder=False`, in the current version of the code does not use the processed new coordinates, but the initial ones of the Molecule. This usually results for example in a Molecule not being centered in the box.
Unless I am missing something there should be no reason not to use the modified coordinates in the case of `reorder=False`. This PR switches to use the same coordinates independently of the final sorting.

I added a test based on the coordinates that would have failed in the previous implementation due to the molecule not being centered in the box.


shyuep:
Thanks,.

PR #4422, opened 2025-05-29
[WIP]`Trajectory` -> Gaussview readable ".log" file

benrich37:
From #4420

## Summary

Major changes:

- feature 1: Adds the class function `core.trajectory.Trajectory.write_Gaussian_log`, which writes a Gaussview readable ".log" file from the contents of a `Trajectory` object

## Todos

- feature 1: Fill out the function stub
-- Write each `Structure` in the `Trajectory`
-- Fetch the energies series from `Trajectory.frame_properties["energy"]`
-- Default write from `Trajectory.site_properties[i]`
--- "charges" to "Mulliken charges" and "magmom" to "Mulliken spin"
-- Implement an optional mapping of a site property name to a supported (producible by Gaussian) site property
--- Allows flexible representation of up to (9?) compatible site properties
--- Assumes the requested site property maps each site to a single float

### DIFF
--- a/src/pymatgen/core/trajectory.py
+++ b/src/pymatgen/core/trajectory.py
@@ -477,6 +477,21 @@ class Trajectory(MSONable):
         with zopen(filename, mode="wt", encoding="utf-8") as file:
             file.write(xdatcar_str)  # type:ignore[arg-type]
 
+    def write_Gaussian_log(
+        self,
+        filename: PathLike = "calc.log",
+    ) -> None:
+        """Write Trajectory object to a Gaussian-formatted log file.
+
+        The produced file will be readable by Gaussview.
+
+        Args:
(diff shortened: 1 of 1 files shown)


benrich37:
@shyuep For `core.trajectory.Trajectory.write_Gaussian_log`, a handful of helper functions are required - right now they're being held in `core.trajectory` and are being made "private", but I wanted to check with you if that is okay or if there is another location to hold them in that would make more sense. 

There are also some `str` constants being held in `core.trajectory` that are supposed to be used by the user for creating custom `site_property_map` for `core.trajectory.Trajectory.write_Gaussian_log` - is this overly complicated? It also feels silly to have them defined in `core.trajectory`, but I'm not sure where I would move them.

shyuep:
Sorry for the late response. Yes, what you propose is fine. However, we have moved this to pymatgen-core. Please resubmit a PR there when ready.

PR #4431, opened 2025-06-05
Single source of truth for POTCAR directory structure

### LINKED ISSUE #4430: Inconsistent POTCAR naming for PBE_64 between `pmg_config.py` and `inputs.py`
I've noticed a conflict between the POTCAR naming conventions used in `pmg_config.py` and `inputs.py`.

In `pmg_config.py`, the following mapping is defined:
```
"potpaw_PBE.64": "POT_GGA_PAW_PBE_64"
```
However, in `vasp.inputs.py`, the `functional_dir` maps:
```
"PBE_64": "POT_PAW_PBE_64"
```
This means that there's no consistent way to access `"PBE_64"` POTCARs using both `pmg` and `pymatgen.io.vasp.inputs`. Either:

* `"PBE_64"` should resolve to `"POT_GGA_PAW_PBE_64"` (to match the rest of the naming in `pmg_config.py`), or
* `"potpaw_PBE.64"` in `pmg_config.py` should map to `"POT_PAW_PBE_64"`.

I wanted to flag this inconsistency in case it causes issues for other users or workflows relying on `PBE_64`. I’m not sure which mapping is intended to be canonical, but resolving the discrepancy might help avoid confusion or errors.

esoteric-ephemera:
Modifies the pymatgen CLI to use the same POTCAR library directory structure as in `pymatgen.io.vasp.inputs` to close #4430. Possibly breaking from the CLI side (the directory structure will change)

Pinging @mkhorton since #4424 was probably motivated by similar concerns?

### DIFF
--- a/src/pymatgen/cli/pmg_config.py
+++ b/src/pymatgen/cli/pmg_config.py
@@ -17,6 +17,7 @@ from monty.serialization import dumpfn, loadfn
 from pymatgen.core import OLD_SETTINGS_FILE, SETTINGS_FILE, Element
 from pymatgen.io.cp2k.inputs import GaussianTypeOrbitalBasisSet, GthPotential
 from pymatgen.io.cp2k.utils import chunk
+from pymatgen.io.vasp.inputs import POTCAR_FUNCTIONAL_MAP
 
 if TYPE_CHECKING:
     from argparse import Namespace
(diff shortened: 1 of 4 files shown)


mkhorton:
Thanks for notifying!

PR #4438, opened 2025-06-17
Add basic S3 object retrieval to the pymatgen user agent

esoteric-ephemera:
The MP API has changed to accommodate increasingly large datasets. This PR allows for retrieving certain data objects from S3 that are no longer accessible via MongoDB

In the future, communicating which features unexpectedly no longer work, e.g., through matsci.org, is a great way to handle feature changes. [Calling out the team](https://github.com/materialsproject/pymatgen/commit/c1db75380fbf8e94bbb58ce85c10d9dffade0c5e) is not going to help

fyi @mkhorton, @computron

### DIFF
--- a/src/pymatgen/electronic_structure/bandstructure.py
+++ b/src/pymatgen/electronic_structure/bandstructure.py
@@ -18,7 +18,7 @@ from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
 from pymatgen.util.coord import pbc_diff
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping, Sequence
+    from collections.abc import Mapping
     from typing import Any
 
     from numpy.typing import ArrayLike, NDArray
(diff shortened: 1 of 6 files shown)


shyuep:
1. I already communicated the problem to the MP team via email. Got a non helpful response.
2. Breaking an API with zero communication to the community (and no updates to available documentation) is a bad practice and dangerously careless to begin with. 
3. I don't consider the commit message as "calling out the MP team". Merely stating a fact.

esoteric-ephemera:
1. Can't comment other than our python client is kept up to date with the most recent database / S3 changes - if a feature set breaks, looking there is a good idea

2. Let's not be hyperbolic, there's no danger here. Yes, we'd love to be able to update documentation more quickly, but it's much less of a priority given time constraints than working on new data products. 

3. Toeing the line of the code of conduct. The *fact* is that the same data is available, just no longer in the database for reasons of scale

shyuep:
The API and the python client are two separate entities. Breaking the API is even worse than breaking the python client since an end user is technically supposed to pin to a specific python dependency. An API breakage breaks even previous python clients, i.e., a backwards incompatibility that cannot be mitigated even by pinning. There is a whole literature on API versioning best practices, which was not done in this case.

Documentation and features go hand in hand. Documentation should be added/updated as new data/features are added, not patched as an afterthought when someone finds breakages in downstream code. It is 2025. If the belief is that new features are more important than documentation even in this day, we will have to agree to disagree.

PR #4443, opened 2025-06-30
Better comment handling for CP2K input file

### LINKED ISSUE #4442: CP2K input: Error from a slash in the comments
### Python version

Python 3.10.18

### Pymatgen version

2025.6.14

### Operating system version

Red Hat Enterprise Linux 9.5 (Plow)

### Current behavior

Attempting to read an input [file](https://github.com/PolyachenkoYA/lammps_questions/blob/main/pymatgen_parse/hit.inp) via `pymatgen.io.cp2k.inputs.Cp2kInput` results in the error 
```
File "/home/yp1065/.conda/envs/main_10/lib/python3.10/site-packages/pymatgen/io/cp2k/inputs.py", line 538, in by_path
    sec_str = sec_str.get_section(pth)
AttributeError: 'NoneType' object has no attribute 'get_section'
```

I modified `inputs.py` to print `pth` every time, and it breaks at line 79 (in `hit.inp`): `&VELOCITY   ! [bohr / au_t = b ...` on the `/`. I tried removing this `/` - then it goes to the next one.

This input file actually has a `/` on line 64, so before 79. But this line seems to be handled differently because I tried swapping the text of comments on these lines, and it still fails on 79.

Removing all `/`-s from the 79th line resolves the issue.

### Expected Behavior

Any text to be allowed in the comments (i.e. after a `!`)

### Minimal example

```Python
from pymatgen.io.cp2k.inputs import Cp2kInput
print(Cp2kInput.from_file('hit.inp'))
```

A [script](https://github.com/PolyachenkoYA/lammps_questions/blob/main/pymatgen_parse/test.py)
The input [file](https://github.com/PolyachenkoYA/lammps_questions/blob/main/pymatgen_parse/hit.inp)

DanielYang59:
## Summary

- Better comment handling for CP2K input file, to fix #4442

### DIFF
--- a/src/pymatgen/io/cp2k/inputs.py
+++ b/src/pymatgen/io/cp2k/inputs.py
@@ -151,7 +151,7 @@ class Keyword(MSONable):
         )
 
     @classmethod
-    def from_str(cls, s: str, description: str | None = None) -> Self:
+    def from_str(cls, s: str, description: str | None) -> Self:
         """
         Initialize from a string.
 
@@ -745,7 +745,8 @@ class Cp2kInput(Section):
                 current = f"{current}/{alias or name}"
 
             else:
-                kwd = Keyword.from_str(line, description)
+                kwd = Keyword.from_str(line, (description or None))
+
                 if tmp := self.by_path(current).get(kwd.name):
                     if isinstance(tmp, KeywordList):
                         self.by_path(current).get(kwd.name).append(kwd)
--- a/tests/io/cp2k/test_inputs.py
+++ b/tests/io/cp2k/test_inputs.py
@@ -224,6 +224,7 @@ class TestCp2kInput(MatSciTest):
         assert (
             self.ci["FORCE_EVAL"]["DFT"]["MGRID"]["CUTOFF"].description == "Cutoff in [Ry] for finest level of the MG."
         )
+        assert self.ci["FORCE_EVAL"]["METHOD"].description is None
 
     def test_odd_file(self):
         scramble = ""


src/pymatgen/io/cp2k/inputs.py
```diff
@@ -152,7 +151,7 @@ def from_dict(cls, dct: dict) -> Self:
         )
 
     @classmethod
-    def from_str(cls, s: str) -> Self:
+    def from_str(cls, s: str, description: str | None = None) -> Self:
```
DanielYang59, line 154: TODO: This could be breaking as `Keyword` is also used in other CP2K modules

pyproject.toml
```diff
@@ -297,7 +297,6 @@ namespace_packages = true
 no_implicit_optional = false
 disable_error_code = ["annotation-unchecked", "override", "operator", "attr-defined", "union-attr", "misc", "call-overload", "index"]
 exclude = ['src/pymatgen/analysis', 'src/pymatgen/io/cp2k', 'src/pymatgen/io/lammps']
-plugins = ["numpy.typing.mypy_plugin"]
```
DanielYang59, line 300: /Users/yang/.cache/pre-commit/reposlddsc8t/py_env-python3.13/bin/mypy:8:
DeprecationWarning: `numpy.typing.mypy_plugin` is deprecated, and will
be removed in a future release. Please remove `plugins =
numpy.typing.mypy_plugin` in your mypy config.(deprecated in NumPy 2.3)
  sys.exit(console_entry())

PR #4447, opened 2025-07-03
Fix `has_cobicar` when NcICOBILIST is present

tomdemeyere:
When `icobiBetween` is activated lobster prints out the `NcICOBILIST.lobster` file. In the lobster output we find "Writing COBICAR.lobster, ICOBILIST.lobster and NcICOBILIST.lobster..." instead of the regular line. The current parser then returns `False` even if the COBICAR.lobster is present.

Also, it seems that `has_cohpcar` and `has_coopcar` are flipped?

Switching to a regex approach on the whole files (not `\n` split) would allow much greater flexbility, performance and robustness.

@JaGeo @naik-aakash 

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/src/pymatgen/io/lobster/outputs.py
+++ b/src/pymatgen/io/lobster/outputs.py
@@ -1130,10 +1130,9 @@ class Lobsterout(MSONable):
                     "writing COOPCAR.lobster..." in lines and "SKIPPING writing COOPCAR.lobster..." not in lines
                 )
                 self.has_cobicar = (
-                    ("writing COBICAR.lobster..." in lines
-                     or "Writing COBICAR.lobster, ICOBILIST.lobster and NcICOBILIST.lobster..." in lines)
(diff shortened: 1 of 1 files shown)


naik-aakash:
Hi @tomdemeyere, thanks for catching this flip in checks for  COHP and COOP and fixing the has_cobicar logic when multicenter bonding is enabled.

Regarding the suggestion of switching to regex, I do not have any strong opinion. The only thing I feel is that maintaining it can be harder if one is not very familiar with regex syntax.

JaGeo:
I think my opinion on regex is similar to @naik-aakash 's. As long as one can still easily maintain it, I am fine with it. We also, of course, have to check with the main maintainers of this repo :)

And thanks for catching the bugs!

shyuep:
Thanks,

PR #4448, opened 2025-07-04
Remove LOBSTER output file trailing line sensitivity

### LINKED ISSUE #4215: Cannot init `Cohpcar` in `matgenb ` example notebook: `ValueError: setting an array element with a sequence.`
The [example in `matgenb` seems to be failing](https://github.com/materialsvirtuallab/matgenb/blob/master/notebooks/2018-03-14-Plotting%20COHP%20from%20LOBSTER.ipynb).

```python
from pymatgen.io.lobster import Cohpcar


COHPCAR_path = "COHPCAR.lobster.txt"
cohpcar = Cohpcar(filename=COHPCAR_path)
```

Gives:
```
  File "/home/yang/developer/pymatgen/debug/recreate_cohp.py", line 5, in <module>
    cohpcar = Cohpcar(filename=COHPCAR_path)
  File "/home/yang/developer/pymatgen/src/pymatgen/io/lobster/outputs.py", line 138, in __init__
    data = np.array([np.array(line.split(), dtype=float) for line in lines[num_bonds + 3 :]]).transpose()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (402,) + inhomogeneous part.
```

[COHPCAR.lobster.txt](https://github.com/user-attachments/files/18005481/COHPCAR.lobster.txt)

cc @JaGeo @QuantumChemist @naik-aakash

DanielYang59:
## Summary

- Remove LOBSTER output file trailing line sensitivity, fix #4215

[`str.splitlines(keepends=False)`](https://docs.python.org/3/library/stdtypes.html#str.splitlines):
> Unlike [split()](https://docs.python.org/3/library/stdtypes.html#str.split) when a delimiter string sep is given, this method returns an empty list for the empty string, and a terminal line break does not result in an extra line:

cc @JaGeo @naik-aakash

### DIFF
--- a/tests/io/lobster/test_outputs.py
+++ b/tests/io/lobster/test_outputs.py
@@ -30,6 +30,7 @@ from pymatgen.io.lobster import (
     SitePotential,
     Wavefunction,
 )
+from pymatgen.io.lobster.outputs import _get_lines
 from pymatgen.io.vasp import Vasprun
 from pymatgen.util.testing import TEST_FILES_DIR, VASP_IN_DIR, VASP_OUT_DIR, MatSciTest
 
@@ -82,7 +83,7 @@ class TestCohpcar(MatSciTest):
         self.orb = Cohpcar(filename=f"{TEST_DIR}/COHPCAR.lobster.orbitalwise.gz")
         self.orb_notot = Cohpcar(filename=f"{TEST_DIR}/COHPCAR.lobster.notot.orbitalwise.gz")
 
-        # Lobster 3.1 (Test data is from pre-release of Lobster 3.1)
+        # Lobster 3.1 (Test data is from prerelease of Lobster 3.1)
         self.cohp_KF = Cohpcar(filename=f"{TEST_DIR}/COHPCAR.lobster.KF.gz")
         self.coop_KF = Cohpcar(
             filename=f"{TEST_DIR}/COHPCAR.lobster.KF.gz",
@@ -2531,3 +2532,17 @@ class TestPolarization(MatSciTest):
             "abs": 56.14,
             "unit": "uC/cm2",
         }
+
+
+def test_get_lines():
+    """Ensure `_get_lines` is not trailing end char sensitive."""
+    with open("without-end-char", mode="wb") as f:
+        f.write(b"first line\nsecond line")
+
+    with open("with-end-char", mode="wb") as f:
+        f.write(b"first line\nsecond line\n")
+
+    without_end_char = _get_lines("without-end-char")
+    with_end_char = _get_lines("with-end-char")
+
+    assert len(with_end_char) == len(without_end_char) == 2


src/pymatgen/io/lobster/outputs.py
```diff
@@ -405,14 +404,10 @@ def __init__(
...
 
-        # LOBSTER list files have an extra trailing blank line
-        # and we don't need the header.
         if self._icohpcollection is None:
             with zopen(self._filename, mode="rt", encoding="utf-8") as file:
-                all_lines: list[str] = file.read().splitlines()  # type:ignore[assignment]
+                all_lines: list[str] = cast("list[str]", file.read().splitlines())
 
-                # strip *trailing* blank lines only
-                all_lines = [line for line in all_lines if line.strip()]
```
DanielYang59, line 415: Might need to double check this because I don't think this is "strip *trailing* blank lines only" (unless there's no empty line in the middle of the file)

tests/io/lobster/test_outputs.py
```diff
@@ -64,15 +65,24 @@ def setup_method(self):
             filename=f"{TEST_DIR}/COOPCAR.lobster.BiSe.gz",
             are_coops=True,
         )
-        self.cohp_fe = Cohpcar(filename=f"{TEST_DIR}/COOPCAR.lobster.gz")
+
+        # Make sure Cohpcar also works with terminating line ending char
+        gz_path = f"{TEST_DIR}/COOPCAR.lobster.gz"
+        with gzip.open(gz_path, "rt", encoding="utf-8") as f:
+            content = f.read() + "\n"
+        with tempfile.NamedTemporaryFile("w", delete=False, encoding="utf-8", suffix=".lobster") as tmp:
```
DanielYang59, line 73: Unfortunately we cannot use `self.tmp_path` here fixture is injected after setup method

src/pymatgen/io/lobster/outputs.py
```diff
@@ -125,7 +125,7 @@ def __init__(
             else:
                 self._filename = "COHPCAR.lobster"
 
-        lines = _get_lines(filename)
+        lines: list[str] = _get_lines(filename)
```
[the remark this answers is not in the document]
DanielYang59: ah good catch! I fixed this in the ongoing optimizing `Cohpcar` branch but didn't bring the change here, sure could fix this here

tests/io/lobster/test_outputs.py
```diff
@@ -64,15 +65,24 @@ def setup_method(self):
             filename=f"{TEST_DIR}/COOPCAR.lobster.BiSe.gz",
             are_coops=True,
         )
-        self.cohp_fe = Cohpcar(filename=f"{TEST_DIR}/COOPCAR.lobster.gz")
```
DanielYang59, line 67: I just realized why this `Cohpcar` for COHP is reading COOPCAR.lobster? @naik-aakash

JaGeo: Because it is the exact same format 😃

src/pymatgen/io/lobster/outputs.py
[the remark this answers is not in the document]
DanielYang59: Also added a test dca322a04f82a5e2c1847cdcb9dcc01e3e442046

tests/io/lobster/test_outputs.py
DanielYang59: ah okay that explains

naik-aakash:
Hi @DanielYang59 , thanks for this update. Looks good to me.

PR #4454, opened 2025-07-11
Fix unit for `Element` properties: `density_of_solid` and `Molar volume`

### LINKED ISSUE #4453: Units in docstring appear to be wrong for density_in_solid
https://github.com/materialsproject/pymatgen/blame/f9d9fe8e0ce09ef30cc03bcc4e9937d27afd5a6a/src/pymatgen/core/periodic_table.py#L113-L113C80

```python
>>> from pymatgen.core import Element
>>> Element("Ni").density_of_solid
8908.0
```

This is kg/m^3 rather than g/cm^3 as the docstring says. 

https://kg-m3.com/material/nickel#:~:text=Density%20of%20Nickel%20(Ni)%20is,to%2020%C2%B0C%20air)

DanielYang59:
### Summary

- Remove unit in docstring for `Element` properties, to close #4453
- [x] Check other units in docstring


src/pymatgen/core/periodic_table.py
```diff
@@ -110,7 +110,7 @@ def __init__(self, symbol: str) -> None:
             rigidity_modulus (float, optional): Rigidity modulus (GPa).
             mineral_hardness (float, optional): Mohs hardness.
             vickers_hardness (float, optional): Vickers hardness (MPa).
-            density_of_solid (float, optional): Density in solid phase (g/cm³).
+            density_of_solid (float, optional): Density in solid phase (kg/m³).
```
DanielYang59, line 113: I have no strong opinion on this, I originally thought we just remove all units from the docstring if it's a scalar and redirect people to get the unit from  `FloatWithUnit`, but feels like it's not a good idea as most data for ptable wouldn't be updated very frequently, also many units are pretty much "singleton" (e.g. K for temperature):

https://github.com/materialsproject/pymatgen/blob/f9d9fe8e0ce09ef30cc03bcc4e9937d27afd5a6a/src/pymatgen/core/periodic_table.py#L84-L116

PR #4458, opened 2025-07-20
Fix `PBE_64` folder

### LINKED ISSUE #4456: Wrong directory name for the PBE_64 POTCARS?
### Python version

python 3.12

### Pymatgen version

latest

### Operating system version

_No response_

### Current behavior

I installed the pymatgen potcar environment variables through your `pmg config` command following the instruction, and got the directory structure as follows:

POT_GGA_PAW_PBE/
POT_GGA_PAW_PBE_52/
POT_GGA_PAW_PBE_54/
POT_GGA_PAW_PBE_64/
...

However, when I use MPRelaxSet or other VaspInputSet to generate VASP input files setting `user_potcar_functional` as `PBE_64`, I got the following error:

FileNotFoundError: You do not have the right POTCAR with functional='PBE_64' and symbol='Co'
in your PMG_VASP_PSP_DIR='/public5/home/t6s001944/MY_PSP'.
Paths tried:
- /public5/home/t6s001944/MY_PSP/POT_PAW_PBE_64/POTCAR.Co
- /public5/home/t6s001944/MY_PSP/POT_PAW_PBE_64/Co/POTCAR

Obviously, it seems that the setting PBE_64 POTCAR files was set to be installed in the directory named as `POT_PAW_PBE_64`, while the installed directory by the `pmg config` command was `POT_GGA_PAW_PBE_64`. Currently, I simply renamed the `POT_GGA_PAW_PBE_64` folder name as `POT_PAW_PBE_64` to avoid this error, but I am not sure whether this is really simply a bug of wrong directory name and whether the `POT_GGA_PAW_PBE_64` POTCARs are indeed the true PBE_64 functional.

May I get some suggestions on this issue?

Jason

### Expected Behavior

Finding the true PBE_64 POTCARs.

### Minimal example

```Python

```

_No response_

Andrew-S-Rosen:
Closes #4456.

### DIFF
--- a/src/pymatgen/io/vasp/inputs.py
+++ b/src/pymatgen/io/vasp/inputs.py
@@ -65,7 +65,7 @@ POTCAR_FUNCTIONAL_MAP: dict[str, str] = {
     "PBE_52_W_HASH": "POTPAW_PBE_52",
     "PBE_54": "POT_GGA_PAW_PBE_54",
     "PBE_54_W_HASH": "POTPAW_PBE_54",
-    "PBE_64": "POT_PAW_PBE_64",
+    "PBE_64": "POT_GGA_PAW_PBE_64",
     "LDA": "POT_LDA_PAW",
     "LDA_52": "POT_LDA_PAW_52",
     "LDA_52_W_HASH": "POTPAW_LDA_52",


Andrew-S-Rosen:
FYI, letting you know @esoteric-ephemera.

esoteric-ephemera:
Did this not get fixed in [#4431](https://github.com/materialsproject/pymatgen/pull/4431)?

Andrew-S-Rosen:
@esoteric-ephemera: Thanks for the redirect.

Yes, it looks like if the user installs Pymatgen today and runs the CLI entirely from scratch, everything should work due to #4431.

Upon further inspection, it seems like the issue @jinlhr542 and I ran into in #4456 was due to the breaking change. In previous versions of Pymatgen, the CLI would make a folder called `POT_GGA_PAW_PBE_64`. For whatever reason, this is now called `POT_PAW_PBE_64`. So, users who want to use the .64 PAW PBE potentials with Pymatgen or Atomate2 but used the Pymatgen CLI to generate the pseudopotential directory before #4431 was merged will run into compatibility issues.

Andrew-S-Rosen:
@esoteric-ephemera: So I guess the question is --- do we revert the breaking name change, or do we just stick with what we have in `master`?

JaGeo:
(I would, as usual, be happy about avoiding breaking changes. 😅)

esoteric-ephemera:
Either way leads to a breaking change, since a user who set up the directory structure manually and doesn't use the CLI (like me) will have it break the other way

Only non breaking solution would be a check for either directory name

Andrew-S-Rosen:
@esoteric-ephemera: Yes, this is also true.

I am okay with keeping things as is if that's preferred, but I want to document it here in case we get other reports. I will go ahead and close this PR, but anyone can comment if they want it re-opened.

esoteric-ephemera:
Hey sorry I was unclear, I'm happy to add this check for either PBE.64 folder name. It's probably for the best if both are recognized by pymatgen

Andrew-S-Rosen:
@esoteric-ephemera: That works and may reduce the headache in the long-term for MP staff 😉 

Thanks for clarifying and offering to help! Certainly feel free to do so if you feel so inclined!

shyuep:
Looks good to me.

Andrew-S-Rosen:
@shyuep FYI, I wouldn't merge as-is yet. @esoteric-ephemera suggested supporting both names. I'll leave that for others to deal with.

PR #4461, opened 2025-07-22
Handle numpy array for selective dynamics in `Structure`

### LINKED ISSUE #4460: `Error: Type is not JSON serializable` when Exporting POSCAR Structure to JSON
### Python version

Python 3.12.0

### Pymatgen version

2025.6.14

### Operating system version

Mac Os

### Current behavior

Attempting to export a `Structure` object initialized from a POSCAR file containing site_properties["selective_dynamics"] to JSON format using `to(filename="structure.json", fmt="json")` results in a `Error: Type is not JSON serializable: numpy.ndarray`. The error message indicates that `numpy.ndarray` objects within the `site_properties` are not JSON serializable.

### Expected Behavior

The `Structure` object should be successfully serialized to JSON format.

### Minimal example

```Python
import json
import numpy as np

# Create a mock structure with selective_dynamics
structure = Structure(
    lattice=[[10, 0, 0], [0, 10, 0], [0, 0, 10]],
    species=["H", "O"],
    site_properties={"selective_dynamics": [[True, True, True], [False, False, False]]}
)

# Create a POSCAR object from the structure
poscar1 = Poscar(structure)
# write poscar to file
poscar1.write_file("test.poscar")


# read poscar from file
poscar = Poscar.from_file("test.poscar")
# Attempt to export the POSCAR structure to JSON
try:
    poscar.structure.to(filename="structure.json", fmt="json")
except TypeError as e:
    print(f"Error: {e}")
```

test.poscar is created by the minimal example itself.

DanielYang59:
### Summary

- Handle numpy array for selective dynamics in `Structure`, to fix #4460
- [x] Double check consistent behaviour from `json` and `orjson`, also check round trip?
- [x] Unit test

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -325,9 +325,6 @@ reportMissingModuleSource = false
 reportInvalidTypeForm = false
 exclude = ["**/tests"]
 
-[tool.uv]
-default-groups = ["dev", "lint"]
-
 [dependency-groups]
 dev = [
     "invoke>=2.2.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -325,6 +325,9 @@ reportMissingModuleSource = false
 reportInvalidTypeForm = false
 exclude = ["**/tests"]
 
+[tool.uv]
+default-groups = ["dev", "lint"]
+
 [dependency-groups]
 dev = [
     "invoke>=2.2.0",
--- a/src/pymatgen/io/vasp/inputs.py
+++ b/src/pymatgen/io/vasp/inputs.py
@@ -175,7 +175,6 @@ class Poscar(MSONable):
             dim = value.shape
             if dim[1] != 3 or dim[0] != len(self.structure):
                 raise ValueError(f"{name} array must be same length as the structure.")
-            value = value.tolist()
 
         super().__setattr__(name, value)
 
--- a/src/pymatgen/io/vasp/inputs.py
+++ b/src/pymatgen/io/vasp/inputs.py
@@ -160,15 +160,16 @@ class Poscar(MSONable):
         self.structure = structure.copy(site_properties=site_properties)
         if sort_structure:
             self.structure = self.structure.get_sorted_structure()
-        self.true_names = true_names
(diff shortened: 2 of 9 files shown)


src/pymatgen/io/vasp/inputs.py
```diff
@@ -175,6 +175,7 @@ def __setattr__(self, name: str, value: Any) -> None:
             dim = value.shape
             if dim[1] != 3 or dim[0] != len(self.structure):
                 raise ValueError(f"{name} array must be same length as the structure.")
+            # TODO: avoid list -> array -> list just for shape check?
```
DanielYang59, line 178: TODO tag

src/pymatgen/io/vasp/inputs.py
```diff
@@ -700,7 +700,7 @@ def get_str(
 
         return "\n".join(lines) + "\n"
 
-    get_string = get_str
+    get_string = get_str  # TODO: should we add a deprecation deadline?
```
DanielYang59, line 703: should we add a deprecation deadline?

shyuep:
Looks good to me.

pyproject.toml
```diff
@@ -325,6 +325,9 @@ reportMissingModuleSource = false
 reportInvalidTypeForm = false
 exclude = ["**/tests"]
 
+[tool.uv]
+default-groups = ["dev", "lint"]
```
DanielYang59, line 329: [`uv sync` would only install `dev` by default](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups), linting tools then would not be available:
> By default, uv includes the dev dependency group in the environment (e.g., during uv run or uv sync). 
https://github.com/materialsproject/pymatgen/blob/f9d9fe8e0ce09ef30cc03bcc4e9937d27afd5a6a/pyproject.toml#L328-L345

DanielYang59: Better separate this to another PR, looks like out contributing guide also needs update

PR #4479, opened 2025-08-13
Move JDFTx sets from atomate2 to pymatgen

cote3804:
Following a request by @utf in an atomate2 [PR](https://github.com/materialsproject/atomate2/pull/955), I am moving our input set from atomate2 to pymatgen.

This PR adds a `sets.py` file with one new class, `JdftxInputSet`, which does little more than wrap the other JDFTx I/o classes. The only minor change to the original atomate2 code is that the `__init__` method now accepts a pymatgen `Structure` instead of a `JDFTXStructure`. 

A yaml defining the base input set, `BaseJdftxSet.yaml` was also moved from atomate2.

I added two basic tests to check that both loading from an input file and from a `JdftxInfile` are working and that the file written by the input set yields the expected `JdftxInfile`.

### DIFF
--- a/src/pymatgen/io/jdftx/BaseJdftxSet.yaml
+++ b/src/pymatgen/io/jdftx/BaseJdftxSet.yaml
@@ -1,0 +1,67 @@
+# Default JDFTx settings for atomate2 calculations.
+### Functional ###
+elec-ex-corr: gga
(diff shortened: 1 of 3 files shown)


shyuep:
Thanks.

PR #4482, opened 2025-08-16
Simplify and clean up LobsterNeighbors

naik-aakash:
The `LobsterNeighbors` __init__ method is currently messy, with file paths and objects; either of them can be used for initialization. This PR aims to fix this by adding the factory method `from_files` and rewriting the `__init__` method.  

PR will cause a breaking change in the API. The existing init method will be deprecated in January 2026.

# Changes

1. Added a new init method that will be replaced on 31-03-2026
2. Added from_files class method to initialize Lobsterneighbours object
3. Added tests to ensure new init method behaves similar to exisiting implementation
4. Improved error handling for cases where more than 13 neighbors are detected and added tests associated for such cases


naik-aakash:
Hi @shyuep , this PR is ready to be merged if there are no comments.

shyuep:
Thanks.

PR #4484, opened 2025-08-18
Fix jdftx.outputs usage of 3.11+ only syntax

DanielYang59:
## Summary

[Fix the following 3.11+ only syntax](https://docs.python.org/3/reference/expressions.html?utm_source=chatgpt.com#subscriptions):

> Changed in version 3.11: Expressions in an expression list may be starred. See [PEP 646](https://peps.python.org/pep-0646/).

```
E     File "/home/yang/pmg/src/pymatgen/io/jdftx/outputs.py", line 328
E       projections[spins[i]][:, :, *u_to_oa_map[u]] += proj_sjku[i, :, :, u]
E                                   ^
E   SyntaxError: invalid syntax
```

cc @benrich37

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,21 +31,21 @@ jobs:
         # newest version (currently 3.13) on ubuntu (to get complete coverage on unix).
         config:
           - os: windows-latest
-            python: "3.11"
+            python: "3.10"
             resolution: highest
             extras: ci,optional,prototypes
           - os: windows-latest
             python: "3.11"
             resolution: highest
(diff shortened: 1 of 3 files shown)


benrich37:
Sorry about this! I was under the impression when I wrote this that pymatgen was for python 3.12+, thank you for fixing this!

DanielYang59:
no problem at all, https://github.com/materialsproject/pymatgen/blob/12011a16ccc275e894d2eab6795b7d455bea29fa/pyproject.toml#L47-L51

Also noticed we don't test 3.10 and that's the reason this was not caught, maybe we should consider moving one of the windows configs to 3.10

PR #4487, opened 2025-08-25
Fix TranslateSitesTransformation

Rastow:
## Summary

There was a bug in the `TranslateSitesTransformation` class. Both index variables in `apply_transformation` are named identically. The test case did not capture this error. I changed the existing test case to fail when run with the old code and renamed the variables.

## Checklist


shyuep:
Thanks.

PR #4490, opened 2025-09-01
Replace `linear_assignment` with `scipy` `linear_sum_assignment `

DanielYang59:
## 

- Replace `linear_assignment` with [`scipy` `linear_sum_assignment `](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html#linear-sum-assignment)
- [ ] Speed benchmark


shyuep:
Can you confirm that the scipy version actually is faster? The numbers don't seem to bear that out?

DanielYang59:
> Can you confirm that the scipy version actually is faster? The numbers don't seem to bear that out?

As far as I could tell our version performs almost identically as the scipy version (tested both on WSL2 ubuntu 24.04 AMD64 and my MacOS ARM64 machines), this replacement was not really intended for speed but freeing pymatgen from maintaining `linear_assignment` now that it's available from scipy?

PR #4494, opened 2025-09-12
Re-enable mypy `misc` rule and doc/type cleanup

DanielYang59:
## Summary

- Re-enable mypy rule `misc`
- Minor doc, type and test cleanup


src/pymatgen/core/units.py
```diff
@@ -678,20 +689,20 @@ def to(self, new_unit: str | Unit) -> Self:
...
+        return self.to(cast("Unit", self.unit.as_base_units[0]))
 
     @property
-    def supported_units(self) -> dict:
+    def supported_units(self) -> dict[str, dict[str, float]]:
         # TODO abstract base class property?
         """Supported units for specific unit type."""
         if self.unit_type is None:
-            raise RuntimeError("Cannot get supported unit for None.")
+            raise RuntimeError("Cannot get supported unit for known unit_type.")
```
DanielYang59, line 705: emmm a typo here would fix later, should be unknown unit type

PR #4501, opened 2025-09-21
Fix `pre-commit` errors, make `mypy` runnable in `pre-commit` locally, remove `fdint` from optional dependency, doc/test clean up for `matproj`

DanielYang59:
- Minor doc/test cleanup for `matproj`, more descriptive pytest skip message
- `MPRester.request` add default timeout of 60 secs
- [x] Fix `pre-commit` errors after enabling `misc`
- Make `pre-commit`-`mypy` runnable locally
- [**NEED CONFIRMATION**] remove [optional dependency `fdint`](https://github.com/materialsproject/pymatgen/blob/master/pyproject.toml#L94) as it's not installable: https://github.com/scott-maddox/fdint/issues/21


.pre-commit-config.yaml
```diff
@@ -25,6 +25,8 @@ repos:
     rev: v1.17.1
     hooks:
       - id: mypy
+        entry: mypy -p pymatgen
+        pass_filenames: false
```
DanielYang59, line 29: Not sure if there's any better way to resolve way, locally I got:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

src/pymatgen/io/core.py: error: Duplicate module named "core" (also at
"src/pymatgen/core/__init__.py")
src/pymatgen/io/core.py: note: See
https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules
for more info
src/pymatgen/io/core.py: note: Common resolutions include: a) using
--exclude to avoid checking one of them, b) adding __init__.py
somewhere, c) using --explicit-package-bases or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
```

PR #4502, opened 2025-09-22
Fix: indexing error in Procar._read()

finnrk:
## Summary

Major changes:

- Fixes a bug related to indices when skipping repeated k-points while reading `PROCAR` (for example, the double-counting of symmetry points when using a k-path for plotting bandstructure). 

 
Previously, projections, eigenvalues, occupations and weights would be attributed to the wrong k-points. Also, the k-point following the skipped k-point would have complex projections of NaN, as well as an occupation, weight, eigenvalue and squared projections of 0.


shyuep:
Thanks.

PR #4507, opened 2025-09-30
Add references/source to periodic table data

### LINKED ISSUE #4313: More sources/references in periodic table

DanielYang59:
## Summary

- close #4313 
- https://github.com/materialsproject/pymatgen/discussions/4505

Note:
Looks like many data are from https://www.webelements.com/, but this website is per-element (it's how our recording was constructed before recent refactoring), also I didn't find API for it, so a bit hard to automate the validation/update.

--- 

TODO (for a follow up PR): finish TODOs in `elemental_properties.yaml`, many data might need to be updated/validated


shyuep:
Thanks.

PR #4514, opened 2025-10-07
Fix `Ion.as_reduced_dict` doesn't reduce charge

DanielYang59:
## Summary

- Fix `Ion.as_reduced_dict` doesn't reduce charge
- Replace seemingly unnecessary `deepcopy` in `Ion.from_dict`
- Add [missing tests](https://app.codecov.io/gh/materialsproject/pymatgen/blob/master/pymatgen%2Fcore%2Fion.py)

```python
ion = Ion.from_dict({"Mn": 2, "O": 8, "charge": -2})
print(ion.as_reduced_dict()) 
```

```diff
- {'Mn': 1.0, 'O': 4.0, 'charge': -2}
+ {'Mn': 1.0, 'O': 4.0, 'charge': -1.0}
```

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -8,144 +8,33 @@ nav_order: 4
 
 ## v2025.10.7
 
-- PR #4502 Fix: indexing error in Procar._read() by @finnrk
-    Major changes:
-    - Fixes a bug related to indices when skipping repeated k-points while reading `PROCAR` (for example, the double-counting of symmetry points when using a k-path for plotting bandstructure).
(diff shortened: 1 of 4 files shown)


src/pymatgen/core/ion.py
```diff
@@ -279,30 +278,39 @@ def as_dict(self) -> dict[str, float]:
...
 
     @classmethod
-    def from_dict(cls, dct: dict) -> Self:
+    def from_dict(cls, dct: dict[str, float]) -> Self:
         """Generate an ion object from a dict created by as_dict().
 
         Args:
             dct: {symbol: amount} dict.
         """
-        dct_copy = deepcopy(dct)
```
DanielYang59, line 288: I don't think `deepcopy` is needed here? shallow copy should suffice to prevent the original dict being modified in place, test added in 9d9a6033fc1e969e1d181a31cb13c8e479ff28ee

PR #4516, opened 2025-10-09
Python 3.14 support

DanielYang59:
- Build wheels for Python 3.14
- Skip optional dependencies that don't support Python 3.14 yet
- [ ] Waiting for some optional dependencies to support Python 3.14 (at least the actively updated ones)
- [**Need confirm**] remove `[tool.pdm.dev-dependencies]` section from pyproject.toml (seems like the `pdm.lock` was removed in fb6ba71dac3cf6a079aaa2b85c0e4971434ba1ad?)

---

- https://github.com/openai/openai-python/pull/2683
- https://github.com/numba/numba/issues/9957 (also blocked `hiphive`)
- https://github.com/h5py/h5py/issues/2651 (also blocked `phonopy`)
- https://github.com/Unidata/netcdf4-python/issues/1432
- https://pypi.org/project/vtk/#history

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.13.3
+    rev: v0.14.3
     hooks:
       - id: ruff
         args: [--fix, --unsafe-fixes]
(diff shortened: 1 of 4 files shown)


mkhorton:
@DanielYang59 Thank you for this.

I'm comfortable merging given that the only dependencies that do not yet support Python 3.14 are optional and you have cleanly marked these dependencies with `python_version<'3.14'`, so it should not cause any difficulty for people installing pymatgen. These deps can be updated in a subsequent PR. Would you have any objection to this approach?

DanielYang59:
Hi @mkhorton no problem at all.

Sure we could go with this and I would add another PR once `numba` and `vtk` support 3.14

shyuep:
Thanks.

PR #4518, opened 2025-10-11
Enhance INCAR parsing: handle multi-line string and line continuation with backslash, fix handling of `!` as comment marker

### LINKED ISSUE #4517: Support for Multi-Line INCAR Parameters in user_incar_settings
Dear pymatgen developers,

I'm using pymatgen (via atomate2) to construct VASP input sets and have noticed a limitation when trying to include multi-line string INCAR parameters like WANNIER90_WIN.

It seems that the `user_incar_settings` in `MPStaticSet` or related classes) does not currently support multi-line values (an example is given below). I was wondering if there is a recommended way to include such multi-line string parameters in the INCAR file through pymatgen input sets.

I appreciate any thoughts or advice from you. Thanks very much for your time and help.

Best,
Zhenyao

```
WANNIER90_WIN = "
begin projections
Ga: s
As: p
end projections
"
```

DanielYang59:
### Enhancements:
- Allow multi-line strings parsing, close #4517
- Allow line continuation with backslash parsing

### Fixes

- Fix handling of `!` as comment marker for string tags like `PREC = Accurate ! hello`:
  ```python
  print(incar["PREC"])  # >>> Accurate ! hello
  ```

### DIFF
--- a/src/pymatgen/io/vasp/inputs.py
+++ b/src/pymatgen/io/vasp/inputs.py
@@ -961,13 +961,39 @@ class Incar(UserDict, MSONable):
         Returns:
             Incar object
         """
+        string = "\n".join([ln.split("#", 1)[0].split("!", 1)[0].rstrip() for ln in string.splitlines()])
+
         params: dict[str, Any] = {}
-        for line in clean_lines(string.splitlines()):
-            for sline in line.split(";"):
-                if match := re.match(r"(\w+)\s*=\s*(.*)", sline.strip()):
-                    key: str = match[1].strip()
-                    val: str = match[2].strip()
-                    params[key] = cls.proc_val(key, val)
+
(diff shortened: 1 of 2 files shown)


DanielYang59:
@esoteric-ephemera can you please review this maybe? I understand using RE would hurt readability but I didn't find a better way other than using a quite complex state machine, also added unit test to cover as much edge cases as I could think of.

esoteric-ephemera:
@DanielYang59  sure will try to find some time to!

shyuep:
I wish VASP would switch to YAML for INCARs.... I think use of regex is fine. We use it all over Vasp.outputs.

shyuep:
Thanks!

DanielYang59:
no problem

PR #4520, opened 2025-10-16
Avoid dict mutation in VaspInputSet.structure setter

janosh:
`_config_dict["POTCAR"]` was being modified in-place during `Yb` PSP warning check
this caused `user_potcar_settings` to be permanently merged into  `_config_dict["POTCAR"]`. no known downstream repercussions but could in principle lead to incorrect POTCAR settings in subsequent operations


shyuep:
Thanks.

PR #4521, opened 2025-10-16
Add `ISEARCH` parameter to `incar_parameters.json`

kavanase:
## Summary

VASP 6.5 introduces the `ISEARCH` parameter: https://www.vasp.at/wiki/index.php/ISEARCH

This PR add `ISEARCH` to `incar_parameters.json`, so that it is recognised as a valid INCAR flag.

### DIFF
--- a/src/pymatgen/io/vasp/incar_parameters.json
+++ b/src/pymatgen/io/vasp/incar_parameters.json
@@ -402,6 +402,13 @@
       4
      ]
  },
+ "ISEARCH": {
+   "type": "int",
+   "values": [
+      0,
+      1
+     ]
+ },
  "ISIF": {
    "type": "int",
    "values": [


DanielYang59:
~~Also I guess `int_keys` should be updated otherwise it would be treated as string?~~
https://github.com/materialsproject/pymatgen/blob/821893487db356bc4d2b5414de592c2abc32a261/src/pymatgen/io/vasp/inputs.py#L1019

Never mind, I guess it would be correctly inferred as `int` because `int(val)` is the first try for unknown value: https://github.com/materialsproject/pymatgen/blob/821893487db356bc4d2b5414de592c2abc32a261/src/pymatgen/io/vasp/inputs.py#L1087-L1091

Guess those recording in `proc_val` could be removed and reuse `incar_parameters.json`... I would add this to my TODO list... (not sure if I missed any reason we need both)

shyuep:
Thanks!

PR #4526, opened 2025-10-20
`VolumetricData.copy` preserve subclass type

### LINKED ISSUE #4523: Hard-coded type of the returned object of `VolumetricData.copy` breaks polymorphism
Currently the returned object of `VolumetricData.copy` is hard-coded to return `VolumetricData`, as a result any subclass.copy would also return `VolumetricData`, including `Locpot`, `Chgcar` and `Elfcar`

https://github.com/materialsproject/pymatgen/blob/bd29947130c497af7e25888bed7016bf898bed2c/src/pymatgen/io/common.py#L144-L151

DanielYang59:
## Summary

- `VolumetricData.copy` use dynamic type to preserve subclass type, fix #4523
- loosen `linear_slice` type requirement

### DIFF
--- a/src/pymatgen/io/common.py
+++ b/src/pymatgen/io/common.py
@@ -23,9 +23,7 @@ from pymatgen.core.units import ang_to_bohr, bohr_to_angstrom
 from pymatgen.electronic_structure.core import Spin
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


shyuep:
Thanks.

DanielYang59:
no problem ;)

PR #3820, opened 2024-05-11
Improve type annotations and comments for `io.cif`

DanielYang59:
## Summary

- Add type annotations and tweak comments for `io.cif`
- Minor code format tweak
- Simplify "check to see if" to "check if" in comments
- Remove `if __name__ == "__main__"` from `core.libxcfunc` (legacy from another PR)

### Potentially Breaking Changes (very unlikely)
1. Make `parse_oxi_states` and `parse_magmoms` of `CifParser` private methods. **Rationale**:
    - These two parse methods are used exclusively for the private `_get_structure` method to parse corresponding properties from the file.

2. Remove unnecessary `lattice` argument from `parse_magmoms` method (now private). **Rationale**:
     - This argument is only used to check if `Lattice is None`, and could be done outside this method. 
     https://github.com/materialsproject/pymatgen/blob/2e1c3016742d930cbdc5f4129e4036819e7755d3/pymatgen/io/cif.py#L850-L869

PR #4531, opened 2025-10-30
Deprecate `CifParser` init support of `filename` as `StringIO`

DanielYang59:
### Summary

- Move CIF parsing logics from `__init__` of `CifParser` to `from_str` 
- ~~Revert accidentally removed filename as `TextIOWrapper` support (though this may not be intended) in https://github.com/materialsproject/pymatgen/pull/3820, add a deprecation warning~~ reverted, guess this was not intended, also doesn't agree with type/docstring
- Currently the `filename` of `CifParser` constructor could also be the string for the CIF file itself (`StringIO`), it's a bit confusing (cc @ka-sarthak) but I guess this was intended to support `from_str`. After this refactoring this could be deprecated I guess

### DIFF
--- a/src/pymatgen/io/cif.py
+++ b/src/pymatgen/io/cif.py
@@ -10,7 +10,7 @@ import warnings
 from collections import defaultdict, deque
 from functools import partial
 from inspect import getfullargspec
-from io import StringIO, TextIOWrapper
+from io import StringIO
 from itertools import groupby
 from pathlib import Path
 from typing import TYPE_CHECKING, Literal, cast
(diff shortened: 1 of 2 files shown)


ka-sarthak:
Thanks, @DanielYang59, for taking this up. My initial confusion was whether using a file handler is supported in the `filename` parameter, like:
```py
cif = 'my_path.cif'
with open(cif) as file:
    parser = CifParser(filename=file) 

# output: "TypeError: Unsupported file format."
```
I think this used to work at some point, even though it might not have been intended. It's more of my negligence to misinterpret the interface.

I like that you make the error statement explicit now, as `Unsupported file format` alone is confusing, given that the file format (.cif) is correct.

PR #4539, opened 2025-11-07
Breaking: Make `charge_spin_check` a `UserWarning` instead of `ValueError`

Andrew-S-Rosen:
Currently, when a `Molecule` object is instantiated, by default it checks to see if the charge and spin multiplicity are compatible. If it is not, then it raises a `ValueError`. I argue that it is better for Pymatgen to raise a `UserWarning` instead.

With the advent of MLIPs, people are running calculations on all kinds of molecules, like radicals and transition metal complexes. For the MLIPs that are not spin aware, there is no concept of charge or spin. However, by not setting any, Pymatgen crashes. There is a keyword argument `charge_spin_check` that can be set to `False`, but issues still remain even with this; if someone writes out a JSON with a `Molecule` object from an MLIP run, `monty` will try to reserialize the `Molecule` object and crash by default. This is not very user friendly.

This PR is a breaking change since it changes a `ValueError` to a `UserWarning`.

### DIFF
--- a/tests/core/test_structure.py
+++ b/tests/core/test_structure.py
@@ -2319,8 +2319,8 @@ Site: H (-0.5134, 0.8892, -0.3630)"""
         assert self.mol.spin_multiplicity == 1
         assert self.mol.nelectrons == 10
         assert_allclose(self.mol.center_of_mass, [0, 0, 0], atol=1e-7)
-        with pytest.raises(
-            ValueError,
+        with pytest.warns(
+            UserWarning,
             match="Charge of 1 and spin multiplicity of 1 is not possible for this molecule",
         ):
             Molecule(
@@ -2348,8 +2348,8 @@ Site: H (-0.5134, 0.8892, -0.3630)"""
             [1.026719, 0, -0.363000],
             [-0.513360, -0.889165, -0.363000],
         ]
-        with pytest.raises(
-            ValueError,
+        with pytest.warns(
+            UserWarning,
             match="Charge of 0 and spin multiplicity of 1 is not possible for this molecule",
         ):
             mol = IMolecule(["C", "H", "H", "H"], coords, charge=0, spin_multiplicity=1)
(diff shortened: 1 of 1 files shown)


shyuep:
Hmm... wouldn't it be easier to set charge_spin_check to False?

Andrew-S-Rosen:
Yes, there is a keyword argument `charge_spin_check` that can be set to `False`, although issues can still remain even with this. If someone writes out a JSON with a `Molecule` object from an MLIP run, `monty` will try to reserialize the `Molecule` object and crash by default. I'm open to alternate suggestions about that kind of scenario. Should we shift the burden to the code developer/user instead?

shyuep:
I am trying to see how that can happen. In the code, there is a pre-condition `if spin_multiplicity:`. I would think an MLIP run would set spin_multiplicity to None or at worst, 0. That would mean this check is skipped entirely. I wouldn't ahve an objection to do `charge_spin_check = False` as a default. I do think a UserWarning is not good enough. When people specify `charge_spin_check = True`, they would expect a terminal failure for non-conforming molecules and not that they have to catch a warning to deal with it.

@samblau Mr OMol, opinions?

Andrew-S-Rosen:
> I do think a UserWarning is not good enough. When people specify charge_spin_check = True, they would expect a terminal failure for non-conforming molecules and not that they have to catch a warning to deal with it.

I agree with this upon reflection.

>  I wouldn't ahve an objection to do charge_spin_check = False as a default.

I think this would be preferred, although it is a breaking change as well. I am interested to know if @samblau or others would prefer this to be `False` by default or not.

> I am trying to see how that can happen. In the code, there is a pre-condition if spin_multiplicity:. I would think an MLIP run would set spin_multiplicity to None or at worst, 0. That would mean this check is skipped entirely. 

Good point. This original issue for this is arising from other people doing hacky things. I think I am trying to patch Pymatgen to solve other people's problems, so that suggests to me this is not the way to go. 

I am going to close this PR but would still be interested in @samblau's opinion on the default kwarg. (For context, this PR emerged from a debugging discussion with @samblau anyway).

PR #4542, opened 2025-11-11
Fix types for entry processing in phase diagram and compatibility

janosh:
- Updated type hints in `PatchedPhaseDiagram` to use `Entry` instead of `PDEntry`.
- Introduced `TypeVarAnyEntry` for correct `Entry` subclass inference on `Compatibility.process_entries` return types

these changes resolve type errors that started showing up in `ty 0.0.1-alpha.26 (b225fd8b4 2025-11-10)` over at https://github.com/janosh/matbench-discovery


shyuep:
Thanks.

PR #4545, opened 2025-11-15
Chargemol fixes for VASP POTCAR 64

esoteric-ephemera:
DDEC6 / chargemol is incompatible with PBE / LDA 64 POTCARs because of the SHA and COPYR tags. Adds some workarounds to:
1. Rewrite the POTCAR without these tags to a temp file if needed
2. Not store the full POTCAR in memory

@rul048 @kenko911 can you take a look at if this looks OK to you?


kenko911:
Hi @esoteric-ephemera, it looks good to me. Thanks!

PR #4547, opened 2025-11-18
Fix `surface.center_slab` handling of Slab regions separated by vacuum 

DanielYang59:
Fix `surface.center_slab` handling of Slab regions separated by vacuum, close #4543

- [ ] `get_slab_regions` currently cannot handle slab outside cell
- [ ] unit test

PR #4548, opened 2025-11-18
Relocate `surface.get_slab_regions/center_slab` as static methods for `Slab`, mark `SlabGenerator.get_unreconstructed_slabs` as private method

DanielYang59:
### Summary

- Relocate `surface.get_slab_regions/center_slab` as static methods for Slab, part of #4547, rationale: these methods are quite specific to `Slab`, make them methods for `Slab` for better grouping
- Mark SlabGenerator.get_unreconstructed_slabs as private method #4548

### Future TODO

- [ ] It's better if they're just instance methods, however they wouldn't work now if `slab` is `Structure`


DanielYang59:
guess it's better to refactor to instance method at some point and avoid such intermeditate relocations

PR #4553, opened 2025-11-23
Fix `Slab.get_slab_regions` handling of slabs outside cell, and periodic boundary condition

DanielYang59:
Fix `Slab.get_slab_regions` handling of slabs outside cell, and periodic boundary condition


DanielYang59:
closing and continue from the pymatgen-core repo

PR #4556, opened 2025-11-28
Improve type hints in `io.ase`

tomdemeyere:
1. Added overloads to get the correct `Atoms` or `MSONAtoms` type depending on the `msonable` parameter in `get_atoms`
2. Added a generic in `get_structure` to get back the correct type.
3. Changed the return type of `get_molecule` to use the generic present in the parameters.

No runtime changes at all.

@Andrew-S-Rosen @shyuep

### DIFF
--- a/src/pymatgen/io/ase.py
+++ b/src/pymatgen/io/ase.py
@@ -51,6 +51,7 @@ __date__ = "Mar 8, 2012"
 IMoleculeT = TypeVar("IMoleculeT", bound=IMolecule)
 StructOrMolT = TypeVar("StructOrMolT", bound=Structure | Molecule)
 
+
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
(Fixing these type hints seems wise. Thanks!)

shyuep:
Thanks,

shyuep:
There seems to be mypy errors resulting from these changes. Can you submit a separate fix in a PR pls? Thanks.

tomdemeyere:
I don't understand why mypy is being so defensive here? `StructOrMolT` is bound to `Structure` why can't this be used as default? pyright does not complain?

The only reasonable fix I see is a bunch of `type: ignore`; the type hinting is working as expected with pylance/pyright

PR #4567, opened 2025-12-10
Symmetry not longer necessary to have first in geometry directive.

francescalb:
## Summary

Major changes:

Corrected reading of nwchem input string/file. In the geometry directive it should be allowed to add symmetry both at the beginning and at the end.

## Todos

This is just a small addition to the NWChem parser which is by no means complete yet.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
Thanks!

PR #4571, opened 2025-12-15
Fix `Structure.__mul__/make_supercell` ignore `to_unit_cell`

DanielYang59:
## Summary

- Fix `Structure.__mul__/make_supercell` ignore `to_unit_cell`, fix https://github.com/materialsproject/pymatgen/discussions/4562

I guess `__mul__` should use `to_unit_cell=False` by default? https://github.com/materialsproject/pymatgen/blob/7b92f9ab4112d538381f4ee4dd6119295c200245/src/pymatgen/core/structure.py#L1125-L1126


shyuep:
I agree we shouldn't map to unit cell before making a supercell. Maybe add some doc to inform people this is the case?

DanielYang59:
Move to new core repo

PR #4577, opened 2026-01-05
Toggling inversion for lattice mappings in StructureMatcher

mmartirossyan:
Added an argument `allow_inversion` to `StructureMatcher` which toggles the use of inversions for finding lattice mappings inside `_get_lattices`. The default for the `allow_inversion` argument is set to True, which leaves the current `StructureMatcher` output unchanged. This argument is useful if the two structures being compared are the same up to inversion, i.e. enantiomorphic pairs of opposite handedness.

If `allow_inversion` is True, then finding the rotation matrix is skipped, and `rmat` values will be None, however the logic inside `(self._allow_inversion or np.linalg.det(rmat) > 0)` (line 501) will always result in True. If `allow_inversion` is False, then the rotation matrix will be required to have a positive determinant for that same logic to be True, therefore lattice mappings which include inversion will not be allowed.


shyuep:
Thanks. Can you add a unittest for this pls?

shyuep:
Thanks. We have refactored pymatgen to move core packages to a separate [pymatgen-core repo](https://github.com/materialsproject/pymatgen-core). If you don't mind, please submit a new PR in that repo. Sorry about the inconvenience.

mmartirossyan:
Hi, sorry for the slowness in replying! I will make a new PR in the new repo and add a test as well in the coming weeks.

PR #4579, opened 2026-01-20
Allow for alchemical pseudoatoms in Abinit

GkAntonius:
Remove the check on the length of znucl in Abinit,
because it may be different than the number of atom types when using alchemical pseudopotentials.


shyuep:
Thanks.

PR #4581, opened 2026-02-04
Add yaml for smoothPES correction

CompRhys:
Adds minimal addition to incorporate https://arxiv.org/abs/2601.21056 into pmg.


CompRhys:
@flt17 @ThomasWarford

Please check that there's nothing wrong with this PR. I took the values from the table in SI an assumed that you omit all oxide and anion corrections hence why they've been set to zero.

ThomasWarford:
This all looks good to me! Thanks for putting this in

shyuep:
Thanks!

PR #4589, opened 2026-02-16
Pass `process_entries` arguments to `solid_compat` in `MaterialsProjectAqueousCompatibility`

lkkmpn:
## Summary

The `MaterialsProjectAqueousCompatibility` class has a `solid_compat` argument for preprocessing, but currently only a subset of arguments to the aqueous compatibility's `process_entries` method are passed on to that of the `solid_compat`. This PR changes this so that all arguments are passed on, for more clear behavior. For instance, currently, if a user sets `on_error='raise'` on the aqueous compatibility object, the solid entries compatibility object would still ignore all errors, which can be confusing.


shyuep:
Pls add a unit test for the new behavior.

lkkmpn:
I have added a unit test focusing on the `on_error` argument.

shyuep:
Thanks!

PR #4602, opened 2026-03-03
Add `doped` and `ShakeNBreak` to external tools list

kavanase:
Just noticed the external tools list. This tiny PR adds `doped` and `ShakeNBreak` to the list.


shyuep:
Thanks.

PR #4615, opened 2026-03-23
fix: correctly clear energy_adjustments in clean mode (avoid skipped removals)

irishmango:
## Summary
Fixes `clean=True` handling in the mixing scheme logic so all existing `energy_adjustments` are cleared correctly.

## Problem
The previous implementation removed items from `entry.energy_adjustments` while iterating over the same list. When multiple adjustments were present, this caused some entries to be skipped, leaving stale adjustments behind.

This meant `clean=True` could silently fail to fully reset entries before reprocessing.

## Fix
- Replace mutation-during-iteration with direct list reset via `entry.energy_adjustments = []`
- Apply the fix in both `mixing_scheme` modules
- Keep behavior aligned with the existing pattern already used in `Compatibility._process_entry_inplace`

## Tests
- Verified existing `test_clean` still passes
- Added a regression test covering entries with multiple energy adjustments
- Confirmed all pre-existing failures remain unrelated to this change

### DIFF
--- a/src/pymatgen/analysis/compatibility/mixing_scheme.py
+++ b/src/pymatgen/analysis/compatibility/mixing_scheme.py
@@ -175,8 +175,7 @@ class MaterialsProjectDFTMixingScheme(Compatibility):
         # any corrections added by compat_1 or compat_2.
         if clean:
             for entry in entries:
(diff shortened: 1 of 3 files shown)


shyuep:
Thanks.

PR #4625, opened 2026-04-03
Add an partial update method to PPDs 

CompRhys:
Another convenience tool for PPDs where we can just update the relevant patches in the atlas rather than having to recompute everything.


shyuep:
Thanks,

PR #4626, opened 2026-04-06
Move sphinx generation to CI and remove .html files and sphinx artifa…

cogsworth37:
…cts from the repo.

closes #4056 

## Summary

Major changes:

- Move Sphinx doc generation to CI
- Remove HTML files from the repo

PR #4631, opened 2026-04-10
Add preparation step for Jekyll deployment and include static files

cogsworth37:
## Summary

Major changes:

Fixes broken styles/JS introduced by #4626 — the CI migration didn't account for Sphinx asset handling.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


.github/workflows/jekyll-gh-pages.yml
DanielYang59, line 58: Great! thanks for fixing this. I have no experience with sphinx/jekyll, but I think the copy step could just go here for simplicity?

DanielYang59: pymatgen [doesn't have `docs` optional dependency](https://github.com/materialsproject/pymatgen/actions/runs/24220563842/job/70710928777)

> WARNING: pymatgen 2026.3.23 does not provide the extra 'docs'

 it's currently declared via uv dependency group `docs`

- https://github.com/materialsproject/pymatgen/blob/master/pyproject.toml#L287

these two steps (pip install pymatgen and then sphinx) could just be `uv sync --group docs`

DanielYang59: can you also help me remove `docs` from default dependency group (and that TODO)? I believe it's not needed anymore

https://github.com/materialsproject/pymatgen/blob/f475a898cd41785a405171cd1dd1fc895f868d7c/pyproject.toml#L283-L285

DanielYang59: these two sphinx commands should be called via uv instead, i.e. `uv run sphinx-xxx ...`

.github/workflows/jekyll-gh-pages.yml
```diff
@@ -59,6 +59,12 @@ jobs:
       - name: Setup Pages
         uses: actions/configure-pages@v5
 
+      - name: Prepare docs for Jekyll
+        working-directory: docs
+        run: |
+          cp -r html/_static _static
+          touch .nojekyll
```
DanielYang59: Curious what does `.nojekyll` do (as i don't have experience with jekyll at all)? I see [here](https://github.blog/news-insights/bypassing-jekyll-on-github-pages/):
> It is now possible to completely bypass Jekyll processing on [GitHub Pages](http://pages.github.com/) by creating a file named .nojekyll in the root of your pages repo and pushing it to GitHub. 


But I think in the `_config.yml` there is already an explicit `include: _static` tag? does these two serve different purposes?

> This should only be necessary if your site uses files or directories that start with underscores since Jekyll considers these to be special resources and does not copy them to the final site.

.github/workflows/jekyll-gh-pages.yml
cogsworth37: I did it this way to create a clear boundary. the `cp -r` command must be run after the sphinx-build command. while I could put it at the end of the block, I wanted to create no confusion if the order gets messed up.

.github/workflows/jekyll-gh-pages.yml
cogsworth37: It does not. it's an extra extra guard. I usually like to add protections in multiple places as long as it's functionally consistent. just a quirk of mine, happy to remove if you don't feel it's necessary.

.github/workflows/jekyll-gh-pages.yml
cogsworth37: Updated to use uv sync --group docs and removed docs from the default groups

cogsworth37: Updated to use uv commands here

DanielYang59: I think both HTML and static are built by sphinx, looks quite confusing to me to copy HTML first and then in another step copy static?

DanielYang59: great thanks

cogsworth37: That is a fair point. would you prefer a collapse into that block or all rm/mv/cp commands be moved to a more operational block?

DanielYang59: maybe let's combine them? making this workflow shorter would make later reading/debug a bit easier (personally)...

cogsworth37: It's moved

DanielYang59: thanks

.github/workflows/jekyll-gh-pages.yml
DanielYang59: I guess we're not talking about security where we need layers of defense here... personally I think if one config gets the job done then we might not need to duplicate here (less confusion for later readers... people like me would think they're serving different purposes)

DanielYang59:
Great style seems to be working now!

image

DanielYang59:
I noticed the links to sources are not working (for both core and non-core modules), not sure if you would have time for a look? @cogsworth37 

For example for [non-core `ext`](https://pymatgen.org/pymatgen.ext.html#module-pymatgen.ext.optimade):

```diff
# Current URL generated in docs
- https://github.com/materialsproject/src/pymatgen/ext/optimade.py

# Correct URL
+ https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/ext/optimade.py
```
image

---

For core [`Ion`](https://pymatgen.org/pymatgen.core.html#module-pymatgen.core.ion) it points to the wrong repo:
```diff
# Current URL generated in docs
- https://github.com/materialsproject/pymatgen/blob/v2026.4.7/src/pymatgen/core/ion.py

# Correct URL
+ https://github.com/materialsproject/pymatgen-core/blob/main/src/pymatgen/core/ion.py
```

image

shyuep:
The sources are in different locations. The core classes are in the pymatgen-core repo. Actually, we can also just use the submodule since it is already linked.

PR #4617, opened 2026-03-24
Avoid and fix Deprecation Warnings

kavanase:
## Summary

This PR:
- Fixes some incorrectly swapped deprecation warnings (for ``entry_tools`` and ``correction_calculator``).
- Updates usages to match the new sub-package structure, avoiding deprecation warnings (caused by the internal code structure).

I would note that this PR update is likely not exhaustive; there are likely other usages of deprecated code in the ``pymatgen`` codebase following the big restructuring with ``pymatgen-core``. It would be helpful to search through the ``DeprecationWarning``s added in https://github.com/materialsproject/pymatgen/pull/4595 and check that the rest of the codebase has been updated accordingly. @DanielYang59 perhaps you could check this as you have a good understanding of the current code structure?
I imagine some code in ``pymatgen-core`` likely needs to be updated too, in a similar fashion.

There is also some issues with ``MPRester`` related to the code structure changes and deprecations, I will raise an issue for that now.

PR #4633, opened 2026-04-14
Commit to preserving backwards compatible imports

esoteric-ephemera:
Removes deprecation warnings for import shims for critical, base infrastructure in pymatgen. Commits to preserving backwards compatible, if aesthetically unappealing, imports

These classes/functionalities have existed under the same imports for many years at this point, removing those imports would constitute a severely breaking change

This would also lead to a huge impact on database users, since their monty-deocded classes will break without them completely migrating collections


esoteric-ephemera:
Also noting that this would supersede (parts of) [#4617](https://github.com/materialsproject/pymatgen/pull/4617) by preserving these legacy imports

PR #4634, opened 2026-04-16
Possible bug in chemenv

esoteric-ephemera:
Seems like the `StructureEnvironments.ce_list` has changed structure and requires slightly different logic to handle cases where coord envs weren't computed

### DIFF
--- a/src/pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
+++ b/src/pymatgen/analysis/chemenv/coordination_environments/chemenv_strategies.py
@@ -710,7 +710,8 @@ class SimplestChemenvStrategy(AbstractChemenvStrategy):
         if cn_map is None:
             return None
 
-        ce = se.ce_list[se.sites_map[isite]][cn_map[0]][cn_map[1]]
+        ce_sub = se.ce_list[se.sites_map[isite]].get(cn_map[0]) or []
+        ce = ce_sub[cn_map[1]] if len(ce_sub) > cn_map[1] else None
         if ce is None:
             return None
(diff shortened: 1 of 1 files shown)


esoteric-ephemera:
Not fully sure if this is the root of the issue or a bandaid. @lutfia95 I see you're the last to have done a significant refactor on chemenv. Any ideas why `ce_list` would have changed structurally?

lutfia95:
Hi @esoteric-ephemera 
I think the practical difference is that the old code assumes `se.ce_list[se.sites_map[isite]][cn_map[0]][cn_map[1]]` is always present, while the new version explicitly handles cases where no coordination environments were computed for that `cn_map[0]`.

So this:

```python
ce = se.ce_list[se.sites_map[isite]][cn_map[0]][cn_map[1]]
```
will raise if the entry is missing, whereas this:

```python
ce_sub = se.ce_list[se.sites_map[isite]].get(cn_map[0]) or []
ce = ce_sub[cn_map[1]] if len(ce_sub) > cn_map[1] else None
```

treats a missing / empty result as “no CE available” and returns `None`.
So this looks reasonable if `ce_list `can now legitimately omit entries when coordination environments were not computed. In that case, the old indexing logic was too strict for the current data model.

Thanks! the fix looks fine to me.

esoteric-ephemera:
Thanks! Then this should be ready to merge

JaGeo:
@shyuep I will merge this then as all tests pass and this is part of ChemEnv.

esoteric-ephemera:
Thanks @JaGeo!

PR #4635, opened 2026-04-16
Publish Sphinx objects.inv with GitHub Pages docs

elinscott:
## Summary
The docs build already generates `html/objects.inv`, but the GitHub Pages workflow currently only moves the HTML pages, `modules.html`, and `_static` into the published docs tree. This two-line PR also publishes `objects.inv` so downstream projects can use pymatgen with `intersphinx`.


shyuep:
Thanks.

elinscott:
No problem! Thanks for merging :slightly_smiling_face:

PR #4648, opened 2026-04-25
Speed up pmg CLI startup with lazy imports

hheei:
## Summary

This PR reduces `pmg` startup overhead by deferring heavy imports until the selected subcommand is actually executed.

The main change is to lazily import CLI handlers and move several expensive scientific imports from module import time to runtime paths. The goal is to improve responsiveness for `pmg --help`, command discovery, and general CLI invocation, without changing command behavior.

## Changes

- lazily import subcommand handlers from `pmg.py`
- defer heavy imports in:
  - `pmg_plot.py`
  - `pmg_analyze.py`
  - `pmg_structure.py`
  - `pmg_potcar.py`
- resolve some command-specific dependencies only when the relevant command path is used

## Benchmarks

Local measurements showed substantial improvement in cold startup/import cost:

- `pmg` cold import:
  - before: ~0.97 s
  - after: ~0.055 s mean over 8 runs

- `pymatgen.cli.pmg_plot` import:
  - before: ~2.13 s
  - after: ~0.56 s

A few subcommand cold-run samples also improved, for example:

- `pmg plot --chgint ...`
  - before: ~2.214 s
  - after: ~1.824 s

- `pmg structure --symmetry ...`
  - before: ~0.408 s
  - after: ~0.314 s

These are local numbers, but they suggest that most of the previous cost came from eager imports.

## Validation

I ran:

- `uv run pytest tests/cli/test_pmg.py tests/cli/test_pmg_plot.py tests/cli/test_pmg_structure.py tests/cli/test_pmg_analyze.py`
- `uv run pytest tests/cli/test_pmg_plot.py`
- `uv run pytest tests/cli/test_pmg_analyze.py tests/cli/test_pmg_structure.py`

All relevant CLI tests passed locally.

## Question For Maintainers: future direction of `pmg`

This PR is intentionally limited to startup performance, but while working on it I noticed a broader CLI maintenance question and wanted to ask for maintainer guidance before exploring anything further.

At the moment, `pmg` is still quite useful and seems to have room for additional functionality, but the current `argparse` layout also means that:
- top-level parser construction in `pmg.py` keeps growing
- adding new commands/options can become verbose
- shell completion and help formatting are somewhat limited
- some workflows that could grow further, especially I/O / extraction / conversion tasks, do not fit especially cleanly into the current layout

One concrete example I am interested in is adding CLI support for extracting data such as `LOCPOT` from `vaspwave.h5`, with output to VASP volumetric format, XSF, or optionally compressed output.

Before attempting any larger CLI change, I wanted to ask:

- Is `pmg` expected to remain a lightweight convenience CLI only?
- Or is there interest in continuing to expand it as a more structured interface for common workflows?
- If expansion is welcome, would maintainers be open to evaluating a separate `Typer` prototype with feature parity to the current CLI, and comparing:
  - startup cost
  - maintainability
  - shell completion
  - help output
- If such a prototype looked reasonable, would a gradual migration path with compatibility aliases be acceptable?

I am not proposing to mix any of that into this PR. I only wanted to ask whether that direction would be considered in scope before doing exploratory work.

Thanks for taking a look.

### DIFF
--- a/src/pymatgen/cli/pmg.py
+++ b/src/pymatgen/cli/pmg.py
@@ -6,23 +6,25 @@ from __future__ import annotations
 
 import argparse
 import itertools
+from importlib import import_module
 from typing import TYPE_CHECKING
 
 from tabulate import tabulate, tabulate_formats
-
-from pymatgen.cli.pmg_analyze import analyze
-from pymatgen.cli.pmg_config import configure_pmg
-from pymatgen.cli.pmg_plot import plot
-from pymatgen.cli.pmg_potcar import generate_potcar
-from pymatgen.cli.pmg_structure import analyze_structures
-from pymatgen.core import SETTINGS
-from pymatgen.core.structure import Structure
-from pymatgen.io.vasp import Incar, Potcar
 
 if TYPE_CHECKING:
     from argparse import Namespace
     from collections.abc import Sequence
     from typing import Any
+
+
+def _lazy_func(module_name: str, func_name: str):
+    """Lazily import a CLI handler the first time it is invoked."""
+
+    def _runner(args: Namespace):
+        func = getattr(import_module(module_name), func_name)
+        return func(args)
+
+    return _runner
 
 
 def parse_view(args: Namespace) -> int:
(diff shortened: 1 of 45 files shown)


shyuep:
Thanks.

PR #4661, opened 2026-05-28
vectorise get_pattern in xrd module

### LINKED ISSUE #4660: Consider further vectorising XRD get_pattern method across hkl
I work on the materialscloud platform and am fairly interested in computing XRD patterns for a wide variety of materials (around 100k).

The xrd get_pattern module does some np.array calculations however the hkl loop often by far dominates the total calculation time (see table). 

```
hkl loop                         0.017381 s
build output                     0.002513 s
reci lattice search              0.001067 s
lattice filtering                0.000125 s
build site arrays                0.000123 s
symmetry ref                     0.000003 s
```

I think it should be possible to vectorise across the hkl planes and provide a significant speedup. My limited testing seems that following snippet should be 1:1 in function but significantly improve calculation time (3-4x speedup).  This ofcourse comes at some cost of readability but worth discussion 
```
hkl loop                          0.000598 s
build output                      0.002347 s
reci lattice search               0.001087 s
lattice filtering                 0.000153 s
build site arrays                 0.000107 s
symmetry ref                      0.000005 s
peak grouping                     0.000672 s # Peaks are grouped using a bin rather than in loop.
```

``` py
    def get_pattern(self, structure: Structure, scaled=True, two_theta_range=(0, 90)):
        if self.symprec:
            finder = SpacegroupAnalyzer(structure, symprec=self.symprec)
            structure = finder.get_refined_structure()
        

        wavelength = self.wavelength
        latt = structure.lattice
        is_hex = latt.is_hexagonal()

        min_r, max_r = (
            (0, 2 / wavelength)
            if two_theta_range is None
            else [2 * sin(radians(t / 2)) / wavelength for t in two_theta_range]
        )



        recip_latt = latt.reciprocal_lattice_crystallographic
        recip_pts = recip_latt.get_points_in_sphere([[0, 0, 0]], [0, 0, 0], max_r)
        if min_r:
            recip_pts = [pt for pt in recip_pts if pt[1] >= min_r]


        # --- Build per-site arrays ---
        _zs, _coeffs, _fcoords, _occus, _dwfactors = [], [], [], [], []
        for site in structure:
            for sp, occu in site.species.items():
                _zs.append(sp.Z)
                try:
                    c = ATOMIC_SCATTERING_PARAMS[sp.symbol]
                except KeyError:
                    raise ValueError(
                        f"Unable to calculate XRD pattern as there is no scattering "
                        f"coefficients for {sp.symbol}."
                    )
                _coeffs.append(c)
                _dwfactors.append(self.debye_waller_factors.get(sp.symbol, 0))
                _fcoords.append(site.frac_coords)
                _occus.append(occu)

        zs       = np.array(_zs)            # (N,)
        coeffs   = np.array(_coeffs)        # (N, 4, 2)
        fcoords  = np.array(_fcoords)       # (N, 3)
        occus    = np.array(_occus)         # (N,)
        dwfactors = np.array(_dwfactors)    # (N,)


        # --- Unpack reciprocal points & filter g_hkl == 0 ---
        recip_pts_sorted = sorted(recip_pts, key=lambda i: (i[1], -i[0][0], -i[0][1], -i[0][2]))

        hkls_raw = np.array([pt[0] for pt in recip_pts_sorted])   # (M, 3)
        g_hkls   = np.array([pt[1] for pt in recip_pts_sorted])   # (M,)

        nonzero  = g_hkls != 0
        hkls_raw = hkls_raw[nonzero]
        g_hkls   = g_hkls[nonzero]

        hkls_int = np.round(hkls_raw).astype(int)           # (M, 3)

        # --- Fully vectorized computation over all M hkl points ---
        # shapes: (M,)
        theta    = np.arcsin(np.clip(wavelength * g_hkls / 2, -1, 1))
        s2       = (g_hkls / 2) ** 2                        # (M,)

        # Atomic scattering factors: (M, N)
        #   fs[m, n] = zs[n] - 41.78214 * s2[m] * sum_k(coeffs[n,k,0] * exp(-coeffs[n,k,1]*s2[m]))
        # coeffs: (N, 4, 2)  →  broadcast s2: (M, 1, 1)
        s2_mnk   = s2[:, None, None]                        # (M, 1, 1)
        gauss    = np.sum(
            coeffs[None, :, :, 0] * np.exp(-coeffs[None, :, :, 1] * s2_mnk),
            axis=2,
        )                                                    # (M, N)
        fs       = zs[None, :] - 41.78214 * s2[:, None] * gauss  # (M, N)

        # Debye-Waller per atom, per hkl: (M, N)
        dw       = np.exp(-dwfactors[None, :] * s2[:, None])

        # g·r for all hkl and all atoms: (M, N)
        g_dot_r  = hkls_int.astype(float) @ fcoords.T       # (M, N)

        # Structure factors: (M,)
        f_hkl    = np.sum(
            fs * occus[None, :] * np.exp(2j * pi * g_dot_r) * dw,
            axis=1,
        )
        i_hkl    = (f_hkl * f_hkl.conjugate()).real         # (M,)

        # Lorentz-polarization factor: (M,)
        cos2t    = np.cos(2 * theta)
        sint     = np.sin(theta)
        cost     = np.cos(theta)
        lorentz  = (1 + cos2t ** 2) / (sint ** 2 * cost)

        intensities = i_hkl * lorentz                        # (M,)
        two_thetas_arr = np.degrees(2 * theta)               # (M,)

        # --- Merge peaks within TWO_THETA_TOL using rounding-based binning ---
        tol      = AbstractDiffractionPatternCalculator.TWO_THETA_TOL
        bin_keys = np.round(two_thetas_arr / tol).astype(int)

        peaks: dict[int, list] = {}
        for m in range(len(g_hkls)):
            hkl   = tuple(hkls_int[m])
            if is_hex:
                hkl = (hkl[0], hkl[1], -hkl[0] - hkl[1], hkl[2])
            key   = bin_keys[m]
            d_hkl = 1.0 / g_hkls[m]
            if key in peaks:
                peaks[key][0] += float(intensities[m])   # np.float64 -> float
                peaks[key][1].append(hkl)
            else:
                peaks[key] = [intensities[m], [hkl], two_thetas_arr[m], d_hkl]


        # --- Build output ---
        max_intensity = max(v[0] for v in peaks.values())
        tol_scaled    = AbstractDiffractionPatternCalculator.SCALED_INTENSITY_TOL
        x, y, hkls_out, d_hkls_out = [], [], [], []

        for key in sorted(peaks):
            v   = peaks[key]
            fam = get_unique_families(v[1])
            if v[0] / max_intensity * 100 > tol_scaled:
                x.append(float(v[2]))
                y.append(float(v[0]))
                hkls_out.append([{"hkl": hkl, "multiplicity": mult} for hkl, mult in fam.items()])
                d_hkls_out.append(float(v[3]))

        xrd = DiffractionPattern(x, y, hkls_out, d_hkls_out)
        if scaled:
            xrd.normalize(mode="max", value=100)
        return xrd
```

Bud-Macaulay:
Implement vector operation on XRD module for improved performance;

Resolves #4660

### DIFF
--- a/src/pymatgen/analysis/diffraction/xrd.py
+++ b/src/pymatgen/analysis/diffraction/xrd.py
@@ -150,33 +150,22 @@ class XRDCalculator(AbstractDiffractionPatternCalculator):
             structure = finder.get_refined_structure()
 
         wavelength = self.wavelength
-        lattice = structure.lattice
-        is_hex = lattice.is_hexagonal()
+        latt = structure.lattice
+        is_hex = latt.is_hexagonal()
 
-        # Obtained from Bragg condition. Note that reciprocal lattice
-        # vector length is 1 / d_hkl.
         min_r, max_r = (
             (0, 2 / wavelength)
             if two_theta_range is None
             else [2 * math.sin(math.radians(t / 2)) / wavelength for t in two_theta_range]
         )
 
-        # Obtain crystallographic reciprocal lattice points within range
-        recip_lattice = lattice.reciprocal_lattice_crystallographic
-        recip_pts = recip_lattice.get_points_in_sphere([[0, 0, 0]], [0, 0, 0], max_r)
+        recip_latt = latt.reciprocal_lattice_crystallographic
+        recip_pts = recip_latt.get_points_in_sphere([[0, 0, 0]], [0, 0, 0], max_r)
         if min_r:
             recip_pts = [pt for pt in recip_pts if pt[1] >= min_r]
 
-        # Create a flattened array of zs, coeffs, frac_coords and occus. This is used to perform
-        # vectorized computation of atomic scattering factors later. Note that these are not
-        # necessarily the same size as the structure as each partially occupied specie occupies its
-        # own position in the flattened array.
-        _zs = []
-        _coeffs = []
-        _frac_coords = []
-        _occus = []
-        _dw_factors = []
-
+        # --- Build per-site arrays ---
+        _zs, _coeffs, _fcoords, _occus, _dwfactors = [], [], [], [], []
         for site in structure:
             for sp, occu in site.species.items():
                 _zs.append(sp.Z)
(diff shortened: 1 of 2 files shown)


shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)**

Nice perf win — pulls the per-`hkl` Python loop body inside a single set of NumPy ops over all M reciprocal-lattice points. The math is preserved (scattering factor sum, Debye–Waller, structure factor, Lorentz-polarization).

- ✅ Correctly filters `g_hkl == 0` before the vectorized block; preserves the original sort key `(g_hkl, -h, -k, -l)` so the "first" peak in each merge bucket is deterministic.
- ✅ `np.clip(wavelength * g_hkls / 2, -1, 1)` in the `arcsin` is a sensible FP guard the original lacked.
- ⚠️ **Peak-merging semantics drift.** Original code merged peaks via `|Δ2θ| < TWO_THETA_TOL` (a true tolerance check). New code bins by `np.round(2θ / tol).astype(int)`. These disagree at bin boundaries — e.g. with `TWO_THETA_TOL = 1e-5`, peaks at `30.0000049` and `30.0000051` have `|Δ| < 1e-5` (merge in original) but round to different bins (don't merge in new). Probably benign in practice but it's a real behavior change worth either a comment, a tighter bin (e.g. `tol/2`), or — preferred — a regression test that pins the multiplicities/intensities for a known structure (e.g. one of the existing test fixtures) against the pre-PR output.
- ⚠️ **No tests added.** A perf refactor of a physics calculation needs a parity test. `xrd.py` already has tests in `tests/analysis/diffraction/test_xrd.py` — at minimum, add a numerical equivalence check on a couple of representative structures (cubic, hexagonal, low-symmetry with partial occupancies).
- 🔍 **Memory cost.** The intermediate `coeffs[None, :, :, 0] * np.exp(-coeffs[None, :, :, 1] * s2_mnk)` allocates `(M, N, 4)`. For a 1000-site supercell with M ≈ 10⁴, that's ~320 MB just for the gauss tensor — could OOM on large inputs where the original code happily streamed. Consider chunking over M (e.g. `np.array_split(np.arange(M), nchunks)`) for the inner products, or document a recommended size limit. Numbers in [the linked issue #4660](https://github.com/materialsproject/pymatgen/issues/4660) on actual benchmark structures would help quantify the tradeoff.
- 🔍 The trailing per-`m` Python loop building `peaks` is still O(M); not a correctness issue, just noting most of the residual cost is here. Could be replaced with `np.unique(bin_keys, return_inverse=True)` + `np.bincount(weights=intensities)` if you want to push further.
- 🔍 The `latt` / `recip_latt` / `_fcoords` rename is mostly cosmetic — the originals were already fine. Keeping the diff focused on the algorithmic change would make review easier; consider reverting the renames.
- 💡 CI: only `pre-commit.ci` ran (workflow likely awaiting maintainer approval for a first-time contributor). The full lint + ubuntu/macos/windows matrix needs to run before this can land.

LGTM in direction; please add a parity test and address the bin-merge edge case before merge.

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no code executed locally._

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)**

Vectorisation of `XRDCalculator.get_pattern` — replaces the per-`hkl` Python loop with broadcasting over the (M, N) hkl × site grid. The shape work is correct (scattering factors `(M, N)`, Debye-Waller `(M, N)`, `g·r = hkls_int @ fcoords.T`, structure factor `(M,)`).

- ⚠️ **Behavioural change in peak merging — needs a regression test.** The original code merged peaks whose `two_theta` differed by `< TWO_THETA_TOL` (proximity test). The new code keys `peaks` by `int(round(two_theta / tol))` (fixed-grid binning). These differ at bin boundaries: two peaks at ±tol/2 around a grid line get the same bin under the new scheme but would be `2·tol` apart under the old. Likely fine in practice, but please add a test comparing the new `get_pattern` output (peak positions, intensities, hkl multiplicities) against a stored reference for at least one cubic and one hexagonal structure to lock the behavior in.
- ⚠️ **Stored `two_theta` is now the first-seen value in the bin.** Earlier the dict was keyed by the first `two_theta` itself (so subsequent matches landed against that exact float). The new schema stores the first `two_thetas_arr[m]` encountered for a given `bin_key` in `peaks[key][2]`. For peaks near a bin edge this means the reported angle drifts up to `tol/2` from the original. Worth confirming this is within tolerance for downstream consumers.
- 💡 `cos2t = np.cos(2 * theta)` and `(1 + cos2t**2) / (sint**2 * cost)` will produce `inf`/`nan` at `theta = 0`. The `nonzero = g_hkls != 0` filter handles the only realistic origin of `theta = 0`, so this is fine — just noting that `np.clip(..., -1, 1)` is defensive only against floating-point overshoot, not a numerical concern.
- 💡 `recip_pts_sorted = sorted(recip_pts, key=...)` still pays the Python-level sort cost over all M points. Since the new pipeline is fully vectorised after sorting, you could drop the sort and bin first, sorting only the unique keys at the end — but the sort is O(M log M) which is small vs the (M, N) compute, so this is a micro-opt.
- 🔍 No new tests in the diff. Vectorisation rewrites of this scope should have at least one numerical-equivalence test against the old loop (or a stored golden output) — without it, regressions in peak positions/intensities are easy to miss.
- ✅ Pre-commit.ci green. Mergeable status `BLOCKED` because the full test matrix hasn't completed yet — that's the signal to wait for.

Once a regression test and a brief comment about the bin-edge behavior change are added, this looks like a clean speedup. Recommend confirming the (M, N) memory footprint is acceptable for the worst-case structures the test suite exercises (large supercells × large `two_theta_range`).

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no code executed locally._

Bud-Macaulay:
Sweet, I'm in agreement and also noted the slight behavior change on bin merging when running on our dataset.

OOO, but will cleanup when back at a machine.

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)**

Follow-up on the 2026-05-28 reviews — the `a84ec1d` commit addresses two of the three blockers:

- ✅ **Regression test added.** `test_get_pattern_merge_regression` pins peak counts and the first 10 hkl families for LiFePO4 (434 peaks) and Li10GeP2S12 (213 peaks). Good structures — orthorhombic LiFePO4 exercises non-cubic geometry; Li10GeP2S12 has many sites + partial occupancies, so the (M, N) vectorised inner products are well covered.
- ✅ **Bin-merging behaviour documented.** The new inline comment (`# Merge peaks within TWO_THETA_TOL by binning…`) makes the semantic shift from `|Δ| < tol` to `round(2θ / tol)` explicit.
- ⚠️ **Test tolerance is too coarse to catch the bin-edge concern.** The test overrides `TWO_THETA_TOL = 5` (degrees) — five orders of magnitude above the default `1e-5`. At that width essentially everything in the same degree-range collapses into one bin, so the original `|Δ2θ| < tol` vs `round(2θ/tol)` behavioural drift I flagged on 2026-05-28 is not exercised. Consider an additional assertion at the **default** tolerance against a stored golden output (peak positions + intensities) for at least one structure — that's the regime production callers actually use.
- ⚠️ **Snapshot, not parity.** The recorded values were generated *after* the rewrite, so the test will catch *future* algorithmic drift but cannot confirm parity with the pre-PR implementation. If you can checkout master, run `get_pattern` on the same two structures at the default tol, and assert equality against the new output (within float tolerance), that closes the loop on the original concern.
- 🔍 **Test naming / scope nit.** `test_get_pattern_merge_regression` reads as if it's specifically about merging, but at `TOL=5` it's really a coarse-grained \"does anything still work\" test. Either tighten the tol or rename to reflect the broader smoke-test intent.
- 💡 Still outstanding (non-blocking): the `(M, N, 4)` `gauss` tensor allocation noted previously. Not addressed; probably fine for typical structures, but if downstream users hit OOM on large supercells this is the first place to chunk.
- 💡 CI: only `pre-commit.ci` has run (green). The full lint + ubuntu/macos/windows matrix is still awaiting maintainer workflow approval — needs to go green before merge.

Direction is good; please tighten the regression test to default tolerance and ideally add a parity check vs master before this lands.

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no code executed locally._

Bud-Macaulay:
The merge regression test assersion values were generated on the current master results, and thus currently passing indicates no behaviour change on test data.

shyuep:
Thanks!

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)**

Follow-up on the 2026-05-29 review — the `9441659` commit drops the `xrd_calc.TWO_THETA_TOL = 5` override in `test_get_pattern_merge_regression`, so the test now runs at the default `1e-5`. That directly resolves the "tolerance too coarse to exercise bin-edge behaviour" blocker.

- ✅ With @Bud-Macaulay's note that the assertion values were generated against master, the test is now a true parity check against the pre-PR implementation at the production tolerance — both LiFePO4 (434 peaks) and Li10GeP2S12 (213 peaks) lock in peak counts + first 10 hkl families.
- ✅ Full CI matrix is green (lint + ubuntu/macos/windows across Python 3.11/3.13/3.14, `pre-commit.ci`). `mergeable: MERGEABLE`.
- 💡 Still outstanding, non-blocking: the `(M, N, 4)` gauss-tensor allocation in the vectorised inner block could OOM on very large supercells × wide two_theta ranges. Worth keeping in mind if a downstream user reports memory pressure, but not a merge blocker.

LGTM — happy to see this land.

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no code executed locally._

shyuep:
Thanks!

PR #4666, opened 2026-06-13
Fix HeisenbergModel as_dict/from_dict round-trip (#4664)

### LINKED ISSUE #4664: HeisenbergModel.as_dict()/from_dict() round-trip broken: ex_mat serialized as a dict but parsed with literal_eval
### Python version

3.12.13

### Pymatgen version

2026.5.4

### Operating system version

RHEL 8.10 (Linux 4.18.0-553)

### Current behavior

`HeisenbergModel` is not round-trippable through `as_dict()`/`from_dict()`. `as_dict()` serializes the `ex_mat` exchange matrix (a `pandas.DataFrame`) with `jsanitize`, which produces a **dict**, but `from_dict()` reads it back with `ast.literal_eval`, which only accepts a **string**. Deserialization raises:

```
ValueError: malformed node or string: {'E': {}, 'E0': {}, '0-1-nn': {}}
```

This breaks any MSON-based persistence of a `HeisenbergModel` — e.g. passing one between jobflow/atomate2 jobs (the atomate2 exchange workflow hits this handing the fitted model from the Heisenberg-mapping job to the Vampire job).

**Root cause**

`as_dict()` stores `ex_mat` as a jsanitized dict:

```python
# heisenberg.py, as_dict()
"ex_mat": jsanitize(self.ex_mat),   # DataFrame -> dict
```

but `from_dict()` assumes a string and parses it with `literal_eval`:

```python
# heisenberg.py, from_dict()
try:
    ex_mat = literal_eval(dct["ex_mat"])   # expects str; gets a dict -> ValueError
    ex_mat = pd.DataFrame.from_dict(ex_mat)
except SyntaxError:  # if ex_mat is empty
    ex_mat = pd.DataFrame(columns=["E", "E0"])
```

`literal_eval` raises `ValueError` (not `SyntaxError`) on a dict, so the existing `except SyntaxError` guard does not catch it.

**When this was introduced**

The mismatch is long-standing, not new. As far back as **v2022.0.8** (2021), `as_dict` already used `jsanitize(self.ex_mat)` while `from_dict` parsed it with `eval(dct["ex_mat"])`; the round-trip only works if `jsanitize(ex_mat)` returns a *string*, but for a real `DataFrame` `monty.json.jsanitize` returns a **dict**. PR #3739 ("Avoid using `eval`…", merged 2024-04-07) replaced `eval` with `literal_eval` in `from_dict` — this did not fix the mismatch, it only changed the failure mode, so the current exact symptom (`ValueError: malformed node or string`) dates to #3739 while the underlying defect is older. The path is rarely exercised because it only triggers when a `HeisenbergModel` is round-tripped through MSON, which is why it went unnoticed.

**Proposed fix** — make `from_dict` consume what `as_dict` produces, tolerating the legacy string form:

```python
# Reconstitute the exchange matrix DataFrame.
# as_dict() stores ex_mat as a (jsanitized) dict; tolerate the legacy
# string form for backwards compatibility.
ex_mat = dct["ex_mat"]
if isinstance(ex_mat, str):
    ex_mat = literal_eval(ex_mat)
ex_mat = pd.DataFrame.from_dict(ex_mat)
```

Verified against all observed forms: empty/non-empty dict (current output) and both legacy string forms. Happy to open a PR.

Also confirmed present on `master`. monty 2026.5.18, pandas 3.0.3.

### Expected Behavior

HeisenbergModel.from_dict(model.as_dict()) should return an equivalent
HeisenbergModel. Instead it raises ValueError: malformed node or string.

### Minimal example

```Python
import pandas as pd
from pymatgen.core import Structure, Lattice
from pymatgen.analysis.magnetism.heisenberg import HeisenbergMapper, HeisenbergModel

latt = Lattice.cubic(4.45)
species = ["Mn", "Mn", "Mn", "Mn", "O", "O", "O", "O"]
coords = [[0, 0, 0], [0.5, 0.5, 0], [0.5, 0, 0.5], [0, 0.5, 0.5],
          [0.5, 0, 0], [0, 0.5, 0], [0, 0, 0.5], [0.5, 0.5, 0.5]]

def make(m):
    s = Structure(latt, species, coords)
    s.add_site_property("magmom", m + [0, 0, 0, 0])
    return s

cfg = [([5, 5, 5, 5], -100.0), ([5, -5, 5, -5], -100.6),
       ([5, 5, -5, -5], -100.3), ([5, -5, -5, 5], -100.4)]
model = HeisenbergMapper(
    [make(m) for m, e in cfg], [e for m, e in cfg], cutoff=3.5, tol=0.04
).get_heisenberg_model()

# ex_mat is a DataFrame for any non-degenerate fit:
model.ex_mat = pd.DataFrame(columns=["E", "E0", "0-1-nn"])

HeisenbergModel.from_dict(model.as_dict())   # raises ValueError
```

_No response_

vasa-develop:
## Summary

Fixes #4664.

`HeisenbergModel` is not round-trippable through `as_dict()` / `from_dict()`. `as_dict()` serializes the `ex_mat` exchange matrix (a `pandas.DataFrame`) with `jsanitize`, which produces a **dict**, but `from_dict()` parsed it back with `ast.literal_eval`, which only accepts a **string** and raises `ValueError: malformed node or string` on a dict. The existing `except SyntaxError` guard does not catch `ValueError`, so deserialization fails.

This breaks any MSON-based persistence of a `HeisenbergModel` (e.g. passing one between jobflow/atomate2 jobs in the exchange workflow).

Full credit to @Luguza for the detailed root-cause analysis and minimal reproducer in the issue.

## Fix

`from_dict()` now consumes what `as_dict()` produces:

- accepts the **dict** form emitted by `jsanitize(DataFrame)`,
- tolerates the legacy **string** form for backwards compatibility,
- falls back to an empty `["E", "E0"]` matrix when `ex_mat` is empty/unparsable.

```python
ex_mat = dct["ex_mat"]
if isinstance(ex_mat, str):
    try:
        ex_mat = literal_eval(ex_mat)
    except (SyntaxError, ValueError):  # empty or unparsable string
        ex_mat = None
ex_mat = pd.DataFrame.from_dict(ex_mat) if ex_mat else pd.DataFrame(columns=["E", "E0"])
```

## Test plan

- Added `test_as_from_dict_round_trip` to `tests/analysis/magnetism/test_heisenberg.py`. It performs a double MSON round-trip on a fitted model: the first hop turns the JSON-string `ex_mat` into a `DataFrame`, and the second hop exercises the `jsanitize`'d **dict** path that previously raised `ValueError`.
- Verified the new test **fails on `main`** (`ValueError: malformed node or string`) and **passes with this change**.
- Full `tests/analysis/magnetism/test_heisenberg.py` suite passes (8 passed).
- `ruff check` and `ruff format --check` clean on the changed files.

### DIFF
--- a/src/pymatgen/analysis/magnetism/heisenberg.py
+++ b/src/pymatgen/analysis/magnetism/heisenberg.py
@@ -913,12 +913,17 @@ class HeisenbergModel(MSONable):
         # Interaction graph
         igraph = StructureGraph.from_dict(dct["igraph"])
 
-        # Reconstitute the exchange matrix DataFrame
-        try:
-            ex_mat = literal_eval(dct["ex_mat"])
-            ex_mat = pd.DataFrame.from_dict(ex_mat)
-        except SyntaxError:  # if ex_mat is empty
-            ex_mat = pd.DataFrame(columns=["E", "E0"])
+        # Reconstitute the exchange matrix DataFrame. as_dict() serializes
+        # ex_mat with jsanitize, which turns a DataFrame into a dict, while
+        # older serializations may store a (JSON/repr) string instead. Accept
+        # both forms and fall back to an empty matrix when ex_mat is empty.
+        ex_mat = dct["ex_mat"]
+        if isinstance(ex_mat, str):
+            try:
+                ex_mat = literal_eval(ex_mat)
+            except (SyntaxError, ValueError):  # empty or unparsable string
+                ex_mat = None
+        ex_mat = pd.DataFrame.from_dict(ex_mat) if ex_mat else pd.DataFrame(columns=["E", "E0"])
 
         return HeisenbergModel(
             formula=dct["formula"],
--- a/tests/analysis/magnetism/test_heisenberg.py
+++ b/tests/analysis/magnetism/test_heisenberg.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import pandas as pd
 from pytest import approx
 
-from pymatgen.analysis.magnetism.heisenberg import HeisenbergMapper
+from pymatgen.analysis.magnetism.heisenberg import HeisenbergMapper, HeisenbergModel
 from pymatgen.core.structure import Structure
 from tests.testing import TEST_FILES_DIR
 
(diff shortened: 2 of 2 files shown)


Luguza:
Thanks for the PR, I didn't come around to do it.

shyuep:
Thanks!

Issue #4677, opened 2026-06-30
Further vectorise neutron diffraction method to align with XRD


Bud-Macaulay:
Similarly to #4660 better vector ops can be applied to neutron diffraction to get a 3x improvement (in my testing),
It will also align the diffraction methods to be closer aligned in algorithm and use similar naming conventions.

I think in principle this vectorisation could likely abstracted out and used across the classes, but is probably unneccesary  ---
 
Have a working version in my own fork - will submit a PR later.

PR #4678, opened 2026-06-30
vectorise neutron.py + improve tests

Bud-Macaulay:
As eluded to in #4677

Should offer between 2--5x improvement across all systems and uses the equivalent vectorisation that the XRD uses (since commit afab344012e833e8645bffb72edecde9d9d7f886)

also improved neutron test to test complete pattern for hex

### DIFF
--- a/src/pymatgen/analysis/diffraction/neutron.py
+++ b/src/pymatgen/analysis/diffraction/neutron.py
@@ -105,16 +105,8 @@ class NDCalculator(AbstractDiffractionPatternCalculator):
         if min_r:
             recip_pts = [pt for pt in recip_pts if pt[1] >= min_r]
 
-        # Create a flattened array of coeffs, frac_coords and occus. This is
-        # used to perform vectorized computation of atomic scattering factors
-        # later. Note that these are not necessarily the same size as the
-        # structure as each partially occupied specie occupies its own
-        # position in the flattened array.
-        _coeffs = []
-        _frac_coords = []
-        _occus = []
-        _dwfactors = []
-
+        # --- Build per-site arrays ---
+        _coeffs, _frac_coords, _occus, _dw_factors = [], [], [], []
         for site in structure:
             for sp, occu in site.species.items():
                 try:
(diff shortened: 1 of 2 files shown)


shyuep:
🤖 **Automated PR review by Claude** (posted via @shyuep). Based on static analysis of the diff; no code was executed locally.

Nice cleanup — the per-site arrays and structure-factor computation are now fully vectorized and more readable, and adding `np.clip` on the `arcsin` argument is a good robustness fix.

One behavioral point worth confirming: peak merging changed from the original sequential tolerance search (find any existing 2θ within `TWO_THETA_TOL`) to fixed-bin rounding (`round(2θ / tol)`). Binning can assign two reflections that are within tolerance of each other but straddle a bin boundary to *separate* bins, so merged-peak grouping may differ slightly from the old behavior in edge cases. The new regression test guards the common path, so please just confirm the binning semantics are intended.

Note: no CI checks have run on this branch yet — please ensure the diffraction test suite passes before merge.

shyuep:
🤖 **Automated PR review by Claude** (posted via @shyuep). Based on static analysis of the diff; no code was executed locally. Re-review triggered by new commits since the prior automated review.

**Verdict:** Looks good — clean vectorization mirroring the XRD approach. A few points:

- ✅ Fully vectorizes θ, Debye-Waller, `g·r`, and structure factors across all hkls; the `np.clip(..., -1, 1)` guard on `arcsin` is a nice robustness improvement over the old `math.asin` (avoids domain errors on FP overshoot).
- ✅ Regression test now pins the full Graphite pattern (x/y), which is the right way to lock in equivalence with prior behavior.
- ⚠️ **Peak-merging semantics changed.** Old code merged into the first two-theta within `TWO_THETA_TOL`; new code bins via `round(two_theta / tol)`. This is more deterministic but not identical at bin boundaries — two peaks `< tol` apart can land in different bins, or peaks up to `tol` apart merge. Test passes, but worth a one-line comment noting the intentional change.
- ⚠️ **Memory:** builds dense `(M, N)` arrays for `dw` and `g_dot_r`. Fine for typical cells, but a large structure × many reciprocal points is now O(M·N) memory vs. the old per-hkl loop. Acceptable trade-off for the speedup; just flagging.
- 🔹 Consider guarding the empty case: `max(v[0] for v in peaks.values())` raises on an empty `peaks` dict (e.g. all `g_hkl == 0`). Pre-existing behavior, low priority.
- 🔹 No CI checks are reported on this branch — please ensure the test workflow runs before merge.

shyuep:

Nice full vectorization — the (M,N) Debye–Waller matrix and batched structure-factor sum look correct, and the regression test values are updated. One behavioral note: peak merging switched from sequential within-tolerance matching to integer binning via `round(two_theta/TWO_THETA_TOL)`. Two reflections closer than the tolerance but on opposite sides of a bin edge will no longer merge, and the stored two_theta becomes the first hit rather than a merged value. Likely fine for typical patterns but worth a note on the semantics change. CI hasn't run yet (branch blocked) — needs green checks before merge.

shyuep:
Thanks!

PR #4682, opened 2026-07-12
fix(magnetism): respect truncate_by_symmetry count in MagneticStructureEnumerator

steps-re:
## What's broken

In `MagneticStructureEnumerator._generate_ordered_structures`, the low-symmetry pruning step checks `self.truncate_by_symmetry` to decide when to truncate, but the actual slice is hardcoded to `[:5]`. So setting `truncate_by_symmetry=2` (or any value other than the default) has no effect once truncation kicks in, you still get up to 5 structures back instead of the count you asked for. Checked with `git blame`: this hardcoding has been there since the original 2019 commit that added `truncate_by_symmetry`.

## Fix

Use `self.truncate_by_symmetry` in the slice instead of the literal `5`, matching the guard condition right above it (`if len(max_symmetries) > self.truncate_by_symmetry`).

## Test

Added `TestMagneticStructureEnumeratorTruncation.test_truncate_by_symmetry_respects_configured_count` in `tests/analysis/magnetism/test_analyzer.py`. It builds a small set of structures spanning distinct symmetry levels and calls `_generate_ordered_structures` directly (bypassing `__init__`, since full construction needs `enumlib` which isn't available in this environment). Confirmed the test fails on the old code and passes on the fix.


shyuep (commented):

**Correct core fix, but watch the default path.** Using `self.truncate_by_symmetry` in the slice instead of the hardcoded `[:5]` is the right change for explicit integer counts, and the regression test covers those well.

**Likely regression for the default case.** `truncate_by_symmetry` defaults to `True`. Because `bool` is a subclass of `int` in Python, the guard a few lines above:

```python
if not isinstance(self.truncate_by_symmetry, int):
    self.truncate_by_symmetry = 5
```

does **not** fire for `True` (`isinstance(True, int) is True`), so `self.truncate_by_symmetry` stays `True` (== 1). With the old hardcoded `[:5]` that was masked, but your new slice becomes `[:True]` == `[:1]`. Net effect: default runs now keep only 1 symmetry level instead of 5 — a silent behavior change for the common path.

Suggested additional fix so the default coerces to 5:

```python
if self.truncate_by_symmetry is True or not isinstance(self.truncate_by_symmetry, int):
    self.truncate_by_symmetry = 5
```

Please also add a test asserting the `truncate_by_symmetry=True` default keeps 5 (your current test bypasses `__init__` and only sets integer values, so it wouldn't catch this).

Nit: `int` type hint would be clearer than `bool` for the signature given it accepts a count.

shyuep:
🤖 _Automated PR review generated by Claude, posted on behalf of @shyuep. Based purely on static analysis of the diff and CI results — no code was executed locally._

Correct core fix. Because `bool` is a subclass of `int`, the old `isinstance(self.truncate_by_symmetry, int)` guard let the default `True` pass through, so the slice `[:True]` (== `[:1]`) kept only one symmetry level instead of five. Coercing `is True` → `5` and slicing `[: self.truncate_by_symmetry]` (instead of the hardcoded `[:5]`) fixes both the default and the user-configured cases. The two added regression tests cover the default-True path and explicit counts, and cleanly bypass enumlib — good.

- Nit: the signature annotation `truncate_by_symmetry: int = True` is misleading; the parameter is semantically `bool | int`. Suggest `truncate_by_symmetry: bool | int = True` to match the docstring.
- CI: no checks are reported on this branch — the workflows do not appear to have run. Please make sure the full test suite executes before merge so the new tests are actually exercised.

LGTM pending green CI.

shyuep:
🤖 _Automated PR review generated by Claude, posted on behalf of @shyuep. Based purely on static analysis of the diff and CI; no code was run locally._

Re-review after the new commit adding regression tests. The core fix is correct: since `bool` subclasses `int`, the default `truncate_by_symmetry=True` previously survived the `isinstance(..., int)` guard and made the slice `[:True] == [:1]`, keeping only one symmetry level. Explicitly coercing `is True → 5` and slicing with `self.truncate_by_symmetry` instead of the hardcoded `[:5]` fixes both the default and the user-configured path.

The two new regression tests are thorough — they cover the direct truncation path and the `__init__` default path, and assert coercion to integer `5` (not `bool`). Both bypass enumlib cleanly.

One thing to confirm before merge: the PR shows **no CI checks reported** on the branch. Please make sure the test workflow actually runs (fork PR / workflow-approval) so the new tests are exercised. Logic and tests LGTM otherwise.

steps-re:
thanks! looks like the Lint/test runs are stuck in `action_required` since it's my first PR here. one click on "approve and run workflows" and the new regression tests should fire.

shyuep:
Thanks for the fix.

PR #4684, opened 2026-07-15
Diffraction pattern optimizations

mkphuthi:
## Summary

Major changes:

- feature 1: These are purely efficiency optimizations under the hood for predicting XRD and neutron diffraction patterns. The interface remains untouched. These gains allow direct prediction of diffraction patterns from MD trajectories, include Debye-Waller factors in seconds on a PC which was impossible before due to memory and compute constraints. The performance and bit-wise accuracy of the enhancements is shown in the attached plots for each optimization.

- Optimization 1 (XRD only): Per element form factors are computed once and stored for each element rather than computed for each atom

02_per_element_form_factors

- Optimization 2: Chunked structure factors per species, which massively reduces memory requirements
 
03_chunked_structure_factors
05_neutron_chunked

- Optimization 3: Friedel halving i.e. only computing for g and not both g and -g and transforming rather than computing for the full sphere

04_friedel_halving
06_neutron_friedel

## Todos
- pass tests

## Checklist

- [ ] Pass tests

### DIFF
--- a/src/pymatgen/analysis/diffraction/neutron.py
+++ b/src/pymatgen/analysis/diffraction/neutron.py
@@ -49,6 +49,12 @@ class NDCalculator(AbstractDiffractionPatternCalculator):
     Chapter13, Cambridge University Press 2003.
     """
 
+    # Upper bound on the number of entries of the phase matrix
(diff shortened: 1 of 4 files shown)


shyuep:
Very nice.

PR #4685, opened 2026-07-16
fix(diffraction): correct interplanar angle for non-orthogonal cells in TEMCalculator

steps-re:
### Summary

`TEMCalculator.get_interplanar_angle` returns wrong angles for triclinic and monoclinic cells. Two copy-paste errors in the reciprocal-lattice formulas:

1. In both `r1_norm` and `r2_norm`, the `k·l` cross term (`b* · c*`) multiplies by `cos_gamma_star` but the b–c reciprocal angle is `alpha*`, so it should be `cos_alpha_star`.
2. In `r1_dot_r2`, the `h·l` cross term has an index typo: `(p1[0] * p2[2] + p2[0] * p1[1])` should be `(p1[0] * p2[2] + p2[0] * p1[2])`.

Both are invisible on the existing cubic/tetragonal/hexagonal tests because their `cos_alpha_star`/`cos_beta_star` terms are zero. On a general triclinic cell the error is several degrees.

### Verification

For `Lattice.from_parameters(4, 5, 6, 80, 95, 70)`, planes `(1,1,1)` and `(1,0,2)`:

| | angle |
| --- | --- |
| current code | 36.41° |
| independent calc (`reciprocal_lattice_crystallographic` + `np.dot`) | 40.39° |
| this PR | 40.39° |

The added `test_interplanar_angle_triclinic` checks against that independent reciprocal-lattice computation rather than a hardcoded value. Existing cubic (90°) and hexagonal (21.0517°) cases are unchanged.

Fixes the interplanar-angle calculation for all non-orthogonal cells.

---

*Disclosure: this fix was found with LLM-assisted code analysis and prepared with LLM assistance. I (Mike German) am the human contributor, verified the bug and fix by running the calculation against an independent reciprocal-lattice reference before submitting, and am accountable for the PR.*


shyuep:
🤖 **Automated PR review** (generated by Claude on behalf of @shyuep). Based purely on the diff and CI results — no code was executed locally.

Checked all three changes against the reciprocal-lattice metric |G|² = Σ hᵢhⱼ·gᵢ*gⱼ*·cos(angleᵢⱼ*):

- **Both norm cross terms** — the b*·c* (k·l) term pairs with **α\***, so `cos_gamma_star` → `cos_alpha_star` is correct.
- **Dot-product h·l term** — the index fix `p1[1]` → `p1[2]` is correct (h = index 0, l = index 2).

The bug was genuinely masked by the existing cubic/tetragonal/hexagonal tests since their cos_alpha*/cos_beta* terms vanish — good catch. The added `test_interplanar_angle_triclinic` validates against an independent `reciprocal_lattice_crystallographic` computation instead of a hardcoded value, which is the right way to write this regression test.

CI is green (lint + full OS/Python matrix). LGTM — small, correct, well-tested. Recommend merge.

shyuep:
Thanks!

PR #4687, opened 2026-07-27
Clarify label_index deduplication behavior in CoherentInterfaceBuilde…

### LINKED ISSUE #4572: Modify CoherentInterfaceBuilder documentation to clearly state the usage of label_index
In generating coherent interfaces between film tetragonal TiO2(001) and substrate cubic Si(111), one expect Si(111) to have two inequivalent terminations. These two terminations correspond to 3-coordinated surface and 1-coordinated surface with dangling Si atoms, both with R-3m_1 symmetry.

3-coordinated, shift = 0.375, R-3m_1

Image

Image

1-coordinated, shift = 0.875, R-3m_1

Image

Image

However, **if not supplied with parameter label_index=True**,  the current implementation in `pymatgen.analysis.interface.CoherentInterfaceBuilder.__find_terminations()`, can only yield the latter termination ( the 1-coordinated one).

Printing out intermediate values in find_terminations() function further validates this:
```python
        film_slabs = film_sg.get_slabs(ftol=self.termination_ftol, filter_out_sym_slabs=self.filter_out_sym_slabs)
        sub_slabs = sub_sg.get_slabs(ftol=self.termination_ftol, filter_out_sym_slabs=self.filter_out_sym_slabs)
        film_shifts = [slab.shift for slab in film_slabs]

        print("Len(film_slabs):", len(film_slabs))
        print("Len(sub_slabs):", len(sub_slabs))
        if self.label_index:
            film_terminations = [
                label_termination(slab, self.termination_ftol, t_idx) for t_idx, slab in enumerate(film_slabs, start=1)
            ]
        else:
            film_terminations = [label_termination(slab, self.termination_ftol) for slab in film_slabs]
        print("film_terminations:", film_terminations)
        print("film_shifts:", film_shifts)

        sub_shifts = [slab.shift for slab in sub_slabs]
        if self.label_index:
            sub_terminations = [
                label_termination(slab, self.termination_ftol, t_idx) for t_idx, slab in enumerate(sub_slabs, start=1)
            ]
        else:
            sub_terminations = [label_termination(slab, self.termination_ftol) for slab in sub_slabs]
        print("sub_terminations:", sub_terminations)
        print("sub_shifts:", sub_shifts)

...
        print("terminations:", self.terminations)
```

Output:
```
Len(film_slabs): 1
Len(sub_slabs): 2
film_terminations: ['TiO2_Cmmm_3']
film_shifts: [0.24999999999999997]
sub_terminations: ['Si_R-3m_1', 'Si_R-3m_1']
sub_shifts: [0.375, 0.875]
terminations: [('TiO2_Cmmm_3', 'Si_R-3m_1')]
```

This is because when `label_index` is False, two slabs will first be assigned the same name and then regarded equivalent if they have the same space-group symmetry.

The [current](https://github.com/materialsproject/pymatgen/blob/682bfd855dd89264b0ff8d6bee4815f8834cc5ac/src/pymatgen/analysis/interfaces/coherent_interfaces.py#L47) documentation sentence falls short of stating this point clearly to the readers, in which one may fail to understand that `label_index` actually **determines the core termination detection behavior**, and may have to refer to the source code to figure this out.

```python
label_index (bool): If True add an extra index at the beginning of the termination label.
```

I suggest editing this documentation in the next version, clearly stating how `label_index=True` correctly captures all inequivalent terminations.

Thank you!

khanaiml:
## Summary

  Expands the `label_index` docstring in `CoherentInterfaceBuilder` to explain
  that it controls termination deduplication. Without it, terminations sharing
  the same space-group symmetry are treated as identical and merged, which can
  silently drop physically inequivalent terminations from the enumeration.

  Fixes #4572

### DIFF
--- a/src/pymatgen/analysis/interfaces/coherent_interfaces.py
+++ b/src/pymatgen/analysis/interfaces/coherent_interfaces.py
@@ -45,7 +45,11 @@ class CoherentInterfaceBuilder:
             zslgen (ZSLGenerator | None): BiDirectionalZSL if you want custom lattice matching tolerances for coherency.
             termination_ftol (float | tuple[float, float]): tolerances (film, substrate) to distinguish
                 different terminating atomic planes.
-            label_index (bool): If True add an extra index at the beginning of the termination label.
+            label_index (bool): If True, prepend an index to each termination label. This controls
+                deduplication of terminations: when False, terminations sharing the same space-group
+                symmetry are treated as identical and merged, which can silently drop physically
+                inequivalent terminations from the enumeration. Set True to retain all inequivalent
+                terminations.
             filter_out_sym_slabs (bool): If True filter out identical slabs with different terminations.
                 This might need to be set as False to find more non-identical terminations because slab
                 identity separately does not mean combinational identity.
--- a/src/pymatgen/analysis/interfaces/coherent_interfaces.py
+++ b/src/pymatgen/analysis/interfaces/coherent_interfaces.py
@@ -23,6 +23,11 @@ class CoherentInterfaceBuilder:
     """
     This class constructs the coherent interfaces between two crystalline slabs
     Coherency is defined by matching lattices not sub-planes.
+
+    Note:
+        By default, terminations sharing the same space-group symmetry are merged.
+        Set ``label_index=True`` to retain all physically inequivalent terminations
+        (see the ``label_index`` argument for details).
     """
 
     def __init__(


shyuep:

Docs-only change, and a helpful one — the expanded `label_index` docstring makes the deduplication side effect explicit (that `label_index=False` can silently merge symmetry-equivalent-but-physically-distinct terminations). Accurate and clearly worded.

Suggestion: since this documents a subtle footgun, consider whether the default should also be revisited, or add a one-line note in the class-level docstring / user guide. Not blocking. LGTM.

khanaiml:
Thanks for the review! Glad the wording works.

Good point on the class-level docstring, I'll add a brief note there pointing to the `label_index` explanation so the footgun is visible at the class level too.

On revisiting the default: agreed it's worth considering, but since that's a behavior change with backward-compatibility implications, it feels better as a separate issue/PR rather than bundling it here. Happy to open an issue to track that discussion if useful.

shyuep:
Thanks!

PR #4688, opened 2026-07-28
Use dynamic versioning

DanielYang59:
https://github.com/materialsproject/pymatgen-core/pull/107


shyuep:

Sensible move to setuptools-scm dynamic versioning. The workflow changes are consistent with that: `fetch-depth: 0` on all checkouts (lint/test/release) so the tag history is available for version derivation, checking out `github.event.release.tag_name` in the release jobs, and routing `workflow_dispatch` runs to TestPyPI. ADMIN.md updates match the new tag-then-release flow.

The failing `test` matrix is NOT caused by this PR — it's the unrelated `test_tem.py::is_parallel` float-comparison assertion, which is fixed by #4689. Recommend merging #4689 first, then rebasing this to get a green matrix.

One thing to double-check before merge: confirm `setuptools-scm` is configured (`[tool.setuptools_scm]` in `pyproject.toml` and `dynamic = ["version"]`) and that sdist/wheel builds still embed the correct version, since the diff shown here is workflow/docs-only.

shyuep:
Thanks.

PR #4689, opened 2026-07-28
Fix float point compare in TEM test

DanielYang59:
fix https://github.com/materialsproject/pymatgen/actions/runs/30377834893/job/90337936566?pr=4688

### DIFF
--- a/src/pymatgen/analysis/diffraction/tem.py
+++ b/src/pymatgen/analysis/diffraction/tem.py
@@ -354,7 +354,11 @@ class TEMCalculator(AbstractDiffractionPatternCalculator):
             bool: True if the planes are parallel, False otherwise.
         """
         phi = self.get_interplanar_angle(structure, plane, other_plane)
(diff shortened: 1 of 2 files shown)


shyuep:

Correct fix. Replacing the exact `phi in (180, 0)` equality with `np.isclose(..., atol=1e-5)` is the right call — interplanar angles are floating point, so the exact membership test was fragile and could silently return `False` for parallel planes. Wrapping in `bool(...)` also avoids leaking a `np.bool_`.

Note: the failing `lint` check is unrelated to this PR — it's pre-existing repo-wide `PLR0917` (too many positional args) in `chemenv`, not touched here. The substantive change is sound and the `test` matrix passes. This also fixes the `test_tem::is_parallel` failure currently breaking #4688. LGTM.

Minor: consider using the module's existing angle tolerance constant if one exists, rather than a literal `1e-5`, for consistency.

PR #4694, opened 2026-08-02
fix: respect strict_anions and explicit S oxidation states in MP2020 (#4538)

### LINKED ISSUE #4538: Incorrect AnionCorrections obtained using MaterialsProject2020Compatibility for ComputedEntry with sulfur having positive oxidation state.
### Python version

Python 3.11.13

### Pymatgen version

2025.10.7

### Operating system version

OSX

### Current behavior

I get AnionCorrections for S6+ species (it is cation) when I create ComputedEntry from scratch and apply corrections using MaterialsProject2020Compatibility.

```python
from pymatgen.core import Composition 
from pymatgen.entries.compatibility import MaterialsProject2020Compatibility

comp = Composition.from_dict({"Li1+": 4, "Cs1+": 4, "S6+":4, "O2-": 16})
energy = -160.3371 # uncorrected energy for mp-6597

entry = ComputedEntry(composition = comp, energy = energy)
entry.parameters = {"run_type": "GGA"}

compat = MaterialsProject2020Compatibility(check_potcar=False, correct_peroxide=True)
compat.process_entry(entry, on_error="raise")
print(entry)
```
```
None ComputedEntry - Cs4 Li4 S4 O16 (CsLiSO4)
Energy (Uncorrected)     = -160.3371 eV (-5.7263  eV/atom)
Correction               = -13.0040  eV (-0.4644  eV/atom)
Energy (Final)           = -173.3411 eV (-6.1908  eV/atom)
Energy Adjustments:
  MP2020 anion correction (S): -2.0120   eV (-0.0719  eV/atom)
  MP2020 anion correction (oxide): -10.9920  eV (-0.3926  eV/atom)
Parameters:
  run_type               = GGA
Data:
  oxidation_states       = {'Li': 1.0, 'Cs': 1.0, 'S': 6.0, 'O': -2.0}
```

However, when I retrieve this record using mp_api, everything appears correct.

```python
with MPRester(api_key) as mp_rester:
    entry = mp_rester.get_entries(['mp-6597'], additional_criteria={'thermo_types': ['GGA_GGA+U']})
print(entry[0])
```

```
Retrieving ThermoDoc documents: 100%|██████████| 1/1 [00:00<00:00, 19328.59it/s]
mp-6597-GGA ComputedStructureEntry - Cs4 Li4 S4 O16 (CsLiSO4)
Energy (Uncorrected)     = -160.3371 eV (-5.7263  eV/atom)
Correction               = -10.9920  eV (-0.3926  eV/atom)
Energy (Final)           = -171.3291 eV (-6.1189  eV/atom)
Energy Adjustments:
  MP2020 anion correction (oxide): -10.9920  eV (-0.3926  eV/atom)
Parameters:
  potcar_spec            = [{'titel': 'PAW_PBE Cs_sv 08Apr2002', 'hash': '096b53a7d80cc0086976bcda50d536e5', 'summary_stats': None}, {'titel': 'PAW_PBE Li_sv 23Jan2001', 'hash': '4799bab014a83a07c654d7196c8ecfa9', 'summary_stats': None}, {'titel': 'PAW_PBE S 17Jan2003', 'hash': 'd368db6899d8839859bbee4811a42a88', 'summary_stats': None}, {'titel': 'PAW_PBE O 08Apr2002', 'hash': '7a25bc5b9a5393f46600a4939d357982', 'summary_stats': None}]
  run_type               = GGA
  is_hubbard             = False
  hubbards               = None
Data:
  oxide_type             = oxide
  aspherical             = True
  last_updated           = 2024-11-21 20:38:24.276956+00:00
  task_id                = mp-1393723
  material_id            = mp-6597
  oxidation_states       = {'Cs': 1.0, 'Li': 1.0, 'S': 6.0, 'O': -2.0}
  license                = BY-C
  run_type               = GGA
```

### Expected Behavior

No anion corrections for sulfur having positive oxidation state.

### Minimal example

```Python
from pymatgen.core import Composition 
from pymatgen.entries.compatibility import MaterialsProject2020Compatibility

comp = Composition.from_dict({"Li1+": 4, "Cs1+": 4, "S6+":4, "O2-": 16})
energy = -160.3371 # uncorrected energy for mp-6597

entry = ComputedEntry(composition = comp, energy = energy)
entry.parameters = {"run_type": "GGA"}

compat = MaterialsProject2020Compatibility(check_potcar=False, correct_peroxide=True)
compat.process_entry(entry, on_error="raise")
print(entry.energy)
```

_No response_

hoangngochuong24947-gif:
The S anion correction in `MaterialsProject2020Compatibility` was applied
unconditionally whenever S was present, even when S has a positive oxidation
state (e.g. S6+ in Li4Cs4S4O16 / mp-6597). This wrongly penalizes sulfates
and other S cations.

Only apply the S correction when S is an anion (oxidation state < 0),
respecting the `strict_anions` setting exactly like the other anion
corrections (`require_bound` / `no_check` / `require_exact`). An explicit
non-negative S oxidation state is always respected — the electronegativity
fallback only kicks in when S has no oxidation state value (None, missing
key, or empty dict), in which case the correction is applied when S is the
most electronegative element. The oxidation-state guessing is moved before
the sulfide/oxide correction blocks so it is available when needed.

Fixes #4538

## Tests
- Added regression tests: S6+ in Li4Cs4S4O16 receives only the oxide
  correction; S2- in FeS still receives the S correction; explicit
  non-negative S (e.g. {"S": 2.0}) is never corrected even when S is most
  electronegative; None / empty-dict oxidation states fall back to the
  electronegativity behavior; and the S gate honors `strict_anions`
  (`require_bound` vs `no_check` vs `require_exact`).
- `pytest tests/analysis/compatibility/` passes (187 passed).
- `ruff check` and `ruff format --check` pass.


shyuep (commented):

Solid fix for #4538 — gating the S correction on oxidation state with the same `strict_anions` semantics as other anions is the right call, and the test coverage (S6+ sulfate, S2- sulfide, explicit non-negative S, missing/None/empty oxi states, all three strict_anions modes) is thorough. Verified `sorted_elements` is defined before the relocated block, so the electronegativity fallback is safe.

Two minor points:

1. Under `strict_anions="require_exact"` with a *missing* S oxidation state, the correction is now never applied (the electronegativity fallback is overridden). That's arguably the correct strict behavior, but worth confirming intent and a test noting it.
2. The oxi-state guessing block was moved earlier, so `entry.data["oxidation_states"]` is now populated even for entries that short-circuit later — same behavior as before for S-containing entries, but it slightly widens which entries get mutated. Fine, just noting.

CI has not run on this branch — needs workflow approval. Otherwise LGTM.

hoangngochuong24947-gif:
Thanks for the thorough review — and for catching the `require_exact` edge case.

**On point 1** — agreed, and I've made the intent explicit. `S` isn't in `MP2020_ANION_OXIDATION_STATE_RANGES` (O and S are deliberately handled separately), so that `require_exact` branch could never match: the range check was effectively dead code that read as if it *might* apply. I replaced it with an explicit `apply_correction = False` plus a comment stating the intent, and added `test_sulfide_correction_require_exact_with_missing_oxidation_state` to pin the behavior (a missing S oxidation state under `require_exact` → no S correction, even when S is the most electronegative element).

**On point 2** — the move is required: the S gate has to read `oxidation_states`, so the guess has to run first. I went back and checked, and I believe the population scope is actually unchanged — the guess only runs for multi-element entries (the single-element early return sits above it in both old and new), so it's just earlier now rather than wider. If you had a specific case in mind where that differs, happy to dig in.

`tests/analysis/compatibility/` passes locally (104 passed), and `ruff check` / `ruff format --check` are clean. CI still needs workflow approval on my end whenever you get a chance.

Thanks again!

shyuep (commented):
*Automated PR review generated by Claude on behalf of @shyuep.*

Re-review after new commits (2026-08-13). The approach is sound: gating the S correction on oxidation state < 0, moving the oxi-state guess ahead of the sulfide block, and mirroring the `strict_anions` semantics of the other anion corrections. Regression test coverage is thorough (mp-6597 sulfate case, explicit non-negative S, None/missing/empty fallbacks, all three `strict_anions` modes).

Two points before merge:

1. Under `require_exact`, the S correction is now *never* applied — including to genuine sulfides like FeS. Since S has no entry in `MP2020_ANION_OXIDATION_STATE_RANGES` this is internally consistent, but it silently changes energies for all sulfides for `require_exact` users. Please confirm intended and call it out in the docstring/changelog.
2. CI has not run on this branch (first-time contributor workflows need approval) — a maintainer should trigger it before merge.

hoangngochuong24947-gif:
Thanks for the re-review — and you were right to flag point 1: forbidding the S correction under `require_exact` would silently drop the correction for *all* genuine sulfides (e.g. FeS), which is a real behavior regression, not just a strictness detail.

I've changed the approach: since S (like O) has no entry in `MP2020_ANION_OXIDATION_STATE_RANGES` and is classified separately via `sulfide_type`, the exact anion definition for S *is* an oxidation state < 0. So for S, `require_exact` now behaves like `require_bound`:

- genuine sulfide anion (S2-, oxidation state < 0) → S correction is still applied;
- S cations (e.g. S6+ in sulfates such as mp-6597) → never corrected (the #4538 fix);
- missing/None S oxidation state → the electronegativity fallback is *not* used under `require_exact` (keeps the strict semantics).

This is documented in the `strict_anions` docstring and in the inline comment, and I added `test_sulfide_correction_still_applied_under_require_exact` (FeS/S2- under `require_exact` still receives the S correction) plus updated the existing `require_exact` assertions to match. I also added a CHANGES.md entry noting the S-cation fix.

Verified locally: `tests/analysis/compatibility/` passes (105 passed in test_compatibility.py; 142 with test_computed_entries.py), `ruff check` and `ruff format --check` are clean.

**On point 2** — CI still needs a maintainer to approve the workflow run for this first-time contributor branch whenever convenient. Thanks!

shyuep (commented):
*Automated PR review generated by Claude on behalf of @shyuep.*

Re-review of `de54728`. The previous concern is resolved: `require_exact` now still corrects genuine sulfides (oxidation state < 0) and only excludes S cations, the docstring explains why S/O bypass `MP2020_ANION_OXIDATION_STATE_RANGES`, and `test_sulfide_correction_still_applied_under_require_exact` locks it in. Logic and test coverage look correct.

One remaining inconsistency:

1. `strict_anions="no_check"` is documented as restoring "the original behavior described in the associated publication", but with `no_check` an S cation (e.g. S6+ in mp-6597) still gets no correction — `apply_correction` is only ever set for `oxidation_state < 0`. That is the intended bug fix, but the docstring now overstates what `no_check` does. Suggest narrowing it to "restores the original fractional-oxidation-state behavior" or similar.
2. Minor: when the S oxidation state is exactly `0.0` the correction is silently dropped, whereas the old electronegativity fallback would have applied it. Probably fine, but worth a one-line comment since it is a different path from `None`/missing.

CI still has not run on this branch (first-time-contributor workflow approval) — a maintainer should trigger it before merge. Otherwise LGTM.

shyuep:

LGTM overall — this correctly gates the S anion correction on oxidation state with the same `strict_anions` semantics as other anions, fixing the sulfate over-penalization (#4538). Verified that `sorted_elements` is defined before the relocated sulfide block, so hoisting the oxi-state guessing is safe. The documented decision that `require_exact` behaves like `require_bound` for S (no entry in `MP2020_ANION_OXIDATION_STATE_RANGES`) is reasonable and explicitly tested rather than silently dropping the correction for sulfides.

Two minor points: (1) the electronegativity fallback now only applies when the S oxidation state is missing/None — an entry with `sulfide_type` explicitly set to `"sulfide"` but a guessed positive S state will now lose the correction; that's the intent, but it changes results for pre-tagged MP entries, so worth a sanity check against a few known MP sulfides. (2) No CI results available yet (checks likely awaiting approval); tests as written look comprehensive (cation, anion, explicit non-negative, None/empty fallback, all three strict_anions modes).

shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep). Updated review following new commits since the last automated review (2026-08-16).

The approach is correct: gating the S correction on oxidation state < 0 with the same `strict_anions` semantics as the other anion corrections fixes the spurious penalty on sulfates (#4538), and treating "require_exact" as "require_bound" for S (which has no `MP2020_ANION_OXIDATION_STATE_RANGES` entry) is the right call — silently dropping the correction for all sulfides under "require_exact" would have been a worse bug. Test coverage is thorough: the mp-6597 sulfate case, FeS regression, explicit non-negative S, None/missing/empty oxidation states, fractional (-1, 0) states across all three `strict_anions` modes, and the require_exact edge cases.

Comments:

1. Precedence question: when `entry.data["sulfide_type"]` is explicitly set to "sulfide" by the caller, a *guessed* positive S oxidation state can now veto that explicit classification. Arguably explicit `sulfide_type` metadata should take precedence over `oxi_state_guesses` output (only an explicit oxidation state should override it). Worth a maintainer decision before merge.
2. Behavior change: entries previously receiving the S correction as S cations will shift by 0.503 eV/S, changing hull distances for affected chemistries. Deliberate, but downstream MP-compatibility consumers should be flagged.
3. Moving the oxidation-state guessing before the sulfide block is fine — it ran unconditionally later in the same method anyway, so net cost is unchanged.
4. CI has not run on this PR ("no checks reported") — a maintainer needs to approve workflow runs; the full compatibility test suite should pass before merge given the blast radius of `get_adjustments`.

Overall: sound fix; resolve (1) and get CI green before merging.

shyuep:

Correct and carefully scoped fix for #4538. Gating the S correction on `oxidation_state < 0` with the same `strict_anions` semantics as the other anion corrections is the right design, the `require_exact`-behaves-like-`require_bound` rationale for S is well documented in both the docstring and inline comments, and the regression tests cover the important branches (S6+ sulfate, S2- sulfide, explicit non-negative S, missing/None fallback, all three `strict_anions` modes). Moving the oxidation-state guess ahead of the sulfide block is safe — `sorted_elements` is defined before first use, and the later anion block reuses `entry.data["oxidation_states"]` unchanged.

Two points before merge:

1. **This changes corrected energies for deployed MP entries.** Any S-cation compound currently receiving the sulfide correction (e.g. mp-6597) will shift by `comp_correction["S"] * n_S` relative to what the MP database serves today. The fit itself is unaffected (fitting data are genuine sulfides), but this should be flagged to the MP data team and noted as a behavior change in the release notes so downstream Ehull values are understood to move.
2. Minor: repos generally auto-generate `docs/CHANGES.md` at release; the manual entry may create a merge conflict, but no action needed beyond maintainer preference.

CI has not run on this branch yet (first-time contributor); a maintainer should approve the workflow run.

shyuep:
@rkingsbury Can you review this pls? This is related to the correction scheme that you did for MP a while back. Thanks.

shyuep:

Real bug, and the fix is in the right place. Note for context: for entries that carry a structure, `sulfide_type()` already classifies sulfates correctly, so the incorrect S correction only hits entries reaching `get_adjustments` without a structure — where `sf_type` defaults to `"sulfide"`. That is worth stating explicitly in the PR body, because it bounds the blast radius. CI green.

**Main concern: the guard now overrides an explicitly supplied classification**

The oxidation-state check runs even when `entry.data["sulfide_type"] == "sulfide"` was set by the caller (or by an upstream builder). If `oxi_state_guesses` returns a poor guess for a genuine sulfide, the correction is now silently dropped where it previously applied. Since MP2020 is a *fitted* scheme, dropping a correction is as wrong as adding one. Suggestion: short-circuit to `apply_correction = True` when `sulfide_type` was explicitly present in `entry.data`, and only apply the oxidation-state gate on the fallback path.

**Other points**

1. **Reproducibility against published MP2020.** Any change to which entries receive an anion correction shifts formation energies relative to the published parameterization [Wang et al., npj Comput. Mater. 2021 — *BibTeX key needs replacement*]. Worth confirming with the MP side that no entries currently in the database change energy, and saying so in the PR.

2. **Hoisting `oxi_state_guesses`.** It now runs before the S and O blocks. `oxi_state_guesses(max_sites=-20)` is expensive; previously any early exit in the S/O blocks would have skipped it. Please confirm there is no path that used to return before this point — otherwise this is a throughput regression on large entry sets (MP-scale corrections runs are millions of entries).

3. **`require_exact` special case.** Making `require_exact` behave as `require_bound` for S is pragmatic but surprising given the parameter name. It's documented in the docstring, which is good — consider also emitting nothing rather than a warning, and mention it in the class-level docs where users are most likely to look.

4. **Symmetry with O.** The docstring now says O and S are both exceptions, but only S got the oxidation-state gate. Is the equivalent O⁶⁺-style case unreachable in practice (peroxide/superoxide/ozonide classification covers it)? A comment saying why O needs no equivalent guard would prevent this being re-opened later.

5. **Test coverage is strong** — eight new cases covering cation, anion, missing, empty, fractional, and all three `strict_anions` modes. Nice. One gap: no test with `sulfide_type` explicitly set to `"sulfide"` *and* a conflicting positive oxidation state, which is exactly the interaction in point 1.

rkingsbury:
Hi @hoangngochuong24947-gif , thanks for the thoughtful and thorough work on this. Your work here seems solid (cleanly implemented, documented, with unit tests), but I'm unsure whether this is truly the right approach, because (as Claude / @shyuep noted), it sort of bypasses the `sulfide_type` setting.

> no test with sulfide_type explicitly set to "sulfide" and a conflicting positive oxidation state, which is exactly the interaction in point 1.

If I understand correctly, this bug occurs because for some structures, `sulfide_type` defaults to `sulfide` _even though S has a positive oxidation state_. **That** seems like the core problem to be addressed. As I noted in #4538 :

> As with oxidation states, sulfide_type is normally populated (correctly) during the MP build process. The fundamental bug here is that sulfide_type defaults to sulfide (see line 1110). So, when you built your entries manually, you got sulfide (since there was no Structure or `.data["sulfide_type"] key) and hence got the erroneous correction.

> ### Recommended solution
> This should be changed so that sulfide_type either defaults to None, or that there is a basic oxidation state check similar to what we do for the other anions, and only if S is anionic do we assume sulfide.

It seems to me that defaulting `sulfide_type` to `None` when there is no structure supplied could address this bug in a less intrusive way. That way, with `sulfide_type=None`, `oxi_state_guesses` should still be triggered, allowing the more reliable identification of S cations.

I think this was actually the major purpose behind `oxide_type` and `sulfide_type` - to prevent unnecessary calls to `oxi_state_guesses` during the build process - see next point

### MP production implications

Claude's point 2 is also important for large-scale database builds:

> Hoisting oxi_state_guesses. It now runs before the S and O blocks. oxi_state_guesses(max_sites=-20) is expensive; previously any early exit in the S/O blocks would have skipped it. Please confirm there is no path that used to return before this point — otherwise this is a throughput regression on large entry sets (MP-scale corrections runs are millions of entries).

`oxi_state_guesses` is, in general, expensive to run, so this change could impact the build process. I'm not in a good position to assess whether that's a serious concern at this point or not.

### Utility of `sulfide_type`?

I'm not well-acquainted with the inner workings of how `sulfide_type` is assigned or whether it (still) holds value for MP / pymatgen. As-written, this PR seems to move in the direction of trusting the oxidation states and discarding `sulfide_type`, which may be a defensible solution.

The fundamental tension here is how to take advantage of "expensive, detailed" informatoin like oxi states and `sulfide_type` added during the MP build process (which often requires analysis of the structure) while still returning reasonable results with manually-created, composition-only entries as you identified.

I hope this helps @shyuep and let me know if you have any questions @hoangngochuong24947-gif

hoangngochuong24947-gif:
Thanks @rkingsbury and @shyuep for the insightful feedback! 

Ryan's point makes complete sense — relying on `sulfide_type` precedence and defaulting to `None` for unclassified entries is much cleaner than trying to force oxidation-state checks everywhere.

I've pushed an update addressing all points:

1. **Strict precedence for `sulfide_type`**: If `entry.data["sulfide_type"]` is explicitly present or derived via `entry.structure`, we respect it directly and apply the correction without touching `oxi_state_guesses`. This ensures upstream MP build classifications are never vetoed and keeps production throughput intact.
2. **Fallback only for composition-only entries**: Only when both structure and explicit metadata are absent does it fall back to oxidation states, correctly filtering out S cations like S⁶⁺ in sulfates (#4538).
3. **Lazy oxidation state evaluation**: `oxi_state_guesses` is now called on demand, avoiding upfront overhead when structures or explicit tags are present.
4. **Test coverage & docs**: Added regression tests verifying that explicit `sulfide_type="sulfide"` and structure-derived classifications take precedence even with conflicting positive oxidation states. Also clarified the `no_check` docstring and added an inline note regarding O.

All 107 compatibility tests are passing locally. Let me know if anything else needs tweaking!

shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff only, CI has not run on this branch — advisory).

Review of `def8e0e5`. This follows @rkingsbury's direction and resolves the main concern from previous rounds: an explicit `entry.data["sulfide_type"]` or a structure-derived classification is now terminal and is never vetoed by `oxi_state_guesses`; the oxidation-state gate fires only on the composition-only fallback. Making `_get_oxidation_states()` lazy also answers the throughput concern — it is now called strictly less often than on `main` (only from the S fallback or when a listed anion is present). The two new precedence tests are the right additions.

One real issue with the new default:

1. **`sulfide_type(structure)` returns `None` for sulfates** (docstring: "sulfide/polysulfide or None if structure is a sulfate"). With `sf_type = None` now meaning "unclassified", a `ComputedStructureEntry` that the structure analysis positively identified as a sulfate drops into the composition-only fallback and runs `oxi_state_guesses`. On `main`, that `None` was terminal (no S correction). The fallback will usually still land on "no correction" because O/F/Cl outrank S in electronegativity, but (a) it forfeits the lazy-evaluation benefit for exactly the structure-bearing entries this revision was meant to protect, and (b) a bad composition-level guess (S < 0 on an unusual mixed-valence formula) can now apply the correction to a structurally identified sulfate. Suggest tracking whether a classification source existed (e.g. `classified = True` in the explicit/structure branches) and entering the fallback only when `not classified`, plus a test with a sulfate `ComputedStructureEntry` asserting no S correction and no `oxi_state_guesses` call.

Minor:

2. The "Failed to guess oxidation states" warning now fires only when `_get_oxidation_states()` is actually consulted, so an entry with a pre-populated empty `oxidation_states` dict and no S or listed anion is silent where `main` warned. Reasonable, but it is a behavior change worth a one-line comment.

CI: no checks have run on this branch (first-time contributor) — a maintainer needs to approve the workflow run; the full `tests/analysis/compatibility` suite must be green before merge given the blast radius of `get_adjustments`.

hoangngochuong24947-gif:
Thanks @shyuep for the sharp review! That's a great catch regarding `sulfide_type` returning `None` for sulfates.

I've pushed an update addressing this:
1. **Explicit `classified` flag**: Introduced a `classified = True` flag when either `entry.data.get("sulfide_type")` or `entry.structure` is present. When `classified`, we directly set `apply_correction = (sf_type == "sulfide")`. For structured sulfates (`sf_type is None`), this immediately short-circuits with `apply_correction = False` without falling back to composition-level oxidation-state guessing.
2. **Preserved laziness & zero guessing on structured entries**: Structured sulfates now never trigger `_get_oxidation_states()`, avoiding any unnecessary `oxi_state_guesses` overhead or potential misclassification.
3. **Regression test**: Added `test_sulfide_correction_structure_sulfate_no_fallback` in `test_compatibility.py` verifying that a sulfate `ComputedStructureEntry` receives no S correction and has zero calls to `_get_oxidation_states()` (asserting `"oxidation_states"` remains absent from `entry.data`).
4. **Clarified warning note**: Added an inline comment in `_get_oxidation_states()` explaining that entries without S or listed anions skip the helper, so unguessable oxidation states are only warned when an anion correction actually requires them.

All 108 compatibility tests are passing locally.

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep; not a human review — please sanity-check before acting.)

The latest revision (`b930e494`) looks solid. The `classified` short-circuit is the right structure: explicit `sulfide_type` metadata and structure-derived classification take precedence, and the oxidation-state gate only runs on the composition-only path that actually caused #4538. `sorted_elements` is defined at the top of `get_adjustments` well before the new use at line 1175, so no ordering problem. Hoisting the oxidation-state guess into `_get_oxidation_states()` is a net win — `oxi_state_guesses(max_sites=-20)` is now skipped entirely for entries with no S and no corrected anion. Test coverage is thorough (S6+/S2-/explicit non-negative/None/empty-dict/fractional × all three `strict_anions` modes). Full CI is green.

Two things I would want the maintainer to sign off on explicitly, since MP2020 is a fitted scheme and this changes which entries get corrected:

**1. `strict_anions="no_check"` no longer restores publication behavior for S.** Under `no_check`, an S cation now receives no correction. The docstring was updated to "disable the fractional-oxidation-state filter", which describes the new semantics accurately but quietly retires the previous "restore the original behavior described in the associated publication" guarantee. This is the intended fix, not a defect — but anyone using `no_check` to reproduce published MP2020 numbers on composition-only sulfate entries will now get different formation energies. Worth an explicit note.

**2. The "Failed to guess oxidation states" warning is now conditional.** Previously it fired for any multi-element entry with unguessable states; now only when an S or listed-anion correction actually needs them. Better signal-to-noise, but it is a silent change for downstream code that filters on that warning.

Minor, non-blocking:

- `_get_oxidation_states() -> dict` could be `-> dict[str, float]`.
- The `nonlocal warned_empty_oxi` closure inside an already-long method is a little awkward; a small private helper taking the entry and returning `(states, guessed_empty)` would read better. Not worth churn if you prefer to keep the diff tight.
- A structured entry whose `sulfide_type()` heuristic returns `None` is treated as `classified` and never reaches the oxidation-state fallback. That is deliberate and documented in the comment, but it does mean the fix is inert for `ComputedStructureEntry` — fine, since those were never affected by #4538.

No correctness objections. Ready to merge from my side once the `no_check` semantics are acknowledged.

shyuep:
@kavanase Do you mind taking a look at this since you have been working on the compatibility scheme recently? Thanks!

kavanase:
Thanks @hoangngochuong24947-gif for raising this!
This looks like a useful fix for the sulfate anion correction bug. From looking through, the one issue I could spot is that it ignores the anion correction for polysulfides where the effective S oxidation state is in the (-1, 0) range, which I would guess is not intended? (Though maybe it is?); e.g. Na2S3, BaS3, Na2S5, K2S6 etc.
This would have a significant effect on the predicted formation energies, so worth thinking about whether this is desired.

Also seems like the lazy `_get_oxidation_states` function could just be the existing block moved up, since it already guesses oxidation states for every multi-element entry?

PR #4700, opened 2026-08-11
 [breaking] Support mixed-supercell orderings in `HeisenbergMapper `via a shared parent-cell sublattice definition

### LINKED ISSUE #4667: `HeisenbergMapper` silently returns wrong exchange parameters when orderings live in different-sized supercells
### Python version

Python 3.12.13

### Pymatgen version

2026.5.4 (reproduced on `master` @ `d6b4ea50a0`, 2026-06-02)

### Operating system version

Rocky Linux 8.10

### Current behavior

`HeisenbergMapper` is documented to take a set of low-energy magnetic orderings,
map them onto a classical Heisenberg model, and solve for the exchange
parameters `J_ij`. It implicitly assumes **every ordering lives in the same
supercell**. When the input orderings instead live in *different-sized*
supercells of a common parent cell — which is the normal output of a magnetic
enumeration workflow — `get_exchange()` returns numbers that are silently and
badly wrong. No exception is raised.

This is caused by two independent bugs that both surface only once the supercell
sizes differ:

1. **Site identifiers are built from a single ordering.** In `__init__`:

   ```python
   self.unique_site_ids, self.wyckoff_ids = self._get_unique_sites(ordered_structures[0])
   ```

   `_get_unique_sites(structure)` runs a `SpacegroupAnalyzer` on **only the first
   ordering** and returns one dict mapping that ordering's site indices to
   identifiers. `_get_exchange_df` then reuses that single dict for **every**
   graph:

   ```python
   for k, v in unique_site_ids.items():
       if idx in k:        # idx is a site index in *this* ordering's supercell
           i_index = v
   ```

   For any ordering larger than `ordered_structures[0]`, site indices beyond the
   first cell's size are simply absent from the dict, so `i_index`/`j_index`
   stay `None`, the interaction is mislabeled (`"None-..."`) and dropped. The
   identifiers are also not shared across orderings — the same crystallographic
   orbit can get different ids in different supercells — so the `J_ij` columns
   are not comparable from one row of the fit matrix to the next.

2. **The fit mixes intensive and extensive quantities.** `HeisenbergScreener._do_cleanup`
   normalizes the energies to *per magnetic ion* (intensive):

   ```python
   energies = [e / len(s) for (e, s) in zip(energies, ordered_structures, strict=True)]
   ```

   and `_get_exchange_df` fixes the `E0` (nonmagnetic) column coefficient at `1`.
   But the spin-product interaction columns in the same matrix are accumulated as
   a sum over **the whole supercell** (extensive) — every site, every bond, only
   a global `.div(2)` at the end:

   ```python
   ex_row.loc[sgraph_index, j_ij] -= s_i * s_j   # summed over all sites in the cell
   ...
   ex_mat[j_columns] = ex_mat[j_columns].div(2)
   ```

   With one fixed supercell size this is relatively benign (the extra factor of `N` is a
   constant). With mixed sizes the left-hand side (per ion) and the columns (per
   supercell) are on different footings, the linear system is inconsistent, and
   the recovered `J`/`E0` are wrong and scale-dependent.

### Why the current test suite misses this

Every ordering in `tests/files/analysis/magnetic_orderings/Mn3Al.json` already
lives in the same supercell, so both bugs cancel (constant `N`, and every
ordering's site indices are present in the single dict). The failure mode only
appears once the input orderings differ in size, which the existing tests never
exercise.

### Suggested direction for a fix

- Compute one `unique_site_ids` map **per ordering**, anchored to a common
  parent cell, so a given crystallographic orbit receives the same identifier in
  every ordering regardless of supercell size; have `_get_exchange_df` select the
  map for the ordering it is currently iterating. Orderings that are not a
  supercell of the parent should raise rather than be silently mislabeled.
- Put the interaction columns on the **same per-ion footing as the energies**
  (e.g. normalize each row by its site count) so orderings in different-sized
  supercells contribute consistent rows to the linear fit.

I am currently writing a PR for this on my fork and would appreciate any feedback once I publish it.

### Expected Behavior

The recovered exchange parameters should be independent of the supercell sizes
of the input orderings and should equal the parameters of the underlying
Heisenberg model. Orderings in different supercells of one parent cell should be
a fully supported input (it is the natural output of a magnetic-ordering
enumeration), not an unstated precondition.

### Minimal example

```Python
"""
A clean test on a known Hamiltonian: a 2D square lattice of `Fe` with
`E0 = -5.0 eV/ion`, NN coupling `J_nn = +0.012 eV`, NNN coupling
`J_nnn = -0.004 eV`. Energies are assigned analytically from
`E = N * E0 - sum_<ij> J_ab s_i s_j`, so the fit must recover `E0`, `J_nn`,
`J_nnn` exactly, regardless of supercell size. The three orderings deliberately
live in different cells: FM in the 1×1 (1 ion), stripe-AFM in 1×2 (2 ions),
Néel-AFM in 2×2 (4 ions).
"""

import numpy as np
from pymatgen.analysis.magnetism.heisenberg import HeisenbergMapper

E0, J_NN, J_NNN = -5.0, 0.012, -0.004
NN  = ((1, 0), (-1, 0), (0, 1), (0, -1))
NNN = ((1, 1), (1, -1), (-1, 1), (-1, -1))

def make_structure(spins):
    spins = np.atleast_2d(np.asarray(spins, float))
    n_y, n_x = spins.shape
    lattice = Lattice.from_parameters(n_x, n_y, 10, 90, 90, 90)
    coords = [[x / n_x, y / n_y, 0.5] for y in range(n_y) for x in range(n_x)]
    magmoms = [spins[y, x] for y in range(n_y) for x in range(n_x)]
    return Structure(lattice, ["Fe"] * (n_x * n_y), coords, site_properties={"magmom": magmoms})

def energy(spins):
    spins = np.atleast_2d(np.asarray(spins, float))
    n_y, n_x = spins.shape
    e_ex = 0.0
    for y in range(n_y):
        for x in range(n_x):
            s_i = spins[y, x]
            nn  = sum(s_i * spins[(y + dy) % n_y, (x + dx) % n_x] for dx, dy in NN)
            nnn = sum(s_i * spins[(y + dy) % n_y, (x + dx) % n_x] for dx, dy in NNN)
            e_ex -= 0.5 * (J_NN * nn + J_NNN * nnn)   # 1/2: each bond counted from both ends
    return n_x * n_y * E0 + e_ex

all_spins = [[[1]], [[1], [-1]], [[1, -1], [-1, 1]]]   # FM 1x1, stripe 1x2, Neel 2x2
structures = [make_structure(s) for s in all_spins]
energies   = [energy(s) for s in all_spins]

hm = HeisenbergMapper(structures, energies, cutoff=1.5, tol=0.02)
print("unique_site_ids:", hm.unique_site_ids)
print("exchange:", hm.get_exchange())


**Observed (master):**


unique_site_ids: {(0,): 0}                 # built from the 1-site FM cell only
exchange: {'E0': -4.968, '0-0-nn': 40.0, '0-0-nnn': -16.0}


**Expected:**


exchange: {'E0': -5.0, '0-0-nn': 12.0, '0-0-nnn': -4.0}   # J in meV


`E0` collapses onto the per-ion energy of one ordering and the `J` values are off
by factors of ~3–4. The result is also unstable under changing which supercells
are supplied.
```

- `src/pymatgen/analysis/magnetism/heisenberg.py`
  - `HeisenbergMapper.__init__` — `_get_unique_sites(ordered_structures[0])`
  - `HeisenbergMapper._get_unique_sites`
  - `HeisenbergMapper._get_exchange_df`
  - `HeisenbergScreener._do_cleanup` — per-ion energy normalization

The tests I built for investigating this are attached:

[test_heisenberg_fit.py](https://github.com/user-attachments/files/28960296/test_heisenberg_fit.py)
[test_heisenberg_mno.py](https://github.com/user-attachments/files/28960297/test_heisenberg_mno.py)

Luguza:
### Summary
Refactors `HeisenbergMapper` so magnetic sublattices are defined once on a common parent cell, and every DFT-relaxed ordering is mapped onto it by structure matching rather than by direct index comparison. This decouples exchange-parameter fitting from any single ordering's cell size/shape, so orderings that were relaxed in different-sized supercells of the same parent can now be combined in one fit.

Closes #4667.

### Motivation
The old implementation identified "unique sites" from the geometry of `ordered_structures[0]` and assumed every other ordering shared that exact indexing/cell. Orderings enumerated by a `MagneticOrderingsWF`-style workflow frequently don't: different orderings can relax into different (but commensurate) supercells of the same parent, and the mapper had no way to reconcile them. This branch introduces an explicit parent cell as the single source of truth for sublattice identity, with each ordering's sites matched onto it via `StructureMatcher(attempt_supercell=True)`.

### Key changes

1. New MagneticOrdering / ParentOrdering / RelaxedOrdering classes (heisenberg.py) - each ordering now owns its magnetic-only structure, its NN graph, and its parent-sublattice labels. ParentOrdering derives sublattices from the parent's Wyckoff orbits (nonmagnetic ions kept in for correct site symmetry); RelaxedOrdering maps onto the parent via get_s2_like_s1 and raises a clear ValueError if it isn't a supercell of the parent.
2. HeisenbergMapper(..., parent=None, ...) - new optional parent argument. If omitted, the parent is inferred as the primitive cell of the lowest-energy ordering (documented caveat: only safe if that ordering hasn't broken the paramagnetic symmetry).
3. SublatticeMinimumDistanceNN - new NN strategy that takes the nearest shell per sublattice pair instead of one global nearest-neighbor window, so intra-sublattice bonds longer than the shortest inter-sublattice bond no longer silently drop out of the graph.
4. Interaction labeling by nearest-shell-of-a-pair, not fixed ±tol windows.
5. Least-squares fit over all supplied orderings (get_exchange) replacing the old square-system solve - surplus orderings now average out DFT-energy noise instead of being discarded; added an ill-conditioned/rank-deficient warning and an RMS residual (meV/magnetic ion) as a fit-quality metric.
6. threshold_nonmag handling so induced moments on nominally nonmagnetic ions don't get miscounted as magnetic sites and change site counts/graph topology between orderings.
7. Removed <J> dependency from the core fit path; estimate_exchange and get_mft_temperature are now @deprecated (deadline 2027-08-01) in favor of get_exchange's shell-resolved J_ij and an external Monte Carlo solver, respectively.
8. HeisenbergModel.from_dict fixed to handle unique_site_ids serialized as lists, and ex_mat serialization corrected (per PR #4664).
9. Units/docstring corrections throughout (J_ij in meV/μ_B² on raw moments vs. the meV-normalized-spin convention used by get_interaction_graph/VAMPIRE/TB2J).

### Testing
All tests pass locally, including new coverage added on this branch:

`TestHeisenbergMapperKnownHamiltonian `- exchange recovery, RMS residual, degenerate-ordering dropping, Wyckoff-orbit sublattices, ill-conditioned-fit warning, shell counting per sublattice pair, interaction-graph consistency, incompatible (non-supercell) ordering rejection
`TestHeisenbergMeanFieldTemperature `- single/multi-sublattice MFT, exchange sign vs. ground state
`TestHeisenbergMapperFullCellSymmetry `- nonmagnetic ions splitting sublattices, symmetry-equivalent sites sharing one
`TestHeisenbergMapperZeroMomentIon `- quenched-moment species staying on the magnetic lattice

### DIFF
--- a/src/pymatgen/analysis/magnetism/heisenberg.py
+++ b/src/pymatgen/analysis/magnetism/heisenberg.py
@@ -7,9 +7,9 @@ model.
 from __future__ import annotations
 
 import logging
+from abc import ABC, abstractmethod
 from ast import literal_eval
 from typing import TYPE_CHECKING
-from abc import ABC
 
 import numpy as np
 import pandas as pd
@@ -26,14 +26,177 @@ from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
 if TYPE_CHECKING:
     from typing import Self
 
-__author__ = "ncfrey"
-__version__ = "0.1"
-__maintainer__ = "Nathan C. Frey"
-__email__ = "ncfrey@lbl.gov"
+# __author__ = "ncfrey"
+# __version__ = "0.1"
+# __maintainer__ = "Nathan C. Frey"
+# __email__ = "ncfrey@lbl.gov"
+# __status__ = "Development"
+# __date__ = "June 2019"
+
+__author__ = "Luca Frey"
+__version__ = "0.2"
+__maintainer__ = "Luca Frey"
+__email__ = "luca.frey@student.kit.edu"
 __status__ = "Development"
-__date__ = "June 2019"
+__date__ = "July 2026"
 
 logger = logging.getLogger(__name__)
+
+DEFAULT_TOL = 0.05  # Angstrom tolerance
+
+
+def _analyzer(structure: Structure, **kwargs) -> CollinearMagneticStructureAnalyzer:
+    """CollinearMagneticStructureAnalyzer factory with the settings used throughout this module.
+
+    ``make_primitive=False`` keeps the cell as given, ``threshold=0.0`` retains any nonzero
+    moment, and ``threshold_nonmag=100.0`` always zeroes out nonmagnetic ions, so induced
+    magnetic moments don't change the number of magnetic sites per unit cell between orderings.
+
+    Extra keyword arguments are passed through to the analyzer.
+    """
+    return CollinearMagneticStructureAnalyzer(
+        structure, **{"make_primitive": False, "threshold": 0.0, "threshold_nonmag": 100.0, **kwargs} # last value wins if there are multiple conflicting values for the same key -> defaults overridden by kwargs
+    )
+
+
+
+class MagneticOrdering(ABC):
+    """A single collinear magnetic configuration on a common parent lattice.
+
+    Base class holding one ordering's structure, its magnetic-only reduction,
+    its NN graph, and the parent-sublattice labels of its magnetic sites.
+    """
+
+    def __init__(self, structure: Structure, cutoff: float = 0, tol: float = DEFAULT_TOL):
+        self.structure = structure
+        self.analyzer = _analyzer(structure)
+        self.cutoff = cutoff
+        self.tol = tol
+        self.nn_graph = self._get_nn_graph()
+        self.magn_species: set[str] = {site.specie.symbol for site in self.magn_substructure}
+
+        self.sublattice_ids_dict: dict[tuple[int, ...], int] = {}
+        self.sublattice_labels: list[int] | None = None
+        self.sublattice_wyckoff_symbols: dict[int, str] = {}
+
+    @property
+    def magn_substructure(self) -> Structure:
+        return self._magnetic_only(self.structure)
+
+    @staticmethod
+    def _nonmagnetic(structure) -> Structure:
+        """Nonmagnetic copy of a structure (moments zeroed, all ions kept)."""
+        s0 = _analyzer(structure).get_nonmagnetic_structure(make_primitive=False)
+        if "wyckoff" in s0.site_properties:
+            s0.remove_site_property("wyckoff")
+        return s0
+
+    @staticmethod
+    def _magnetic_only(structure) -> Structure:
+        """Magnetic-only copy (nonmagnetic ions removed).
+
+        Whether a site is magnetic is decided by its *species*: only species listed in
+        CollinearMagneticStructureAnalyzer's default magmoms are treated as magnetic.
+
+        - Magnetic-species sites keep their existing magmom (any nonzero value is retained,
+          since ``threshold=0.0``). If the input has no magmom site property, the default magmoms are assigned (``replace_all_if_undefined``).
+        - Nonmagnetic-species sites always get a magmom of zero. Because ``threshold_nonmag=100.0``,
+          even induced moments on nonmagnetic ions are zeroed out, so the number of magnetic sites
+          per cell stays constant across orderings.
+
+        Finally, all sites with zero magmom are dropped (even if magnetic-species sites with magmom=0. or magmom=None) and only sites with nonzero magmom are kept.
+        """
+        # overwrite_magmom_mode: if the structure has no magmom site property, assign default magmoms to all magnetic species
+        return _analyzer(structure).get_structure_with_only_magnetic_atoms(make_primitive=False)
+
+    @staticmethod
+    def _sublattice_ids_dict(labels: list[int]) -> dict[tuple[int, ...], int]:
+        """Group site indices by sublattice id: {tuple(site indices): sublattice id}."""
+        groups: dict[int, list[int]] = {}
+        for idx, sub_id in enumerate(labels):
+            groups.setdefault(sub_id, []).append(idx)
+        return {tuple(indices): sub_id for sub_id, indices in groups.items()}
+    
+    def _get_nn_graph(self) -> StructureGraph:
+        """
+        NN StructureGraph of the magnetic-only structure.
+        
+        If self.cutoff is set, the graph includes all neighbors within that distance. 
+        Otherwise, only the nearest neighbors of each site are included.
+        """
+
+        strategy = MinimumDistanceNN(cutoff=self.cutoff, get_all_sites=True) if self.cutoff else MinimumDistanceNN()  # only NN
+        return StructureGraph.from_local_env_strategy(self.magn_substructure, strategy=strategy)
+    
+    @abstractmethod
+    def set_sublattice_ids(self) -> dict[tuple[int, ...], int]:
+        """
+        Sets the following properties for this ordering:
+            - `self.sublattice_ids_dict` = {tuple(site indices): sublattice id}
+            - `self.sublattice_wyckoff_symbols` = {sublattice id: wyckoff symbol}
+            - `self.sublattice_ids` = [sublattice id for each magnetic site in the ordering or None for nonmagnetic sites]
+
+        All nonmagnetic (`CollinearMagneticStructureAnalyzer` default_magmom=0) sites are grouped into one sublattice with label `None`.
+        
+        For `ParentOrdering`, this defines the sublattice ids by grouping symmetrically equivalent sites of ions into sublattices (nonmagnetic ions present for symmetry analysis). 
+        
+        For `RelaxedOrdering`, this is a mapping from the ordering's site indices to the sublattice ids via the finding the corresponding site in the parent for each site in the relaxed ordering.
+        
+        Finally, the sublattice ids are stored as a site property in the ordering's structure for easy access.
+        """
+
+
+class ParentOrdering(MagneticOrdering):
+    """The nonmagnetic parent cell that defines the sublattices.
+
+    Sublattices are its symmetry orbits (computed with nonmagnetic ions present),
+    restricted to the magnetic species.
+
+    """
+
+    def __init__(self, structure: Structure, cutoff: float = 0, tol: float = DEFAULT_TOL):
+        super().__init__(structure, cutoff, tol)
+        self.structure = self._nonmagnetic(structure)  # parent is always nonmagnetic
+        self.set_sublattice_ids()
+
+    def set_sublattice_ids(self):
+        symmetrized_parent = SpacegroupAnalyzer(self.structure).get_symmetrized_structure()
+        self.sublattice_ids = [None] * len(self.structure)
+        for indices, wyckoff_symbol in zip(symmetrized_parent.equivalent_indices, symmetrized_parent.wyckoff_symbols, strict=True):
+            assert all(symmetrized_parent[i].specie == symmetrized_parent[indices[0]].specie for i in indices)
+            if symmetrized_parent[indices[0]].specie.symbol not in self.magn_species:
+                continue
+            sub_id = len(self.sublattice_wyckoff_symbols)
+            self.sublattice_wyckoff_symbols[sub_id] = wyckoff_symbol # one sublattice always has one wyckoff symbol, but multiple sublattices can share the same wyckoff symbol (e.g. two species on the same wyckoff position)
+            for index in indices:
+                self.sublattice_ids[index] = sub_id
+
+        self.structure.add_site_property("sublattice_id", self.sublattice_ids)
+
+        self.sublattice_ids_dict = self._sublattice_ids_dict(self.sublattice_ids)
+
+
+class RelaxedOrdering(MagneticOrdering):
+    """One DFT-relaxed magnetic ordering with its energy and parent back-reference."""
+
+    def __init__(self, structure: Structure, energy: float, parent_ordering: ParentOrdering, cutoff: float = 0, tol: float = DEFAULT_TOL):
+        super().__init__(structure, cutoff, tol)
+        self.parent_ordering = parent_ordering
+        self.energy = energy
+        self.set_sublattice_ids()
+
+    def set_sublattice_ids(self):
+        matcher = StructureMatcher(primitive_cell=False, attempt_supercell=True)
+
+        matched_parent = matcher.get_s2_like_s1(self._nonmagnetic(self.structure), self.parent_ordering.structure)
+        if matched_parent is None:
+            raise ValueError(
+                "The RelaxedOrdering is not a supercell of the ParentOrdering cell; it "
+                "cannot be mapped onto the parent sublattices."
+            )
+        self.sublattice_ids = [int(matched_parent[i].properties["sublattice_id"]) for i in range(len(self.structure))]
+
+
 
 
 class HeisenbergMapper:
@@ -52,7 +215,7 @@ class HeisenbergMapper:
         parent (Structure): Nonmagnetic parent cell that defines the sublattices.
         site_labels (list[list[int]]): site_labels[k][i] is the parent
             sublattice id of site i in ordering k.
-        wyckoff_ids (dict): Maps each sublattice id to its wyckoff symbol.
+        sublattice_wyckoff_symbols (dict): Maps each sublattice id to its wyckoff symbol.
         nn_interactions (dict): {i: j} pairs of NN interactions between sublattices.
         dists (dict): NN, NNN, and NNNN interaction distances.
         ex_mat (DataFrame): Heisenberg Hamiltonian (per magnetic ion) for each ordering.
@@ -60,7 +223,7 @@ class HeisenbergMapper:
             included 'E0' offset stays in eV.
     """
 
-    def __init__(self, ordered_structures, energies, parent=None, cutoff=0, tol: float = 0.02):
+    def __init__(self, ordered_structures, energies, parent, cutoff=0, tol: float = DEFAULT_TOL):
         """Exchange parameters are computed by mapping to a classical Heisenberg
         model. n+1 unique orderings are required to compute n exchange parameters.
 
@@ -74,16 +237,16 @@ class HeisenbergMapper:
             ordered_structures (list): Structure objects with magmoms.
             energies (list): Total energies of each relaxed magnetic structure.
             parent (Structure): Paramagnetic parent cell whose symmetry defines the
-                magnetic sublattices. If None, it is inferred as the primitive cell
-                of the lowest-energy ordering. Defaults to None.
+                magnetic sublattices. 
             cutoff (float): Cutoff in Angstrom for nearest neighbor search.
                 Defaults to 0 (only NN, no NNN, etc.).
             tol (float): Tolerance (in Angstrom) on nearest neighbor distances
-                being equal.
+                being equal and structure matching.
         """
         # Save original copies of inputs
         self.ordered_structures_ = ordered_structures
         self.energies_ = energies
+        self.parent_ = parent
 
         # Sanitize inputs: standardize moments, convert energies to per magnetic
         # ion, drop duplicates and sort by energy. The screener keeps both a
@@ -103,7 +266,7 @@ class HeisenbergMapper:
         # every ordering with the parent sublattice it belongs to.
         self.parent = self._get_parent(parent, self.full_structures)
         self.parent_sgraph = self._get_graph(cutoff, self._magnetic_only(self.parent))
-        self.wyckoff_ids, self.site_labels, self.parent_site_labels = self._label_orderings(self.parent, self.full_structures)
+        self.sublattice_wyckoff_symbols, self.site_labels, self.parent_site_labels = self._label_orderings(self.parent, self.full_structures)
 
         # These attributes are set by internal methods
         self.nn_interactions = self.dists = self.ex_mat = self.ex_params = None
@@ -112,7 +275,7 @@ class HeisenbergMapper:
             raise SystemExit("We need at least 2 unique orderings.")
 
         self._get_nn_dict()
-        self._get_exchange_df()
+        self._build_exchange_matmat()
 
     @staticmethod
     def _get_graph(cutoff, structure):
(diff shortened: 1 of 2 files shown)


shyuep (commented):

Substantial, well-motivated refactor (parent-cell sublattice definition + StructureMatcher mapping + least-squares fit) with unusually thorough docstrings and new test coverage. Main comments:

1. **Attribution**: don't comment out the original `__author__` block (ncfrey). Either keep both authors or drop the dunder metadata entirely — git history carries attribution.
2. `raise SystemExit(...)` was carried over from the old code — please use `ValueError` instead; `SystemExit` kills interactive sessions and notebooks.
3. This is a significant API break for `HeisenbergMapper` (attributes like `sgraphs`, `unique_site_ids`, `wyckoff_ids` are gone; `estimate_exchange`/`get_mft_temperature` deprecated). A short migration note in the PR description / docs would help downstream users.
4. CI has not run on this branch (first-time contributor — needs a maintainer to approve workflows). Please rebase on master and a maintainer should trigger CI before merge.
5. Default-parent inference from the lowest-energy ordering is a documented footgun; consider emitting a `UserWarning` when `parent=None` so users can't miss it.

Nice work overall — the shell-midpoint interaction labeling and pooled magnetic species are clear improvements.

Luguza:
@shyuep @mkhorton As per the discussion in https://github.com/materialsproject/atomate2/pull/1528, should I restore the vampire caller into pymatgen, because it is written to be specific to this workflow and relevant to local execution?

shyuep:

Substantial and well-motivated refactor — defining sublattices once on a parent cell and mapping orderings via `StructureMatcher(attempt_supercell=True)` is the right architecture for mixed-supercell fits, and the migration guide in the module docstring is excellent. Points to address:

1. **Breaking changes without deprecation shims.** Constructor arg order changes (`parent` in third position), `get_exchange` now returns a tuple, and `sgraphs`/`unique_site_ids`/`wyckoff_ids`/`ordered_structures` are removed outright. Consider deprecated property aliases for one release cycle where cheap, and mark the PR title `[breaking]` so the compatibility log picks it up.
2. **`pymatgen-core` submodule pointer bump** rides along in the diff — drop it unless this PR actually requires it.
3. **Attribution:** `__author__`/`__maintainer__` replace @ncfrey entirely; please append rather than overwrite the original authorship.
4. `SublatticeMinimumDistanceNN` is a nice generalization; the cached `nn_graph` guard raising before `sublattice_ids` are set is good defensive design.
5. No CI results available yet (checks likely awaiting approval) — review based on code analysis only.

Luguza:
Agreed on the title, I'll mark it [breaking] so it shows up in the compatibility log.

On the deprecation shims, I'd rather not add them here, and it's not only about cost. Taking them one at a time:

`sgraphs `→ `nn_graphs `would be an actively harmful alias. The old graphs were built on the full ordered structures, with every ion as a node; the new ones are built on the magnetic-only structure. The node sets differ, so any surviving` sgraph.structure[i] `or index-based loop would keep running and silently address a different site. Today that code fails loudly with an `AttributeError `on the first line that touches `sgraphs`. An alias would trade that for wrong numbers.

`unique_site_ids `→ `sublattice_ids `isn't an alias at all. The old attribute was a `dict[tuple[int, ...], int]` mapping tuples of equivalent supercell indices to an id; the new one is a `list[list[int]]` giving the parent-sublattice label of each magnetic site in each ordering. Bridging them means reconstructing the per-supercell orbit map — which is precisely the single-supercell assumption this PR exists to remove.` wyckoff_ids` → `sublattice_wyckoff_symbols` does happen to keep its `dict[int, str]` shape, but the ids underneath now index parent orbits rather than supercell orbits, so a bare alias would hand back keys that don't mean what the caller thinks.

get_exchange is breaking regardless of the return type. The `{"<J>": ...}` fallback for under-determined systems is gone and now raises `ValueError`, and the solve became a least-squares fit over all orderings rather than an exact inverse — so numbers move even for callers who never touch `residual`. Keeping the dict return would hide the signature change without making any of that safe.

There's also one break no shim can cover: `energies `keeps its name but now holds energy per magnetic ion instead of total energy. That's a silent numeric change in an attribute that still exists, which to me is the argument that a partial soft landing here is worse than a clean one — it lets callers get most of the way through a migration while quietly wrong.

Given that, I think the honest version is a hard break with a good migration path rather than a half-compatible surface. The module docstring already carries a migration guide covering every renamed and removed attribute, the constructor order, the `get_exchange `change, and `HeisenbergScreener `— worth a look if it didn't come through in the diff view. Happy to expand it, add a `CHANGES.md` entry, or move it somewhere more discoverable if you'd prefer.

shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep; scheduled review routine).

This is a well-executed refactor. The test suite is a highlight: round-tripping against a synthetic Hamiltonian with known J's, ground-truth energies computed independently of the mapper's own graph, and explicit mixed-supercell (1x1/1x2/2x2) recovery. The migration notes in the module docstring are thorough.

Comments:

1. **Legacy `HeisenbergModel` deserialization is broken.** `from_dict` now requires `magnetic_structures`, `nn_graphs`, `sublattice_ids`, `sublattice_wyckoff_symbols`, and `residual`. Any previously stored MSON dict (with `sgraphs`, `unique_site_ids`, `wyckoff_ids`, `javg`) will `KeyError`. You handle old `ex_mat` string forms, so partial back-compat is clearly intended — either extend `from_dict` to accept the old key set (even if lossy) or raise a clear "serialized with HeisenbergModel <0.2" error.
2. **`tol` default silently changed** from 0.02 to 0.05 (`DEFAULT_TOL`). Not mentioned in the migration section — please add it, since it changes shell-merging behavior for existing callers.
3. **Positional-arg shift**: `HeisenbergMapper(structs, energies, 5.0)` previously meant `cutoff=5.0`; now `5.0` lands in `parent`. It fails loudly in `_nonmagnetic`, but a friendlier `isinstance(parent, Structure)` check with a pointed error message would help migrating users.
4. **`self.residual` is only set in `get_exchange`** — initialize to `None` in `__init__` alongside `ex_params` so attribute access before fitting doesn't raise `AttributeError` (the class docstring implies it's an attribute).
5. **Explicit `parent` is silently primitivized** via `get_primitive_structure()`. Worth a docstring note, since a user supplying a deliberate supercell parent may not expect orbit analysis on the primitive cell.
6. The `pymatgen-core` submodule bump is included — confirm it's intentional and points at merged main (it matches merged #130, so likely fine, just flagging).
7. **CI has not run on the latest commits** ("no checks reported" on the branch). A maintainer needs to approve/trigger workflows before this can be evaluated for merge.

The deprecations of `estimate_exchange`/`get_mft_temperature` with concrete alternatives (shell-resolved fit; MC solvers) are the right call, and pinning the flawed multi-sublattice MFT formula in a test rather than silently "fixing" it is good hygiene.

shyuep:
@mkhorton I suggest you review this since I am not familiar with this code.

shyuep:

This is a well-motivated redesign — defining sublattices once on an explicit parent cell and mapping each relaxed ordering via `StructureMatcher` is the right fix for #4667, and the move from a square solve to least squares over all orderings (with rank-deficiency warning + RMS residual) is a clear improvement. The test coverage (known-Hamiltonian recovery, Wyckoff-orbit sublattices, non-supercell rejection) is thorough, and deprecating `estimate_exchange`/`get_mft_temperature` rather than removing them is the right call.

Requested changes:

1. **Revert the `pymatgen-core` submodule pointer bump** — it's unrelated to this PR and appears accidental (also present in #4699).
2. **CI has not run** on this branch (needs maintainer approval for first-time contributor workflows); once the submodule change is reverted, we'll approve the run.
3. **`HeisenbergScreener` signature is now `(orderings, screen=False)`** rather than `(structures, energies, screen)` — that's a second breaking public API beyond `HeisenbergMapper` itself. Please call it out explicitly in the PR description/CHANGES so downstream users (e.g. atomate2 magnetism workflows, which also touch `sgraphs`/`unique_site_ids`) can migrate.
4. Minor: `SystemExit` → `ValueError` is appreciated.

Luguza:
@shyuep @mkhorton Rebased onto `main` and addressed the outstanding review points:

1. **Rebased cleanly onto `main`.** Dropped the `pymatgen-core` submodule pointer bump (unrelated/accidental, flagged in both reviews) — it now matches `main` exactly, no diff.
2. **`self.residual` is now initialized to `None` in `__init__`** alongside `ex_params`, so reading it before a fit returns `None` instead of raising `AttributeError`.
3. **Documented the `tol` default change** (0.02 → 0.05) in the module's migration guide — it changes shell-merging behavior for existing callers.
4. **Documented that `parent` is always reduced to its primitive cell**, whether inferred or explicitly supplied — a caller passing a deliberate supercell parent should know it's analyzed on the primitive cell.
5. **Added a pointed `TypeError`** when `parent` isn't a `Structure`, so the positional `cutoff`→`parent` argument shift (from the signature change) fails immediately with a clear message instead of deep inside unrelated `Structure`/symmetry code.
6. **`HeisenbergModel.from_dict` now rejects pre-0.2 serializations with a clear `ValueError`** instead of a bare `KeyError`. Old dicts carry `unique_site_ids`/`wyckoff_ids`/`sgraphs`, which don't map onto the new parent-cell sublattices — there's no safe lossy reconstruction, so this raises rather than silently returning something wrong, consistent with the rest of this PR's approach to the breaking changes.

Could you approve the CI workflows on the latest commits?

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff and GitHub state only, no code executed).

Re-review after the 2026-09-03 rebase. Everything from the last round is addressed: clean rebase with no `pymatgen-core` submodule diff, `self.residual` initialized in `__init__`, the `tol` 0.02 -> 0.05 change and the always-primitive `parent` reduction documented in the migration guide, a pointed `TypeError` for the positional `cutoff` -> `parent` shift, and `from_dict` rejecting 0.1 serializations explicitly. The migration guide also now covers the `HeisenbergScreener` signature change, which was my point 3 last time.

Remaining items, roughly in order of importance:

1. **Row deduplication in `_build_exchange_mat` discards the wrong thing.** `key = tuple(round(row[c], 10) for c in j_columns)` drops any ordering whose Hamiltonian row duplicates an earlier one — but the *energies* are not part of the key. Two orderings with the same Heisenberg row and different DFT energies is precisely the signal that the Heisenberg model is incomplete for this system, and it is what `residual` exists to report. Dropping the second one silently deflates the residual and destroys the diagnostic. Least squares handles duplicated rows without trouble; I would keep them, or at minimum warn with the energy spread before dropping.

2. **`RelaxedOrdering.set_sublattice_ids` assumes species alignment without checking it.** `matched_parent[i]` is taken to describe the same site as `self.structure[i]`, and the magnetic subset is then selected by two independent tests (non-`None` `sublattice_id` on the parent side, species membership on the ordering side). With `StructureMatcher` defaults (`ltol=0.2`, `stol=0.3`) on relaxed cells, a bad match misaligns every sublattice label and every downstream `J_ij` with no error. A cheap `if matched_parent[i].specie.symbol != self.structure[i].specie.symbol: raise ValueError(...)` turns a silent wrong answer into a loud one.

3. **`np.linalg.cond(H) > 1e5` is scale-dependent.** The `E0` column is exactly 1.0 while the `J` columns are sums of `m_i * m_j` (O(10-100) for typical moments). The condition number therefore tracks the column scaling as much as the degeneracy of the orderings, so the fixed 1e5 threshold will fire or not depending on moment magnitude. Normalize the columns (or use the singular values `lstsq` already returns, which you currently discard as `_singular`) before thresholding.

4. **Warning channel is inconsistent.** The missing-`parent` case uses `warnings.warn`, while rank deficiency and ill-conditioning — the two a caller most needs to catch, filter, or turn into errors programmatically — go to `logger.warning`. Recommend `warnings.warn` for all three.

5. Minor: `j_ij[1:] *= 1000` relies on `E0` being column 0 of `ex_mat`. True by construction today (`ex_mat[["E", "E0", *j_columns]]`), but implicit; indexing by name would survive a future reordering.

6. CI still has not run on this branch (first-time-contributor workflow approval). Given the size (+1408/-699 across the module and its tests) and that this is a `[breaking]` change to a public API, I would not merge until the full suite is green.

Overall the design is sound and the documentation is unusually thorough. Items 1 and 2 are the ones I would want fixed before merge; 3-5 are polish.

shyuep:

The physics motivation is right, and the core insight — define sublattices once on a paramagnetic parent cell so orderings in different supercells can be fitted together — is a real improvement over deriving them from the orderings themselves. `SublatticeMinimumDistanceNN` fixes a genuine defect (sublattice-blind `MinimumDistanceNN` drops long intra-sublattice bonds entirely), and replacing the exactly-determined solve with a least-squares fit plus a reported residual is clearly better practice.

That said, at +1427/−699 in one module with a breaking public surface and **no CI results reported (0 checks)**, this is not mergeable as-is.

## Blocking

1. **No CI.** Please push/rebase so the suite runs. Nothing below can be confirmed without it.

2. **Positional-argument insertion is a silent breakage.** `HeisenbergMapper(structures, energies, parent, cutoff, tol)` puts `parent` where `cutoff` used to be. Existing positional calls will pass a `float` as `parent` and get a confusing downstream failure rather than a `TypeError`. Make `parent` keyword-only, or append it last, or add a type check on the third argument that raises a clear deprecation error.

3. **`energies` changes meaning under the same name** (total → per magnetic ion). Silently returning different numbers from an unchanged attribute is the hardest kind of change to detect downstream. Rename to `energies_per_magnetic_ion` and either drop `energies` or keep it as the total with a deprecation warning.

4. **`get_exchange` return type changed** from `ex_params` to `(ex_params, residual)`. Every existing caller breaks. Consider attaching `residual` only to `self.residual` / `HeisenbergModel` and leaving the return type alone — the information is already stored on both.

5. **Removals with no deprecation path**: `sgraphs`, `unique_site_ids`, `wyckoff_ids`, `ordered_structures`, `HeisenbergScreener.screened_structures` / `screened_energies`, and the `{"<J>": ...}` under-determined fallback. `monty.dev.deprecated` is already imported in this file — please use it for at least one release cycle. atomate2 and emmet both touch this class.

6. **`HeisenbergModel.from_dict` round-trip.** The diff rewrites `from_dict` (−64/+20). Serialized `HeisenbergModel` documents exist in production MongoDB stores without a `residual` field and with the old `ex_params` shape. Please add an explicit test deserializing a pre-0.2 dict.

7. **`__maintainer__` / `__email__`.** Adding yourself as module maintainer is a decision for the pymatgen maintainers, not the PR. `__author__` is the appropriate field for a contribution of this size; please revert the maintainer change and let @shyuep decide.

## Should fix

8. **Default `tol` 0.02 → 0.05 Å** silently changes which distances merge into a shell, and therefore the fitted J values, for every existing user. If the new default is better justified, say why in the PR body; otherwise keep 0.02 and let users opt in.

9. **Split the PR.** As one change this is very hard to review and impossible to bisect. Natural seams: (a) `SublatticeMinimumDistanceNN` + tests, (b) parent-cell sublattice refactor, (c) least-squares fit + residual reporting, (d) deprecations. (a) and (c) are independently valuable and could land first.

10. **`DISTANCE_ROUND_DECIMALS = 2` interacts with `tol`.** Distances are rounded to 0.01 Å and then compared against `tol = 0.05 Å`. Two constants controlling the same shell-merging decision invites inconsistency; consider deriving the rounding from `tol` or documenting the relationship.

11. **`_interaction_label` falls back to `min(labels, key=|dist - self.dists[label]|)`** — an off-parent-lattice bond is assigned to its nearest shell with only a `logger.warning`. For a relaxed ordering that has genuinely drifted, this silently contaminates the fit. Given the PR already routes ill-conditioning through `warnings.warn`, this one should be a `UserWarning` too.

## Positive notes

- The "Migrating from HeisenbergMapper 0.1" section is unusually thorough — please keep it and promote it to the release notes.
- Reporting rank deficiency / ill-conditioning as a `UserWarning` rather than a logger call is the right call.
- Deprecating `get_mft_temperature` in favour of a proper MC solver is correct; mean-field T_c overestimates badly for low-dimensional magnets.
- Test file grows from 95 to 477 lines — good.

shyuep:
**Automated PR review generated by Claude** (sanity-check before acting on it)

The design direction is good: defining sublattices once on a shared parent cell and mapping each relaxed ordering onto it via `StructureMatcher(attempt_supercell=True)` is the right fix for #4667, and the least-squares `get_exchange` (replacing the square-system solve, adding a rank-deficiency warning and RMS residual) is a real improvement over silently discarding surplus orderings. Deprecating `estimate_exchange`/`get_mft_temperature` with a dated sunset is handled the right way.

However, this cannot merge as-is: **CI is failing because the file doesn't parse.** `src/pymatgen/analysis/magnetism/heisenberg.py` around line 856 has a stray fragment appended directly onto the closing `}` of the `ex_params` dict comprehension, with no line break:

```
        }if matched_parent[i].specie.symbol != self.structure[i].specie.symbol: raise ValueError(...
```

Ruff reports this as `invalid-syntax` (`Expected `else`, found `:`` / `Expected `,`, found name` / `Expected `)`, found newline`) — this is a genuine syntax error, not a style nit, and it's why lint fails and every test matrix job (`macos`, `ubuntu`, `windows`) fails too (the module can't be imported). `matched_parent` and `i` also aren't otherwise bound in this scope, and the `raise ValueError(...)` looks like an unfinished edit (literal `...`) rather than intended code — looks like a merge/edit artifact that needs to be either completed properly or removed before anything else here can be evaluated.

Given the size of this diff (1427/-699 across 2 files) and that it's marked `[breaking]`, I'd hold off on a deeper design/correctness pass until the file at least imports and CI is green, since right now none of the new test classes described in the PR body (`TestHeisenbergMapperKnownHamiltonian`, etc.) can actually run.

PR #4702, opened 2026-08-19
Fix MagneticStructureEnumerator duplicate-ordering pruning never running

### LINKED ISSUE #4697: `MagneticStructureEnumerator` duplicate-pruning logic never removes duplicate ordered structures
### Python version

Python 3.12.13

### Pymatgen version

2026.5.4

### Operating system version

_No response_

### Current behavior

In `MagneticStructureEnumerator._generate_ordered_structures` (`src/pymatgen/analysis/magnetism/analyzer.py`, around line 1017), after identifying which ordered structures are duplicates via `matches_ordering`, the pruning block is gated on the wrong condition:

```python
if len(structures_to_remove) == 0:
    self.logger.info(f"Removing {len(structures_to_remove)} duplicate ordered structures")
    ordered_structures = [s for idx, s in enumerate(ordered_structures) if idx not in structures_to_remove]
    ordered_structures_origins = [
        o for idx, o in enumerate(ordered_structures_origins) if idx not in structures_to_remove
    ]
```

### Expected Behavior

When duplicate ordered structures are detected, they should be pruned from `ordered_structures` / `ordered_structures_origins` — i.e. the condition should be `if len(structures_to_remove) > 0:`.

### Minimal example

```Python
from pymatgen.analysis.magnetism.analyzer import MagneticStructureEnumerator

structure = Structure.from_file("POSCAR")  # any structure prone to symmetry-equivalent orderings
enumerator = MagneticStructureEnumerator(structure)
print(len(enumerator.ordered_structures)) # test with and without bugfix applied
```

_No response_

fronzi:
Fixes #4697.

### The bug

`MagneticStructureEnumerator._generate_ordered_structures` identifies duplicate magnetic orderings via `matches_ordering`, then guards the removal block with an inverted condition:

```python
if len(structures_to_remove) == 0:
    self.logger.info(f"Removing {len(structures_to_remove)} duplicate ordered structures")
    ordered_structures = [...]
```

The block therefore only runs when there is nothing to remove. Duplicates are detected, logged as `Removing 0 duplicate ordered structures`, and then kept. The surrounding comment (`# in case we've introduced duplicates, let's remove them`) and the log message both make the intent clear, so this looks like a typo rather than a deliberate disable.

### The fix

```diff
-        if len(structures_to_remove) == 0:
+        if structures_to_remove:
```

### This is a user-visible behaviour change

Worth flagging explicitly: the pruning has been dead since it was written, so enabling it means `MagneticStructureEnumerator` now returns **fewer** orderings than before for any input that produces symmetry-equivalent orderings. That is the documented intent, but downstream workflows that depend on the current ordering count will see a difference.

Two consequences I found:

1. **Two existing truncation tests asserted the unpruned counts.** `test_truncate_by_symmetry_respects_configured_count` and `test_truncate_by_symmetry_default_true_keeps_five` build six Ni/O structures that carry no magmoms and differ only in lattice. Because `StructureMatcher` defaults to `scale=True` and `primitive_cell=True`, `matches_ordering` judges two of those pairs equivalent, so pruning removes two and the tests' `== 6` assertions fail. Since those tests were written to exercise *truncation*, I've stubbed `matches_ordering` in both so they isolate that logic, rather than rewriting their assertions around a new premise.

2. **Cost.** The pruning is O(N²) in `CollinearMagneticStructureAnalyzer` construction and `matches_ordering` calls. It was free while dead; real enumerations now pay it. At the default `num_orderings=64` this is not significant, but it is no longer zero.

Happy to gate this behind a keyword argument instead if you'd prefer a strictly non-breaking change.

### Test

`test_duplicate_orderings_are_pruned` feeds three orderings, two of which are identical, and asserts the duplicate is removed and no two survivors match. It follows the existing enumlib-free pattern in `TestMagneticStructureEnumeratorTruncation` (`object.__new__` to bypass `__init__`), so it runs in CI without the enumlib binaries.

Verified in both directions: the test fails on master (`assert 3 < 3`) and passes with the fix.

### DIFF
--- a/src/pymatgen/analysis/magnetism/analyzer.py
+++ b/src/pymatgen/analysis/magnetism/analyzer.py
@@ -1010,7 +1010,7 @@ class MagneticStructureEnumerator:
                     ):
                         structures_to_remove.append(check_idx)
 
-        if len(structures_to_remove) == 0:
+        if structures_to_remove:
             self.logger.info(f"Removing {len(structures_to_remove)} duplicate ordered structures")
             ordered_structures = [s for idx, s in enumerate(ordered_structures) if idx not in structures_to_remove]
             ordered_structures_origins = [
--- a/tests/analysis/magnetism/test_analyzer.py
+++ b/tests/analysis/magnetism/test_analyzer.py
@@ -431,9 +431,7 @@ class TestMagneticStructureEnumeratorTruncation:
         enumerator.transformations = {}
         enumerator.sanitized_structure = structures[0]
         enumerator.num_orderings = 64
-        enumerator.input_analyzer = CollinearMagneticStructureAnalyzer(
-            structures[0], overwrite_magmom_mode="none"
-        )
+        enumerator.input_analyzer = CollinearMagneticStructureAnalyzer(structures[0], overwrite_magmom_mode="none")
         enumerator.ordered_structures = list(structures)
         enumerator.ordered_structure_origins = list(origins)
 
--- a/tests/analysis/magnetism/test_analyzer.py
+++ b/tests/analysis/magnetism/test_analyzer.py
@@ -313,13 +313,21 @@ class TestMagneticStructureEnumeratorTruncation:
     # Not gated on ENUMLIB_PRESENT: these tests call _generate_ordered_structures
     # directly on a manually-built MagneticStructureEnumerator, so they don't need
     # enumlib to generate orderings.
-    def test_truncate_by_symmetry_respects_configured_count(self):
+    def test_truncate_by_symmetry_respects_configured_count(self, monkeypatch):
         # Regression test: truncate_by_symmetry used to be ignored by the actual
         # truncation step, which had a hardcoded cap of 5 regardless of what the
         # user configured. Build a set of structures with 5 distinct symmetries
         # and confirm that requesting truncate_by_symmetry=2 actually keeps 2,
         # not 5. Bypasses __init__ (and enumlib) since only the truncation logic
         # in _generate_ordered_structures is under test.
+        #
+        # These structures carry no magmoms and differ only in lattice, so
+        # StructureMatcher (scale=True, primitive_cell=True) considers some of
+        # them equivalent orderings. Stub out matches_ordering so the duplicate
+        # pruning earlier in _generate_ordered_structures cannot interfere with
+        # the truncation behavior actually under test here.
+        monkeypatch.setattr(CollinearMagneticStructureAnalyzer, "matches_ordering", lambda self, other: False)
+
         base = Structure.from_spacegroup(225, Lattice.cubic(4.2), ["Ni", "O"], [[0, 0, 0], [0.5, 0.5, 0.5]])
         lattices = [
             Lattice.from_parameters(4.2, 4.2, 4.6, 90, 90, 90),  # tetragonal
@@ -374,6 +382,8 @@ class TestMagneticStructureEnumeratorTruncation:
         # structures directly so the real truncation logic in
         # _generate_ordered_structures runs on our controlled set.
         monkeypatch.setattr(MagneticStructureEnumerator, "_generate_transformations", lambda self, struct: {})
+        # As above, isolate truncation from the duplicate-ordering pruning.
+        monkeypatch.setattr(CollinearMagneticStructureAnalyzer, "matches_ordering", lambda self, other: False)
 
         real_generate = MagneticStructureEnumerator._generate_ordered_structures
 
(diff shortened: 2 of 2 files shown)


shyuep (commented):
**🤖 Automated PR review generated by Claude** (posted on behalf of @shyuep)

Nice catch — the inverted guard (`if len(structures_to_remove) == 0:`) clearly contradicts the comment and log message, and the one-line fix is correct. The regression test feeding an exact duplicate through `_generate_ordered_structures` is well constructed, and stubbing `matches_ordering` in the truncation tests correctly isolates them from the now-active pruning.

Two notes: (1) as you flag, this is a user-visible behavior change — `MagneticStructureEnumerator` will now return fewer orderings for inputs with symmetry-equivalent orderings; please add a line to the changelog so downstream users aren't surprised. (2) CI hasn't run on this branch yet; a maintainer needs to approve the workflow run before merge.

fronzi:
Thanks for the review. Changelog entry added under a new `## Unreleased` heading, flagging the behavior change and the O(N²) pruning cost that is no longer skipped.

On CI: the run is blocked by the first-time-contributor gate rather than anything on the branch — the workflows will need a maintainer to approve them. Locally `tests/analysis/magnetism/` is green, and the regression test fails on `main` (`assert 3 < 3`) and passes with the fix.

Also happy to gate the pruning behind a keyword argument if you'd prefer to avoid the behavior change in a patch release.

shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep).

- Fix is correct: the `len(structures_to_remove) == 0` guard inverted the obvious intent, so the pruning block has been dead since it was written. `if structures_to_remove:` is the right minimal change. Well-diagnosed writeup.
- Good call stubbing `matches_ordering` in the two truncation tests so they keep testing truncation rather than inheriting the pruning change; the new `test_duplicate_orderings_are_pruned` is enumlib-free and verified to fail on master.
- On the kwarg question: no — the behavior change (fewer orderings, plus the O(N²) pruning cost) is the documented intent. Ship it as a plain fix.
- Please drop the hand-edited `docs/CHANGES.md` "Unreleased" entry — that file is regenerated from PR metadata at release time, so hand-edits get duplicated or overwritten. The PR body already carries the needed detail.
- No CI reported on the head commit. Mark ready for review to trigger the test matrix; happy to merge once green.

fronzi:
Thanks — dropped the docs/CHANGES.md edit (didn't realise it's generated from PR metadata) and marked ready for review to trigger the matrix. Keeping it as a plain fix, no kwarg. Let me know if anything comes back red.

Luguza:
I have also adressed this already in #4699. Since @fronzi already implemented tests however, I will revert my dedup changes so we seperate the PRs into #4702 being the fix and #4699 adding the enumeration strategy.

fronzi:
Thanks for flagging, and for splitting them — that split makes sense to me. Happy to go in whichever order suits; if #4699 lands first I'm glad to rebase #4702 onto it, or vice versa. Shout if you'd like the regression test moved or extended to cover anything on your side.

Luguza:
I ended up needing to include a few of your changes to pass the tests, but they are identical copies and should merge cleanly. I think this PR should be resolved first though.

fronzi:
Sounds good — identical hunks should merge cleanly either way. This one is only waiting on the workflow approval; tests passed locally and on the last approved run (macOS/Ubuntu/Windows), with just a formatting commit since. @shyuep could you approve the run when you get a chance? Happy to see #4699 go straight in after.

shyuep:

LGTM. Clean minimal fix (`if structures_to_remove:`) for a clearly inverted guard, and the PR body does a good job flagging the user-visible consequence (fewer orderings for symmetry-equivalent inputs). The regression test builds the enumerator via `object.__new__` to bypass enumlib — pragmatic — and the `monkeypatch` of `matches_ordering` in the two truncation tests correctly isolates them from the newly-live pruning.

Two notes: (1) #4699 contains this exact same one-line fix bundled with a new feature — please coordinate so one of the two rebases after the other lands; this fix-only PR with regression tests is the natural one to land first. (2) No CI results are available yet (checks likely awaiting maintainer approval) — review is based on code analysis only.

fronzi:
Thanks. Coordination is already sorted — @Luguza confirmed above he'll rebase #4699 after this one lands.

On CI: the run is blocked by the first-time-contributor gate, so it needs a maintainer to click "Approve and run workflows" on the yellow banner at the top of this PR — nothing I can trigger from my side. Locally tests/analysis/magnetism/ is green (13 passed), the regression test fails on main (assert 3 < 3) and passes with the fix, and ruff check / ruff format --check are clean.

Luguza:
@shyuep Could you please approve the CI workflows?

shyuep:
Thanks!

PR #4705, opened 2026-09-04
Reuse first termination slabs in `CoherentInterfaceBuilder._find_matches` 

### LINKED ISSUE #4651: Singular film supercell transformation matrix when calling get_interfaces for CoherentInterfaceBuilder
### Python version

Python 3.11.4

### Pymatgen version

2023.9.10

### Operating system version

Mac Tahoe 26.2 (25C56)

### Current behavior

Not sure if this is a known pymatgen issue when building interfaces but I can’t find a GitHub issue for it. Printing some information from CoherentInterfaceBuilder.get_interfaces shows the following:

film_shift 0.08333336181879059
sub_shift 0.12499996873087585
film_slab.lattice.matrix[:2]
[[-1.96204495  5.88613486 -6.26018476]
 [ 5.88613486  5.88613486  6.26018476]]
match.film_sl_vectors
[[ -7.84817982   0.           0.        ]
 [ -3.92408991 -11.77226973 -12.52036953]]
super_film_transform
[[ 0  0  0]
 [ 0 -2  0]
 [ 0  0  1]]

I think that last transform, a film supercell transform, is the issue since there should never be a zero along the diagonal.

Here is a script to reproduce the issue:

```
import numpy
from pymatgen.analysis.interfaces.zsl import ZSLGenerator
from pymatgen.core.structure import Structure as PMGStructure
from pymatgen.analysis.interfaces.coherent_interfaces import CoherentInterfaceBuilder

pbc1 = [
    [3.816817045211792, 0.0, 0.0],
    [0.0, 3.816817045211792, 0.0],
    [0.0, 0.0, 3.816817045211792]]
ans1 = [57, 27, 8, 8, 8]
xyz1 = numpy.array([
    [1.908409, 1.908409, 1.908409],
    [0.0, 0.0, 0.0],
    [1.908409, 0.0, 0.0],
    [0.0, 1.908409, 0.0],
    [0.0, 0.0, 1.908409]])
sub_st = PMGStructure(pbc1, ans1, xyz1, charge=0,
    validate_proximity=False, to_unit_cell=False,
    coords_are_cartesian=True, site_properties=None)

pbc2 = [
    [3.9240899085998535, 0.0, 0.0],
    [0.0, 3.9240899085998535, 0.0],
    [0.0, 0.0, 12.520369529724121]]
ans2 = [57, 57, 57, 57, 28, 28, 8, 8, 8, 8, 8, 8, 8, 8]
xyz2 = numpy.array([
    [0.0, 0.0, 4.550816],
    [1.962045, 1.962045, 1.709368],
    [1.962045, 1.962045, 10.811001],
    [0.0, 0.0, 7.969553],
    [0.0, 0.0, 0.0],
    [1.962045, 1.962045, 6.260185],
    [0.0, 0.0, 2.241059],
    [1.962045, 1.962045, 4.019126],
    [1.962045, 0.0, 0.0],
    [0.0, 1.962045, 0.0],
    [1.962045, 1.962045, 8.501243],
    [0.0, 0.0, 10.279311],
    [0.0, 1.962045, 6.260185],
    [1.962045, 0.0, 6.260185]])
film_st = PMGStructure(pbc2, ans2, xyz2, charge=0,
    validate_proximity=False, to_unit_cell=False,
    coords_are_cartesian=True, site_properties=None)

kwargs = {'max_area': 450.0}
film_hkl = (3, 2, 3)
sub_hkl = (2, 1, 0)
term = ('La_C2/m_1', 'CoO2_P2/m_3')

zslgen = ZSLGenerator(**kwargs)
builder = CoherentInterfaceBuilder(sub_st, film_st, film_hkl, sub_hkl, zslgen=zslgen)
inters = builder.get_interfaces(term, gap=1.0, vacuum_over_film=10.0,
    film_thickness=4.0, substrate_thickness=4.0, in_layers=True)
print(len(list(inters)))
```

Throws the following traceback:

```
  File "/Users/hughes/pymatgen_test/test.py", line 54, in <module>
    print(len(list(inters)))
              ^^^^^^^^^^^^
  File "/Users/hughes/scr/bld_26-3/internal/lib/python3.11/site-packages/pymatgen/analysis/interfaces/coherent_interfaces.py", line 200, in get_interfaces
    film_sl_slab.make_supercell(super_film_transform)
  File "/Users/hughes/scr/bld_26-3/internal/lib/python3.11/site-packages/pymatgen/core/structure.py", line 4115, in make_supercell
    supercell = struct * scaling_matrix
                ~~~~~~~^~~~~~~~~~~~~~~~
  File "/Users/hughes/scr/bld_26-3/internal/lib/python3.11/site-packages/pymatgen/core/structure.py", line 1320, in __mul__
    f_lat = lattice_points_in_supercell(scale_matrix)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/hughes/scr/bld_26-3/internal/lib/python3.11/site-packages/pymatgen/util/coord.py", line 295, in lattice_points_in_supercell
    frac_points = np.dot(all_points, np.linalg.inv(supercell_matrix))
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<__array_function__ internals>", line 200, in inv
  File "/Users/hughes/scr/bld_26-3/internal/lib/python3.11/site-packages/numpy/linalg/linalg.py", line 538, in inv
    ainv = _umath_linalg.inv(a, signature=signature, extobj=extobj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/hughes/scr/bld_26-3/internal/lib/python3.11/site-packages/numpy/linalg/linalg.py", line 89, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")
numpy.linalg.LinAlgError: Singular matrix
```

### Expected Behavior

I expect to get an interface back with no traceback.

### Minimal example

```Python

```

_No response_

Sanftperlig:
Currently, CoherentInterfaceBuilder._find_matches always uses shift 0. As discussed in https://github.com/materialsproject/pymatgen/issues/4651 and https://github.com/materialsproject/pymatgen-core/issues/105 this will lead to some primitivizations differing from the nonzero shifts in other functions.
I believe that to be the source of issues such as https://github.com/materialsproject/pymatgen/issues/4651 (and thus it closes #4651), but I do not have a test case, given the one in the issue works fine already (I do believe that this will cause errors in other cases where primitivization is shift-dependent though).
Additionally, passing the slabs to find_matches saves creating the SlabGenerator and generating the two slabs.

## Checklist

### DIFF
--- a/src/pymatgen/analysis/interfaces/coherent_interfaces.py
+++ b/src/pymatgen/analysis/interfaces/coherent_interfaces.py
@@ -11,7 +11,7 @@ from scipy.linalg import polar
 from pymatgen.analysis.elasticity.strain import Deformation
 from pymatgen.analysis.interfaces.zsl import ZSLGenerator, fast_norm
 from pymatgen.core.interface import Interface, label_termination
-from pymatgen.core.surface import SlabGenerator
+from pymatgen.core.surface import Slab, SlabGenerator
 
 if TYPE_CHECKING:
     from collections.abc import Iterator, Sequence
@@ -68,37 +68,16 @@ class CoherentInterfaceBuilder:
         self.termination_ftol = termination_ftol
         self.label_index = label_index
         self.filter_out_sym_slabs = filter_out_sym_slabs
-        self._find_matches()
-        self._find_terminations()
+        self._find_matches(self._find_terminations())
 
-    def _find_matches(self) -> None:
-        """Find and stores the ZSL matches."""
+    def _find_matches(self, film_slab: Slab, sub_slab: Slab) -> None:
+        """Find and stores the ZSL matches.
+
+        Args:
+            film_slab (Slab): A slab of the film structure, transformed as used in the other functions.
+            sub_slab (Slab): Associated slab of the substrate structure.
+        """
         self.zsl_matches = []
-
-        film_sg = SlabGenerator(
-            self.film_structure,
-            self.film_miller,
-            min_slab_size=1,
-            min_vacuum_size=3,
-            in_unit_planes=True,
-            center_slab=True,
-            primitive=True,
-            reorient_lattice=False,  # This is necessary to not screw up the lattice
-        )
-
-        sub_sg = SlabGenerator(
-            self.substrate_structure,
-            self.substrate_miller,
-            min_slab_size=1,
-            min_vacuum_size=3,
-            in_unit_planes=True,
-            center_slab=True,
-            primitive=True,
-            reorient_lattice=False,  # This is necessary to not screw up the lattice
-        )
-
-        film_slab = film_sg.get_slab(shift=0)
-        sub_slab = sub_sg.get_slab(shift=0)
 
         film_vectors = film_slab.lattice.matrix
         substrate_vectors = sub_slab.lattice.matrix
@@ -119,8 +98,11 @@ class CoherentInterfaceBuilder:
                     "Substrate lattice vectors changed during ZSL match, check your ZSL Generator parameters"
                 )
 
-    def _find_terminations(self):
-        """Find all terminations."""
+    def _find_terminations(self) -> tuple[Slab, Slab]:
+        """Find all terminations.
+        Returns:
+            first_pair (tuple[Slab, Slab]): The slab pair of the first termination, useful to check the primitivization.
+        """
         film_sg = SlabGenerator(
             self.film_structure,
             self.film_miller,
@@ -179,6 +161,8 @@ class CoherentInterfaceBuilder:
             )
         }
         self.terminations = list(self._terminations)
+
+        return film_slabs[0], sub_slabs[0]
 
     def get_interfaces(
         self,
--- a/src/pymatgen/analysis/interfaces/coherent_interfaces.py
+++ b/src/pymatgen/analysis/interfaces/coherent_interfaces.py
@@ -68,7 +68,7 @@ class CoherentInterfaceBuilder:
         self.termination_ftol = termination_ftol
         self.label_index = label_index
         self.filter_out_sym_slabs = filter_out_sym_slabs
-        self._find_matches(self._find_terminations())
+        self._find_matches(*self._find_terminations())
 
     def _find_matches(self, film_slab: Slab, sub_slab: Slab) -> None:
         """Find and stores the ZSL matches.


shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff + CI only — advisory, not a human review).

The diagnosis is right: `_find_matches` built its own `SlabGenerator` and took `get_slab(shift=0)`, while `_find_terminations` uses `get_slabs(...)` at physically determined shifts. With `primitive=True` the Niggli/primitive reduction is shift-dependent, so ZSL matching could be done on a lattice that no constructed interface ever uses. Reusing a real termination slab is the correct direction, and dropping the duplicate generators is a free win.

Three things before this can go in:

1. **`film_slabs[0]` is still only *one* termination.** If primitivization is genuinely shift-dependent, slabs at other shifts can have different reduced lattices, so the matches remain inconsistent for terminations 2..N — the bug is narrowed, not removed. Preferred fixes, in order: (a) assert all slabs share the same lattice (`np.allclose` on the matrices) and raise/warn if not, so the silent inconsistency becomes loud; or (b) match on a non-primitivized generator so the lattice is shift-independent by construction. At minimum, add a comment stating the assumption that all terminations share a reduced lattice.

2. **`film_slabs[0]` / `sub_slabs[0]` can `IndexError`.** `get_slabs(filter_out_sym_slabs=...)` is not guaranteed non-empty; the old `get_slab(shift=0)` always returned a slab. Guard and raise a clear error naming the miller index.

3. **No test.** Understood that the #4651 reproducer already passes, but a regression guard is cheap and does not need a shift-dependent case: assert that the lattice `_find_matches` consumes is exactly `next(iter(...))` of the termination slabs (e.g. via `film_slabs[0].lattice.matrix`), plus a characterization test on a standard pair (Si/GaAs) showing `zsl_matches` is unchanged. That pins the invariant this PR establishes.

Smaller points:

- `__init__` now runs `_find_terminations()` before `_find_matches()`. Order inversion means the "Substrate lattice vectors changed during ZSL match" `ValueError` surfaces only after the expensive `get_slabs()` calls, and `self.zsl_matches` is now set last. Both private, both fine — just worth a line in the release notes since `_find_matches`'s signature changed and downstream subclasses overriding it will break.
- Docstring formatting: `_find_terminations` needs a blank line between the summary and `Returns:` for the Google convention, and the Returns entry should describe the tuple rather than name it `first_pair` (`tuple[Slab, Slab]: film and substrate slabs of the first termination`). Same for the `Args:` block, which reads fine otherwise.
- Nit: `Slab` is annotation-only — it belongs in the `TYPE_CHECKING` block rather than the runtime import.

**CI:** no checks have reported on `shift-change` and `mergeable_state` is `blocked` — workflows need maintainer approval for this fork. Please approve the run; the review above is static-analysis only, with no test evidence behind it.

shyuep:

The diagnosis is right: matching against `get_slab(shift=0)` while `get_interfaces` uses shift-dependent primitivized slabs is inconsistent, and reusing the slabs already built in `_find_terminations` removes two redundant `SlabGenerator` constructions. Points to address:

1. **`film_slabs[0]` is still an arbitrary choice.** The fix replaces "shift 0" with "first termination", but if primitivization is shift-dependent then different terminations can yield different reduced lattices, and the ZSL match will still be inconsistent for every termination but the first. Either document this assumption explicitly in the docstring, or verify that all slabs share the same reduced lattice and raise/warn otherwise. Also, `film_slabs[0]` will `IndexError` if `get_slabs` returns empty, where the old code would not.
2. **A test is feasible even without a reproducer.** You cannot reproduce #4651, but you can lock in the new invariant: assert that the lattices in `zsl_matches` match those of the first-termination slabs from `_find_terminations` for an existing interface test case. That prevents silent regression back to `shift=0`.
3. **Docstring will fail ruff/pydocstyle.** `_find_terminations` needs a blank line between the summary and the `Returns:` section (`D205`), and the Google convention does not name the return variable — drop `first_pair (tuple[Slab, Slab])` and just describe the tuple.
4. Minor readability: prefer `film_slab, sub_slab = self._find_terminations()` followed by `self._find_matches(film_slab, sub_slab)` over `self._find_matches(*self._find_terminations())`. `_find_terminations` now both mutates state and returns a value, which is worth a one-line comment.

No CI has run on this branch (workflows awaiting approval), so correctness against the existing interface tests is unverified.

shyuep:
Thanks for fixing this issue. Can you add a unittest pls? Using a new example is fine. There is no need to rely on the existing examples.

Sanftperlig:
I sadly don't have time to work on this more significantly right now, but I will see if I can create a small test - I'm not too familiar with the `CoherentInterfaceBuilder` compared to the `SlabGenerator`, so the test might not be too great.
I'm not familiar with semiconductors, so it would probably be a test around the calcite case where I know this issue is a thing.

To address the Claude comments: To my knowledge, `get_slabs` always returns slabs if valid terminations can be found. That should always be the case (unless bonds are defined, which they aren't in this case)? Thus it would be fine to use [0].
As far as I am aware any produced terminations result in the same slab lattice, thus checking the other indices shouldn't be necessary.

Sorry for the delay and thank you for running the CI.

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep; not a human review — please sanity-check before acting.)

Good catch on the root cause: hard-coding `shift=0` in `_find_matches` while `get_interfaces` uses the termination-specific shifts is exactly the kind of inconsistency behind #4651, and reusing the slabs from `_find_terminations` also removes a duplicate `SlabGenerator` pass. Three things to fix before merge.

**1. The new test will error, not pass.** `Interface.lattice` is a `Lattice`, which does not define `__getitem__`:

```python
assert np.isclose(inter.lattice[0, 0], 5, atol=1e-4)   # TypeError: 'Lattice' object is not subscriptable
```

Use `inter.lattice.matrix[0, 0]` / `[1, 1]`. Since no CI has run on this PR yet (zero status checks on `3ff6ad2a`), this was not caught. While there, prefer the already-imported `assert_allclose(inter.lattice.matrix[0, 0], 5, atol=1e-4)` over `assert np.isclose(...)` for the file's style and better failure messages — that also drops the new `import numpy as np`.

**2. `film_slabs[0]` / `sub_slabs[0]` is an unguarded, arbitrary choice.**

- `IndexError` if `get_slabs()` returns an empty list. The old `get_slab(shift=0)` could not fail this way, so this is a new failure mode for edge-case Miller indices. Worth an explicit error message.
- More substantively: the ZSL matches are now computed from the *first* termination's primitivized cell but applied to *every* termination in `get_interfaces`. That is only sound if the in-plane lattice is termination-independent. If that invariant is assumed, please state it in the docstring; if it is not guaranteed, a cheap `assert`/raise comparing `film_slabs[i].lattice.matrix[:2]` across terminations would surface the violation instead of silently producing wrong matches. The current docstring ("A slab of the film structure, transformed as used in the other functions") glosses over this.

**3. Behavior change is user-visible and undocumented.** `zsl_matches`, and hence the strains and supercells returned by `get_interfaces`, can change for existing users whose structures primitivize shift-dependently. Please say so in the PR description and flag it for release notes — silently different interface geometries are hard to debug downstream (atomate2 uses this).

Minor: the hard-coded termination label `"CaCO_Pmma_6"` makes the test brittle against any change in `label_termination` or spglib version; consider selecting `cib.terminations[0]` instead. Also the checklist still shows "Tests added" unchecked although a test is present.

Nothing else concerning in the refactor — the `__init__` reordering is safe since `_find_terminations` does not read `zsl_matches`, and the added type annotations are correct.

shyuep:
**Automated PR review generated by Claude** (sanity-check before acting on it)

Nice, well-scoped fix. `_find_matches` previously rebuilt its own `shift=0` slabs via a fresh `SlabGenerator`, which could primitivize differently than the slabs actually enumerated in `_find_terminations`, so the ZSL lattice-matching could run against a lattice inconsistent with the interface later returned by `get_interfaces`. Reusing `film_slabs[0]/sub_slabs[0]` from `_find_terminations` closes that gap, and `test_shiftdependent_primitive` (calcite, space group 167) is a good regression test that actually exercises the shift-dependent primitivization case rather than just re-asserting the old behavior.

Two things worth a look before merge:
- `_find_terminations` returns `film_slabs[0], sub_slabs[0]` unconditionally. If `filter_out_sym_slabs` (or `ftol`) ever yields an empty `film_slabs`/`sub_slabs` for a given miller index, this is now an `IndexError` in `__init__` instead of the old independent `get_slab(shift=0)` call, which always succeeded. Worth a guard/clearer error if that's reachable in practice.
- No CI has run on this PR (`no checks reported on the 'shift-change' branch`) — it's a fork PR (`Sanftperlig/pymatgen`), so this is likely just pending workflow approval rather than a CI problem, but flagging since there's no green signal yet to weigh against the diff.

Diff itself (interfaces + reordered `__init__`) looks correct and minimal.

shyuep:
**Automated PR review generated by Claude** (posted on behalf of @shyuep; not a human review — please sanity-check before acting.)

Solid fix, well-targeted at the root cause. `_find_terminations` already builds `film_slabs`/`sub_slabs` via `get_slabs()`; reusing `film_slabs[0]`/`sub_slabs[0]` in `_find_matches` instead of spinning up two more `SlabGenerator`/`get_slab(shift=0)` calls removes duplicated, inconsistent slab generation and directly addresses #4651 (shift-0 primitivization sometimes disagreeing with other shifts). The new `np.allclose` guard on `matrix[:2]` (in-plane a/b vectors only, ignoring the vacuum-dependent c-vector) is the right check, and turning a previously-silent wrong-lattice bug into a loud `ValueError` is the correct call for a scientific library — much better than quietly returning bad ZSL matches. `itertools.product` ordering confirms `self.terminations[0]` does correspond to `(film_slabs[0], sub_slabs[0])`, so the docstring's "first termination" claim checks out. The new `test_shiftdependent_primitive` calcite case is a good regression test and the PR body is upfront about the behavior change (interfaces/ZSL matches can shift for structures where shift-0 primitivization previously failed).

Two things worth a look before merge:
- No CI has run on this PR yet (zero status checks on the latest commit) — given the new `ValueError` guards are stricter than before, it'd be worth confirming the full `analysis/interfaces` suite (and anything downstream, e.g. `pymatgen-core`) still passes rather than newly raising on existing fixtures.
- Minor nit: the film/sub lattice-consistency blocks in `_find_terminations` are near-identical (empty-list check + `allclose` loop + `ValueError`); could be factored into a small helper to avoid the duplication, though not blocking.

